PR #87, opened 2025-05-20
feat: Add label selectors to listing tools

eranco74:
This PR introduces the ability to filter Kubernetes resources by label using a labelSelector parameter for the following tools:

 * pods_list
 * pods_list_in_namespace
 * resources_list

This enhancement allows users to retrieve a more specific set of resources based on their labels, improving the flexibility and utility of these tools.

The labelSelector parameter accepts standard Kubernetes label selector syntax, such as app=myapp,env=prod or app in (myapp,yourapp).

Resolves: https://github.com/manusa/kubernetes-mcp-server/issues/54

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,7 +241,9 @@ Get a Kubernetes Pod in the current or provided namespace with the provided name
 
 List all the Kubernetes pods in the current cluster from all namespaces
 
-**Parameters:** None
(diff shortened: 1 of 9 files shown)


manusa:
Tested in Goose CLI with `gemini-2.0-flash`

Issue #83, opened 2025-05-16
Support for multiple contexts and switching between them


dbsanfte:
Currently there doesn't seem to be a way to handle multiple contexts. This is a pretty tough limitation.

In my agent flows, I typically need to access several clusters in a group. There is a k8s mcp that has implemented this quite well:

https://github.com/razvanmacovei/k8s-multicluster-mcp

Might be worth a look and implementing the same.

manusa:
We can try to add an **optional** `context` parameter for each function/tool.

- The parameter is only added in case the MCP Server is a local deployment based on a kubeconfig file that has more than one context (to avoid confusing smaller LLMs)
- The parameter description will include a list of contexts that the LLM can select (this might be problematic in case the kubeconfig file has hundreds of contexts)

/cc @alexeykazakov

alexeykazakov:
> The parameter description will include a list of contexts that the LLM can select (this might be problematic in case the kubeconfig file has hundreds of contexts)

Right. Or, as you said, it could be also a new tool: list-contexts. But in this case it would always require to do this extra tool call to list all available contexts before calling the actual tool.

ardaguclu:
In multiple context environment, it is clear that we can't use MCP Server's kubeconfig which is tied to a single cluster. Another point is that user has to pass `bearer` to authenticate the MCP Server. 

Under these two facts I think: Enriching the `audience` field of bearer token with the cluster API Server url that is tried to be interacted with can resolve this. As complementary, MCP Server can ask the cluster url via Elicitation.

EDITED: Delegating the context management to bearer token or elicitation will remove the unexpected context usages triggered by LLM.

ardaguclu:
> We can try to add an optional context parameter for each function/tool.

This brings about the requirement to pass context parameter in each request. Also what will happen if user modifies context parameter in each tool call within the same session?.

ardaguclu:
As far as I understand, https://github.com/razvanmacovei/k8s-multicluster-mcp supports defining multiple kubeconfigs in that MCP Server. If I am unwanted actor who can reach to this MCP Server, there are plenty of kubeconfigs that are usable for any unwanted operation.

alexeykazakov:
Imaging a situation when an agent needs to get access to multiple clusters. In different sessions of with the same session. For example it has a task to do something in a multi-cluster system.
IMO, the MCP Server should not be in the business of policing what the cluster the client can access and what it can't. It's up to the kubeconfig to dictate it.
[...]

PR #97, opened 2025-05-28
fix(auth): ensure OIDC auth plugin is included in compiled binary

ziadmoubayed:
### Summary:

This PR fixes startup failures when the server is pointed at clusters that authenticate with OIDC.
We add a single side-effect import:

```go
import _ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
```

which registers the OIDC plug-in at init() time so that client-go can recognise the auth-provider: oidc stanza found in many kubeconfigs.

### Motivation: 

Running `npx kubernetes-mcp-server@0.0.30` against an OIDC-backed cluster produced:

```
Failed to initialize MCP server: no Auth Provider found for name "oidc"
```

### DIFF
--- a/pkg/kubernetes-mcp-server/cmd/root.go
+++ b/pkg/kubernetes-mcp-server/cmd/root.go
@@ -17,6 +17,10 @@ import (
 	"strings"
 )
 
+import (
+	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
+)
+
 var rootCmd = &cobra.Command{
 	Use:   "kubernetes-mcp-server [command] [options]",
 	Short: "Kubernetes Model Context Protocol (MCP) server",


ziadmoubayed:
I am not a go expert. Please let me know if the import should be place somewhere else

manusa (approved):
LGTM, thx!

ziadmoubayed:
> Thx! This is awesome (maybe we want to add other plugins too).
> 
> > I am not a go expert. Please let me know if the import should be place somewhere else
> 
> To keep things in the same domain package, maybe it's better if we add the plugin import in between these lines:
> 
> https://github.com/manusa/kubernetes-mcp-server/blob/2bd12f8a64f9eecfd7e5d9287f3dc3f2cc209ff0/pkg/kubernetes/kubernetes.go#L14-L15
> 
> and remove it from the cmd/root.go
> 
> could you please check if this works too?

~~Thanks for the review. I did a test today. It works now. Facing #83 unfortunately.~~

Continued the support in https://github.com/manusa/kubernetes-mcp-server/pull/105

PR #102, opened 2025-06-01
(chore): steps to deploy in kubernetes cluster

chiragkyal:


manusa:
Once we completely implement OAuth, we can create Kubernetes resource manifests and Helm packages to easily deploy this into a cluster (in a safe way).

manusa:
Chirag, thanks for your attempt to contribute this. It was never merged because when it was originally proposed, deploying to Kubernetes without adding a security layer would have been sending a wrong message from our side.
Since then, a helm chart is already provided for this purpose.
Closing as the PR has become stale.

PR #125, opened 2025-06-17
refactor(kubernetes): force usage of Derived kubernetes

manusa:
Prevents consumers of the kubernetes package the usage of public methods on a non-derived config instance.

/cc @ardaguclu

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -6,20 +6,17 @@ import (
 	"github.com/manusa/kubernetes-mcp-server/pkg/helm"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
-	clientgokubernetes "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/klog/v2"
-	"k8s.io/metrics/pkg/apis/metrics"
 	"strings"
 )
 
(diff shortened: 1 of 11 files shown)


pkg/kubernetes/kubernetes.go
```diff
@@ -26,23 +29,51 @@ const (
...
-type Kubernetes struct {
+type Kubernetes interface {
+	WatchKubeConfig(onKubeConfigChange func() error)
+	Close()
+	Derived(ctx context.Context) DerivedKubernetes
+	ConfigurationView(minify bool) (runtime.Object, error)
+	IsOpenShift(ctx context.Context) bool
+}
+
+type DerivedKubernetes interface {
```
ardaguclu: This interface is too rigid to be used to implement different `DerivedKubernetes` objects. In my opinion, having a struct is better, probably after renaming the `Kubernetes` (which hides the package names) to `KubernetesManager`.

pkg/kubernetes/kubernetes.go
```diff
@@ -26,23 +29,51 @@ const (
 
 type CloseWatchKubeConfig func() error
 
-type Kubernetes struct {
+type Kubernetes interface {
```
ardaguclu: Same, I don't think we need this for now

pkg/kubernetes/kubernetes.go
```diff
@@ -26,23 +29,51 @@ const (
...
+	PodsTop(ctx context.Context, options PodsTopOptions) (*metrics.PodMetricsList, error)
+	PodsExec(ctx context.Context, namespace, name, container string, command []string) (string, error)
+	ProjectsList(ctx context.Context, options ResourceListOptions) (runtime.Unstructured, error)
+	ResourcesList(ctx context.Context, gvk *schema.GroupVersionKind, namespace string, options ResourceListOptions) (runtime.Unstructured, error)
+	ResourcesGet(ctx context.Context, gvk *schema.GroupVersionKind, namespace, name string) (*unstructured.Unstructured, error)
+	ResourcesCreateOrUpdate(ctx context.Context, resource string) ([]*unstructured.Unstructured, error)
+	ResourcesDelete(ctx context.Context, gvk *schema.GroupVersionKind, namespace, name string) error
+}
+
+type kubernetes struct {
```
ardaguclu: I'd recommend renaming it to `KubernetesManager`. I don't think, exporting the fields in here hurts us. However, in my opinion, maybe we can get rid of some of them (e.g. I bet we don't need parameterCodec by using an upstream functionality.).

pkg/kubernetes/kubernetes.go
```diff
@@ -159,6 +189,19 @@ func (k *Kubernetes) Derived(ctx context.Context) *Kubernetes {
 	if err != nil {
 		return k
 	}
-	derived.Helm = helm.NewHelm(derived)
 	return derived
 }
+
+func (k *kubernetes) CacheInvalidate() {
```
ardaguclu: Why do we need invalidating the cache?. Doesn't MCP server interact with the same cluster throughout its lifetime?. Stdio would require this. But maybe when stdio is used, we fully use a different path.

manusa: This is currently used for testing purposes only.
In this case, the cache becomes stale because the metrics-api is made available after an initial check.

As stated, this isn't used as of now in production code, but I assume that similar circumstances might happen in a real cluster (cluster is the same, available resources might change).

ardaguclu: DiscoveryClient has etag mechanism that compares the cached hash values against the API Server each time. If it detects any change in etags, it internally invalidates the cache.

manusa: We should check that together, the test doesn't validate that behavior (I think). (I'll add a TODO)

ardaguclu: Yeah, this invalidation mechanism requires to have an actual API server. So I wouldn't be surprised if unit tests fails due to this. Would it be an issue to add this invalidation in unit test?.

ardaguclu: https://github.com/kubernetes/enhancements/issues/3352 in case you are interested in

manusa: The test using this is here: https://github.com/manusa/kubernetes-mcp-server/blob/54d564a54e38febc63c56c60b9ccb76501adf55f/pkg/mcp/pods_top_test.go#L10

There's a fake API server behind which does reply. Only one request is performed.
I'll set the TODO and we can check later

manusa:
@ardaguclu, please check if the subsequent commit aligns with your expectations

ardaguclu:
This looks good to me
/lgtm

Issue #51, opened 2025-04-18
Allow specifying the KUBECONFIG to use vs. in-cluster


humphd:
I have a specific use-case: I'm running in an EKS cluster, but I want to connect the MCP server to *another* EKS cluster.  My plan had been to authenticate and generate a kube config, set `KUBECONFIG` in the env, then have the MCP Server use that.  This works locally; but when I run it in production, it looks like it prefers the in-cluster config.

I'd really like to use this MCP server for k8s vs. others I've seen, but this is a deal-breaker for my use case.  Any way around it, or possible to fix via arg/env I can pass through from the MCP client?

manusa:
> This works locally; but when I run it in production, it looks like it prefers the in-cluster config.

Yes, this is something I didn't anticipate.
When used in-cluster, the server always prefer the cluster-provided configuration.

---

add `--kubeconfig` flag as described in https://kubernetes.io/docs/reference/kubectl/generated/kubectl_config/

manusa:
Fixed in https://github.com/manusa/kubernetes-mcp-server/commit/fa5bb81fe56ae9794c101f14f16280ab7c1b4574

manusa:
I just released https://github.com/manusa/kubernetes-mcp-server/releases/tag/v0.0.28 with the fix.
You should now be able to provide an explicit `--kubeconfig` flag to start the server and it will use that file or fail (no autoconfig).

manusa:
Please, let me know if it works so I can close the issue.

humphd:
@manusa thanks for jumping on this.  Yes, I can confirm that `--kubeconfig <file>` is working for me.

However, I've noticed another issue now that I can connect: in the HTTP logs I can see your server trying to hit `/apis/project.openshift.io/v1/projects` on my cluster, which fails (401) not least because I'm connecting to an Amazon EKS cluster, not OpenShift.

Is it possible to not assume OpenShift, or is that a key part of what you're trying to offer with this (e.g., should I find a different MCP Server)?

manusa:
This is because you enabled fine-logging.
It's hitting that endpoint (should be only on start) to see if you're running an OpenShift cluster and then provide enhanced functionality.
I assumed this shouldn't bother users too much (in fact I was planning to add enhanced functionality for other kinds of clusters too).

I could add an additional flag to completely disable this, but I'm not sure it is worth the effort.

humphd:
No, that all makes sense.  I'll close this as "completed," great work.

PR #128, opened 2025-06-18
Introduce version as separate command

ardaguclu:
This PR;
* Introduces version as a separate command instead of a flag
* Unifies port number flag to be used by sse and streamable and updates it's type from int to string (i.e. `:8080`)
* Adds more validations
* Introduces `type` flag to be used to select the transport method instead of trying to extract from the port usages.

Note: kubeconfig is not passed to sse and streamable options. Assuming that this won't break.

### DIFF
--- a/pkg/kubernetes-mcp-server/cmd/root.go
+++ b/pkg/kubernetes-mcp-server/cmd/root.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"fmt"
 	"regexp"
@@ -153,7 +152,7 @@ func (m *MCPServerOptions) Run() error {
 
 	switch m.MCPType {
 	case "streamable":
-		mcpServer, err := mcp.NewSever(mcp.Configuration{
+		mcpServer, err := mcp.NewServer(mcp.Configuration{
 			Profile:            m.profileObj,
 			ListOutput:         m.listOutputObj,
 			ReadOnly:           m.ReadOnly,
@@ -170,7 +169,7 @@ func (m *MCPServerOptions) Run() error {
 			return fmt.Errorf("failed to start streaming HTTP server: %w\n", err)
 		}
 	case "sse":
-		mcpServer, err := mcp.NewSever(mcp.Configuration{
+		mcpServer, err := mcp.NewServer(mcp.Configuration{
 			Profile:            m.profileObj,
 			ListOutput:         m.listOutputObj,
 			ReadOnly:           m.ReadOnly,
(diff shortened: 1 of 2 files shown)


manusa:
Some comments:
- I'd rather keep both options to retrieve the version (flag + command)
- For port, I think it's also better to keep separate flags as opposed to a single `port` flag.
  - Why?
    - Backwards compatibility
    - Option to start the server in both modes (this is not currently supported but might be interesting)
      For example, the container image could start both servers which would simplify the distribution.
      Other alternatives include creating a [backwards compatible endpoint](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#backwards-compatibility) that would select one server or the other, but it will require more effort to implement.
- kubeconfig might be needed by those options too, it shouldn't be limited to the stdio mode

ardaguclu:
> Some comments:
> 
> * I'd rather keep both options to retrieve the version (flag + command)
> * For port, I think it's also better to keep separate flags as opposed to a single `port` flag.
>   
>   * Why?
>     
>     * Backwards compatibility
>     * Option to start the server in both modes (this is not currently supported but might be interesting)
>       For example, the container image could start both servers which would simplify the distribution.
>       Other alternatives include creating a [backwards compatible endpoint](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#backwards-compatibility) that would select one server or the other, but it will require more effort to implement.

I understand the backwards compatibility concerns in this repo and it makes sense. The one in the openshift, we don't have backwards compatibility concerns initially. Everyone should move to newer protocol, I couldn't envision that an MCP Server would initiate sse and streamable simultaneously. 

> * kubeconfig might be needed by those options too, it shouldn't be limited to the stdio mode

In my opinion, using the passed kubeconfig rather than the user's token is fundamental flaw in any scenario. In my opinion, it would be safe not passing at all. Do you have any use case by passing this? (again I'm more asking about OpenShift one, as the maintainer of this repository probably some people use it).

ardaguclu:
cc'ing @mrunalp

manusa:
> Do you have any use case by passing this?

This are _some of the_ common deployment scenarios and the applicable transport modes:
- MCP server running in the same machine as the MCP client (**local**) connecting to a **remote cluster**, typically **STDIO** maybe **SSE** and **HTTP**.
- MCP server running in a **remote** machine exposed by an MCP gateway connecting to a **remote cluster** or if running in a Pod to the **local cluster**, **STDIO** (or the others?) reverse-proxied by the gateway (for example https://github.com/supercorp-ai/supergateway)
- MCP server running in a **remote** Kubernetes cluster connecting to the same cluster (in-cluster), **SSE** or **HTTP**
- MCP server running in a **remote** Kubernetes cluster connecting to a **different cluster** (kubeconfig) #51, **SSE** or **HTTP**
- MCP server running in a **remote** machine (not kubernetes) connecting to a **remote cluster**, **SSE** or **HTTP**

The _problem_ is that there are many deployment scenarios hence the current flexibility. IMO most of this can be applicable for a downstream product too. But we should probably have a more thorough discussion and analysis.

Issue #129, opened 2025-06-18
Sensitive information redacting


manusa:
## Description

As a user, I want to be able to start the Kubernetes MCP Server providing a configuration option that will redact sensitive information from responses (tool calls, resources, etc.), so that sensitive information is not shared with the LLM or the MCP client.

- Configuration flag? `--redact`??
- Resources to be redacted (TBD):
  - Secrets
  - ...
- If possible, reuse built-in Kubernetes tooling?

manusa:
Implemented in #629 (logging capability with secret redaction) and further expanded in #792 (centralized K8s error logging). The redaction layer covers passwords, bearer tokens, basic auth, API keys, AWS credentials, GitHub/GitLab tokens, GCP API keys, OpenAI/Anthropic API keys, JWT tokens, SSH private keys, and database connection strings.

PR #130, opened 2025-06-18
Prohibit any operation for Secret resource

ardaguclu:
This PR prohibits any operation with regards to `Secret` resource.


manusa:
This should be configurable.

Somehow related to #129
Might require #131

manusa:
Original idea was to provide an allolist and denylist #132

Which also relates to providing configuration through toml file

ardaguclu:
Thanks for the pointers. It is better to have general approach as you mentioned. Closing...

manusa:
> Thanks for the pointers. It is better to have general approach as you mentioned. Closing...

np, I think it should be better to see how to tackle #131 first to be able to easily provide a deny list (for example).
Implementation after that should then be extremely easy, it could even be done with a generic interceptor

Issue #132, opened 2025-06-18
Allow/Deny resources list


manusa:
## Description

As a user/mcp admin, I want to be able provide a list of allowed or denied resources, so that I can control and block access to a given set of Kubernetes resources.

Requires #131 
Relates to #129

manusa:
> ```toml
> # A list of denied Kubernetes resources in Group/Version/Kind format.
> # If a resource is in this list, your MCP server should deny all operations
> # on that resource type across all namespaces.
> [[denied_resources]]
> ```
> Would this format make sense to you?.

https://github.com/manusa/kubernetes-mcp-server/pull/133#discussion_r2156551595

PR #133, opened 2025-06-19
Introduce toml configuration file with a set of deny list

### LINKED ISSUE #131: Configuration file
## Description

As a user/mcp admin, I want to be able to start the server using a configuration file, so that I can easily share configurations and simplify the command entry-point.

Configuration command flags should remain available to be able to override some of the configurations provided in the file for specific command invocations (one of).

Initially discussed providing the configuration as a toml file (similar to what other Kubernetes ecosystem tools accept).

- Config file based on TOML format
- Config file can be provided to the MCP server through the `--config` command flag

ardaguclu:
Fixes #131 
Introduces PoC for #132 

This PR introduces a new static configuration file in toml format that defaults to `conf.toml` but can be modified via flag `config`. 

Additionally, this PR adds predefined set of resources to not allow any operations in MCP Server.

This PR is supposed to fix https://github.com/manusa/kubernetes-mcp-server/issues/132 and https://github.com/manusa/kubernetes-mcp-server/issues/131

### DIFF
--- a/pkg/mcp/common_test.go
+++ b/pkg/mcp/common_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/manusa/kubernetes-mcp-server/pkg/config"
 	"github.com/manusa/kubernetes-mcp-server/pkg/output"
 	"github.com/mark3labs/mcp-go/client"
 	"github.com/mark3labs/mcp-go/client/transport"
@@ -100,6 +101,7 @@ type mcpContext struct {
 	listOutput         output.Output
 	readOnly           bool
 	disableDestructive bool
+	staticConfig       *config.StaticConfig
 	clientOptions      []transport.ClientOption
 	before             func(*mcpContext)
 	after              func(*mcpContext)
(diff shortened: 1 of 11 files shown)


pkg/config/conf.toml
```diff
@@ -0,0 +1,16 @@
+# A list of denied Kubernetes resources in Group/Version/Kind format.
+# If a resource is in this list, your MCP server should deny all operations
+# on that resource type across all namespaces.
+[[denied_resources]]
```
ardaguclu: Would this format make sense to you?.

pkg/config/conf.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "v1"
+kind = "ServiceAccount"
+
+[[denied_resources]]
+group = ""
+version = "v1"
+kind = "Secret"
+
+[[denied_resources]]
+group = "rbac.authorization.k8s.io"
```
ardaguclu: We allow group based prohibitions.

pkg/mcp/mcp.go
```diff
@@ -2,6 +2,7 @@ package mcp
 
 import (
 	"context"
+	"github.com/manusa/kubernetes-mcp-server/pkg/config"
```
ardaguclu, line 5: nit: Missing go import

pkg/mcp/mcp_test.go
```diff
@@ -2,6 +2,7 @@ package mcp
 
 import (
 	"context"
+	"github.com/manusa/kubernetes-mcp-server/pkg/config"
```
ardaguclu: nit: missing go import

ardaguclu:
@manusa @mrunalp

pkg/mcp/common_test.go
```diff
@@ -125,11 +127,26 @@ func (c *mcpContext) beforeEach(t *testing.T) {
...
+					Version: "v1",
+					Kind:    "Secret",
+				},
+				{
+					Group:   "rbac.authorization.k8s.io",
+					Version: "v1",
+				},
+			},
+		}
+	}
```
manusa, line 143: > Additionally, this PR adds predefined set of resources to not allow any operations in MCP Server.

I wouldn't set any opinionated defaults (at least for the default _upstream_ full profile)

Maybe we want to discuss if we want to provide an upstream `safe` profile or maybe even a mode of operation that users could activate and added the opinionated denied resources.

Everything else looks good :)

pkg/config/conf.toml
manusa: Let's move this discussion to #132

manusa (approved):
LGTM, thx!

PR #134, opened 2025-06-19
Introduce new full-safe profile with a default resource deny list

ardaguclu:
After the https://github.com/manusa/kubernetes-mcp-server/pull/133 has merged, we'll have an option to deny operations for some resources.

This PR introduces new profile, namely `full-safe` that brings in with a set of resources that will be denied. Additionally, this profile will not serve a tool for viewing kubeconfig. 

There won't be any change in default behaviors.

### DIFF
--- a/pkg/kubernetes-mcp-server/cmd/root_test.go
+++ b/pkg/kubernetes-mcp-server/cmd/root_test.go
@@ -74,7 +74,7 @@ func TestProfile(t *testing.T) {
 		rootCmd := NewMCPServer(ioStreams)
 		rootCmd.SetArgs([]string{"--help"})
 		o, err := captureOutput(rootCmd.Execute) // --help doesn't use logger/klog, cobra prints directly to stdout
-		if !strings.Contains(o, "MCP profile to use (one of: full) ") {
+		if !strings.Contains(o, "MCP profile to use (one of: full, full-safe) ") {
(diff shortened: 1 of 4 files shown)


ardaguclu:
@manusa what is your opinion about this one?.

manusa:
> @manusa what is your opinion about this one?.

Sorry for the late reply, it's been an intense morning.
It looks good.
My only doubt is about the name plus if this is aligned with @mrunalp expectations.

Since the changes are very isolated, I think we can hold it until he checks it out, I want to merge the dependabot PRs and maybe cut a release today.

ardaguclu:
@mrunalp could you please have a look at and drop your thoughts?.

ardaguclu:
This PR needs to be updated based on the latest changes.
/hold

ardaguclu:
I think, we don't need another profile (i.e. `full-safe`) any more, after the latest changes in the code base.

Explicitly passing this configuration via `--config` flag satisfies what new profile tries to achieve;

```toml
denied_resources = [
    {group = "apps", version = "v1", kind = "ServiceAccount"},
    {group = "apps", version = "v1", kind = "Secret"},
    {group = "rbac.authorization.k8s.io", version = "v1"}
]

disabled_tools = ["configuration_view"]
```

So that I'm closing this PR.

PR #146, opened 2025-06-20
Gracefully cast objects and return err instead of panic

ardaguclu:
I noticed the panic while testing due to the unexpected types of objects. This PR checks the type first and returns error gracefully to prevent panic (which results in the crash of application)


manusa:
This seems more like a bad client implementation sending types which don't adhere to the "contract".
In any case, better safe than sorry :) :rocket:

manusa (approved):
LGTM, thx!

PR #149, opened 2025-06-23
feat(config): deny resources by using RESTMapper as an interceptor

### LINKED ISSUE #132: Allow/Deny resources list

manusa:
Fixes #132

This approach ensures that resources in the deny list are **always** processed regardless of the implementation.

The RESTMapper takes care of verifying that the requested Group Version Kind complies with the deny list while checking for the REST endpoint.

/cc @ardaguclu

### DIFF
--- a/pkg/kubernetes/accesscontrol.go
+++ b/pkg/kubernetes/accesscontrol.go
@@ -1,0 +1,40 @@
+package kubernetes
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/manusa/kubernetes-mcp-server/pkg/config"
+)
+
+// isAllowed checks the resource is in denied list or not.
+// If it is in denied list, this function returns false.
+func isAllowed(
+	staticConfig *config.StaticConfig, // TODO: maybe just use the denied resource slice
+	gvk *schema.GroupVersionKind,
+) bool {
+	if staticConfig == nil {
+		return true
+	}
+
+	for _, val := range staticConfig.DeniedResources {
+		// If kind is empty, that means Group/Version pair is denied entirely
+		if val.Kind == "" {
+			if gvk.Group == val.Group && gvk.Version == val.Version {
+				return false
+			}
+		}
+		if gvk.Group == val.Group &&
+			gvk.Version == val.Version &&
+			gvk.Kind == val.Kind {
+			return false
+		}
+	}
+
+	return true
+}
+
+func isNotAllowedError(gvk *schema.GroupVersionKind) error {
+	return fmt.Errorf("resource not allowed: %s", gvk.String())
+}
--- a/pkg/kubernetes/accesscontrol_clientset.go
+++ b/pkg/kubernetes/accesscontrol_clientset.go
@@ -5,13 +5,16 @@ import (
 	"fmt"
 
 	authorizationv1api "k8s.io/api/authorization/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/httpstream"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	authorizationv1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
 	"k8s.io/metrics/pkg/apis/metrics"
 	metricsv1beta1api "k8s.io/metrics/pkg/apis/metrics/v1beta1"
 	metricsv1beta1 "k8s.io/metrics/pkg/client/clientset/versioned/typed/metrics/v1beta1"
@@ -23,6 +26,7 @@ import (
 // Only a limited set of functions are implemented with a single point of access to the kubernetes API where
 // apiVersion and kinds are checked for allowed access
 type AccessControlClientset struct {
+	cfg             *rest.Config
 	delegate        kubernetes.Interface
 	discoveryClient discovery.DiscoveryInterface
 	metricsV1beta1  *metricsv1beta1.MetricsV1beta1Client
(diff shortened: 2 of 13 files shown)


pkg/kubernetes/accesscontrol_restmapper.go
```diff
@@ -0,0 +1,106 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/restmapper"
+
+	"github.com/manusa/kubernetes-mcp-server/pkg/config"
+)
+
+type AccessControlRESTMapper struct {
+	delegate     *restmapper.DeferredDiscoveryRESTMapper
+	staticConfig *config.StaticConfig // TODO: maybe just store the denied resource slice
```
ardaguclu, line 13: In my opinion, we can pass `staticConfig` in here instead of denied resource slice. Maybe in the future, we want to extend this based on allowed resources similar to denied resources.

pkg/kubernetes/accesscontrol_restmapper.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+var _ meta.RESTMapper = &AccessControlRESTMapper{}
+
+// isAllowed checks the resource is in denied list or not.
+// If it is in denied list, this function returns false.
+func (a AccessControlRESTMapper) isAllowed(gvk *schema.GroupVersionKind) bool {
+	if a.staticConfig == nil {
```
ardaguclu: This logic is true. But if staticConfig is accidentally passed as nil, this would mean we allow all resources. Just brainstorming, would it be too disruptive if we return false, if staticConfig is nil? or can we return error in `NewAccessControlRESTMapper` if staticConfig is nil?. WDYT?

pkg/kubernetes/kubernetes.go
```diff
@@ -13,6 +13,7 @@ import (
 	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
```
ardaguclu: nit: this should be at the end of all imports, in order to emphasize it is in here just for dependencies.

ardaguclu (commented):
This change looks nice. Dropped a few comments.

pkg/kubernetes/accesscontrol_restmapper.go
ardaguclu: I think, I'm leaning towards your TODO comment. We can pass only denyList instead of an entire staticConfig (as it includes different type of resources)

pkg/kubernetes/accesscontrol_restmapper.go
manusa: Note: the method is the same as the original method, I just moved it here.

It's very likely that this method will be deeply refactored in the future.

Currently, `staticConfig` can be nil (in fact will be nil unless a config is explicitly loaded.

If we go with https://github.com/manusa/kubernetes-mcp-server/issues/151#issuecomment-3018150180 or similar, then probably staticConfig will always have a value and we can probably change this.

pkg/kubernetes/accesscontrol_restmapper.go
manusa: That's why I left a TODO comment, I think that we can leave it and reconsider it once we have more pieces of the puzzle.

pkg/kubernetes/kubernetes.go
manusa: seems the IDE moved it, changing

pkg/kubernetes/accesscontrol.go
```diff
@@ -0,0 +1,40 @@
...
+			gvk.Kind == val.Kind {
+			return false
+		}
+	}
+
+	return true
+}
+
+func isNotAllowedError(gvk *schema.GroupVersionKind) error {
+	return fmt.Errorf("resource not allowed: %s", gvk.String())
```
ardaguclu, line 39: (not a blocker for this PR):
We have to return 403: Forbidden HTTP Status. But I'm not really sure that mcp library we use support this. If it doesn't support returning http code we have, we'll have major issue in oauth side as well.

pkg/kubernetes/accesscontrol_clientset.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (a *AccessControlClientset) Pods(namespace string) (corev1.PodInterface, error) {
+	gvk := &schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Pod"}
+	if !isAllowed(a.staticConfig, gvk) {
+		return nil, isNotAllowedError(gvk)
+	}
+	return a.delegate.CoreV1().Pods(namespace), nil
+}
+
+func (a *AccessControlClientset) PodsExec(namespace, name string) (*rest.Request, error) {
```
ardaguclu: What about moving `createExecutor` also in here and this function returns `remotecommand.Executor`?

pkg/kubernetes/accesscontrol.go
ardaguclu: It just simply sets http 200 statically https://github.com/mark3labs/mcp-go/blob/1eddde7bd69b760f745a1b4064969cffcf97e935/server/streamable_http.go#L347 :(

manusa: Yes, I don't think the library (or even the [protocol](https://modelcontextprotocol.io/specification/draft/basic/authorization)) is prepared for fine-grained operation status codes yet.
As I understand it, the protocol negotiates authentication and authorization first, and then enables

image

pkg/kubernetes/accesscontrol_clientset.go
manusa: This should be done now, Schema and ParameterCodec are now reused from client-go globals

pkg/kubernetes/resources.go
```diff
@@ -228,7 +188,11 @@ func (k *Kubernetes) supportsGroupVersion(groupVersion string) bool {
 }
 
 func (k *Kubernetes) canIUse(ctx context.Context, gvr *schema.GroupVersionResource, namespace, verb string) bool {
-	response, err := k.manager.clientSet.AuthorizationV1().SelfSubjectAccessReviews().Create(ctx, &authv1.SelfSubjectAccessReview{
+	accessReviews, err := k.manager.accessControlClientSet.SelfSubjectAccessReviews()
+	if err != nil {
+		return false
```
ardaguclu, line 193: very nice!

pkg/kubernetes/resources.go
```diff
@@ -113,15 +101,15 @@ func (k *Kubernetes) resourcesListAsTable(ctx context.Context, gvk *schema.Group
 	}
 	url = append(url, gvr.Resource)
 	var table metav1.Table
-	err := k.manager.clientSet.CoreV1().RESTClient().
+	err := k.manager.discoveryClient.RESTClient().
```
ardaguclu, line 104: I think, now we support all the resources

pkg/kubernetes/resources.go
```diff
@@ -113,15 +101,15 @@ func (k *Kubernetes) resourcesListAsTable(ctx context.Context, gvk *schema.Group
...
 		Get().
 		SetHeader("Accept", strings.Join([]string{
 			fmt.Sprintf("application/json;as=Table;v=%s;g=%s", metav1.SchemeGroupVersion.Version, metav1.GroupName),
 			fmt.Sprintf("application/json;as=Table;v=%s;g=%s", metav1beta1.SchemeGroupVersion.Version, metav1beta1.GroupName),
 			"application/json",
 		}, ",")).
 		AbsPath(url...).
-		SpecificallyVersionedParams(&options.ListOptions, k.manager.parameterCodec, schema.GroupVersion{Version: "v1"}).
```
ardaguclu, line 124: I don't want to block this PR for this. So we can handle it in a subsequent PR. Can we now safely remove `parameterCodec` from `manager`? or we still require it?

pkg/kubernetes/kubernetes.go
```diff
@@ -37,24 +36,27 @@ type Kubernetes struct {
...
+	discoveryClient         discovery.CachedDiscoveryInterface
+	accessControlClientSet  *AccessControlClientset
+	accessControlRESTMapper *AccessControlRESTMapper
+	dynamicClient           *dynamic.DynamicClient
+
+	staticConfig         *config.StaticConfig
+	CloseWatchKubeConfig CloseWatchKubeConfig
 }
 
+var Scheme = scheme.Scheme
```
ardaguclu, line 51: (not a blocker for this PR)
It is unfortunate that we are tied to `scheme.Scheme` (as cluster is a dynamic environment, we'll miss some new resources that is not defined in our static scheme). But I understand the reason of this. I'd recommend moving this to its own scheme.go file.

pkg/kubernetes/resources.go
manusa: parameterCodec should no longer be in Manager.

pkg/kubernetes/resources.go
manusa: Not sure what you mean with this.
AFAIU, before all resources were supported too, this was/is just a lazy way to pass share the RESTClient.

pkg/kubernetes/kubernetes.go
manusa: Let's merge this in first and reiterate on that.
In this case, I just needed a quick way to access the Scheme and ParameterCodec. I recalled you mentioned about using a global Scheme at some other PR so went on this direction.

pkg/kubernetes/resources.go
ardaguclu: NVM

manusa:
> I really like the way of how this PR handles the problem

I think it's a good start, but I'm still not fully convinced about the solution.
I like the RESTMapper because it ensures the access control is verified for absolutely each request.
However, this is not possible for the ClientSets. The only way I envisioned to do this was by using a RoundTripper, but it added a lot of complexity and didn't cover the problem completely.

For now, I think we need to keep adding Denied Tests for each new feature we implement.

Issue #151, opened 2025-06-28
Adding namespace and tools control arguments


N0YU0:
## Description

Adding an argument (like --namespace) to select a namespace to limit all commands to will be great as a security feature to limit the access to the mcp server to a single namespace even if the kubeconfig has access to multiple (or cluster level access).
Another argument that will help with access control is one to limit tools that can be used with the mcp server (like --exclude-tools or --include-tools). This argument will limit the tools the mcp server can access (like --read-only) but it will enable each user to choose which tools to expose.

My use case is that I want to create an API that will allow users to send questions about their namespaces and will return an answer. The namespace is checked with their permissions but the can ask in the question body to get another namespace's resources. Another thing is I would want to exclude the configuration_view tool as well add the --read-only flag.

ardaguclu:
@N0YU0 I think this new profile proposal https://github.com/manusa/kubernetes-mcp-server/pull/134 would mitigate some of your concerns (i.e. excluding `configuration_view` tool).

After thinking on this, maybe instead of adding a new profile that are mostly a set of static rules per each requirement. Maybe we can provide a `custom` profile that is fully up to the passed configuration file.
* included tools list
* namespace (default all)
* denied resources

@manusa WDYT?

manusa:
We need to think about it.

It might be interesting to deal with all this in the scope of configuration files #131 + #137

If we check the GitHub MCP server, they introduced the concept of [_Dynamic tools_](https://github.com/github/github-mcp-server/issues/189) ([implementation](https://github.com/github/github-mcp-server/pull/188)) which is very similar to some of the features we want to tackle with **profiles** and **configuration file**.

My initial idea would be to create a set of well-defined tool groups for example:
- resources_readonly
- resources_destructive
- pods_readonly
- pods_destructive
- helm_readonly
- ...

The `StaticConfig` can then have a field toolSets which defines the tools available.
In addition, and in scope of #137, every other configuration flag should be settable via the config file too. ---> StaticConfig is the single source of truth for the MCP Server configuration+behavior.

Finally, continuing on #134, we can provide a set of predefined profiles with defaults (full, full-safe) for all values.

Still (see #137), **some** cmd flags are preserved for one-off overrides.

Config inheritance/extension should also be taken into account. i.e. a user wants to have the `full` behavior, but just limit the namespace.

ardaguclu:
[...]

PR #145, opened 2025-06-20
test(config): additional test cases for config errors

manusa:
Relates to #131

PR #152, opened 2025-06-30
Define flags in configuration file

ardaguclu:
This PR adds support of passing flag values in configurations files as described in https://github.com/manusa/kubernetes-mcp-server/issues/137. Flags are always higher priority than the configuration definitions.

This PR fixes https://github.com/manusa/kubernetes-mcp-server/issues/137

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -8,6 +8,15 @@ import (
 
 type StaticConfig struct {
 	DeniedResources []GroupVersionKind `toml:"denied_resources"`
+
+	LogLevel           int    `toml:"log_level,omitempty"`
+	SSEPort            int    `toml:"sse_port,omitempty"`
+	HTTPPort           int    `toml:"http_port,omitempty"`
+	SSEBaseURL         string `toml:"sse_base_url,omitempty"`
+	KubeConfig         string `toml:"kubeconfig,omitempty"`
+	ListOutput         string `toml:"list_output,omitempty"`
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -8,6 +8,15 @@ import (
 
 type StaticConfig struct {
```
ardaguclu, line 9: Should we define `--profile` in here as well?

ardaguclu:
I think, windows-latest failure is not related to this PR?.

manusa:
> I think, windows-latest failure is not related to this PR?.

doesn't seem related.

PTAL #145 and see if we can put it in first

pkg/config/config.go
manusa: In scope of #151, I'm not sure if profile is of a higher order than config.

e.g.
We have `full` profile, `safe` profile, `tekton` profile. Each provides a default configuration that can then be overridden.
Config/StatocConfig has a filed toolsets where you define the tools that are provided

ardaguclu: So we should add `profile` in here or not?

manusa: Sorry, my previous reply was unclear.
I think that **for the moment** we SHOULD NOT add the profile here.

ardaguclu: Sounds good. Thanks for the comment. 

Once I rebase main branch on this PR, it will be ready for review.

ardaguclu:
@manusa could you PTAL?.

PR #155, opened 2025-07-01
Introduce enabled/disabled tool list in configuration file

### LINKED ISSUE #156: Enabled and Disabled tools configuration
## Description

As a user/mcp admin, I want to be able provide a list of allowed or denied tools, so that I can configure which tools are accessible to the AI assistant/agent.

- Configuration (toml) contains the following fields
 - ~`allowed_tools`~ `enabled_tools` (slice/array)
   - If empty, all tools are available
   - If NOT empty, only tools configured are available
 - ~`denied_tools`~ `disabled_tools`
   - If empty, NO OP
   - If NOT empty, tools configured are explicitly removed from the list of available tools

_Originally defined by @ardaguclu in #155_

ardaguclu:
Fixes #156 

This PR introduces enabled/disabled tool list in the configuration file. 

Users will be able to add enabled and disabled tools in the configuration file. e.g.

```toml
enabled_tools = ["configuration_view", "events_list", "namespaces_list", "pods_list", "resources_list", "resources_get", "resources_create_or_update", "resources_delete"]
disabled_tools = ["pods_delete", "pods_top", "pods_log", "pods_run", "pods_exec"]
```

This list will be applied to the given profile (full is the default);
* If enabled list is empty, all the tools given in the profile will be populated
* If enabled list is not empty, only the enabled tools will be populated
* If disabled list is not empty, specifically disabled tools will be removed from the tool list.

This PR partially fixes https://github.com/manusa/kubernetes-mcp-server/issues/151

In a follow up PR, I'll add validations to prevent duplicate entries in both enable and disabled list.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -11,16 +11,18 @@ import (
 type StaticConfig struct {
 	DeniedResources []GroupVersionKind `toml:"denied_resources"`
 
-	LogLevel           int      `toml:"log_level,omitempty"`
-	SSEPort            int      `toml:"sse_port,omitempty"`
-	HTTPPort           int      `toml:"http_port,omitempty"`
-	SSEBaseURL         string   `toml:"sse_base_url,omitempty"`
-	KubeConfig         string   `toml:"kubeconfig,omitempty"`
-	ListOutput         string   `toml:"list_output,omitempty"`
-	ReadOnly           bool     `toml:"read_only,omitempty"`
+	LogLevel   int    `toml:"log_level,omitempty"`
+	SSEPort    int    `toml:"sse_port,omitempty"`
+	HTTPPort   int    `toml:"http_port,omitempty"`
+	SSEBaseURL string `toml:"sse_base_url,omitempty"`
+	KubeConfig string `toml:"kubeconfig,omitempty"`
+	ListOutput string `toml:"list_output,omitempty"`
+	// When true, expose only tools annotated with readOnlyHint=true
+	ReadOnly bool `toml:"read_only,omitempty"`
+	// When true, disable tools annotated with destructiveHint=true
 	DisableDestructive bool     `toml:"disable_destructive,omitempty"`
-	AllowedTools       []string `toml:"allowed_tools,omitempty"`
-	DeniedTools        []string `toml:"denied_tools,omitempty"`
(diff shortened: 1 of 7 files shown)


ardaguclu:
/cc @manusa 
I don't want to cause any merge conflicts with your https://github.com/manusa/kubernetes-mcp-server/pull/149. We can wait to merge this https://github.com/manusa/kubernetes-mcp-server/pull/149 first

manusa:
Currently implementing your review suggestions. Will ping you there when ready for a last review

pkg/config/config.go
```diff
@@ -6,17 +6,21 @@ import (
...
-	ListOutput         string `toml:"list_output,omitempty"`
-	ReadOnly           bool   `toml:"read_only,omitempty"`
-	DisableDestructive bool   `toml:"disable_destructive,omitempty"`
+	LogLevel           int      `toml:"log_level,omitempty"`
+	SSEPort            int      `toml:"sse_port,omitempty"`
+	HTTPPort           int      `toml:"http_port,omitempty"`
+	SSEBaseURL         string   `toml:"sse_base_url,omitempty"`
+	KubeConfig         string   `toml:"kubeconfig,omitempty"`
+	ListOutput         string   `toml:"list_output,omitempty"`
+	ReadOnly           bool     `toml:"read_only,omitempty"`
```
manusa: I think it would be good to move the description comment here since to clarify its purpose:
```
// When true, expose only tools annotated with readOnlyHint=true
```

Same for DisableDestructive

manusa (commented):
LGTM, however, I think I'd rename `allowed_tools` to `enabled_tools` and `denied_tools` to `disabled_tools`.

pkg/config/config.go
ardaguclu: Good point. I'll add these comments.

ardaguclu:
> LGTM, however, I think I'd rename `allowed_tools` to `enabled_tools` and `denied_tools` to `disabled_tools`.

enabled/disabled is more descriptive than allowed/denied. I'll update with that.

pkg/config/config.go
ardaguclu: Updated.

PR #157, opened 2025-07-02
Introduce middleware for audit logs and authentication checks

ardaguclu:
This PR is the continuation of https://github.com/manusa/kubernetes-mcp-server/pull/153. First commit basically is from https://github.com/manusa/kubernetes-mcp-server/pull/153. Second commit introduces the middleware.

This middleware;
* logs requests if the log-level is 5 or higher
* will check the authorization header and return accordingly, if bearer token is not found.

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -98,6 +98,7 @@ func (s *Server) ServeHTTP(httpServer *http.Server) *server.StreamableHTTPServer
 	options := []server.StreamableHTTPOption{
 		server.WithHTTPContextFunc(contextFunc),
 		server.WithStreamableHTTPServer(httpServer),
+		server.WithStateLess(true),
 	}
 	return server.NewStreamableHTTPServer(s.server, options...)
 }
--- a/pkg/middleware/middleware.go
+++ b/pkg/middleware/middleware.go
@@ -1,0 +1,61 @@
+package middleware
+
+import (
+	"bufio"
+	"net"
+	"net/http"
+	"time"
+
+	"k8s.io/klog/v2"
+)
+
+func RequestMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+
+		lrw := &loggingResponseWriter{
+			ResponseWriter: w,
+			statusCode:     http.StatusOK,
+		}
+
+		next.ServeHTTP(lrw, r)
+
+		duration := time.Since(start)
+		klog.V(5).Infof("%s %s %d %v", r.Method, r.URL.Path, lrw.statusCode, duration)
+	})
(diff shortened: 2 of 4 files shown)


pkg/mcp/mcp.go
```diff
@@ -98,6 +98,7 @@ func (s *Server) ServeHTTP(httpServer *http.Server) *server.StreamableHTTPServer
 	options := []server.StreamableHTTPOption{
 		server.WithHTTPContextFunc(contextFunc),
 		server.WithStreamableHTTPServer(httpServer),
+		server.WithStateLess(true),
```
ardaguclu, line 101: This is not related to the changes in this PR. I updated this, because I think we want stateless streamable http server. If it is suggested that this change should be in a different PR, I can remove it from here.

manusa:
> * logs requests if the log-level is 5 or higher

I'm not sure if this is aligned with the `--log-level` config description: _Sets the logging level (values [from 0-9](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md)). Similar to [kubectl logging levels](https://kubernetes.io/docs/reference/kubectl/quick-reference/#kubectl-output-verbosity-and-debugging)._

According to the `kubectl` quick reference table shouldn't the value be 6 instead?

ardaguclu:
I think, it is not related but hard to say for sure.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,61 @@
+package middleware
```
manusa: Not sure if this scope is too narrow, what do you think about a broader `http` package to include everything related to our customized http server?

manusa (commented):
besides the nitty comments, everything else looks good :)

pkg/middleware/middleware.go
ardaguclu: Firstly, I renamed it to `http` to be honest. But this name collides with the native Go http package. But I can rename it to, if you want to. And we can use package name as internalhttp.

ardaguclu: I'm fine whatever you decide.

manusa: Your call, we can always change the name in the future.

For me it's uncomfortable to see the `http` server logic in `root.go` and I think that we should probably move that somewhere else (hence the broader scope for this package, to cover everything related to http/internalhttp).

In any case, this can be covered in the future.

ardaguclu: Renamed to http, because it is indeed better. We can move server logic under http in a followup PR.

ardaguclu: I'll move server logic in http package, when I wire the sig terms to the server.

ardaguclu: Opened this https://github.com/manusa/kubernetes-mcp-server/pull/164

PR #158, opened 2025-07-02
Accept standard oauth header by keeping the current header

ardaguclu:
In order to be OAuth compliant, we'll need to accept bearer tokens from `Authorization` header. This PR passes the token in Authorization header to the context with the same name. 

In order to be backwards compatible, this PR still supports the custom header.

### DIFF
--- a/pkg/kubernetes/impersonate_roundtripper.go
+++ b/pkg/kubernetes/impersonate_roundtripper.go
@@ -8,7 +8,7 @@ type impersonateRoundTripper struct {
 
 func (irt *impersonateRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	// TODO: Solution won't work with discoveryclient which uses context.TODO() instead of the passed-in context
-	if v, ok := req.Context().Value(AuthorizationHeader).(string); ok {
+	if v, ok := req.Context().Value(OAuthAuthorizationHeader).(string); ok {
 		req.Header.Set("Authorization", v)
 	}
 	return irt.delegate.RoundTrip(req)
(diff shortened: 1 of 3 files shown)


pkg/kubernetes/impersonate_roundtripper.go
```diff
@@ -8,7 +8,7 @@ type impersonateRoundTripper struct {
 
 func (irt *impersonateRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	// TODO: Solution won't work with discoveryclient which uses context.TODO() instead of the passed-in context
-	if v, ok := req.Context().Value(AuthorizationHeader).(string); ok {
+	if v, ok := req.Context().Value(OAuthAuthorizationHeader).(string); ok {
```
manusa, line 11: Just a note that this is dead code at the moment.

ardaguclu: I was planning to ask you this. Why don't we remove it for clarity?

manusa: See https://github.com/manusa/kubernetes-mcp-server/pull/96

At the moment, the usage of the RoundTripper approach wasn't completely discarded.
I can't remember now, but back then, the only piece missing the context was the discovery client.
I think you've been involved on some of the issues, so you might better now what are the possible outcomes in the future.

- https://github.com/kubernetes/kubernetes/pull/129109
- https://github.com/kubernetes/kubernetes/issues/110810

Anyway, maybe now, with the more complex auth strategies and enabled/disabled resources this might be completely off.

ardaguclu: Sounds good. We can leave it for now. I don't think, we'll be able to pass context to DiscoveryClient anytime soon.

manusa (approved):
LGTM, thx!

PR #163, opened 2025-07-03
Isolate bearer token config from kubeconfig

ardaguclu:
Copying the template from kubeconfig has security implications, as it also sets the certificates from kubeconfig that can be used as a fallback mechanism, if the bearer token is invalid. 

We have to generate config from scratch by explicitly setting the required fields from kubeconfig, such as ca bundle, host, api path, qps, etc.

This PR generates derived config from scratch by setting the token manually, custom user-agent to be identified in the api server audit logs and required fields from kubeconfig.

Based on my tests, this works properly and expected.

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -28,6 +28,8 @@ import (
 const (
 	CustomAuthorizationHeader = "kubernetes-authorization"
 	OAuthAuthorizationHeader  = "Authorization"
+
+	CustomUserAgent = "kubernetes-mcp-server/bearer-token-auth"
 )
 
 type CloseWatchKubeConfig func() error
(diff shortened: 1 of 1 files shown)


manusa:
A few tests to ensure behavior would be helpful.

manusa (approved):
LGTM, thx!

ardaguclu:
> A few tests to ensure behavior would be helpful.

Good point. I think, I'll have to add some tests for Derived config not only cover this case but also ensuring that the staticConfig is wired (as we discussed previously). Thank you.

manusa:
> > A few tests to ensure behavior would be helpful.
> 
> Good point. I think, I'll have to add some tests for Derived config not only cover this case but also ensuring that the staticConfig is wired (as we discussed previously). Thank you.

:heart: 

I added the comment as a reminder.
It'd be good to have tests that ensure the expected behavior.
- Derived config is deep equal to original config in case derived doesn't apply -> entry-point for further `--require-auth` refinement
- ////////
- Derived config has host matching original config
- Derived config has APIPath matching original config
- Derived config has ... matching original config
- Derived config overrides Bearer token in original config

PR #164, opened 2025-07-03
Add graceful shutdown of http server by catching interruption signals

ardaguclu:
As promised in https://github.com/manusa/kubernetes-mcp-server/pull/157#discussion_r2179952511, this PR moves http serving logic under its own specific directory. 

Additionally, this PR registers to interruption signals to catch and gracefully shutdown the http server by draining in-flight requests before actual shutting down.

### DIFF
--- a/pkg/http/http.go
+++ b/pkg/http/http.go
@@ -1,15 +1,20 @@
 package http
 
 import (
+	"context"
 	"errors"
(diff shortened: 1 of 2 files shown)


ardaguclu:
@manusa would you please have a look at, when you have a chance?. Thanks.

PR #172, opened 2025-07-10
Introduce OIDC token verification if authorization-url is specified

ardaguclu:
This is continuation of https://github.com/manusa/kubernetes-mcp-server/pull/170 that provides foundational work.  Currently many changes in this PR is coming from there. Basically last 3 commits belong to this PR.

Sending TokenReview request to API Server works with an assumption that this token will be used in the same cluster that MCP Server is running on. This is valid for most of cases where MCP Server and users are on the same cluster.

However, multi-tenant environments where token can belong to another API server other than the MCP Server is running on needs different approach. Because host name difference causes the failure of TokenReview request, as the token is issued for another cluster. 

In order to mitigate the issue (not fully fixing, because we still use the MCP Server's host name) this PR proposes if the authorization url is specified in MCP Server, the token passed in Authorization header can/should be verified against the given OIDC Provider instead of sending TokenReview request to API Server.

### DIFF
--- a/pkg/http/authorization_test.go
+++ b/pkg/http/authorization_test.go
@@ -183,7 +183,7 @@ func TestAuthorizationMiddleware(t *testing.T) {
 		handlerCalled = false
 
 		// Create middleware with OAuth disabled
-		middleware := AuthorizationMiddleware(false, "", nil)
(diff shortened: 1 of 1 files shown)


ardaguclu:
This one has became obsolete. I'll open a new one.

PR #175, opened 2025-07-14
test(mcp): refactor tool filtering tests

manusa:
- Prevent declaring tools that are both read-only and destructive
- Remove redundant tests and preserve those behavioral and semantic


pkg/mcp/mcp_tools_test.go
```diff
@@ -0,0 +1,118 @@
...
+			if err != nil {
+				t.Fatalf("call ListTools failed %v", err)
+			}
+		})
+		t.Run("Destructive tools ARE NOT read only", func(t *testing.T) {
+			for _, tool := range tools.Tools {
+				readOnly := ptr.Deref(tool.Annotations.ReadOnlyHint, false)
+				destructive := ptr.Deref(tool.Annotations.DestructiveHint, false)
+				if readOnly && destructive {
+					t.Errorf("Tool %s is read-only and destructive, which is not allowed", tool.Name)
```
ardaguclu, line 25: that will prevent accidental manifestation of a tool as readonly **and** destructive (which is semantically incorrect). Nice!

PR #176, opened 2025-07-14
Introduce OIDC token verification if authorization-url is specified

ardaguclu:
This PR is revamped version of https://github.com/manusa/kubernetes-mcp-server/pull/172.

If Authorization URL is set, that means MCP Server is asked to validate the tokens from the given issuer url.

This PR adds support this. Currently, the expectation is to get one token regardless of the authorization url is specified or not. And this token is used against API Server after validated.

### DIFF
--- a/pkg/http/authorization.go
+++ b/pkg/http/authorization.go
@@ -31,37 +31,41 @@ func AuthorizationMiddleware(requireOAuth bool, serverURL string, mcpServer *mcp
 				return
 			}
 
+			audience := Audience
+			if serverURL != "" {
+				audience = serverURL
+			}
+
 			authHeader := r.Header.Get("Authorization")
 			if authHeader == "" || !strings.HasPrefix(authHeader, "Bearer ") {
 				klog.V(1).Infof("Authentication failed - missing or invalid bearer token: %s %s from %s", r.Method, r.URL.Path, r.RemoteAddr)
 
-				w.Header().Set("WWW-Authenticate", fmt.Sprintf(`Bearer realm="Kubernetes MCP Server", audience=%s, error="invalid_token"`, Audience))
+				if serverURL == "" {
+					w.Header().Set("WWW-Authenticate", fmt.Sprintf(`Bearer realm="Kubernetes MCP Server", audience="%s", error="invalid_token"`, audience))
+				} else {
+					w.Header().Set("WWW-Authenticate", fmt.Sprintf(`Bearer realm="Kubernetes MCP Server", audience="%s"", resource_metadata="%s/.well-known/oauth-protected-resource", error="invalid_token"`, audience, serverURL))
+				}
 				http.Error(w, "Unauthorized: Bearer token required", http.StatusUnauthorized)
 				return
 			}
 
 			token := strings.TrimPrefix(authHeader, "Bearer ")
 
-			audience := Audience
-			if serverURL != "" {
-				audience = serverURL
-			}
-
 			err := validateJWTToken(token, audience)
 			if err != nil {
 				klog.V(1).Infof("Authentication failed - JWT validation error: %s %s from %s, error: %v", r.Method, r.URL.Path, r.RemoteAddr, err)
 
-				w.Header().Set("WWW-Authenticate", fmt.Sprintf(`Bearer realm="Kubernetes MCP Server", audience=%s, error="invalid_token"`, Audience))
+				w.Header().Set("WWW-Authenticate", fmt.Sprintf(`Bearer realm="Kubernetes MCP Server", audience=%s, error="invalid_token"`, audience))
 				http.Error(w, "Unauthorized: Invalid token", http.StatusUnauthorized)
 				return
 			}
 
 			// Validate token using Kubernetes TokenReview API
-			_, _, err = mcpServer.VerifyToken(r.Context(), token, Audience)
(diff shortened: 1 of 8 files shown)


pkg/mcp/mcp.go
```diff
@@ -122,6 +124,13 @@ func (s *Server) GetKubernetesAPIServerHost() string {
 	return s.k.GetAPIServerHost()
 }
 
+func (s *Server) GetOIDCProvider() *oidc.Provider {
```
ardaguclu, line 127: OIDCProvider is exported. Maybe we don't need this function.

pkg/http/authorization.go
```diff
@@ -31,37 +32,83 @@ func AuthorizationMiddleware(requireOAuth bool, serverURL string, mcpServer *mcp
...
+					} else {
+						w.Header().Set("WWW-Authenticate", fmt.Sprintf(`Bearer realm="Kubernetes MCP Server", audience="%s"", resource_metadata="%s%s", error="invalid_token"`, audience, serverURL, oauthProtectedResourceEndpoint))
+					}
+					http.Error(w, "Unauthorized: Invalid token", http.StatusUnauthorized)
+					return
+				}
+			}
+
+			// Scopes are likely to be used for authorization.
+			scopes := claims.GetScopes()
```
ardaguclu, line 88: This is the foundational work for scoped based authorization. This can also be used for logging.

ardaguclu:
Just a summary of what the flow looks like if `--require-oauth` is enabled;
* If authorization url is empty, that means our source of truth is the API Server defined in MCP Server. Every token will be validated against it.
* If authorization url is not empty and one token is passed in Authorization header, that means our source of truth is Authorization URL **and** API Server defined in MCP Server. This token will be validated in Authorization url and API Server.
* If authorization url is not empty and two tokens are passed in different headers, that means the token in the Authorization header will be validated via authorization url and the other will be validated via API Server.

pkg/http/authorization.go
```diff
@@ -72,32 +119,60 @@ func AuthorizationMiddleware(requireOAuth bool, serverURL string, mcpServer *mcp
...
+	if c.Scope == "" {
+		return nil
+	}
+	return strings.Fields(c.Scope)
 }
 
-// validateJWTToken validates basic JWT claims without signature verification
-func validateJWTToken(token, audience string) error {
+func (c *JWTClaims) ContainsAudience(audience string) bool {
+	switch aud := c.Audience.(type) {
```
ardaguclu, line 136: audience is relatively arbitrary fields. So that we have to support different data structures.

ardaguclu:
@manusa this PR is ready for review. Regardless of the ways we decide to support in terms of auth, the changes in this PR will likely be relevant to keep us OAuth compliant.

pkg/mcp/mcp.go
manusa: we can clean up everything in a follow-up PR once usage is clarified

pkg/http/authorization.go
manusa: This will probably need clarification and an enumeration of different scenarios in a specific issue or user story

ardaguclu: Agree. Once we have a clear picture about the role of the scopes, we'll need this ^^

pkg/kubernetes-mcp-server/cmd/root.go
```diff
@@ -235,10 +245,21 @@ func (m *MCPServerOptions) Run() error {
...
 	}
+
+	var oidcProvider *oidc.Provider
+	if m.StaticConfig.AuthorizationURL != "" {
+		provider, err := oidc.NewProvider(context.TODO(), m.StaticConfig.AuthorizationURL)
+		if err != nil {
+			return fmt.Errorf("unable to setup OIDC provider: %w", err)
+		}
+		oidcProvider = provider
+	}
```
manusa, line 256: It seems that oidc.Provider is basically passed along to the mcp.Server instance and in turn is then passed on to the http AuthorizationMiddleware function.

Eventually there's also a call to VerifyToken which in turn performs a call to the Kubernetes VerifyToken function.

There seems to be a lot of coupling here that **might** be possible to extract.

Since the http package is providing auth, IMO mcp should have nothing else to do in this area.

Anyway, this should be done after merging this PR and only if it's possible.

ardaguclu: yes, that is correct that there are a couple of tight couplings. I'm open to improve this. 

Maybe we can directly pass oidcprovider instead of carrying via McpServer. That would be better. VerifyToken needs to access kubernetes which is managed in McpServer, this one might be difficult.

ardaguclu: What about this https://github.com/manusa/kubernetes-mcp-server/pull/179?

ardaguclu: With respect to VerifyToken, I'm open to suggestions.

manusa: > What about this https://github.com/manusa/kubernetes-mcp-server/pull/179?

Not sure yet, but it's a good start.

I'll be creating some more tests for further clarification of the auth flows.
I'll try to provide an alternate proposal (or incremental to #179) once I have a clearer picture.

ardaguclu: I appreciate it. Thank you.

PR #179, opened 2025-07-16
Carry oidc provider directly instead of mcpServer

ardaguclu:
Carrying oidcProvider in McpServer is redundant and unnecessarily causing tight coupling. Instead we can pass oidc provider directly to the authorization middleware.


manusa (approved):
LGTM, thx!

PR #189, opened 2025-07-17
fix(auth): delegate JWT parsing to github.com/golang-jwt/jwt

manusa:
I tried to use the jose library (required by oidc) but it's not as lenient as golang-jwt.

Encapsulated JWT validation logic (client-side) in `(c *JWTClaims) Validate(audience string)`. Initial step to #179 / https://github.com/manusa/kubernetes-mcp-server/pull/176#discussion_r2210288722

Tests verify all local JWT parsing and validations. More tests coming in follow-up PR.

/cc @ardaguclu


go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/BurntSushi/toml v1.5.0
 	github.com/coreos/go-oidc/v3 v3.14.1
 	github.com/fsnotify/fsnotify v1.9.0
+	github.com/golang-jwt/jwt/v4 v4.5.2
```
ardaguclu: I agree that this simplifies the code. On the other hand, we simply require some field based validations after extracting the JWT token. We don't need sophisticated security based processes which is supposed to be handled by OIDC or API Server. 

Adding a new dependency  just for this simple sanity check brings about some downsides. For example, https://github.com/golang-jwt/jwt/security mentions about the security posture. https://github.com/golang-jwt/jwt/security/advisories/GHSA-mh63-6h87-95cp looks like an high severity. This was fixed. But who knows what else.

In my opinion, we don't need to rely on external dependency for something that we can handle. Maybe we can move JWT based functions into its dedicated directory. WDYT?

ardaguclu: Since we have already an indirect dependency to go-jose due to https://github.com/coreos/go-oidc/blob/a7c457eacb849c163a496b29274242474a8f44ab/go.mod#L8C2-L8C31, maybe it makes sense to use it. But I see that you have tried this as first choice.

manusa: > Adding a new dependency just for this simple sanity check brings about some downsides. For example, https://github.com/golang-jwt/jwt/security mentions about the security posture. [GHSA-mh63-6h87-95cp](https://github.com/golang-jwt/jwt/security/advisories/GHSA-mh63-6h87-95cp) looks like an high severity. This was fixed. But who knows what else.

Oh crap, I added the wrong version :sweat: 

> In my opinion, we don't need to rely on external dependency for something that we can handle.

My original plan was to reuse the jose library (brought in by oidc). I'm not happy bringing in dependencies either.

So the additional benefit is that signatures are checked plus we don't need to rely on our own parsing which might be missing something else (e.g. you added padding for base64, but it's also difficult to track what else might be missing).


Let me give it another spin

manusa: Parsing and validating is now delegated to go-jose.
This should cover both concerns:
- The dependency is reused from go-oidc, so we're not bringing in a new one
- Whatever downsides we might face would already be brought in by relying on go-oidc and its side-effects

ardaguclu: This sounds reasonable. I'll review in depth tomorrow.

pkg/http/authorization.go
```diff
@@ -55,7 +54,10 @@ func AuthorizationMiddleware(requireOAuth bool, serverURL string, mcpServer *mcp
 			// Validate the token offline for simple sanity check
 			// Because missing expected audience and expired tokens must be
 			// rejected already.
-			claims, err := validateJWTToken(token, audience)
+			claims, err := ParseJWTClaims(token)
```
ardaguclu, line 57: What about we only have one function that manages parsing, validating, extracting. This function returns scopes and error. Like this;

```go
scopes, err := validateJWTAndGetScopes(token, audience, "") // 3rd parameter is authorization url to check the issuer, if it is set
```

So that we don't need this;

```go
			if err == nil && claims != nil {
				err = claims.Validate(audience)
			}
```

pkg/http/authorization.go
```diff
@@ -132,66 +148,20 @@ func (c *JWTClaims) GetScopes() []string {
...
-		return nil, fmt.Errorf("token expired")
-	}
-
-	if !claims.ContainsAudience(audience) {
-		return nil, fmt.Errorf("token audience mismatch: %v", claims.Audience)
-	}
-
-	return claims, nil
+// Validate Checks if the JWT claims are valid and if the audience matches the expected one.
+func (c *JWTClaims) Validate(audience string) error {
```
ardaguclu, line 152:
```suggestion
func validateJWTAndGetScopes(token string, audience string, issuerURL string) ([]string, error) {
	tkn, err := jwt.ParseSigned(token, allSignatureAlgorithms)
	if err != nil {
		return nil, fmt.Errorf("failed to parse JWT token: %w", err)
	}

	claims := &JWTClaims{}
	err = tkn.UnsafeClaimsWithoutVerification(claims)
	if err != nil {
		return nil, fmt.Errorf("failed to extract claims: %w", err)
	}

	expected := jwt.Expected{
		AnyAudience: jwt.Audience{audience},
		Time:        time.Now(), // internally it is set to time.Now(), maybe explicitly setting is better
	}
	if issuerURL != "" { // maybe not the scope of this PR, you can add it or skip it
		expected.Issuer = issuerURL
	}

	err = claims.Claims.Validate(expected)
	if err != nil {
		return nil, fmt.Errorf("JWT validation failed: %w", err)
	}

	var scopes []string
	if claims.Scope != "" {
		scopes = strings.Fields(claims.Scope)
	}

	return scopes, nil
}
```

pkg/http/authorization.go
```diff
@@ -132,66 +148,20 @@ func (c *JWTClaims) GetScopes() []string {
...
 }
 
-func parseJWTClaims(payload string) (*JWTClaims, error) {
-	// Add padding if needed
-	if len(payload)%4 != 0 {
-		payload += strings.Repeat("=", 4-len(payload)%4)
-	}
-
-	decoded, err := base64.URLEncoding.DecodeString(payload)
+func ParseJWTClaims(token string) (*JWTClaims, error) {
```
ardaguclu, line 158: We won't need this, as validateJWTAndGetScopes handles it.

pkg/http/authorization_test.go
```diff
@@ -1,187 +1,222 @@
 package http
```
ardaguclu, line 1: Unit tests will also be simplified, because they'll have to exercise only the `validateJWTAndGetScopes` function.

pkg/http/authorization.go
```diff
@@ -118,11 +120,25 @@ func AuthorizationMiddleware(requireOAuth bool, serverURL string, mcpServer *mcp
 	}
 }
 
+var allSignatureAlgorithms = []jose.SignatureAlgorithm{
+	jose.EdDSA,
+	jose.HS256,
```
ardaguclu, line 125: It looks like go-oidc does not easily support verifying symmetric tokens (the ones contain `HS` prefixes). But it is better to add this in here, in case in the future we want to support this.

pkg/http/authorization.go
manusa: I think that would be the next step.
We need a function or an entity that orchestrates all of the validation

pkg/http/authorization.go
manusa: I'd prefer to keep JWTToken as a single entity for further refactoring

pkg/http/authorization.go
manusa: I'd prefer to keep it separate for now

pkg/http/authorization.go
manusa: ~OK, I copied the list from them (IIRC)~

manusa: I copied the list from https://github.com/go-jose/go-jose/blob/3a80e136a96e747bf44049414eadc02828df4d33/jose-util/crypto.go#L48-L62

Since the original implementation was completely lenient on the signature I assumed we wanted to allow anything, hence reusing a list.

manusa:
> Overall looks good to me. Dropped a few comments.
> 
> This is totally your preference but maybe #179 PR can be added in this PR as well?.

I'd prefer to do the refactor in very small steps.
There are also areas of the auth "module" that aren't tested that I'd like to test before changing more things.

This is should be considered as an initial step to simplify the JWTClaims struct.
This entity can then evolve to orchestrate all validation or be merged into another one.

e.g. it could be renamed to something else and then be the OIDC provider holder or even the kubernetes holder to perform the verifyToken operation (now in kubernetes)

pkg/http/authorization.go
ardaguclu: I should have been clearer. I think, this list is correct and nice. The problem is in oidcProvider side which does not support symmetric algorithms such as `HS256`. So I think, I need to change oidcProvider side and we should keep this list as is.

PR #193, opened 2025-07-18
fix(lint): add golangci-lint make target + lint

manusa:
/cc @ardaguclu

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -26,9 +26,11 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
 )
 
+type HeaderKey string
+
 const (
-	CustomAuthorizationHeader = "kubernetes-authorization"
-	OAuthAuthorizationHeader  = "Authorization"
+	CustomAuthorizationHeader = HeaderKey("kubernetes-authorization")
+	OAuthAuthorizationHeader  = HeaderKey("Authorization")
 
 	CustomUserAgent = "kubernetes-mcp-server/bearer-token-auth"
 )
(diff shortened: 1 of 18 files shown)


pkg/kubernetes/kubernetes.go
```diff
@@ -26,9 +26,11 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
 )
 
+type HeaderKey string
+
 const (
-	CustomAuthorizationHeader = "kubernetes-authorization"
-	OAuthAuthorizationHeader  = "Authorization"
+	CustomAuthorizationHeader = HeaderKey("kubernetes-authorization")
```
ardaguclu, line 32: Not sure why linter forces us to do this

ardaguclu:
Maybe it is better to define set of linters like https://github.com/openshift/lws-operator/commit/1f2f698c9538c47d8bdb420ee42d9cc5227aa87c#diff-9917ddc9f1c3304218f7269265b746d997c5c0615478177b5fceecd33ef47cb5. Not a blocker for this PR though

pkg/kubernetes/kubernetes.go
manusa: The problem is not at the header level, but at the usage within context.WithValue:

https://github.com/manusa/kubernetes-mcp-server/blob/b4d73d54c5fe711da26d5f53a0b2280505dcc182/pkg/mcp/mcp.go#L166

This was the first thing I thought of to solve it.

ardaguclu: Makes sense

manusa:
> Maybe it is better to define set of linters like [openshift/lws-operator@1f2f698#diff-9917ddc9f1c3304218f7269265b746d997c5c0615478177b5fceecd33ef47cb5](https://github.com/openshift/lws-operator/commit/1f2f698c9538c47d8bdb420ee42d9cc5227aa87c#diff-9917ddc9f1c3304218f7269265b746d997c5c0615478177b5fceecd33ef47cb5). Not a blocker for this PR though

Whatever you think is more suitable, I added the linter because you suggested it.

Once this is in, we should also enable the linter on the CI, we an define the set in the follow-up too.

ardaguclu:
> > Maybe it is better to define set of linters like [openshift/lws-operator@1f2f698#diff-9917ddc9f1c3304218f7269265b746d997c5c0615478177b5fceecd33ef47cb5](https://github.com/openshift/lws-operator/commit/1f2f698c9538c47d8bdb420ee42d9cc5227aa87c#diff-9917ddc9f1c3304218f7269265b746d997c5c0615478177b5fceecd33ef47cb5). Not a blocker for this PR though
> 
> Whatever you think is more suitable, I added the linter because you suggested it.
> 
> Once this is in, we should also enable the linter on the CI, we an define the set in the follow-up too.

Thank you. That looks good to me.

PR #194, opened 2025-07-18
feat(mcp): log tool call (function name + arguments)

### LINKED ISSUE #174: [MCP]: log tool calls
_TO BE DESCRIBED_

manusa:
Fixes #174

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -3,6 +3,7 @@ package mcp
 import (
 	"context"
 	"fmt"
+	"k8s.io/klog/v2"
 	"net/http"
 	"slices"
 
@@ -56,6 +57,7 @@ func NewServer(configuration Configuration) (*Server, error) {
 			server.WithPromptCapabilities(true),
 			server.WithToolCapabilities(true),
 			server.WithLogging(),
+			server.WithToolHandlerMiddleware(toolCallLoggingMiddleware),
 		),
 	}
 	if err := s.reloadKubernetesClient(); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/mcp/mcp.go
```diff
@@ -165,3 +167,10 @@ func contextFunc(ctx context.Context, r *http.Request) context.Context {
 
 	return ctx
 }
+
+func toolCallLoggingMiddleware(next server.ToolHandlerFunc) server.ToolHandlerFunc {
+	return func(ctx context.Context, ctr mcp.CallToolRequest) (*mcp.CallToolResult, error) {
```
ardaguclu, line 172: Based on this commit https://github.com/mark3labs/mcp-go/commit/baa715326faba8f95aa371486d72f7665c152892, `CallToolRequest` contains headers. Would you think it would be useful logging some of the values in here like user-agent, etc.?. Would it be useful for us or unnecessary noise?

manusa: Maybe with more verbosity, similar to [kube-ctl](https://kubernetes.io/docs/reference/kubectl/quick-reference/#kubectl-output-verbosity-and-debugging).
1. log tool call with 5
2. log request headers with 7

But I'm not even sure if this would be that useful for users

ardaguclu: Actually this middleware https://github.com/manusa/kubernetes-mcp-server/blob/775fa21bd1345eee52fea6089c119821c0d95c9e/pkg/http/middleware.go#L12 logs all the requests (if verbosity is set to 5 or higher). Maybe we can log the headers in there (only the ones that are not carrying sensitive data).

In which scenario this middleware will be useful?. Request is seen in the audit logs and verifying that it actually results in tool call?

manusa: > Actually this middleware...
This works for HTTP but not for STDIO (for example).

IMO is good to have both and also keep them separate.
An HTTP call might eventually become a tool call or not, especially now that we're working on auth (e.g. we get an HTTP request but auth blocks it).

> In which scenario this middleware will be useful?. 

The user who made the request was building an MCP proxy and wasn't able to see any effects on the MCP server. They wanted to know if the tool was actually being called.
There are multiple places where the calls can be logged (MCP Client, MCP Host, etc.), I guess that this is yet another tracing point.
I understand it can be interesting to trace where the tool call chain was broken.

ardaguclu: Good points and make sense to me.

manusa: I'm checking, but the Header field hasn't been added yet in v0.0.34:
https://github.com/mark3labs/mcp-go/compare/v0.34.0...main

We can add the finer trace once v0.35.0 is out

PR #199, opened 2025-07-23
chore(doc): update README.md Configuration options

manusa:
Added missing entry for --port and removed deprecated options (should not be documented on the README)


ardaguclu:
/lgtm

This is a reminder for me that I shouldn't forget to update here, when we mark the oauth related flags as visible.

PR #217, opened 2025-07-28
PoC: Agentic scopes

ardaguclu:
This PR is used to fall back to in-cluster config of MCP Server to by pass the token which is issued for authentication/authorization not k8s usage.

PR #224, opened 2025-07-31
Introduce scoped based authorization

ardaguclu:
This PR is derivation of this https://github.com/containers/kubernetes-mcp-server/pull/217. However, https://github.com/containers/kubernetes-mcp-server/pull/217 contains some hacky code to make agentic flow working such as intentional fall back to in-cluster config, etc.

This PR introduces scoped based authorization for tool calls. If it detects the tool name is not included in the token scopes,  mcp server fails with 403 forbidden error. 

Currently this PR uses tool names as scoped names. But in the future, we may want to create a mapping to customize scope names.

### DIFF
--- a/pkg/kubernetes-mcp-server/cmd/root.go
+++ b/pkg/kubernetes-mcp-server/cmd/root.go
@@ -227,18 +227,6 @@ func (m *MCPServerOptions) Validate() error {
 			klog.Warningf("authorization-url is using http://, this is not recommended production use")
 		}
 	}
-	if m.StaticConfig.ServerURL != "" {
-		u, err := url.Parse(m.StaticConfig.ServerURL)
-		if err != nil {
-			return err
-		}
-		if u.Scheme != "https" && u.Scheme != "http" {
-			return fmt.Errorf("--server-url must be a valid URL")
-		}
-		if u.Scheme == "http" {
-			klog.Warningf("server-url is using http://, this is not recommended production use")
-		}
-	}
 	if m.StaticConfig.JwksURL != "" {
 		u, err := url.Parse(m.StaticConfig.JwksURL)
 		if err != nil {
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -19,6 +19,8 @@ import (
 	"github.com/containers/kubernetes-mcp-server/pkg/version"
 )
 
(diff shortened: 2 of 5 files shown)


pkg/kubernetes-mcp-server/cmd/root.go
```diff
@@ -227,18 +227,6 @@ func (m *MCPServerOptions) Validate() error {
 			klog.Warningf("authorization-url is using http://, this is not recommended production use")
 		}
 	}
-	if m.StaticConfig.ServerURL != "" {
```
ardaguclu, line 230: Currently, server url field is used for arbitrary audience rather than a structured URL format. I'm removing this validation (and its test). But if we decide to force URL format in the future, we'll need to revert this change (and its test).

pkg/mcp/mcp.go
```diff
@@ -181,3 +197,16 @@ func toolCallLoggingMiddleware(next server.ToolHandlerFunc) server.ToolHandlerFu
 		return next(ctx, ctr)
 	}
 }
+
+func toolScopedAuthorizationMiddleware(next server.ToolHandlerFunc) server.ToolHandlerFunc {
+	return func(ctx context.Context, ctr mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+		scopes, ok := ctx.Value(TokenScopesContextKey).([]string)
```
ardaguclu, line 203: @manusa although we agreed that we'll add this scoped based validation in authorization.go rather than here, mcp clients all failed to handle the flow. I think, apart from here all the rest is a clear indication of model context protocol violation.

After adding the scope based check in tool call middleware, it simply works.

ardaguclu:
I have no idea what are these verification errors in here https://github.com/containers/kubernetes-mcp-server/pull/224/checks?check_run_id=47091334553. So I'll have to ignore them.

@manusa would you please have a look at this PR, when you have a moment?. Thanks.

manusa:
> I have no idea what are these verification errors in here https://github.com/containers/kubernetes-mcp-server/pull/224/checks?check_run_id=47091334553. So I'll have to ignore them.

No idea either. We need to migrate to our own pipelines.

> @manusa would you please have a look at this PR, when you have a moment?. Thanks.

:+1:

manusa (approved):
LGTM, thx!

PR #244, opened 2025-08-06
feat(auth): implement proxied /.well-known/oauth-authorization-server

manusa:


pkg/http/http.go
```diff
@@ -44,6 +44,7 @@ func Serve(ctx context.Context, mcpServer *mcp.Server, staticConfig *config.Stat
 	mux.HandleFunc(healthEndpoint, func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
 	})
+	mux.HandleFunc(oauthAuthorizationServerEndpoint, OAuthAuthorizationServerHandler(staticConfig))
```
ardaguclu, line 47: @manusa this endpoint is served by oidc providers not mcp servers. Is there any reason for this?. If you implemented based on my suggestion on Slack, sorry it seems that I wasn't clear.

ardaguclu: I think, you implemented this because mcp/inspector looks for this endpoint which is served by Keycloak in a different endpoint?.

manusa: Exactly, this is looked for by the MCP clients, I'm currently working on an implementation to proxy the following endpoints:
```
	oauthAuthorizationServerEndpoint = "/.well-known/oauth-authorization-server"
	oauthProtectedResourceEndpoint   = "/.well-known/oauth-protected-resource"
	openIDConfigurationEndpoint      = "/.well-known/openid-configuration"
```
This delegates everything to the auth server (Keycloak).
This seems to work with the inspector and very likely with VSCode.

I'll create a PR shortly

PR #246, opened 2025-08-07
feat(auth): .well-known endpoints delegated to auth server

manusa:
/cc @ardaguclu


pkg/http/authorization.go
```diff
@@ -32,9 +33,6 @@ func AuthorizationMiddleware(requireOAuth bool, serverURL string, oidcProvider *
 			}
 
 			audience := Audience
-			if serverURL != "" {
```
ardaguclu, line 35: Removal of this makes sense for demo purposes. But in the future, we need to have server url specific audience, otherwise, any user can generate a token for mcp-server audience and sends here.

manusa: Was thinking of providing the audience via config flag (considering our internal document for the Keycloak integration)

ardaguclu: Getting audience via flag is a good idea.

PR #249, opened 2025-08-07
Feature/kube context name

igoichuk:
Adds kube context name cli parameter, if not specifies uses current context (as before).


Cali0707:
@igoichuk do the changes in #348 accomplish what you needed? I know it's not 1:1 the same, as that feature enables context name to be configured at each tool call, rather than globally when starting the server. But, maybe this solves the problem you were facing?

igoichuk:
Thank you. Will give a try. I needed multi-cluster support. Context name was a simplified way to achieve that.

matzew:
@igoichuk I think we can close this as we had merged #348 since.

If there is anything missing - feel free to raise a GH issue.

Thanks for the contribution!

PR #270, opened 2025-08-15
Initial stab at adding OLMv1 functionality

bentito:
Adds capabilities for OLMv1 (assumes OLMv1 APIs installed on cluster):
For a `ClusterExtension` (OLMv1 speak for an operator, or controller-like, or other resources) do the following:
- install
- list
- uninstall
- upgrade


bentito:
I think the ubuntu test fail is not due to this PR's changes? (I think that mostly because the failing test ran fine on other platforms)

Issue #273, opened 2025-08-19
Rigid binding to KUBECONFIG env variable causes unit tests failures in some environments


ardaguclu:
As seen in here https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_openshift-mcp-server/21/pull-ci-openshift-openshift-mcp-server-main-test/1957772596984418304 (PR: https://github.com/openshift/openshift-mcp-server/pull/21), CI pipeline in OpenShift automatically sets its KUBECONFIG environment variable. Relying on KUBECONFIG in unit tests cause unexpected failures. 

we need to pass fake kubeconfig via StaticConfig instead of modifying KUBECONFIG (i.e. `os.Setenv("KUBECONFIG", kubeConfig)`) to have a reliable unit tests that work on various environments.

ardaguclu:
I'm trying to fix it.

ardaguclu:
This WIP PR https://github.com/openshift/openshift-mcp-server/pull/21 tries to fix the KUBECONF env var dependencies by moving to kubeConfig arguments passed by. Seems to be passing. 

However, tests that is about empty namespaces (trying to exercise the default namespace) are still failing. 

As a result, `tests` currently broken in openshift-mcp-server repository.

ardaguclu:
Tests passed https://github.com/openshift/openshift-mcp-server/pull/21. Resolution is to skip all these tests. I'll open a PR in downstream to filter out these tests only for OpenShift CI.

PR #280, opened 2025-08-25
Removal of rigid binding to kubeconfig env var and skip some tests in OCP

ardaguclu:
Relying on KUBECONFIG env variable causes issues in cases where env var is statically set by the environments. So that default namespace is modified to something different than default with less permissions. That causes failures of the tests.

Additionally, all namespaces tests are broken in CI environments that the given temporary namespace can't access to query the all namespaces in the cluster.

This PR skips these tests.

As proven in here https://github.com/openshift/openshift-mcp-server/pull/24, after the changes in here, `test` passes.


ardaguclu:
/cc @manusa 
for when you have a chance to look at.

ardaguclu:
Fixes https://github.com/containers/kubernetes-mcp-server/issues/273

ardaguclu:
@manusa has found a better solution. Closing this one

PR #305, opened 2025-09-08
add mcp endpoint related parameters

hevervie:
Add mcp endpoint related parameters, to support customize sse and streamable http access path:
- **HealthEndpoint**: customize health check endpoint
- **SSEEndpoint**: customize sse access endpoint
- **SSEMessageEndpoint**: customize sse message access endpoint
- **StreamableHttpEndpoint**: customize streamable http access endpoint

### DIFF
--- a/pkg/kubernetes-mcp-server/cmd/root.go
+++ b/pkg/kubernetes-mcp-server/cmd/root.go
@@ -70,8 +70,8 @@ type MCPServerOptions struct {
 
 	HealthEndpoint         string
 	StreamableHttpEndpoint string
-	SseEndpoint            string
-	SseMessageEndpoint     string
+	SSEEndpoint            string
+	SSEMessageEndpoint     string
 
 	ConfigPath   string
 	StaticConfig *config.StaticConfig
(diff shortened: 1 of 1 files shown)


pkg/config/config.go
```diff
@@ -48,6 +48,14 @@ type StaticConfig struct {
 	StsScopes            []string `toml:"sts_scopes,omitempty"`
 	CertificateAuthority string   `toml:"certificate_authority,omitempty"`
 	ServerURL            string   `toml:"server_url,omitempty"`
+	// HealthEndpoint is the health check endpoint
+	HealthEndpoint string `toml:"health_endpoint,omitempty"`
```
ardaguclu, line 52: Every new flag complicates the maintenance and user experience. Why do we need to modify all these endpoints?. Why do we need to add a functionality to modify SSE endpoints, as it is already a deprecated functionality?.

matzew: I'd agree on these points. Let's not do that

pkg/config/config_test.go
```diff
@@ -65,6 +65,11 @@ denied_resources = [
 
 enabled_tools = ["configuration_view", "events_list", "namespaces_list", "pods_list", "resources_list", "resources_get", "resources_create_or_update", "resources_delete"]
 disabled_tools = ["pods_delete", "pods_top", "pods_log", "pods_run", "pods_exec"]
+
+health_endpoint = "/k8s/healthz"
+streamable_http_endpoint = "/k8s/mcp"
```
matzew, line 70: what's wrong with the standard default `/mcp` ? Well established, and known

manusa:
Closing since there's a lack of feedback from the original author and it doesn't seem to provide anything that's not already available.

PR #319, opened 2025-09-11
refactor(mcp): toolset Tools definition is agnostic of MCP impl

manusa:
Initial PR to make the toolsets agnostic of the usd MCP implementation (migration to go-sdk). 

The decoupling will also be needed to move the different toolsets to separate nested packages (toolsets).

/cc @mrunalp @ardaguclu


ardaguclu:
LGTM

PR #323, opened 2025-09-12
feat(toolsets): add support for multiple toolsets in configuration

### LINKED ISSUE #115: README.md: Automatically generate tool/function descriptions
## Description

To simplify the maintenance and documentation of the server, the list of available tools, descriptions, and parameters should be automatically generated.

### LINKED ISSUE #227: Toolsets
## Description

As a Kubernetes MCP server **user** or **administrator**, I want to be able to **enable or disable** specific groups of functionalities (tools, resources, prompts, and so on) by providing a configuration option or a command-line flag, so that the server only exposes these functionalities to the MCP clients.

As a Kubernetes API or extension **maintainer**, I want to be able to **define** a group of functionalities (tools, resources, prompts, and so on) by **implementing** them in the Kubernetes MCP server, so that the users of my extension are able to leverage this MCP server to interact with it.

### For Kubernetes MCP server users or administrators

A new command-line flag `--toolsets` is available.
The flag accepts a comma-separated list of toolsets to be enabled.

A TOML config field is available to be able to define toolsets in the `config.toml` file (`--config config.toml`):

```toml config.toml
toolsets = ["core", "config"]
```

By default, all toolsets are enabled (this might change in the future).

The list of available toolsets and their description is available in the [README.md](https://github.com/containers/kubernetes-mcp-server/blob/main/README.md).

## For Kubernetes API extension maintainers

Maintainers who seek adding support for their extension in this MCP server can do so by providing a toolset for their extension.

New toolsets are defined under the `pkg/toolsets` package in a dedicated, nested package.

The package should provide an implementation of `api.Toolset` and be registered using the `toolsets.Register` function in the `init()` function. 

Finally, an import to the new toolset package should be added to the `pkg/mcp/modules.go` to enable it.

manusa:
Fixes #227
Fixes #115

Users can now enable or disable different toolsets either by providing a command-line flag or by setting the toolsets array field in the TOML configuration.

Downstream Kubernetes API developers can declare toolsets for their APIs by creating a new nested package in pkg/toolsets and registering it in pkg/mcp/modules.go

/cc @mrunalp @ardaguclu


pkg/config/config.go
```diff
@@ -50,22 +51,32 @@ type StaticConfig struct {
 	ServerURL            string   `toml:"server_url,omitempty"`
 }
 
+func Default() *StaticConfig {
+	return &StaticConfig{
+		ListOutput: "table",
+		Toolsets:   []string{"core", "config", "helm"},
```
mrunalp, line 57: I would leave helm out of the default set.

manusa: In <= v0.0.50 it's included by default.
I would do that after the first release with the toolsets feature so that we don't make it a breaking change.
We can also make it more palatable for upstream users if we remove it from the default set of toolsets but in exchange we add a few more features (e.g. `helm toolset is no longer activated by default since it now includes multiple tooling specific to helm, you can enable it by starting the kubernetes MCP server with the following flag --toolsets core,config,helm`).
IMO, the plan should be release v0.0.51 with toolsets without the breaking changes, then, release v0.0.52 with deactivated and specialized toolsets such as helm.

Similarly, for the config one, we could enable it only in stdio mode which is the only deployment+runtime mode where it makes sense.

ardaguclu:
This looks good to me.

PR #331, opened 2025-09-17
fix(cmd): disable klog in STDIO mode

manusa:
Prevents protocol issues with some clients.

Kubernetes tooling uses klog to log messages.
Some of these messages end up logged in the stderr or stdout which breaks some of the clients that expect jsonrpc messages.

/cc @ardaguclu

### DIFF
--- a/pkg/kubernetes-mcp-server/cmd/root.go
+++ b/pkg/kubernetes-mcp-server/cmd/root.go
@@ -207,6 +207,12 @@ func (m *MCPServerOptions) loadFlags(cmd *cobra.Command) {
 func (m *MCPServerOptions) initializeLogging() {
 	flagSet := flag.NewFlagSet("klog", flag.ContinueOnError)
 	klog.InitFlags(flagSet)
+	if m.StaticConfig.Port == "" {
+		// disable klog output for stdio mode
+		// this is needed to avoid klog writing to stderr and breaking the protocol
+		_ = flagSet.Parse([]string{"-logtostderr=false", "-alsologtostderr=false", "-stderrthreshold=FATAL"})
+		return
+	}
 	loggerOptions := []textlogger.ConfigOption{textlogger.Output(m.Out)}
 	if m.StaticConfig.LogLevel >= 0 {
 		loggerOptions = append(loggerOptions, textlogger.Verbosity(m.StaticConfig.LogLevel))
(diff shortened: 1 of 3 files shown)


pkg/kubernetes-mcp-server/cmd/root.go
```diff
@@ -207,6 +207,12 @@ func (m *MCPServerOptions) loadFlags(cmd *cobra.Command) {
 func (m *MCPServerOptions) initializeLogging() {
 	flagSet := flag.NewFlagSet("klog", flag.ContinueOnError)
 	klog.InitFlags(flagSet)
+	if m.StaticConfig.Port == "" {
```
ardaguclu, line 210: Why don't we simply by pass `klog.InitFlags` if `m.StaticConfig.Port` is empty?

manusa: because we need to set the other flags (logtostderr, alsologtostderr, stderrthreshold)  which have defaults that enable logging of some messages

manusa: the fix is not for our logs (which there aren't many at the moment) but rather for the upstream kubernetes logs which aren't controlled easily.

ardaguclu: Does that mean our logs do not break the jsonrpc?

manusa: Yes, they would break the jsonrpc too.
This statement also disables them (as asserted in the provided tests).

ardaguclu: Just for clarification; by running `klog.InitFlags` in SDTIO, which logs that do not break jsonrpc will be logged?

manusa: Ok, let me try to summarize.

If you run kubernetes-mcp-server in STDIO mode (i.e. no `--port` flag) with an unauthorized cluster you can immediately trigger an error that will be logged to STDERR:
```
E0917 16:47:01.678632   85672 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: ...
```

I understand that there might be other scenarios where this might happen.

I've started to see that some MCP Hosts(->Clients) start failing because they've become ore strict and will mark the MCP server as not working.

The goal here is to disable **all** logging either to stderr or stdout in order not to break the protocol for these more strict clients.

By running klog.InitFlags **and then providing the three log configuration flags** we prevent klog from logging to stderr which is what happens by default.

Not sure if it's clearer now.

ardaguclu: Thank you for explanation. It makes sense. 

LGTM

PR #339, opened 2025-09-24
fix(mcp): add missing properties field to namespace tools

bluek1te:
Properties field is required when defining Type: "object" in input schema. Because this is missing, some clients fail to import tools on schema validation.


manusa:
@bluek1te could you please confirm if the issue has been fixed in [v0.0.52](https://github.com/containers/kubernetes-mcp-server/releases/tag/v0.0.52) that was released a few minutes ago?

PR #345, opened 2025-09-26
chore(security): try to fix snyk security issues

manusa:
Local snyk execution is still reporting issues 😞, but these should be false positives.

I can't find a way to make snyk believe this is safe 😅. 
If the issue is still reported downstream we'll have to [suppress the issue](https://docs.snyk.io/manage-risk/prioritize-issues-for-fixing/ignore-issues).

/cc @ardaguclu

### DIFF
--- a/internal/tools/update-readme/main.go
+++ b/internal/tools/update-readme/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"maps"
 	"os"
+	"path/filepath"
 	"slices"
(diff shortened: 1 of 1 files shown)


ardaguclu:
If this is false positive, we can exclude that part in here https://github.com/openshift/openshift-mcp-server/blob/main/.snyk with a good justification. We don't need to fix everything. Thanks for looking this :)

PR #325, opened 2025-09-16
Support for multiple contexts in kubeconfig

alexeykazakov:
Fixes https://github.com/containers/kubernetes-mcp-server/issues/83

1. New `contexts_list` tool for listing all available contexts in the cubeconfig.
2. All tools now have an optional `contex` param. If not specified then the default context is used.

Here is an example of `context_list` result:
```
Available Kubernetes contexts (3 total, current: cluster-1):

Format: [*] CONTEXT_NAME -> CLUSTER_SERVER_URL
        (* indicates the current active context)

Contexts:
─────────
* cluter-1 -> https://api.cluster-1.com:443
  cluter-2 -> https://api.cluster-2.com:443
  cluter-3 -> https://api.cluster-2.com:443
─────────

Usage:
To use a specific context with any tool, add the 'context' parameter:
Example: {"name": "pods_list", "arguments": {"context": "cluster-1"}}`
```

PR #348, opened 2025-09-26
feat: Multi Cluster Support

### LINKED ISSUE #83: Support for multiple contexts and switching between them

Cali0707:
Fixes #83 
Supersedes closes #325

This PR builds on the approach taken in #325, but makes it more generic so that we can configure how clients select the cluster based on the cluster provider strategy (currently supporting in-cluster and kubeconfig).

There is a `contexts_list` tool added as well, which is conditionally enabled when there are more than 15 contexts in the kubeconfig, as that is when we switch from using an enum parameter to a general string parameter.

### DIFF
--- a/pkg/kubernetes/configuration.go
+++ b/pkg/kubernetes/configuration.go
@@ -98,17 +98,22 @@ func (k *Kubernetes) ConfigurationContextsDefault() (string, error) {
 
 // ConfigurationContextsList returns the list of available context names
 // TODO: Should be moved to the Provider level ?
-func (k *Kubernetes) ConfigurationContextsList() ([]string, error) {
+func (k *Kubernetes) ConfigurationContextsList() (map[string]string, error) {
 	if k.manager.IsInCluster() {
-		return []string{inClusterKubeConfigDefaultContext}, nil
+		return map[string]string{inClusterKubeConfigDefaultContext: ""}, nil
 	}
 	cfg, err := k.manager.clientCmdConfig.RawConfig()
 	if err != nil {
 		return nil, err
 	}
-	contexts := make([]string, 0, len(cfg.Contexts))
-	for context := range cfg.Contexts {
-		contexts = append(contexts, context)
+	contexts := make(map[string]string, len(cfg.Contexts))
+	for name, context := range cfg.Contexts {
+		cluster, ok := cfg.Clusters[context.Cluster]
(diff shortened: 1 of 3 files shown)


Cali0707:
So far, this PR only adds the ClusterProvider interface and the implementation for a kubeconfig provider.

The idea is that the provider can be used to get a Manager instance for the cluster requested in the tool provider.

Barring any feedback against this approach, I will continue with adding the cluster parameter to tools whenever the number of clusters is > 1, and using the provider interface to get the correct manager to each tool.

Cali0707:
Testing now, it seems like there is a NPE when using the non-default cluster - adding a fix for that

Cali0707:
@manusa I tried to keep the changes separated out by commit to make reviewing easier, let me know if I should squash them

manusa:
It's looking good, let me review it locally and propose some changes (if applicable) as a PR to your branch.

manusa:
I'm adding a few changes to the branch and restructuring a bit. I'll try to have a PR to your branch ready later today.

manusa:
We can squash everything together once the feature is ready.

manusa:
Hui, note I'm working on this: https://github.com/Cali0707/kubernetes-mcp-server/compare/multi-cluster-auth-config...marcnuri-forks:kubernetes-mcp-server:review/mutli-cluster-config?expand=1
I'll send a PR very soon.

Cali0707:
Thanks @manusa !

manusa:
@alexeykazakov would you mind giving a look and see if this would work for your current use-case? 🙏

alexeykazakov (approved):
I tested the PR. Works great!
The readme needs to be updated though.

sabre1041:
@Cali0707 @manusa Latest enhancements work great! Only comment is what @alexeykazakov already mentioned related to the documentation

manusa:
I'll merge now since the PR is quite complex as it is

> The readme needs to be updated though.

To be updated in a follow-up. Also maybe add an opt-out option for multi-cluster before cutting the release.

PR #354, opened 2025-10-02
Add local development environment with Kind and Keycloak for OIDC

matzew:
Adds local development environment using KIND with Keycloak as an OIDC provider and cert-manager for automated TLS certificate provisioning for testing the Kubernetes-mcp-server server with OIDC authentication flows against a (local) Kubernetes cluster.

The PR adds make targets for KIND cluster setup with Keycloak OIDC provider, including automated realm configuration with test user (mcp/mcp) and RBAC authorization.

Testing can be done with these commands:

```bash
make local-env-setup       # Setup of KIND, Keycloak (as oidc-issuer) and Cert-Manager 
make keycloak-setup-realm  # Creates a realm and adds RBAC for "mcp" user
```

The realm has one test user: `mcp / mcp`

Remove the environment with:

```bash
make local-env-teardown  # tears it all down...
```


build/keycloak.mk
```diff
@@ -0,0 +1,340 @@
...
+	else \
+		echo "Keycloak is not installed. Run: make keycloak-install"; \
+	fi
+
+.PHONY: keycloak-logs
+keycloak-logs: ## Tail Keycloak logs
+	@kubectl logs -n $(KEYCLOAK_NAMESPACE) -l app=keycloak -f --tail=100
+
+.PHONY: keycloak-setup-realm
+keycloak-setup-realm: ## Setup OpenShift realm with token exchange support
```
matzew: @sabre1041 Does this match your setup of realm, clients and scopes - what you did on the previous "managed" KC? (btw. I am on KC 26.x).

Cali0707:
@matzew can you add some kind of wait for pods to be ready? I ran into this error:
```bash
make keycloak-forward
Forwarding Keycloak to http://localhost:8090
Login: admin / admin
kubectl port-forward -n keycloak svc/keycloak 8090:80
error: unable to forward port because pod is not running. Current status=Pending
make: *** [keycloak-forward] Error 1
```

manusa:
This looks great @matzew, thx!

Leaving here some comments and **nitpicks**:

1. I'm not sure why the realm is not set up as part of the `keycloak-install` or more generic `local-env-setup` target.
  What's the reasoning behind the isolated/granular target?
2. I think there are ~13 new _**publicly documented (make help)**_ make targets, I find it kind of overwhelming user-wise.
   It's my understanding that most of these targets are used internally by other more global targets.
   I think we should hide/undocument the targets that are not relevant for developer flows or that will be called by others.
  Maybe just keep:
   - `local-env-setup`
   - `local-env-teardown`
   - `keycloak-status`
   - `keycloak-logs`
   - Regarding `keycloak-setup-realm` see (1) and only expose if necessary
3. Not sure of the current purpose of `keycloak-forward`, given the current implementation (just an informative message), maybe we should just remove it.
4. Directory structure:
    - `build` seems fine for the nested makefiles (At some point we could extract some of the current targets in the main file to this directory too).
    - `config` is used to host dev environment files, which might be confusing, I'd propose to move them somewhere else, or maybe move this `config` directory under a `dev`/`dev-env` (or something like that) directory
    - `bin`, `hack/cert-manager-ca` maybe move those to `_output` (`golangci-lint` is currently downloaded there) -- The directory can be renamed to `.work` or something more generic if needed.
5. Considering (4), the TOML configuration files could be created for the user within the `_output` directory instead of just printing their contents.


The local test with `@modelcontextprotocol/inspector@latest` and a browser with disabled CORS is working great 🎉

matzew:
Thx for review, I pushed the updates!

PR #363, opened 2025-10-08
chore(docs): add multi-cluster to readme

manusa:
Relates to #83
Relates to #348 

Adds multi-cluster feature to README and documents configuration flag to opt-out.

_I'm sure this can be improved, just adding it to not forget_

/cc @Cali0707 @matzew

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,8 @@ The following sets of tools are available (all on by default):
 
 ### Tools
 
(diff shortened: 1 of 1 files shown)


Cali0707:
Thanks for adding this @manusa !

The only thing I would add on top of this is a little note in the Tools section of the README explaining how the tools will be mutated if there are multiple contexts available in the kubeconfig

manusa:
> The only thing I would add on top of this is a little note in the Tools section of the README explaining how the tools will be mutated if there are multiple contexts available in the kubeconfig

Absolutely. I added the note, this will need updating once other strategies are allowed.

PR #364, opened 2025-10-08
make AccessControlClientset public for creating a kubernetes client outside pkg/kubernetes

harche:
With this change, the clients outside pkg/kuberntes, such as the one getting developed for openshift-mcp-server, can assemble a statically typed client. 

With this change we can assemble a typed client like this, [https://github.com/openshift/openshift-mcp-server/compare/main...harche:kubernetes-mcp-server:exec_tool?expand=1#diff-b880f18e8c0[…]405fe48bac8bR110](https://github.com/openshift/openshift-mcp-server/compare/main...harche:kubernetes-mcp-server:exec_tool?expand=1#diff-b880f18e8c099373d7fbac298d4ce4ca4120111987890ff1d4ef405fe48bac8bR110)
and then create an object (a pod in this case) like this, [https://github.com/openshift/openshift-mcp-server/compare/main...harche:kubernetes-mcp-server:exec_tool?expand=1#diff-b880f18e8c0[…]405fe48bac8bR195](https://github.com/openshift/openshift-mcp-server/compare/main...harche:kubernetes-mcp-server:exec_tool?expand=1#diff-b880f18e8c099373d7fbac298d4ce4ca4120111987890ff1d4ef405fe48bac8bR195)

Even if we are making the rest config public, this doesn't increase the attack surface because the user would have been able to create arbitrary CRs allowed by their authorization using  existing `ResourcesCreateOrUpdate` anyway.

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -41,6 +41,12 @@ type Kubernetes struct {
 	manager *Manager
 }
 
+// AccessControlClientset returns the access-controlled clientset
+// This ensures that any denied resources configured in the system are properly enforced
+func (k *Kubernetes) AccessControlClientset() *AccessControlClientset {
+	return k.manager.accessControlClientSet
+}
+
 type Manager struct {
 	cfg                     *rest.Config
 	clientCmdConfig         clientcmd.ClientConfig


mrunalp:
This seems reasonable to me.

manusa:
This is **partially** OK.

However, if the config has configured denied access to a given resource, access WILL NOT be restricted and would be up to the toolset implementor to check the config (which won't happen)

Considering the following config:
```toml
[[denied_resources]]
version = "v1"
```

The implementation in the linked PR will still be able to execute.

I'm unaware of the downstream possible scenarios but I'd still suggest that we don't open it that much (i.e. completely exposing the RESTConfig).

My suggestion would be to give access to the [AccessConstolClientset](https://github.com/containers/kubernetes-mcp-server/blob/cad863ff2246cacb9f2964f3b8ee35c02127736f/pkg/kubernetes/accesscontrol_clientset.go#L30)
And then, exposing whatever clientset is needed donwstream.

Unless I missed anything, the current implementation already provides access to every API that's needed.

So:
```go
func (k *Kubernetes) AccessControlClientset() {
  return k.manager.accessControlClientSet
}
```

Should allow for:
```go
k.AccessControlClientset().Pods($namespace).xxxWhateverInCoreV1PodInterface
```

If an interface is missing (Nodes or whatever), we can simply add a new method to the [AccessConstolClientset](https://github.com/containers/kubernetes-mcp-server/blob/cad863ff2246cacb9f2964f3b8ee35c02127736f/pkg/kubernetes/accesscontrol_clientset.go#L30)


I believe that this serves as the main reasoning as to why the access to the underlying APIs should remain centralized in a central place and the API should only expose the computed clients.
If at any point we change access-control-related logic implementors automatically benefit from it and we won't have to struggle with changing implementations downstream.

harche:
Thanks @manusa, I have updated the PR to make `AccessControlClientset` public instead.

PR #372, opened 2025-10-10
refactor(kubernetes): keep Provider as only external Kubernetes interface

manusa:
Initial phase to unify-merge the Provider interface with the Manager struct.

- Renamed ManagerProvider to Provider (i.e. kubernets.Provider)
- Moved Manager related logic to specific files
- Exposed relevant method through Provider interface (GetDerivedKubernetes, IsOpenShift, VerifyToken)

/cc @Cali0707

### DIFF
--- a/pkg/kubernetes/provider_kubeconfig.go
+++ b/pkg/kubernetes/provider_kubeconfig.go
@@ -94,8 +94,8 @@ func (p *kubeConfigClusterProvider) VerifyToken(ctx context.Context, context, to
 
 func (p *kubeConfigClusterProvider) GetTargets(ctx context.Context) ([]string, error) {
 	contextNames := make([]string, 0, len(p.managers))
-	for cluster := range p.managers {
-		contextNames = append(contextNames, cluster)
+	for contextName := range p.managers {
+		contextNames = append(contextNames, contextName)
 	}
 
 	return contextNames, nil


pkg/kubernetes/provider_kubeconfig.go
```diff
@@ -87,23 +75,56 @@ func (k *kubeConfigClusterProvider) GetManagerFor(ctx context.Context, context s
...
+	if err != nil {
+		return nil, nil, err
+	}
+	return m.VerifyToken(ctx, token, audience)
+}
+
+func (p *kubeConfigClusterProvider) GetTargets(ctx context.Context) ([]string, error) {
+	contextNames := make([]string, 0, len(p.managers))
+	for cluster := range p.managers {
+		contextNames = append(contextNames, cluster)
```
Cali0707: nit: I think this variable name mistake came from some refactors in my original PR, let's try and clean it up as part of this
```suggestion
	for contextName := range p.managers {
		contextNames = append(contextNames, contextName)
```

pkg/kubernetes/provider_kubeconfig.go
```diff
@@ -87,23 +75,56 @@ func (k *kubeConfigClusterProvider) GetManagerFor(ctx context.Context, context s
...
+	p.managers[context] = m
 
 	return m, nil
 }
 
-func (k *kubeConfigClusterProvider) GetDefaultTarget() string {
-	return k.defaultContext
+func (p *kubeConfigClusterProvider) IsOpenShift(ctx context.Context) bool {
+	return p.managers[p.defaultContext].IsOpenShift(ctx)
+}
```
Cali0707, line 85: I'm not convinced is the `IsOpenShift` method makes sense here - what if there are a mix of openshift and non openshift clusters in the kubeconfig?

Cali0707: Maybe we can either:
1. check if _any_ of the clusters are openshift
2. figure out a way to set the "contexts" parameter for the openshift specific tools to only include the contexts which _are_ openshift

pkg/mcp/m3labs.go
```diff
@@ -39,15 +39,9 @@ func ServerToolToM3LabsServerTool(s *Server, tools []api.ServerTool) ([]server.S
...
+			// get the correct derived Kubernetes client for the target specified in the request
 			cluster := request.GetString(s.p.GetTargetParameterName(), s.p.GetDefaultTarget())
-			m, err := s.p.GetManagerFor(ctx, cluster)
-			if err != nil {
-				return nil, err
-			}
-
-			// derive the manager based on auth on top of the settings for the cluster
-			k, err := m.Derived(ctx)
+			k, err := s.p.GetDerivedKubernetes(ctx, cluster)
```
Cali0707, line 44: this change is great

pkg/kubernetes/provider_kubeconfig.go
manusa: 👍

pkg/kubernetes/provider_kubeconfig.go
manusa: This is part of the _legacy_ behavior where the Kubernetes MCP server is supposed to adapt to the underlying Kubernetes clusters and provide flavor-specific tools.

In the current state, the toolsets GetTools method receive this interface and can perform some logic to register tools based on the provided information.

With the Multi-Cluster approach and the segregation of toolsets, this might not be applicable any more.

I was going to add a TODO comment in the Provider interface definition beside the `OpenShift` reference indicating that this was going to be removed.
But since I'm not sure of how we want to move forward, I didn't.

> Maybe we can either:
>
> 1. check if any of the clusters are openshift
> 2. figure out a way to set the "contexts" parameter for the openshift specific tools to only include the contexts which are openshift

These are valid options, the other alternative is to just remove the interface and provide a base openshift toolset.

Since it's unclear, and there are a few more refactors pending from my side, I think that the // TODO comment would be enough for now.

Thoughts?

Cali0707: Yeah that makes sense - if we have a comment tracking that we should revisit this logic I'm happy to leave it as is for now

manusa: I've added the TODO comment so that we eventually deal with this.

PR #377, opened 2025-10-15
refactor(config): introduce provider-specific config registry

Cali0707:
This PR introduces provider-specific configs, which the providers can define themselves.

This will allow us (and others), to add specific configs for providers in downstream repos without introducing any git conflicts.


manusa (approved):
LGTM, thx!

PR #378, opened 2025-10-16
refactor(kubernetes): streamline provider configuration and in-cluster detection

manusa:
- Removed Manager.IsInCluster method and created a function scoped to the runtime environment.
  As a method, the implementation was not correct.
- Removed GetAPIServerHost method from Manager which is no used.
- **Temporarily** added an `inCluster` field to the Manager struct but should be eventually removed since it doesn't really make sense to hava a Manager in-cluster or out-of-cluster in the multi-cluster scenario.
- Provider resolution (resolveStrategy) is now clearer, added complete coverage for all scenarios.
- Added additional coverage for provider and manager.
- Removes Provider.newForContext(context string) method to simplify Manager creation.

The next follow-up would be to ideally remove the Manager parameter from the ProviderFactory function type.
It doesn't make sense for all scenarios to have a base Kubernetes Manager, Provider implementations should be responsible of creating their own Manager instances if needed, when needed.

/cc @Cali0707

### DIFF
--- a/pkg/kubernetes/kubernetes_derived_test.go
+++ b/pkg/kubernetes/kubernetes_derived_test.go
@@ -47,7 +47,7 @@ users:
 			kubeconfig = "` + strings.ReplaceAll(kubeconfigPath, `\`, `\\`) + `"
 		`)))
 		s.Run("without authorization header returns original manager", func() {
-			testManager, err := NewManager(testStaticConfig)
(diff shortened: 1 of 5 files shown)


Cali0707 (approved):
LGTM - thanks for continuing to refactor and clean up this code @manusa !

PR #379, opened 2025-10-17
refactor(kubernetes): Provider implementations deal with Manager instantiations

manusa:
Removed `*Manager` parameter from `ProviderFactory`.

Provider implementations should deal with the appropriate (base) Manager instantiation if needed at all.

Manager creation function divided into two explicit functions:
- NewKubeconfigManager: to be used when using KubeConfig files
- NewInClusterManager: to be used inside a cluster

New functions contain validations to ensure they are used in the expected places. This ensures that the right manager is used by the provider implementation.

Fake kubeconfig for in-cluster Manager is now generated when the Manager is created. This kubeconfig has the "magic" strings (inClusterKubeConfigDefaultContext) that are used by the MCP server and tool mutators.

### DIFF
--- a/pkg/kubernetes/provider_kubeconfig.go
+++ b/pkg/kubernetes/provider_kubeconfig.go
@@ -28,7 +28,9 @@ func init() {
 }
 
 // newKubeConfigClusterProvider creates a provider that manages multiple clusters
-// via kubeconfig contexts. Returns an error if the manager is in-cluster mode.
+// via kubeconfig contexts.
+// Internally, it leverages a KubeconfigManager for each context, initializing them
+// lazily when requested.
 func newKubeConfigClusterProvider(cfg *config.StaticConfig) (Provider, error) {
 	m, err := NewKubeconfigManager(cfg, "")
 	if err != nil {
--- a/pkg/kubernetes/provider_single.go
+++ b/pkg/kubernetes/provider_single.go
@@ -25,7 +25,8 @@ func init() {
 }
 
 // newSingleClusterProvider creates a provider that manages a single cluster.
-// Validates that the manager is in-cluster when the in-cluster strategy is used.
+// When used within a cluster or with an 'in-cluster' strategy, it uses an InClusterManager.
+// Otherwise, it uses a KubeconfigManager.
 func newSingleClusterProvider(strategy string) ProviderFactory {
 	return func(cfg *config.StaticConfig) (Provider, error) {
 		if cfg != nil && cfg.KubeConfig != "" && strategy == config.ClusterProviderInCluster {


manusa:
This should be the last of the changes I had in mind for the Povider-Manager problem 😅 

It breaks https://github.com/openshift/openshift-mcp-server/pull/58 a little, but I think that it simplifies its ProviderFactory implementations since many of the checks won't be necessary :)

pkg/kubernetes/provider_single.go
```diff
@@ -26,12 +27,23 @@ func init() {
 // newSingleClusterProvider creates a provider that manages a single cluster.
 // Validates that the manager is in-cluster when the in-cluster strategy is used.
 func newSingleClusterProvider(strategy string) ProviderFactory {
-	return func(m *Manager, cfg *config.StaticConfig) (Provider, error) {
+	return func(cfg *config.StaticConfig) (Provider, error) {
```
matzew, line 31: should the comment for the func be updated ? It states that the manager is being validated, but I guess that's more based on the old behavior where a manager is passed in. 

Now it is handled internally

pkg/kubernetes/provider_kubeconfig.go
```diff
@@ -28,10 +29,13 @@ func init() {
 
 // newKubeConfigClusterProvider creates a provider that manages multiple clusters
 // via kubeconfig contexts. Returns an error if the manager is in-cluster mode.
-func newKubeConfigClusterProvider(m *Manager, cfg *config.StaticConfig) (Provider, error) {
-	// Handle in-cluster mode
-	if IsInCluster(cfg) {
-		return nil, fmt.Errorf("kubeconfig ClusterProviderStrategy is invalid for in-cluster deployments")
+func newKubeConfigClusterProvider(cfg *config.StaticConfig) (Provider, error) {
```
matzew, line 34: the above is same/similar to the `prodvider_single.go` 

IMO the comment assumes that a passed in argument is evaluated about its mode

matzew:
LGTM me, overall 

I like the cleaner approach of this

pkg/kubernetes/provider_single.go
manusa: It's actually validated (for even more specific scenarios), however this is mostly handled by the manager instantiation (which is not part of this function).
I've updated the comment to reflect what's actually being done within the function (The tests do verify everything else).

pkg/kubernetes/provider_kubeconfig.go
manusa: I've updated the comment here too.
Validation is still happening, but it's at the other layer.

PR #384, opened 2025-10-20
add support for nodes query

blublinsky:
This PR is adding supports for getting information about kubernetes cluster nodes

### DIFF
--- a/pkg/toolsets/core/nodes.go
+++ b/pkg/toolsets/core/nodes.go
@@ -63,10 +63,8 @@ func nodesLog(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
 		switch v := tail.(type) {
 		case float64:
 			tailInt = int64(v)
-		case int:
+		case int: case int64:
 			tailInt = int64(v)
-		case int64:
-			tailInt = v
 		default:
 			return api.NewToolCallResult("", fmt.Errorf("failed to parse tail parameter: expected integer, got %T", tail)), nil
 		}


manusa:
Hi, Boris.

Thanks for contributing!

As discussed internally, at the moment, we're trying to minimize the number of tools the MCP server exposes to avoid problems with smaller models or taking up too much of the token context window budget.

In this PR you're introducing multiple CRUD tools that the LLM should not need since it should be able to leverage the generic resource tools.
Similarly, for events, there is specific tooling for event retrieval that should not be necessary. In case there is a specific prompt+model combination that doesn't work with the current tooling, but works with the one proposed, we should evaluate how to deal with the specific case.

IMHO the only tool that brings something new and would be impossible to replace with current tooling is the one to retrieve the node logs. We should only keep this one.

@Cali0707 @matzew @mrunalp thoughts?

Cali0707:
> IMHO the only tool that brings something new and would be impossible to replace with current tooling is the one to retrieve the node logs. We should only keep this one.

+1 here

blublinsky:
removed all the tools except log

pkg/mcp/nodes_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+		t.Run("node_log returns successfully or with expected error", func(t *testing.T) {
+			if err != nil {
+				t.Fatalf("call tool failed: %v", err)
+			}
+			// Node logs might not be available in test environment
+			// We just check that the tool call completes
+			if toolResult.IsError {
+				content := toolResult.Content[0].(mcp.TextContent).Text
+				// Expected error messages in test environment
```
matzew, line 44: not sure, why are they expected?

pkg/toolsets/core/nodes.go
```diff
@@ -0,0 +1,81 @@
...
+	var tailInt int64
+	if tail != nil {
+		// Convert to int64 - safely handle both float64 (JSON number) and int types
+		switch v := tail.(type) {
+		case float64:
+			tailInt = int64(v)
+		case int:
+			tailInt = int64(v)
+		case int64:
+			tailInt = v
```
matzew: since JSON unmarshaling typically produces float64 for numbers, how about 

```
  case int, int64:
      tailInt = int64(v)
```

pkg/mcp/nodes_test.go
blublinsky: They are not. I am explaining why we only need tools completion, not the value

pkg/toolsets/core/nodes.go
blublinsky: fixed

matzew:
@blublinsky thanks for the PR and addressing our comments
I will test it before I merged

matzew:
I tested w/ the `kind-control-plane` node on my machine, and changed the log-path to `alternatives.log`, and got results

PR #385, opened 2025-10-21
feat(mcp): refactor to use go-sdk

### LINKED ISSUE #219: [NFR] Migrate to Go-SDK
## Description

The first stable release of Model Context Protocol Go SDK is projected for August 2025.

To be able to keep up with the latest spec changes, we should migrate the **server** side to the new SDK as soon as it's GAd.

## References

- Original Proposal: https://github.com/orgs/modelcontextprotocol/discussions/364
- Go-SDK 2025-06-18 spec compliance:
  https://github.com/modelcontextprotocol/go-sdk/issues/2
- Tool Binding APIs
  https://github.com/modelcontextprotocol/go-sdk/discussions/37

manusa:
Fixes #219


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/go-jose/go-jose/v4 v4.1.3
 	github.com/google/jsonschema-go v0.3.0
 	github.com/mark3labs/mcp-go v0.43.0
```
matzew, line 11: some of the tests still have reference to this. Will that clean-up follow in a separate PR?

pkg/mcp/mcp.go
```diff
@@ -139,38 +146,48 @@ func (s *Server) reloadKubernetesClusterProvider() error {
 			s.enabledTools = append(s.enabledTools, tool.Tool.Name)
 		}
 	}
-	m3labsServerTools, err := ServerToolToM3LabsServerTool(s, applicableTools)
-	if err != nil {
-		return fmt.Errorf("failed to convert tools: %v", err)
+
+	// TODO: No option to perform a full replacement of tools.
+	// Remove tools that are no longer applicable
```
matzew, line 151: this is also to follow up, w/ upstream?

pkg/mcp/mcp.go
```diff
@@ -139,38 +146,48 @@ func (s *Server) reloadKubernetesClusterProvider() error {
...
-		server.WithStreamableHTTPServer(httpServer),
-		server.WithStateLess(true),
-	}
-	return server.NewStreamableHTTPServer(s.server, options...)
+func (s *Server) ServeHTTP() *mcp.StreamableHTTPHandler {
+	return mcp.NewStreamableHTTPHandler(func(request *http.Request) *mcp.Server {
+		return s.server
+	}, &mcp.StreamableHTTPOptions{
+		// For clients to be able to listen to tool changes, we need to set the server stateful
+		// https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#listening-for-messages-from-the-server
```
matzew, line 188: wondering: was the old impl than not "compliant" to latest spec?

matzew:
I want to test with w/ the checked in keycloak scripts

go.mod
manusa: The tests are still running using the Mark3Labs MCP client.
We can eventually move to just use the official go-sdk mcp client, but I find it more reliable this way though.

pkg/mcp/mcp.go
manusa: yup, I think we discussed this earlier today or yesterday in a different PR or chat.
Once this is in I'll open an issue upstream and see if they are receptive to a SetTools implementation (comments in their code don't suggest they might want it, so better to provide a real use-case/scenario).

pkg/mcp/mcp.go
manusa: I think mark3labs was more lenient in this regard.
I'm not sure that the MCP protocol states that listening for tool changes is something that requires a session id, but in the go-sdk implementation it does.

matzew:
```
I1113 08:35:06.331469   44005 middleware.go:38] "mcp tool call: namespaces_list(map[])"
I1113 08:35:06.331516   44005 manager.go:244] "Authorization header found (Bearer), using provided bearer token"
I1113 08:35:06.342283   44005 middleware.go:29] "POST /mcp 200 20.15874ms"
```

tested the keycloak setup from the `docs` folder and it worked!

matzew:
LGTM :tada:

manusa:
Merging 🚀

PR #392, opened 2025-10-22
feat: add basic CORS support

Cali0707:
This PR adds basic support for CORS.

This is needed for:
1. Testing with auth setups (e.g. through mcp inspector), we don't want to require CORS disabled browsers
2. Any clients that build a browser application that want to communicate to the MCP server

### DIFF
--- a/build/keycloak.mk
+++ b/build/keycloak.mk
@@ -423,6 +423,8 @@ keycloak-setup-realm:
 	echo "  sts_audience = \"openshift\""; \
 	echo "  sts_scopes = [\"mcp:openshift\"]"; \
 	echo "  certificate_authority = \"_output/cert-manager-ca/ca.crt\""; \
+	echo "  [cors]"; \
+	echo "  origins = [\"http://localhost:6274\"]"; \
 	echo "========================================"; \
 	echo ""; \
 	echo "Note: The Kubernetes API server is configured with:"; \
@@ -444,5 +446,7 @@ keycloak-setup-realm:
 		'sts_audience = "openshift"' \
 		'sts_scopes = ["mcp:openshift"]' \
 		'certificate_authority = "_output/cert-manager-ca/ca.crt"' \
+		'[cors]' \
+		'origins = ["http://localhost:6274"]' \
 		> _output/config.toml; \
 	echo "✅ Configuration written to _output/config.toml"


manusa:
I'm not sure about this one. I'll better check once we merge the others.

Note that all of the CORS issues are caused by a bad implementation of the modelcontextprotocol inspector. The main problem is that it's performing operations that should be handled by the backend from the browser (maybe due to a limitation on how it was implemented in the first place).

pkg/http/cors.go
```diff
@@ -0,0 +1,102 @@
...
+				w.WriteHeader(http.StatusNoContent)
+				return
+			}
+
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+func setCORSHeaders(w http.ResponseWriter, origin string, corsConfig *config.CORSConfig) {
```
matzew, line 67: not really sure we want that all the time.

Is it perhaps worth also following up on the cors issue, w/ the inspector tool?

Cali0707:
@manusa @matzew from https://github.com/modelcontextprotocol/inspector/issues/817#issuecomment-3393500089 it looks like we should at least be setting CORS headers to allow all origins for the `.well-known/oauth-protected-resource` endpoint

manusa:
> @manusa @matzew from [modelcontextprotocol/inspector#817 (comment)](https://github.com/modelcontextprotocol/inspector/issues/817#issuecomment-3393500089) it looks like we should at least be setting CORS headers to allow all origins for the `.well-known/oauth-protected-resource` endpoint

In any case, this should be handled by the proxied keycloak instance.
i.e. the header should be set by keycloak and not overridden by us which are only acting as a man in the middle.

manusa:
Just to clarify:
We're proxying the .well-known endoints to be able to override certain values.
This is all implemented in https://github.com/containers/kubernetes-mcp-server/blob/49afbad502e5d97248f3c0d4b7b324311a91d021/pkg/http/wellknown.go

In this case, the implementation would be as simple as adding the header somewhere within:

https://github.com/containers/kubernetes-mcp-server/blob/49afbad502e5d97248f3c0d4b7b324311a91d021/pkg/http/wellknown.go#L49

But still, this should be responsibility of the proxied auth provider.

Cali0707:
> In any case, this should be handled by the proxied keycloak instance.

Ah, I had missed that - I am good with closing this then, but ideally I don't think we should require CORS disabled browsers for testing

manusa:
> > In any case, this should be handled by the proxied keycloak instance.
> 
> Ah, I had missed that - I am good with closing this then, but ideally I don't think we should require CORS disabled browsers for testing

Agreed. AFAIU the changes in #393 should be enough (if not there might be some other underlying issue).

For me #393 is fine regardless if what it's stated in https://github.com/modelcontextprotocol/inspector/issues/817#issuecomment-3393500089 is a common practice or not (I'd really like to see the sources that back what's stated in that screen shot).
In the end, it's just a keycloak environment for development purposes.

Cali0707:
Closing this as per discussion above - let's go with just the changes in #393 for now

PR #390, opened 2025-10-22
cleanup: refactor all the keycloak setup to json files

Cali0707:
This PR refactors all of the keycloak JSON into JSON files, to make it easier to change keycloak settings in the future

PR #393, opened 2025-10-22
Keycloak cors setup

Cali0707:
This builds on top of the refactor introduced in #390 and should not be merged until after that one.

With these changes, the local keycloak will not run into CORS issues

### DIFF
--- a/dev/config/keycloak/clients/mcp-client.json
+++ b/dev/config/keycloak/clients/mcp-client.json
@@ -7,6 +7,7 @@
   "serviceAccountsEnabled": false,
   "authorizationServicesEnabled": false,
   "redirectUris": ["*"],
+  "webOrigins": ["*"],
   "defaultClientScopes": ["profile", "email"],
   "optionalClientScopes": ["mcp-server"]
 }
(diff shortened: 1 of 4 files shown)


matzew:
while these (`"webOrigins": ["*"],`) are generally fine for development. Does it still make sense merging just this?

I guess #392 has some implications on the subject - but this here would be only part of the solution (until the MCP inspector is fixed? If ever?)

I guess I am OK merging this, but not sure if that really makes sense.

Cali0707:
> I guess #392 has some implications on the subject - but this here would be only part of the solution (until the MCP inspector is fixed? If ever?)

Based on the conversation I saw there, it's not clear to me that inspector will proxy the auth as well as the mcp bits, IMO this is fine since these scripts are local dev only

Cali0707:
Hey @manusa @matzew - fixed the merged conflicts here

manusa:
I identified the issue in our proxy service, since we're not propagating the original headers (including Origin), keycloak doesn't return the Access-Control-Allow-Origin header.

I'll fix in a follow-up PR.

Related references:
- https://github.com/keycloak/keycloak/issues/33434

PR #394, opened 2025-10-22
fix: do not use in-use port in example run command

Cali0707:
Currently, if you run `make local-env-setup` it will tell you:

```
Run the MCP server with:
  ./kubernetes-mcp-server --port 8080 --conifg _output/config.toml
```

However, because of the way we set up keycloak, this does not work as keycloak is listening on that port.

This PR updates the port we recommend from 8080 to 8008


manusa (approved):
LGTM, thx!
The port could even be added to the toml file.

PR #395, opened 2025-10-22
refactor(test): toolset tests do not rely on composite JSON files

### LINKED ISSUE #391: Refactor tests to merge multiple json files
To make it easier to extend toolsets in forks, like the openshift-mcp-server, we should refactor the mcp/toolset tests to:
1. Have tools from each toolset defined in it's own json file
2. Merge those tools into the final expected object based on which toolsets should be enabled in the scenario
3. Make assertions about equality

Cali0707:
Resolves #391 

This PR removes the need for larger "composite" files of different toolsets. Instead, it merges those and provides the appropriate transformations.


manusa:
Check https://github.com/containers/kubernetes-mcp-server/issues/391#issuecomment-3435013548

Cali0707:
Closing in favour of #396

PR #396, opened 2025-10-23
feat(config): default configuration with merge support for downstream overrides

### LINKED ISSUE #391: Refactor tests to merge multiple json files

manusa:
Creates the required infrastructure for downstream forks to be able to provide default config overrides without modifying the original source code.

Downstream forks should be able to create merge/rebase scripts that automatically accepted downstream merge conflicts in config_default_overrides.go since this file will never change upstream.

Example usage for downstream forks:

To customize defaults, simply populate fields in the returned StaticConfig:

```go
func defaultOverrides() *StaticConfig {
  return &StaticConfig{
    ListOutput: "json",           // Override default list output format
    Toolsets:   []string{"core"}, // Override default enabled toolsets
    Port:       "9000",           // Override default port
  }
}
```

Any fields specified here will override the base defaults defined in config_default.go. Fields not specified will preserve their base default values.

--- 

The second commit fixes #391 (AFAIU)

The full toolset metadata tests are skipped in case there are config overrides.

If downstream follows the convention:
- Not to mutate the original toolsets
- Provide config overrides in the appropriate pkg/config/config_default_overrides.go file

The tests should not fail.

### DIFF
--- a/pkg/config/config_default.go
+++ b/pkg/config/config_default.go
@@ -16,6 +16,18 @@ func Default() *StaticConfig {
 	return &mergedConfig
 }
 
+// HasDefaultOverrides indicates whether the internal defaultOverrides function
+// provides any overrides or an empty StaticConfig.
+func HasDefaultOverrides() bool {
+	overrides := defaultOverrides()
+	var buf bytes.Buffer
+	if err := toml.NewEncoder(&buf).Encode(overrides); err != nil {
+		// If marshaling fails, assume no overrides
+		return false
+	}
+	return len(bytes.TrimSpace(buf.Bytes())) > 0
+}
+
 // mergeConfig applies non-zero values from override to base using TOML serialization
 // and returns the merged StaticConfig.
 // In case of any error during marshalling or unmarshalling, it returns the base config unchanged.
--- a/pkg/mcp/toolsets_test.go
+++ b/pkg/mcp/toolsets_test.go
@@ -65,6 +65,9 @@ func (s *ToolsetsSuite) TestNoToolsets() {
 }
 
 func (s *ToolsetsSuite) TestDefaultToolsetsTools() {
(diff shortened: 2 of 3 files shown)


pkg/config/config_default.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/BurntSushi/toml"
+)
+
+func Default() *StaticConfig {
+	defaultConfig := StaticConfig{
+		ListOutput: "table",
+		Toolsets:   []string{"core", "config", "helm"},
+	}
+	overrides := defaultOverrides()
+	mergedConfig := mergeConfig(defaultConfig, overrides)
```
matzew, line 15: I like this approach

pkg/mcp/toolsets_test.go
```diff
@@ -65,6 +65,9 @@ func (s *ToolsetsSuite) TestNoToolsets() {
 }
 
 func (s *ToolsetsSuite) TestDefaultToolsetsTools() {
+	if configuration.HasDefaultOverrides() {
```
matzew, line 68: I guess this is a good hook, for downstream.

pkg/config/config_default_overrides.go
```diff
@@ -0,0 +1,8 @@
+package config
+
+func defaultOverrides() StaticConfig {
+	return StaticConfig{
+		// IMPORTANT: this file is used to override default config values in downstream builds.
+		// This is intentionally left blank.
+	}
```
Cali0707, line 7: This is great @manusa !

PR #399, opened 2025-10-23
fix(nodes): reviewed kubernetes.nodes implementation

manusa:
Follows up on #384

- Changed node_log tool name to nodes_log for consistency
- Added access control check for nodes and configured denied resources
- Migrated tests to testify
- Added complete test coverage for nodes_log

/cc @blublinsky

It also seems that the tailLines param is not placed in the expected location. Will look into it as a subsequent fix: (might be an issue with the cluster itself)
https://kubernetes.io/docs/concepts/cluster-administration/system-logs/#log-query

### DIFF
--- a/pkg/mcp/nodes_test.go
+++ b/pkg/mcp/nodes_test.go
@@ -158,6 +158,23 @@ func (s *NodesSuite) TestNodesLog() {
 					"expected log content '%s', got %v", expectedMessage, toolResult.Content[0].(mcp.TextContent).Text)
 			})
 		})
+		s.Run("nodes_log(name=existing-node, log_path=kubelet.log, tail=-1)", func() {
+			toolResult, err := s.CallTool("nodes_log", map[string]interface{}{
+				"name":     "existing-node",
+				"log_path": "kubelet.log",
+				"tail":     -1,
+			})
+			s.Require().NotNil(toolResult, "toolResult should not be nil")
+			s.Run("no error", func() {
+				s.Falsef(toolResult.IsError, "call tool should succeed")
+				s.Nilf(err, "call tool should not return error object")
+			})
+			s.Run("returns full log", func() {
+				expectedMessage := "Line 1\nLine 2\nLine 3\nLine 4\nLine 5\n"
+				s.Equalf(expectedMessage, toolResult.Content[0].(mcp.TextContent).Text,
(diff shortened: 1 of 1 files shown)


blublinsky:
Thanks @manusa, One other thing that I have noticed is that there is no default value for the  tailLines. Should we use something like 500?

manusa:
> Thanks @manusa, One other thing that I have noticed is that there is no default value for the tailLines. Should we use something like 500?

At the moment there's a default value of 100 (you set this in the initial impl):

https://github.com/containers/kubernetes-mcp-server/blob/acf465b5ed75c792df534b4ed6dc1e0f5d169d6c/pkg/toolsets/core/nodes.go#L33

Clients should preset this value. 

However, the value is not enforced. Maybe the minimum is not checked either (I need to verify this).

manusa:
I added an additional test to verify the current behavior with negative (unsupported) tail arguments.

In this case, we're not validating and simply defaulting to show the entire log.
I'm not sure if this behavior will change with #385, but at least we'll be aware of any changes in the behavior.

pkg/toolsets/core/nodes.go
```diff
@@ -63,13 +63,14 @@ func nodesLog(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
 		switch v := tail.(type) {
 		case float64:
 			tailInt = int64(v)
-		case int: case int64:
-			tailInt = int64(v)
+		case int:
+		case int64:
+			tailInt = v
```
Cali0707, line 68: @manusa I feel like I've seen this code/similar code in other places in the codebase - is it worth extracting to a util?

Or with the sdk migration I guess if we define the paramaters to a tool using a `struct` with json tags, this would be handled by the sdk

manusa: It's in the same function for pod logs (although it's not exactly the same),

With our approach towards the go-sdk and decoupled toolsets, it might not be the case that the SDK takes care of this, but I need to evaluate first.

If it's not the case, and we do want to support reading number values with diverse input format or types, we definitely need to extract this to some util.
Similarly to extract strings, etc.

manusa:
> Overall LGTM - left one comment I wasn't sure about @manusa
> 
> Feel free to merge if you want

Merging, I'll take care of argument parsing in the scope of #385

PR #400, opened 2025-10-23
test(pods): update PodsExec tests to use testify and improve readability

manusa:
Will facilitate #385

### DIFF
--- a/pkg/mcp/pods_exec_test.go
+++ b/pkg/mcp/pods_exec_test.go
@@ -67,7 +67,7 @@ func (s *PodsExecSuite) TestPodsExec() {
 	}))
 	s.InitMcpClient()
 
-	s.Run("pods_exec(name=pod-to-exec, command=[ls -l])", func() {
+	s.Run("pods_exec(name=pod-to-exec, namespace=nil, command=[ls -l]), uses configured namespace", func() {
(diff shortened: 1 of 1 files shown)


pkg/mcp/pods_exec_test.go
```diff
@@ -7,125 +7,132 @@ import (
...
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: "default",
+				Name:      "pod-to-exec",
+			},
+			Spec: v1.PodSpec{Containers: []v1.Container{{Name: "container-to-exec"}}},
+		})
+	}))
+	s.InitMcpClient()
+
+	s.Run("pods_exec(name=pod-to-exec, command=[ls -l])", func() {
```
Cali0707: nit, can we call out that `namespace=""` here - I liked how explicit the old test was, especially since that is the key for why the tool call fails

manusa: > especially since that is the key for why the tool call fails

Absolutely, I added more clarity to the description too.
In this case, the tool call won't fail and will be executed in the configured namespace.
Given your comment, the test description was not doing a great job.

Cali0707: Looking now, I understand what's supposed to happen - thanks!

PR #413, opened 2025-10-27
Developer doc for keycloak environment

matzew:
Adding new doc for how to use the Keycloak setup for "development environment" and testing it w/ the mcp inspector

I kept the screenshots to a minimum, while still hoping to provide value


manusa:
Note that what's shown here and the config flags and fields are not yet stable.
I would add an admonition or something stating that this is still preview.
We don't want users complaining that their configuration is not working anymore in case we change some of this in the future.

matzew:
Ok, thats fair enough

PR #425, opened 2025-10-31
feat(toolsets): add kiali support

aljesusg:
Adding kiali toolset to kubernetes-mcp

- Adds Kiali toolset with a mesh status tool.
- Requires `kiali-url` when Kiali is enabled (via flag or config).
- Safer URL building and proper token forwarding to Kiali.(We use the same token to validate permissions and access, so in the local environment we also check the kubeconfig to ensure that the provided context includes valid authentication.)
- Adds focused tests for Kiali manager, client, and mesh.
- Updates help text to list Kiali as an available toolset.

We'll add others later after this first PR

Screenshot From 2025-10-31 13-07-48

cc @hhovsepy

### DIFF
--- a/docs/KIALI_INTEGRATION.md
+++ b/docs/KIALI_INTEGRATION.md
@@ -4,7 +4,7 @@ This server can expose Kiali tools so assistants can query mesh information (e.g
 
 ### Enable the Kiali toolset
 
-You can enable the Kiali tools via config or flags.
+Enable the Kiali tools via the server TOML configuration file.
 
 Config (TOML):
 
@@ -13,19 +13,14 @@ toolsets = ["core", "kiali"]
 
 [toolset_configs.kiali]
 url = "https://kiali.example"
-# insecure = true  # optional: allow insecure TLS
+# insecure = true  # optional: allow insecure TLS (not recommended in production)
+# certificate_authority = """-----BEGIN CERTIFICATE-----
+# MIID...
+# -----END CERTIFICATE-----"""
+# When url is https and insecure is false, certificate_authority is required.
 ```
 
-Flags:
-
-```bash
-kubernetes-mcp-server \
-  --toolsets core,kiali \
-  --kiali-url https://kiali.example \
-  [--kiali-insecure]
-```
-
-When the `kiali` toolset is enabled, a Kiali toolset configuration is required. Provide it via `[toolset_configs.kiali]` in the config file or by passing flags (which populate the toolset config). If missing or invalid, the server will refuse to start.
+When the `kiali` toolset is enabled, a Kiali toolset configuration is required via `[toolset_configs.kiali]`. If missing or invalid, the server will refuse to start.
 
 ### How authentication works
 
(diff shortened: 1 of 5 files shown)


pkg/kubernetes-mcp-server/cmd/root.go
```diff
@@ -94,6 +97,8 @@ type MCPServerOptions struct {
 	CertificateAuthority string
 	ServerURL            string
 	DisableMultiCluster  bool
+	KialiUrl             string
```
matzew: since this list is growing and growing, should we group or structure those options?

Just a general thought

aljesusg: I think it’s a good idea, since we’ll have to add more options like CA, so grouping everything under something like kiali: seems like a good idea to me.

pkg/kubernetes/kubernetes.go
```diff
@@ -37,3 +38,12 @@ func (k *Kubernetes) NewHelm() *helm.Helm {
...
 }
+
+// CurrentBearerToken returns the bearer token that the Kubernetes client is currently
+// configured to use, or empty if none is set in the underlying rest.Config.
+func (k *Kubernetes) CurrentBearerToken() string {
+	if k == nil || k.manager == nil || k.manager.cfg == nil {
+		return ""
+	}
+	return strings.TrimSpace(k.manager.cfg.BearerToken)
+}
```
manusa, line 46: Ideally, we should generate the kiali instance here just like we generate the one for helm (NewKiali).

This way, kiali shouldn't need to worry about auth.
In the same way, there won't be any kiali-related login in the mcp package.
The toolset impl could then do params.NewKiali().MeshStatus(params.Context)

aljesusg: Let me check, because we tried it that way at first and it didn’t work for us — maybe we did it wrong.

manusa: Let me know if you need assistance.
I can try to assist you if it doesn't work (maybe something else needs changing beforehand).

pkg/kubernetes-mcp-server/cmd/root.go
manusa: Maybe we can do something similar to what we do for providers:

Having a `toolset_configs` or similar could allow for:
```toml
[toolset_configs.kiali]
url = "http://example.com"
```

This could also be leveraged by downstream toolset implementors to add their own config options.

Cali0707: +1 on toolset_configs - we can probably reuse most (maybe all) of the logic from https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/config/provider_config.go

pkg/kubernetes/kubernetes.go
aljesusg: Worked ! Thanks

pkg/kubernetes-mcp-server/cmd/root.go
aljesusg: I moved kiali related under key: kiali . What do you think?

matzew:
It would make sense to provide a little document on this, e.g. in here https://github.com/containers/kubernetes-mcp-server/tree/main/docs ?

For instance, when setting the configuration toml to `toolsets = ["kiali"]` the MCP server would serve those kiali tools, and require the `URL` where it is running.

aljesusg:
Sure! I’ll take care of the docs part — I just wanted to get the code reviewed first.
By the way, there’s an error message when the Kiali toolset is enabled without a defined URL.

pkg/config/config.go
```diff
@@ -68,6 +74,9 @@ type StaticConfig struct {
 	// This map holds raw TOML primitives that will be parsed by registered provider parsers
 	ClusterProviderConfigs map[string]toml.Primitive `toml:"cluster_provider_configs,omitempty"`
 
+	// KialiOptions is the configuration for the kiali toolset.
+	KialiOptions KialiOptions `toml:"kiali,omitempty"`
```
Cali0707: @aljesusg do you think you could move this into a toolset specific config, following the pattern we used for the `ClusterProviderConfigs` above? 

IMO this will make managing different toolset configurations easier + clearer, and more extensible in downstream forks

aljesusg: Ok let me do it !

aljesusg: Done !

pkg/kiali/kiali.go
```diff
@@ -0,0 +1,109 @@
...
+func (k *Kiali) authorizationHeader() string {
+	if k == nil || k.manager == nil {
+		return ""
+	}
+	token := strings.TrimSpace(k.manager.BearerToken)
+	if token == "" {
+		return ""
+	}
+	lower := strings.ToLower(token)
+	if strings.HasPrefix(lower, "bearer ") {
```
matzew: I think the rest of the code is using like `strings.HasPrefix(authorization, "Bearer ")`, followed by a `TrimPrefix`

Just lowering why lowering, than checking and trim space?

matzew: Is there something special here?

aljesusg: Let me check it, because we inherited this from one of the cases we had.

aljesusg: Well we saw from other tools that the code is passed without "Bearer". We know it’s not the right way, but we wanted to consider it.

aljesusg:
Reviewed and merged. thanks !

aljesusg:
I’ll remove the flags related with Kiali next day. 
Since I see there’s consensus regarding the configuration, I’ll also add the rest of the tools we have.

Thanks

aljesusg:
Hi! I’ve removed the Kiali configuration and the flags. I’ve also added the rest of the tools that are available so far.

docs/KIALI_INTEGRATION.md
```diff
@@ -0,0 +1,170 @@
...
+When the `kiali` toolset is enabled, a Kiali toolset configuration is required via `[toolset_configs.kiali]`. If missing or invalid, the server will refuse to start.
+
+### How authentication works
+
+- The server uses your existing Kubernetes credentials (from kubeconfig or in-cluster) to set a bearer token for Kiali calls.
+- If you pass an HTTP Authorization header to the MCP HTTP endpoint, that is not required for Kiali; Kiali calls use the server's configured token.
+
+### Available tools (initial)
+
+<details>
```
manusa, line 32: This essentially duplicates what's shown in the main README.md file but it's not automatically generated.
As a follow-up we might want to either generate this automatically, or remove it from here and link to the main readme.

manusa (approved):
LGTM, thx.

However, a few considerations for follow-up PRs.

This is adding close to 20 new tools which might be sort of overwhelming for many LLMs. Especially because it's likely that this toolset will be enabled along with the standard/core kubernetes toolset.
All in all, only with this domain, ~40 tools will be available for the LLM to choose from.

There is also collision of tools (namespaces_list for example), these are already defined.

We might want to consider some of the tools to avoid it. (e.g. the xxx_traces tools could be easily combined into a single tool, probably the metrics too)

The tool descriptions are kind of vague:
- we might want to prepend the tool names with `kiali_` or `istio_`. e.g. `health` tool might be very confusing for the LLM
- Some descriptions mention "the mesh", maybe we want to add the Istio word too.

`workload_logs`: Doesn't this provide similar features as `pods_log`?

aljesusg:
> LGTM, thx.
> 
> However, a few considerations for follow-up PRs.
> 
> This is adding close to 20 new tools which might be sort of overwhelming for many LLMs. Especially because it's likely that this toolset will be enabled along with the standard/core kubernetes toolset.
> All in all, only with this domain, ~40 tools will be available for the LLM to choose from.
> 
> There is also collision of tools (namespaces_list for example), these are already defined.
> 
> We might want to consider some of the tools to avoid it. (e.g. the xxx_traces tools could be easily combined into a single tool, probably the metrics too)
> 
> The tool descriptions are kind of vague:
> - we might want to prepend the tool names with `kiali_` or `istio_`. e.g. `health` tool might be very confusing for the LLM
> - Some descriptions mention "the mesh", maybe we want to add the Istio word too.
> 
> `workload_logs`: Doesn't this provide similar features as `pods_log`?

Thanks!

Totally agree — we’re happy to simplify things in a follow-up PR. For now, we’ve been experimenting with which tools can provide the most useful information to the LLM, but we’ll definitely review and consolidate them where possible.

Regarding the namespaces overlap, our implementation only returns the namespaces that belong to the mesh, not all Kubernetes namespaces. This distinction is intentional, since the LLM uses it specifically to inspect mesh-related namespaces rather than the full cluster scope.

About the logs — good point! I’ll discuss it with the team, since it’s true that workload_logs already aggregates the logs from the pods, so we could probably remove the pod-level one.

In fact, we’ve noticed that the LLM sometimes makes multiple calls in sequence, so we were already considering grouping some of them. That will definitely be our next step.

Thanks again for the feedback!

aljesusg:
@matzew @Cali0707 Are you ok with this?

manusa:
> @matzew @Cali0707 Are you ok with this?

Once reviewed, please don't merge. I want to cut a last release before moving to go-sdk and adding the new toolset.

Issue #416, opened 2025-10-28
How to handle timeouts


iamsudip:
How to handle tool calls which takes time to return, for example, helm_list with all namespaces true in a cluster with 100s of services will take some time, but the mcp server send a timeout response after 10 seconds`

How do we overcome that?

<img width="679" height="121" alt="Image" src="https://github.com/user-attachments/assets/8d3bb33b-3796-43f8-a1a4-4e6407774276" />

`

manusa:
As a follow up of #219 we'll be looking into long-running tasks and support for Progress.
https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/progress

Cali0707:
@iamsudip if you are using the mcp inspector (from the screenshot it looks like you might be), that also sets a timeout on the requests. 

<img width="306" height="268" alt="Image" src="https://github.com/user-attachments/assets/b8cedc75-a612-4cce-aff6-fe1a5fb94157" />

iamsudip:
Thanks. Got it now, I can manage this from mcp client timeout.

PR #430, opened 2025-11-03
Add node_files tool for privileged node filesystem operations

ngopalak-redhat:
This PR adds a new node_files MCP tool that enables AI assistants like Claude to perform file operations on Kubernetes node filesystems through temporary privileged pods.

###   Motivation

  AI tools are effective at identifying the specific node files needed for debugging (e.g., kernel configs, system logs, cgroup settings). However, the manual workflow of creating privileged debug pods and copying files is time-consuming and repetitive. While kubectl cp and kubectl debug achieve the same goal, they require constant monitoring and are slower when fetching multiple files (10-15+) for analysis.

###   Key Features

  - Three operations: put (copy to node), get (copy from node), list (directory listing)
  - Auto-cleanup: Temporary privileged pod is automatically deleted after operation
  - Configurable:
    - namespace: Defaults to default (not all clusters allow privileged pods in all namespaces)
    - image: Defaults to busybox (organizations can specify approved images)
    - privileged: Can be toggled if needed

###   Use Cases

  1. Rapid multi-file debugging: Fetch and analyze 10-15 files together for comprehensive troubleshooting
  2. Verification: Quickly verify configuration changes applied to nodes
  3. Monitoring: Use list operation to monitor file creation/changes
  4. Privileged file access: Access files requiring privileged permissions (note: use nodes_log for standard logging via kubelet proxy)

Note: I didn't add any zipping feature as I observed in most cases the files fetched need to analyzed by other AI tool. Zipping slows it down. It can be considered later.

###   Performance

  Significantly faster than manual debug pod creation - the MCP server handles pod lifecycle automatically, allowing Claude to focus on analysis rather than cluster operations.

### Examples
```
Get node system information:
  Get /proc/cpuinfo and /proc/meminfo from node ip-10-0-119-110.ec2.internal

  Debug cgroup settings:
  Get /sys/fs/cgroup/memory.max from node ip-10-0-50-25.ec2.internal
  Verify memory limits applied at the cgroup level.

  Monitor kubelet configuration changes:
  List /etc/kubernetes/kubelet.conf on all nodes
  Verify if kubelet config updates were rolled out correctly.
```


ngopalak-redhat:
@manusa thank you very much for your reviews on the previous PRs. I've rebased this PR and moved the test to a different file, similar to the earlier PR.

ngopalak-redhat:
@manusa I still believe this provides value. Would you be open to reconsidering this?

pkg/kubernetes/nodes.go
```diff
@@ -96,3 +108,266 @@ func (k *Kubernetes) NodesTop(ctx context.Context, options NodesTopOptions) (*me
...
+	Image      string
+	Privileged bool
+}
+
+// NodesFiles handles file operations on a node filesystem by creating a privileged pod
+func (k *Kubernetes) NodesFiles(ctx context.Context, opts NodeFilesOptions) (string, error) {
+	// Set defaults
+	if opts.Namespace == "" {
+		opts.Namespace = "default"
+	}
```
manusa, line 128: For other operations we're using the `k.NamespaceOrDefault` method.
I'm not sure if this would apply here too.

pkg/kubernetes/nodes.go
```diff
@@ -96,3 +108,266 @@ func (k *Kubernetes) NodesTop(ctx context.Context, options NodesTopOptions) (*me
...
+	if opts.Image == "" {
+		opts.Image = "busybox"
+	}
```
manusa, line 131: Since the purpose of the tool/feature is to perform file operations is there any value in being able to customize the used image?

I understand, that changing the image might also break the implementation. Wouldn't it be better to use a specific image that can be freely pulled and that worked on any cluster?

pkg/kubernetes/nodes.go
```diff
@@ -96,3 +108,266 @@ func (k *Kubernetes) NodesTop(ctx context.Context, options NodesTopOptions) (*me
...
+					HostPath: &v1.HostPathVolumeSource{
+						Path: "/",
+					},
+				},
+			}},
+		},
+	}
+
+	// Create the pod
+	pods, err := k.AccessControlClientset().Pods(opts.Namespace)
```
manusa, line 173: Note to self: this is a good candidate to try out MCP progress notifications (Not in this PR)

- https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/progress
- https://github.com/containers/kubernetes-mcp-server/issues/416
- https://github.com/containers/kubernetes-mcp-server/pull/496#discussion_r2557155286

pkg/kubernetes/nodes.go
```diff
@@ -96,3 +108,266 @@ func (k *Kubernetes) NodesTop(ctx context.Context, options NodesTopOptions) (*me
...
+	default:
+		return "", fmt.Errorf("unknown operation: %s", opts.Operation)
+	}
+
+	_ = createdPod
+	return result, opErr
+}
+
+// nodeFilesPut copies a file from local filesystem to node filesystem
+func (k *Kubernetes) nodeFilesPut(ctx context.Context, namespace, podName, sourcePath, destPath string) (string, error) {
```
manusa, line 214: For the copy operations we should probably replicate what kubectl does to ensure the behavior is consistent:

https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/cp/cp.go

Personally, I've had some bad experiences with the implementation of these operations in the [Fabric8 Kubernetes Client](https://github.com/fabric8io/kubernetes-client/)

manusa:
> @manusa I still believe this provides value. Would you be open to reconsidering this?

Hi, @ngopalak-redhat 
Sorry for the lack of replies. The last couple of weeks have been crazy.
I added a few comments to the current implementation. Feature-wise sound like something useful to me, not sure what the rest of the team thinks.

I added a few comments.

PR #440, opened 2025-11-04
feat(helm): add Helm chart for Kubernetes MCP Server deployment

### LINKED ISSUE #435: A Helm chart to install the MCP server on the cluster
Helm is the package manager of Kubernetes.

It is about time we had a Helm chart of Kubernetes MCP server.

Cluster administrators will be able to install the MCP server with a simple `helm install k8s-mcp-server . -f values.yaml` command

I recommend we put this link [https://github.com/frenoid/kubernetes-mcp-server-helmchart](https://github.com/frenoid/kubernetes-mcp-server-helmchart) on the README.md to make this the official helm chart.

sabre1041:
Added a Helm chart for the MCP Server

Future enhancements include automating the publishing of the chart to an OCI repository

resolves #435

### DIFF
--- a/charts/kubernetes-mcp-server/templates/deployment.yaml
+++ b/charts/kubernetes-mcp-server/templates/deployment.yaml
@@ -1,0 +1,85 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "kubernetes-mcp-server.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "kubernetes-mcp-server.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      {{- include "kubernetes-mcp-server.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+      {{- with .Values.podAnnotations }}
+        {{- tpl (toYaml .) . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "kubernetes-mcp-server.labels" . | nindent 8 }}
+        {{- with .Values.podLabels }}
+        {{- tpl (toYaml .) . | nindent 8 }}
+        {{- end }}
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- tpl (toYaml .) . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "kubernetes-mcp-server.serviceAccountName" . }}
+      securityContext:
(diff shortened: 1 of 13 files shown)


charts/kubernetes-mcp-server/templates/deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+          {{- end }}
+          volumeMounts:
+            - name: config
+              mountPath: {{ .Values.configFilePath | dir }}
+          {{- with .Values.extraVolumeMounts }}
+          {{- tpl (toYaml .) . | nindent 12 }}
+          {{- end }}
+          {{- with .Values.extraVolumeMounts }}
+            {{- tpl (toYaml .) . | nindent 12 }}
+          {{- end }}
```
Cali0707, line 66: this seems duplicated

charts/kubernetes-mcp-server/values.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+# -- This will set the replicaset count more information can be found here: https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/
+replicaCount: 1
+
+# -- This sets the container image more information can be found here: https://kubernetes.io/docs/concepts/containers/images/
+image:
+  registry: quay.io
+  repository: manusa/kubernetes_mcp_server
+  # -- This sets the tag or sha digest for the image.
+  version: latest
```
Cali0707, line 12: should this go to `latest`? or do we want to pin it to the most recent release?

charts/kubernetes-mcp-server/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: {{ include "kubernetes-mcp-server.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "kubernetes-mcp-server.labels" . | nindent 4 }}
+  {{- with .Values.serviceAccount.annotations }}
+  annotations:
+    {{- tpl (toYaml .) . | nindent 4 }}
+  {{- end }}
+{{- end }}
```
Cali0707, line 13: @sabre1041 maybe I'm missing something but don't we need a `ClusterRole` + `ClusterRoleBinding` if `require_oauth` is not set to `true` (`false` is the default)?

Cali0707: In this scenario, if a request does not have a bearer token in the request, the underlying kubernetes manager will be used (using the service account credentials), which has no permissions

charts/kubernetes-mcp-server/values.yaml
matzew: I think generally, we should have image somewhere else published, instead of `manusa` repo :sweat_smile:

manusa: I will work on that next :)
Should be released to quay/containers org

PR #441, opened 2025-11-05
fix(toolsets): remove "redundant" annotations that have defaults

manusa:
Will facilitate #385

The removed annotations have default values and are not needed. I personally prefer to keep them because it makes it easier for reviewers to understand the tool behaviors.
However, this is problematic with go-sdk since these values aren't marshalled/serialized (omitempty).


pkg/toolsets/helm/helm.go
```diff
@@ -39,9 +39,8 @@ func initHelm() []api.ServerTool {
 			},
 			Annotations: api.ToolAnnotations{
 				Title:           "Helm: Install",
-				ReadOnlyHint:    ptr.To(false),
 				DestructiveHint: ptr.To(false),
-				IdempotentHint:  ptr.To(false), // TODO: consider replacing implementation with equivalent to: helm upgrade --install
+				IdempotentHint:  nil, // TODO: consider replacing implementation with equivalent to: helm upgrade --install
```
Cali0707, line 43: Is there a reason why this is nil here and just removed elsewhere?

manusa: To preserve the TODO note

Cali0707: ah makes sense

PR #443, opened 2025-11-07
test(authorization): improved tests to use real MCP clients

manusa:
Will facilitate #385

After these changes, the entire test suite should pass with go-sdk, any failure is related to a missing implementation.

### DIFF
--- a/internal/test/mcp.go
+++ b/internal/test/mcp.go
@@ -12,6 +12,13 @@ import (
 	"golang.org/x/net/context"
 )
 
+func McpInitRequest() mcp.InitializeRequest {
+	initRequest := mcp.InitializeRequest{}
+	initRequest.Params.ProtocolVersion = mcp.LATEST_PROTOCOL_VERSION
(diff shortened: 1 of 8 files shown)


internal/test/mcp.go
```diff
@@ -12,6 +12,13 @@ import (
 	"golang.org/x/net/context"
 )
 
+func McpInitRequest() mcp.InitializeRequest {
+	initRequest := mcp.InitializeRequest{}
+	initRequest.Params.ProtocolVersion = mcp.LATEST_PROTOCOL_VERSION
```
matzew, line 17: just curiois: new sdk requires `2025-06-18` ?

manusa: No, that was already there:

https://github.com/containers/kubernetes-mcp-server/blob/6342379505a0726821c5d1420102a8969e99aadb/internal/test/mcp.go#L31-L33

Just moved it around and made it reusable.

matzew:
nice Authorization tests!

PR #444, opened 2025-11-07
feat(cmd)!: remove deprecated http/sse port from cli

matzew:
The CLI still has the deprecated flags for `http`/`sse` port. Removing just those deprecated flags since we are close to migrate to new SDK.

I guess somewhat related, and likely different issue, but isn't the SSE transport is deprecated? Or is that being tackled with the migration to the new, "official", SDK, @manusa ?


manusa:
/hold
Let's hold this until we do a final release before changing to go-sdk
We can introduce some of the breaking changes then.

PR #449, opened 2025-11-11
test(mcp): add tests for WatchKubeConfig tools reload

manusa:
Ensures #385 / #219 preserves current behavior.

Verifies that tools are added and **removed** when kubeconfig file changes.

Go SDK lacks a SetTools method, AddTool only upserts tools. This test verifies that internally we handle the case that no-longer applicable tools are removed from the list of tools.

### DIFF
--- a/internal/test/mock_server.go
+++ b/internal/test/mock_server.go
@@ -59,6 +59,10 @@ func (m *MockServer) Handle(handler http.Handler) {
 	m.restHandlers = append(m.restHandlers, handler.ServeHTTP)
 }
 
+func (m *MockServer) ResetHandlers() {
+	m.restHandlers = make([]http.HandlerFunc, 0)
+}
+
 func (m *MockServer) Config() *rest.Config {
 	return m.config
 }
--- a/pkg/mcp/mcp_watch_test.go
+++ b/pkg/mcp/mcp_watch_test.go
@@ -1,0 +1,103 @@
+package mcp
+
+import (
+	"context"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/containers/kubernetes-mcp-server/internal/test"
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/stretchr/testify/suite"
+)
+
+type WatchKubeConfigSuite struct {
+	BaseMcpSuite
+	mockServer *test.MockServer
+}
+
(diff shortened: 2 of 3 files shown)


pkg/mcp/mcp_watch_test.go
```diff
@@ -0,0 +1,103 @@
...
+				found = true
+				break
+			}
+		}
+		s.Truef(found, "expected OpenShift tool to be available")
+	})
+
+	s.Run("OpenShift tool is removed after kubeconfig change", func() {
+		// Reload Config without OpenShift
+		s.mockServer.ResetHandlers()
```
matzew, line 88: It took me a bit, to see that it actually "reset" the handler**s** (while the other one accepts one handler).

I guess no major issue

internal/test/mock_server.go
```diff
@@ -59,6 +59,10 @@ func (m *MockServer) Handle(handler http.Handler) {
 	m.restHandlers = append(m.restHandlers, handler.ServeHTTP)
 }
 
+func (m *MockServer) ResetHandlers() {
+	m.restHandlers = make([]http.HandlerFunc, 0)
```
matzew, line 63: there is similar code above when creating the server

matzew:
> Go SDK lacks a SetTools method, AddTool only upserts tools.

is this worth to call out as an issue on the repo, or was already a similar issue?

matzew:
LGTM

pkg/mcp/mcp_watch_test.go
manusa: I can try to make it clearer.
This was the easiest path I found to test the functionality.

I will probably improve the test with a custom provider to verify other behaviors such as the Provider close if exists:

https://github.com/containers/kubernetes-mcp-server/blob/6342379505a0726821c5d1420102a8969e99aadb/pkg/mcp/mcp.go#L107-L110

and so on.

Bu It don't want to spend too much time on this at the moment since it's blocking #385

manusa:
> > Go SDK lacks a SetTools method, AddTool only upserts tools.
> 
> is this worth to call out as an issue on the repo, or was already a similar issue?

Once I implement the workaround here I'll open a PR upstream so it's clear what's the problem.

I was the one who actually implemented SetTools in the mark3labs repo https://github.com/mark3labs/mcp-go/pull/24 in the first place :)

PR #469, opened 2025-11-17
chore(readme): clarify default enabled toolsets in documentation

manusa:
Follows-up on #425

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,16 +204,16 @@ Enabling only the toolsets you need can help reduce the context size and improve
 
 ### Available Toolsets
 
-The following sets of tools are available (config, core, and helm are enabled by default):
+The following sets of tools are available (toolsets marked with ✓ in the Default column are enabled by default):
 
 <!-- AVAILABLE-TOOLSETS-START -->
 
-| Toolset | Description                                                                                                                                                                                  |
-|---------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| config  | View and manage the current local Kubernetes configuration (kubeconfig)                                                                                                                      |
-| core    | Most common tools for Kubernetes management (Pods, Generic Resources, Events, etc.)                                                                                                          |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -204,21 +204,19 @@ Enabling only the toolsets you need can help reduce the context size and improve
...
+|---------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| config  | View and manage the current local Kubernetes configuration (kubeconfig)                                                                                                                      |
+| core    | Most common tools for Kubernetes management (Pods, Generic Resources, Events, etc.)                                                                                                          |
+| helm    | Tools for managing Helm charts and releases                                                                                                                                                  |
+| kiali   | Most common tools for managing Kiali, check the [Kiali integration documentation](https://github.com/containers/kubernetes-mcp-server/blob/main/docs/KIALI_INTEGRATION.md) for more details. |
```
Cali0707: @manusa can we add a column to this table about enabled/disabled by default?

Cali0707 (changes requested):

README.md
manusa: At the moment this is summarized in line 207 (managed manually).
By next version, I think we'll only enable core and config by default.
Do you still think it makes sense to add the column? (i.e. at some point there might be 10 rows and only two of them will have the check)

Cali0707: Yeah I get your point

My thinking was that it can be clearer to understand what is enabled while looking at the table (that has the descriptions), than going between the comment above and the table.

Maybe we can add to the descriptions for the enabled by default toolsets "Enabled by default" ? That way there's no extra column but it's also easy to tell what is enabled just from the table

manusa: Let me try different alternatives and see what looks better.

manusa: I let Claude work freely on this one. It seems it did a good job, please recheck

Cali0707 (approved):
LGTM, this looks great @manusa !

Issue #467, opened 2025-11-17
[TOOLSETS] Kiali tool names should have prefix


manusa:
## Description

#425 [follow-up](https://github.com/containers/kubernetes-mcp-server/pull/425#pullrequestreview-3447621487)

Toolset tool names should be prepended with the toolset name  to avoid confusion.

After the kiali toolset implementation, some tools have very similar names that will cause confusion to the LLM if are enabled simultaneously.

To avoid this issue, tools such as the ones provided by the kiali toolset should have a prefix to prevent the LLM confusion (something like `kiali_`, `istio_`, or similar should work).

/cc @aljesusg

PR #471, opened 2025-11-17
fix(kiali): Tool names with prefix kiali_

aljesusg:
Fix https://github.com/containers/kubernetes-mcp-server/issues/467

Add prefix _kiali__ to all tools related with Kiali

### DIFF
--- a/docs/KIALI.md
+++ b/docs/KIALI.md
@@ -12,7 +12,7 @@ Config (TOML):
 toolsets = ["core", "kiali"]
 
 [toolset_configs.kiali]
-url = "https://kiali.example"
+url = "https://kiali.example" # Endpoint/route to reach Kiali console
 # insecure = true  # optional: allow insecure TLS (not recommended in production)
 # certificate_authority = """-----BEGIN CERTIFICATE-----
 # MIID...


aljesusg:
cc @manusa

docs/KIALI_INTEGRATION.md
```diff
@@ -27,138 +27,6 @@ When the `kiali` toolset is enabled, a Kiali toolset configuration is required v
 - The server uses your existing Kubernetes credentials (from kubeconfig or in-cluster) to set a bearer token for Kiali calls.
 - If you pass an HTTP Authorization header to the MCP HTTP endpoint, that is not required for Kiali; Kiali calls use the server's configured token.
 
-### Available tools (initial)
```
Cali0707, line 30: @aljesusg is there a reason for getting rid of all the tool docs here?

If we want to de-duplicate the tool docs, maybe we can link to the section of the README which describes these tools from this file?

manusa: I was actually going to try that :) I wasn't aware Alberto had removed the docs already.

Let me take care of adding the link in a different PR, I want to see if I can create `a` tags with names that can be linked and expand the collapsed section or at least point the user exactly to the collapsed section summary

aljesusg: There are some duplicates in the README, so I can update the documentation script for the rest of the tools if you want and keep the methods here.
It makes more sense to include a link from the README to the toolset documentation instead of duplicating the information.

aljesusg: I did some changes in docs for toolsets out the "core"

aljesusg:
- Update Links
- And remove tools from KIALI.md in favor README.md

docs/KIALI.md
```diff
@@ -0,0 +1,37 @@
...
+
+Enable the Kiali tools via the server TOML configuration file.
+
+Config (TOML):
+
+```toml
+toolsets = ["core", "kiali"]
+
+[toolset_configs.kiali]
+url = "https://kiali.example"
```
Cali0707: @aljesusg maybe it is because I haven't used Kiali much but it's not clear to me from the docs here what value should go in this url/how I would find it.

Would you mind adding a sentence somewhere about what this value should be?

aljesusg: Of course. It’s the endpoint/route to Kiali; I’ll add a comment. I’ll also try to implement autodiscovery for this downstream.

Cali0707: thanks!

PR #472, opened 2025-11-17
feat: add AccessControl dynamicclient

Cali0707:
To make it easier for people adding toolsets to complete arbitrary actions against arbitrary resources while maintaining the access control, this creates a `dynamic.Interface` that implements our access control model.


Cali0707:
@manusa not totally sure if we want this or not, just trying to prototype some ways to make it easier for toolset authors to do what they want while enforcing the access control. Lmk what you think!

manusa:
Let's hold this one in favor of #473

PR #473, opened 2025-11-17
feat: add AccessControl rest client

Cali0707:
To make it easier for people adding toolsets to complete arbitrary actions against arbitrary resources while maintaining the access control, this creates a `rest.Interface` that is configured with a roundtripper that will ensure that only requests against GVRs that are allowed can be made.

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -1,13 +1,16 @@
 package kubernetes
 
 import (
+	"net/http"
+
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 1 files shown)


Cali0707:
@manusa not totally sure if we want this or not, just trying to prototype some ways to make it easier for toolset authors to do what they want while enforcing the access control. Lmk what you think!

manusa:
This is exactly the approach I wanted to follow: Move it up the stack as much as possible so that we don't need to check for access at the specific clients.
If all clients are using the RoundTripper or the approved REST Client, then we should be safe.

What I'm not sure is about the side-effects of this approach, neither if there are edge cases that need further consideratio.

Let me work on the base you created here and see if we can actually free ourselves from the other client implementations and fixes.

PR #474, opened 2025-11-17
fix: restmapper checks permissions on resources, not just kinds

Cali0707:
While working on #472 and #473 I realized that we don't ensure that the user is allowed to access resources, only kinds. To make our enforcement of access control more consistent I've added that here


manusa:
Let's hold this one in favor of #473

PR #475, opened 2025-11-18
feat(config)!: consolidate custom config parsing into config.Extended

### LINKED ISSUE #470: [CONFIG] Unify implementations for ProvierConfig and ToolsetConfig
## Description

#425 [follow-up](https://github.com/containers/kubernetes-mcp-server/pull/425#pullrequestreview-3447621487)

Config has now ToolsetConfig and ProviderConfig which do essentially the same.

We want to consolidate this into a single interface (e.g. `Extended`) which provides the functionality.
Analyze if there's specific functionality to any of the approaches, but initially it seems that they are doing exactly the same. If not, check what the best approach is to provide the common functionality along with the specialized one for each approach.

This will likely create breaking changes downstream. Follow up with those.

manusa:
Fixes #470 

This might imply breaking changes downstream if there are references to the interfaces.

I think the breaking changes tradeoff is OK because there is now a single simple config.Extended interface which avoids code duplication.
This will also allow to better reuse the pattern for other config extensions if needed.

### DIFF
--- a/pkg/config/extended.go
+++ b/pkg/config/extended.go
@@ -15,17 +15,17 @@ type Extended interface {
 
 type ExtendedConfigParser func(ctx context.Context, primitive toml.Primitive, md toml.MetaData) (Extended, error)
 
-type ExtendedConfigRegistry struct {
+type extendedConfigRegistry struct {
 	parsers map[string]ExtendedConfigParser
 }
 
-func NewExtendedConfigRegistry() *ExtendedConfigRegistry {
-	return &ExtendedConfigRegistry{
+func newExtendedConfigRegistry() *extendedConfigRegistry {
+	return &extendedConfigRegistry{
 		parsers: make(map[string]ExtendedConfigParser),
 	}
 }
 
-func (r *ExtendedConfigRegistry) Register(name string, parser ExtendedConfigParser) {
+func (r *extendedConfigRegistry) register(name string, parser ExtendedConfigParser) {
 	if _, exists := r.parsers[name]; exists {
 		panic("extended config parser already registered for name: " + name)
 	}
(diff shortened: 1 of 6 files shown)


pkg/config/extended.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/BurntSushi/toml"
+)
+
+// Extended is the interface that all configuration extensions must implement.
+// Each extended config manager registers a factory function to parse its config from TOML primitives
+type Extended interface {
```
Cali0707, line 12: nit: I feel like `ExtendedConfig` is a bit clearer here
```suggestion
type ExtendedConfig interface {
```

pkg/config/extended.go
```diff
@@ -0,0 +1,61 @@
...
+
+		if err = extendedConfig.Validate(); err != nil {
+			return nil, fmt.Errorf("failed to validate extended config for '%s': %w", name, err)
+		}
+
+		parsedConfigs[name] = extendedConfig
+	}
+
+	return parsedConfigs, nil
+}
```
Cali0707: @manusa what do you think about making these private? As far as I can tell, we currently expose public interfaces to these through the `RegisterProviderConfig` and `RegisterToolsetConfig` methods

pkg/config/extended.go
manusa: ExtendedConfig is explicitly marked as problematic (redundant) by **some** linters since it contains the package name, you'd usually refer to this as `config.Extended`

pkg/config/extended.go
manusa: yup, sounds good

manusa:
Hi @Cali0707 

Addressed the comments regarding ExtendedConfigRegistry and making it private.

I didn't change https://github.com/containers/kubernetes-mcp-server/pull/475#discussion_r2539506680 yet.
I don't mind changing if you still consider it unclear with regards to my comment (`config.Extended` should be clearer than `config.ExtendedConfig` IMHO).
Please, let me know about the last point and we'll proceed accordingly.

Cali0707 (approved):
LGTM, thanks for this cleanup @manusa 

I'll leave it up to you to decide when to merge this, since as you note it may break some things downstream

manusa:
Merging now, please let me know if you run into trouble when merging downstream or if you need any help with that.

Issue #476, opened 2025-11-19
[TOOLSET] Error checking Kiali when the url includes /kiali


josunect:
The toolset kiali returns 404 when Kiali is configured with /kiali. 
It appears it is not using the path /kiali in the API call, returning a 404. 

Ex. url = "http://localhost:20001/kiali" in the config.toml.
 

<img width="1915" height="764" alt="Image" src="https://github.com/user-attachments/assets/3eb13277-41ce-4139-8ab0-ae6e305e4b9d" />

npx @modelcontextprotocol/inspector@latest $(pwd)/kubernetes-mcp-server -- --config config.toml

PR #477, opened 2025-11-19
fix(kiali): Allow kiali url with path

josunect:
Fixes https://github.com/containers/kubernetes-mcp-server/issues/476 

Using the url: 

`url = "http://localhost:20001/kiali/"`

image

### DIFF
--- a/pkg/kiali/kiali.go
+++ b/pkg/kiali/kiali.go
@@ -60,25 +60,16 @@ func (k *Kiali) validateAndGetURL(endpoint string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("invalid endpoint path: %w", err)
 	}
-	// Get the base path and endpoint path
-	basePath := strings.TrimSuffix(baseURL.Path, "/")
-	endpointPath := strings.TrimPrefix(endpointURL.Path, "/")
-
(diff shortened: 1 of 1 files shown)


pkg/kiali/kiali.go
```diff
@@ -53,11 +53,32 @@ func (k *Kiali) validateAndGetURL(endpoint string) (string, error) {
...
+		fullPath = basePath + "/" + endpointPath
+	}
+
+	// Reconstruct the URL with the concatenated path, preserving query and fragment
+	resultURL := *baseURL
+	resultURL.Path = fullPath
+	resultURL.RawQuery = endpointURL.RawQuery
+	resultURL.Fragment = endpointURL.Fragment
+
+	return resultURL.String(), nil
```
Cali0707: @josunect I think we can simplify this and leverage the `url` package a bit more for this:

```suggestion
	resultURL, err := url.JoinPath(baseURL.String(), endpointURL.Path)
	if err != nil {
		return "", fmt.Errorf("failed to join kiali base URL with endpoint path: %w", err)
	}

	u, _ := url.Parse(resultURL)
	u.RawQuery = endpointURL.RawQuery
	u.Fragment = endpointURL.Fragment

	return u.String(), nil
```

PR #478, opened 2025-11-19
feat(kiali): add Kiali trace details

josunect:
Add toolset to check information about a trace: including all spans, timing details, and metadata for in-depth analysis. The trace ID can be obtained by any response from workload_traces, service_traces or app_traces.


aljesusg:
Thanks! Just to clarify — before we add any new features, we really need to focus on reducing the number of tools and checking how the MCP behaves with the current setup. So for now, we should pause on adding anything else until we finish that cleanup and performance review.

manusa:
> Thanks! Just to clarify — before we add any new features, we really need to focus on reducing the number of tools and checking how the MCP behaves with the current setup. So for now, we should pause on adding anything else until we finish that cleanup and performance review.

Should this be set on hold until we merge #496? @aljesusg @josunect

aljesusg:
Yes, it will be integrated into one of the existing tools after merge

aljesusg:
Hi @josunect . WE did the consolidation, these changes need to be applied in the tool **kiali_get_traces**

PR #482, opened 2025-11-19
feat(kubernetes): public new manager to enable external providers

mbillow:
In order to allow third parties to build their own cluster providers, I'd like to propose a very small change to make the `Manager` initialization function `newManager` public. This allows people to easily extend the upstream package without having to fork it.

Currently, we can create our own implementation of `Provider` and a register a provider configuration, but the provider implementation itself needs to be able to initialize `Manager`s for each cluster it wants to use, which is impossible since only the `NewInClusterManager` and `NewKubeconfigManager` functions are public. 

A super simple example would look like:
### pkg/custom_config.go
```go
import (
	// ...others...
	"github.com/containers/kubernetes-mcp-server/pkg/config"
)

type Config struct {
	DemoConfig string `toml:"demo"`
}

func init() {
	// Register config parser with the upstream MCP server
	config.RegisterProviderConfig(ProviderStrategy, parseConfig)
}

func parseConfig(ctx context.Context, primitive toml.Primitive, md toml.MetaData) (config.ProviderConfig, error) {
	// ...impl specific config parsing...
}
```

### pkg/custom_provider.go
```go
import (
	// ...others...
	"github.com/containers/kubernetes-mcp-server/pkg/config"
	"github.com/containers/kubernetes-mcp-server/pkg/kubernetes"
)

const ProviderStrategy = "custom"

type Provider struct {
	clusters map[string]*ClusterInfo
	managers map[string]*kubernetes.Manager
}
var _ kubernetes.Provider = &Provider{}

func init() {
	// Register this provider with the upstream MCP server
	kubernetes.RegisterProvider(ProviderStrategy, NewProvider)
}

func NewProvider(cfg *config.StaticConfig) (kubernetes.Provider, error) {
	p := &Provider{}
	p.fetchClusters()
	return p, nil
}

func (p *Provider) fetchClusters() error {
	p.clusters = company.magicClusterDiscovery()
}

func (p *Provider) managerForCluster(shortname string) (*kubernetes.Manager, error) {
	if m, ok := p.managers[shortname]; ok && m != nil {
		return m, nil
	}
	cluster, ok := p.clusters[shortname]
	if !ok {
		return nil, fmt.Errorf("cluster %s not found", shortname)
	}

	// we'd build the necessary clientcmd configs, etc. here
	// then we can call our newly public function
	m, err := kubernetes.NewManager(...)
	p.managers[shortname] = m
	return m, nil
}

// rest of the provider implementation where functions like `VerifyToken` call `p.managerForCluster`...
```

### main.go
```go
import (
	// ...others...

	// import mcp server
	"github.com/containers/kubernetes-mcp-server/pkg/kubernetes-mcp-server/cmd"

	// import our custom provider - this triggers init() which registers the provider
	_ "github.com/foo/bar/pkg/custom_provider"
)

func main() {
	flags := pflag.NewFlagSet("kubernetes-mcp-server", pflag.ExitOnError)
	pflag.CommandLine = flags

	root := cmd.NewMCPServer(genericiooptions.IOStreams{In: os.Stdin, Out: os.Stdout, ErrOut: os.Stderr})
	if err := root.Execute(); err != nil {
		os.Exit(1)
	}
}
```

### config.toml
```toml
cluster_provider_strategy = "custom"

[cluster_provider_configs.custom]
demo = "hello, world"
```

### DIFF
--- a/pkg/kubernetes/kubernetes_derived_test.go
+++ b/pkg/kubernetes/kubernetes_derived_test.go
@@ -158,7 +158,7 @@ users:
 			badConfig := test.Must(config.ReadToml([]byte(`
 				kubeconfig = "` + strings.ReplaceAll(badKubeconfigPath, `\`, `\\`) + `"
 			`)))
-			badManager, _ := newManager(badConfig, testManager.accessControlClientset.cfg, testManager.accessControlClientset.clientCmdConfig)
+			badManager, _ := NewManager(badConfig, testManager.accessControlClientset.cfg, testManager.accessControlClientset.clientCmdConfig)
 			// Replace the clientCmdConfig with one that will fail
 			pathOptions := clientcmd.NewDefaultPathOptions()
 			pathOptions.LoadingRules.ExplicitPath = badKubeconfigPath
@@ -194,7 +194,7 @@ users:
 			s.T().Cleanup(testManager.Close)
 
 			// Now create a bad manager with RequireOAuth=true
-			badManager, _ := newManager(testStaticConfig, testManager.accessControlClientset.cfg, testManager.accessControlClientset.clientCmdConfig)
+			badManager, _ := NewManager(testStaticConfig, testManager.accessControlClientset.cfg, testManager.accessControlClientset.clientCmdConfig)
 			// Replace the clientCmdConfig with one that will fail
 			pathOptions := clientcmd.NewDefaultPathOptions()
 			pathOptions.LoadingRules.ExplicitPath = badKubeconfigPath
--- a/pkg/kubernetes/manager_test.go
+++ b/pkg/kubernetes/manager_test.go
@@ -200,28 +200,28 @@ func (s *ManagerTestSuite) TestNewKubeconfigManager() {
 
 func (s *ManagerTestSuite) TestNewManager() {
 	s.Run("with nil config returns error", func() {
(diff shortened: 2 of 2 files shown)


manusa:
Hi @mbillow 
Thanks for sending this, your use case looks very interesting and aligned with our goal of making this MCP server a good base to build other toolsets on top.
Let me evaluate how does this intersect with the work I'm doing in #473 and what are other alternatives we might want to consider to add support for your use case.

mbillow:
> I think it does no harm to expose the NewManager function but it would be great if you could describe a little bit more how you're using this.

We are running the MCP server in a single cluster and want our users to be able to use it against our entire fleet. We have a central API that exposes connection info and metadata about our clusters. I wrote a `ClusterProvider` that queries this API and builds `Manager`s for each cluster. This allows us to enumerate our clusters without having to inject some kind of generated kubeconfig or something similar. The code example in the PR description provides a pretty simple example of what we’re doing.

Generally, I’m trying to do it this way to avoid having to maintain a fork of the server. This way we can just use it as a package and our repo is just the internal plugins (cluster provider, config for that, etc.).

manusa:
Thanks for sharing @mbillow :heart:

We'll keep you in the loop for any changes to the `Provider` API

PR #386, opened 2025-10-21
feat(kubevirt): Add basic VM creation toolset

lyarwood:
## Summary

This PR introduces an initial KubeVirt toolset for the kubernetes-mcp-server, enabling AI agents to create virtual machines through MCP tools.

### `vm_create`

  Provides VM creation with automatic resource resolution:
  - **Workload Matching**: Auto-resolves OS names (fedora, ubuntu, rhel, etc.) to container disks or DataSources
  - **Resource Selection**:
    - Auto-selects instancetypes based on size/performance hints
    - Auto-matches preferences based on workload names
    - Supports DataSource defaults
  - **Flexible Configuration**: Supports both cluster-scoped and namespaced instancetypes/preferences
  - **Configurable Storage**: Customizable root disk size (default: 30Gi)

### Basic VM with auto-selected resources

```json
  {
    "name": "vm_create",
    "arguments": {
      "namespace": "vms",
      "name": "my-vm",
      "workload": "fedora"
    }
  }
```

### VM with specific size and performance hints

```json
  {
    "name": "vm_create",
    "arguments": {
      "namespace": "vms",
      "name": "my-vm",
      "workload": "rhel",
      "size": "large",
      "performance": "compute-optimized",
      "storage": "100Gi"
    }
  }
```

### VM with explicit instancetype and preference

```json
  {
    "name": "vm_create",
    "arguments": {
      "namespace": "vms",
      "name": "my-vm",
      "workload": "ubuntu",
      "instancetype": "u1.medium",
      "preference": "ubuntu",
      "autostart": true
    }
  }
```

### VM with custom container disk image

```json
  {
    "name": "vm_create",
    "arguments": {
      "namespace": "vms",
      "name": "my-vm",
      "workload": "quay.io/myrepo/custom-os:v1.0",
      "size": "xlarge",
      "storage": "200Gi"
    }
  }
```

PR #483, opened 2025-11-20
feat(kubernetes)!: simplified Kubernetes client access for toolsets

manusa:
Supersedes closes #472
Supersedes closes #473
Supersedes closes #474

Built on top of the changes proposed by @Cali0707 on #473

Should simplify things for #386 and others

This was one of the original plans when implementing the denied resources (#132), but never had the chance to visit appropriately.

Thanks to the work that Calum got started in #473, I think that with the additional changes, we provide robust, resilient access to all client APIs while ensuring that denied resources are not exposed.

If needed this can be decomposed into 2 different pull requests, one for the access control and another for the Kubernetes client set simplification.


pkg/kubernetes/accesscontrol_clientset.go
```diff
@@ -3,71 +3,103 @@ package kubernetes
...
+	discoveryClient discovery.CachedDiscoveryInterface
+	dynamicClient   *dynamic.DynamicClient
+	restMapper      meta.ResettableRESTMapper
 	metricsV1beta1  *metricsv1beta1.MetricsV1beta1Client
-	staticConfig    *config.StaticConfig // TODO: maybe just store the denied resource slice
 }
 
-func (a *AccessControlClientset) DiscoveryClient() discovery.DiscoveryInterface {
+func NewAccessControlClientset(staticConfig *config.StaticConfig, restConfig *rest.Config) (*AccessControlClientset, error) {
+	configCopy := *restConfig
```
Cali0707: Should we use `rest.CopyConfig` here? https://pkg.go.dev/k8s.io/client-go/rest#CopyConfig

pkg/kubernetes/accesscontrol_clientset.go
```diff
@@ -3,71 +3,103 @@ package kubernetes
...
 func (a *AccessControlClientset) Nodes() (corev1.NodeInterface, error) {
-	gvk := &schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Node"}
-	if !isAllowed(a.staticConfig, gvk) {
-		return nil, isNotAllowedError(gvk)
-	}
-	return a.delegate.CoreV1().Nodes(), nil
+	return a.CoreV1().Nodes(), nil
 }
 
 func (a *AccessControlClientset) NodesLogs(ctx context.Context, name string) (*rest.Request, error) {
```
Cali0707: Is there a reason for not deprecating this one as well? I get that it has a specific `AbsPath` but IMO that's fine to be set in a toolset...

pkg/kubernetes/accesscontrol_clientset.go
```diff
@@ -3,71 +3,103 @@ package kubernetes
...
 	}
 
 	url := []string{"api", "v1", "nodes", name, "proxy", "logs"}
-	return a.delegate.CoreV1().RESTClient().
+	return a.CoreV1().RESTClient().
 		Get().
 		AbsPath(url...), nil
 }
 
 func (a *AccessControlClientset) NodesMetricses(ctx context.Context, name string, listOptions metav1.ListOptions) (*metrics.NodeMetricsList, error) {
```
Cali0707: Is there a reason for not deprecating this one as well?

pkg/kubernetes/accesscontrol_clientset.go
```diff
@@ -87,37 +119,26 @@ func (a *AccessControlClientset) NodesMetricses(ctx context.Context, name string
 }
 
 func (a *AccessControlClientset) NodesStatsSummary(ctx context.Context, name string) (*rest.Request, error) {
```
Cali0707: Is there a reason for not deprecating this one as well?

pkg/kubernetes/accesscontrol_clientset.go
manusa: good catch!

pkg/kubernetes/accesscontrol_clientset.go
manusa: These ones perform additional logic.
I assume we are the only ones using it, we can simply move the extra-logic to the consumer method at the Kubernetes struct level

Cali0707: that makes sense

pkg/kubernetes/accesscontrol_clientset.go
```diff
@@ -3,71 +3,106 @@ package kubernetes
...
 // Only a limited set of functions are implemented with a single point of access to the kubernetes API where
 // apiVersion and kinds are checked for allowed access
 type AccessControlClientset struct {
-	cfg             *rest.Config
-	delegate        kubernetes.Interface
-	discoveryClient discovery.DiscoveryInterface
+	cfg *rest.Config
+	kubernetes.Interface
+	discoveryClient discovery.CachedDiscoveryInterface
+	dynamicClient   *dynamic.DynamicClient
```
lyarwood: What about using `dynamic.Interface` to allow mocking with `k8s.io/client-go/dynamic/fake` in tests?

manusa: yup, good catch.
Wanted to switch to interfaces as much as possible, this one remained.

_It won't be me who uses mocks for that though._

pkg/kubernetes/pods.go
```diff
@@ -22,7 +23,7 @@ import (
 	"github.com/containers/kubernetes-mcp-server/pkg/version"
 )
 
-// Default number of lines to retrieve from the end of the logs
+// DefaultTailLines default number of lines to retrieve from the end of the logs
```
Cali0707: nit: `DefaultTailLines is the default number of lines ...`

Cali0707 (commented):
This is looking really good @manusa!

pkg/kubernetes/pods.go
manusa: fixed, thx

manusa:
Took a little longer than expected but I think that this should be ready now.
I'm planning on some follow-up changes to further simplify the `kubernetes.Manager` and related, but it'll be better to defer those to further PRs.

If merging the PR, do Rebase and merge as there are two commits I want to preserve in isolation in case we need to rollback.

PR #490, opened 2025-11-21
[WIP] feat: add IsValid check for toolsets

Cali0707:
This PR allows us to dynamically enable/disable toolsets, as well as the targets within tools, based on whether or not prerequisites for the toolset are installed in a given target cluster.

That way, users can enable e.g. `kiali` toolset, and if it is only installed in 3/4 clusters, we will only hint targets for those 3 clusters to the MCP client. Similarly, if none of the clusters have `kiali` installed, we will simply disable the toolset and not present it to any clients.

### DIFF
--- a/pkg/toolsets/kiali/toolset.go
+++ b/pkg/toolsets/kiali/toolset.go
@@ -1,6 +1,7 @@
 package kiali
 
 import (
+	"context"
 	"slices"
 
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
@@ -37,6 +38,19 @@ func (t *Toolset) GetTools(_ internalk8s.Openshift) []api.ServerTool {
 		initLogs(),
 		initTraces(),
 	)
+}
+
+func (t *Toolset) IsValid(k *internalk8s.Kubernetes) bool {
+	// Create a Kiali client
+	kialiClient := k.NewKiali()
+
+	// Check if Kiali is actually accessible by making a lightweight API call
+	// We'll try to get the mesh status as it's a simple endpoint that doesn't require parameters
+	ctx := context.Background()
+	_, err := kialiClient.MeshStatus(ctx)
+
+	// If we can successfully call Kiali, it's valid
+	return err == nil
 }
 
 func init() {
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -44,6 +44,9 @@ type Toolset interface {
 	// Will be used to generate documentation and help text.
 	GetDescription() string
 	GetTools(o internalk8s.Openshift) []ServerTool
+	// IsValid returns whether or not the toolset is valid on the current cluster
+	// Used to identify if toolsets should be disabled despite the config based on what is installed in the cluster
+	IsValid(k *internalk8s.Kubernetes) bool
 }
 
 type ToolCallRequest interface {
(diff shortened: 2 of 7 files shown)


pkg/toolsets/kiali/toolset.go
```diff
@@ -39,6 +40,19 @@ func (t *Toolset) GetTools(_ internalk8s.Openshift) []api.ServerTool {
 	)
 }
 
+func (t *Toolset) IsValid(k *internalk8s.Kubernetes) bool {
```
Cali0707, line 43: @aljesusg when you have time can you take a look at this method and let me know if this makes sense?

It's supposed to validate that there is a working kiali installation in the target cluster

Cali0707:
Will add tests to this next, then it should be good to merge (if people like the approach here)

pkg/toolsets/kiali/toolset.go
aljesusg: Yes… although I would do the following: since the AuthInfoEndpoint endpoint is already created, I would create a method to call it, as it doesn’t require any parameters and is ‘available’ whether the user is authenticated or not.

manusa:
I believe that this entire functionality requires more thought.

This is similar to what's already done in:
https://github.com/containers/kubernetes-mcp-server/blob/0e889354fdc39331bc308141979d258813d20823/pkg/api/toolsets.go#L46

and related methods/functons.

> [!NOTE]
> This was part of the first spike of the MCP server hence it doesn't make sense in the current state of the cluster.

In the **early stages** of the Kubernetes MCP server this was just enough:
- The MCP server was supposed to connect to a single cluster
- The MCP server ran with a _root_ user that could access most cluster resources (allow for inference of an OpenShift cluster)
- The eventually authorized user might have a subset of the available cluster resources but that was fine since the listed tools were tailored for the vanilla Kubernetes or OpenShift experience based on the cluster-wide accessible resources discovered by the root user.

In the **current state** (multi-cluster + auth support), when a user/client performs a list request (tools, prompts, and so on), the MCP server **SHOULD** account for everything that applies to that user in its specific scenario:
- Accessible clusters (might be of different flavors)
- API extensions/Operators available in each cluster (each cluster might have a different set of operators installed)
- APIs accessible by the user (each user might have different RBAC settings that prevent them from accessing a given set of resources)

I'm not convinced about the idea of completely enabling or disabling an **entire toolset** (as this PR proposes).
Consider the `core` toolset which has tools to retrieve metrics, ideally, we should be able to detect if the metrics server is available, and then present the tool to the user only if the resource is in the cluster.

One of the challenges to do this properly is that go-sdk (or any of the other SDKs) require the tools to be loaded at start-up and don't provide callbacks to provide the tool or resource list dynamically or per session/user (AFAIK, maybe this has changed).

In this PR we're getting the derived Kubernetes when reloading the tools, but this will likely fail if there's no global auth for that cluster.

I'm really not sure how the problem can be properly tackled. Maybe checking the internals of the SDK and seeing if the list/tools request can be intercepted or something along those lines.

matzew:
@Cali0707 wanna still do this?

pkg/toolsets/kiali/toolset.go
aljesusg: Furthermore, Kiali supports multicluster deployments, allowing a single Kiali endpoint to communicate with instances in other clusters. However, I am still evaluating whether this specific architecture will meet our requirements.


In some tools, we have a clusterName attribute that defaults to 'local' if it isn't specified by the LLM

Cali0707:
Closing this out, we may want to revisit the problem of how a model can understand which cluster targets are valid for which tools in the future though

cc @manusa

PR #493, opened 2025-11-21
feat(kubernetes): add cluster state monitoring and drop-in config support and SIGHUP reload

nader-ziada:
This PR covers adding Dynamic MCP Server Reload

the first part is to add cluster state monitoring with debounced reload, 
Implement automatic detection of cluster state changes (API groups, OpenShift status) with configurable polling and debounce windows. The cluster state watcher runs in the background, invalidates the discovery cache periodically, and triggers a reload callback when changes are detected after a debounce period.

the second part is to add drop-in configuration directory support with lexical ordering and implement SIGHUP signal handler for live config reloads. Configuration files are merged in order from defaults to main config to drop-ins, with a new --config-dir flag available for specifying the drop-in director


manusa:
This is looking great.
However, I'd like to add a few changes to the first commit to further decouple things from the mcp domain.

would you mind creating a separate PR with only the first commit and we can work on the second part later?

nader-ziada:
> This PR covers adding Dynamic MCP Server Reload
> 
> the first part is to add cluster state monitoring with debounced reload, Implement automatic detection of cluster state changes (API groups, OpenShift status) with configurable polling and debounce windows. The cluster state watcher runs in the background, invalidates the discovery cache periodically, and triggers a reload callback when changes are detected after a debounce period.

thanks for taking a look @manusa 

created a new PR with the first commit only and will close this one #502

PR #496, opened 2025-11-24
feat(kiali): Consolidate kiali tools from 20 to 6

aljesusg:
Consolidate Kiali Tools reduncy the number of tools from 20 to 6

### DIFF
--- a/pkg/kiali/get_mesh_graph.go
+++ b/pkg/kiali/get_mesh_graph.go
@@ -34,6 +34,7 @@ func (k *Kiali) GetMeshGraph(ctx context.Context, namespaces []string, queryPara
 		Errors: make(map[string]string),
 	}
 
+	var errorsMu sync.Mutex
 	var wg sync.WaitGroup
 	wg.Add(4)
 
@@ -42,7 +43,9 @@ func (k *Kiali) GetMeshGraph(ctx context.Context, namespaces []string, queryPara
 		defer wg.Done()
 		data, err := k.getGraph(ctx, cleaned, queryParams)
 		if err != nil {
+			errorsMu.Lock()
 			resp.Errors["graph"] = err.Error()
+			errorsMu.Unlock()
 			return
 		}
 		resp.Graph = data
@@ -53,7 +56,9 @@ func (k *Kiali) GetMeshGraph(ctx context.Context, namespaces []string, queryPara
 		defer wg.Done()
 		data, err := k.getHealth(ctx, cleaned, queryParams)
 		if err != nil {
+			errorsMu.Lock()
 			resp.Errors["health"] = err.Error()
+			errorsMu.Unlock()
 			return
 		}
 		resp.Health = data
@@ -64,7 +69,9 @@ func (k *Kiali) GetMeshGraph(ctx context.Context, namespaces []string, queryPara
 		defer wg.Done()
 		data, err := k.getMeshStatus(ctx)
 		if err != nil {
+			errorsMu.Lock()
 			resp.Errors["mesh_status"] = err.Error()
+			errorsMu.Unlock()
 			return
 		}
 		resp.MeshStatus = data
@@ -75,7 +82,9 @@ func (k *Kiali) GetMeshGraph(ctx context.Context, namespaces []string, queryPara
 		defer wg.Done()
 		data, err := k.getNamespaces(ctx)
 		if err != nil {
+			errorsMu.Lock()
 			resp.Errors["namespaces"] = err.Error()
+			errorsMu.Unlock()
 			return
 		}
 		resp.Namespaces = data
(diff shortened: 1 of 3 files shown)


aljesusg:
@manusa I did the reduction of tools from 20 to 6. I tested then with gevals and it's working with all use cases.

pkg/toolsets/kiali/manage_istio_config.go
```diff
@@ -0,0 +1,117 @@
...
+						Type:        "string",
+						Description: "JSON data to apply or create the object",
+					},
+				},
+				Required: []string{"action"},
+			},
+			Annotations: api.ToolAnnotations{
+				Title:           "Manage Istio Config: List, Get, Create, Patch, Delete",
+				ReadOnlyHint:    ptr.To(true),
+				DestructiveHint: ptr.To(false),
```
manusa:
```suggestion
				ReadOnlyHint:    ptr.To(false),
				DestructiveHint: ptr.To(true),
```

After changing the purpose of the tool, the annotations are now incorrect.

pkg/toolsets/kiali/manage_istio_config.go
```diff
@@ -0,0 +1,117 @@
...
+				IdempotentHint:  ptr.To(true),
```
manusa, line 56: I'm not sure about this one either, I understand that this should be false now, but I'm not 100% sure.

pkg/kiali/get_mesh_graph.go
```diff
@@ -0,0 +1,133 @@
...
+	"strings"
+	"sync"
+)
+
+type GetMeshGraphResponse struct {
+	Graph      json.RawMessage   `json:"graph,omitempty"`
+	Health     json.RawMessage   `json:"health,omitempty"`
+	MeshStatus json.RawMessage   `json:"mesh_status,omitempty"`
+	Namespaces json.RawMessage   `json:"namespaces,omitempty"`
+	Errors     map[string]string `json:"errors,omitempty"`
```
manusa, line 15: Since this is performing concurrent logic, I believe we should add a an `errorsMu` lock to prevent race conditions.

pkg/toolsets/kiali/get_mesh_graph.go
```diff
@@ -77,8 +85,17 @@ func graphHandler(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
 		}
 		namespaces = unique
 	}
+
+	// Extract optional query parameters
+	queryParams := make(map[string]string)
+	if rateInterval, ok := params.GetArguments()["rateInterval"].(string); ok && rateInterval != "" {
+		queryParams["rateInterval"] = rateInterval
+	}
+	if graphType, ok := params.GetArguments()["graph_type"].(string); ok && graphType != "" {
```
manusa:
```suggestion
	if graphType, ok := params.GetArguments()["graphType"].(string); ok && graphType != "" {
```

pkg/toolsets/kiali/manage_istio_config.go
```diff
@@ -0,0 +1,117 @@
...
+		if jsonData == "" {
+			return fmt.Errorf("json_data is required for action %q", action)
+		}
+	}
+	if action == "patch" {
+		if jsonData == "" {
+			return fmt.Errorf("json_data is required for action %q", action)
+		}
+	}
+	return nil
```
manusa, line 133: Should we check if action value is one of list, get, create, patch, delete?

pkg/kiali/get_mesh_graph.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// GetMeshGraph fetches multiple Kiali endpoints in parallel and returns a combined response.
+// Each field in the response corresponds to one API call result.
+// - graph:       /api/namespaces/graph (optionally filtered by namespaces)
+// - health:      /api/clusters/health  (optionally filtered by namespaces and queryParams)
+// - status(mesh):/api/mesh/graph
+// - namespaces:  /api/namespaces
+func (k *Kiali) GetMeshGraph(ctx context.Context, namespaces []string, queryParams map[string]string) (string, error) {
```
manusa, line 24: Note, this is a good candidate to try out MCP progress notifications (Not in this PR)

pkg/toolsets/kiali/manage_istio_config.go
aljesusg: Yes, It's destructive thanks ! Good catch

pkg/toolsets/kiali/manage_istio_config.go
aljesusg: No, It's true due.

If the option to create the object is launched again, it doesn’t have any effect because it’s already created.
If you try to delete an object again, it also has no effect because it no longer exists.
And if you perform a patch, since they are Kubernetes objects, it won’t update unless there is an actual modification.

pkg/kiali/get_mesh_graph.go
aljesusg: Totally

pkg/toolsets/kiali/manage_istio_config.go
aljesusg: Well, I was checking that the arguments required for a specific action are there. The thing here is taht I forgot to add the fields in required. Let me fix it. Thanks !

pkg/toolsets/kiali/manage_istio_config.go
manusa: Sounds right.
Since I don't know the internals of your API I wasn't sure how RESTful it was :)

PR #502, opened 2025-11-25
feat(kubernetes): add cluster state monitoring with reload

nader-ziada:
Implement automatic detection of cluster state changes (API groups, OpenShift status) with configurable polling and debounce windows. The cluster state watcher runs in the background, invalidates the discovery cache periodically, and triggers a reload callback when changes are detected after a debounce period.

- Add clusterStateWatcher to monitor API groups and OpenShift status
- Implement debounced reload to avoid excessive reloads during changes
- Add WatchClusterState method to Manager with configurable intervals
- Integrate cluster state watching in kubeconfig and single cluster providers

closed PR #493 to split into this PR and future follow up

### DIFF
--- a/pkg/kubernetes/manager.go
+++ b/pkg/kubernetes/manager.go
@@ -5,8 +5,11 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
+	"sync"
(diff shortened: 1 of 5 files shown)


nader-ziada:
@manusa

PR #503, opened 2025-11-25
refactor(kubernetes): enhance AccessControlClientset and streamline Manager and Kubernetes structs

manusa:
Simplifies the relationship between the `Manager`, `Kubernetes`, and `AccessControlClientset` types.
The changes reduce indirection and move functionality to more appropriate locations.

`Kubernetes` no longer wraps a `Manager` which could be confusing especially for the derived scenarios.

`Manger` should only concern with the lifecycle management and the derived instance creation (wip).
All access to the cluster should be done from a derived `Kubernetes` instance (wip).


pkg/kubernetes/kubernetes.go
```diff
@@ -20,26 +24,52 @@ const (
...
+}
+
+func (k *Kubernetes) ToRESTMapper() (meta.RESTMapper, error) {
+	return k.AccessControlClientset().RESTMapper(), nil
+}
+
+// ToRESTConfig returns the rest.Config object (genericclioptions.RESTClientGetter)
+func (k *Kubernetes) ToRESTConfig() (*rest.Config, error) {
+	return k.AccessControlClientset().cfg, nil
+}
```
Cali0707, line 71: @manusa do we want to keep the `error` in the return types here? I understand changing that would be a breaking change downstream, but I think long term it could simplify our code somewhat if we don't anticipate changes re-introducing the error value here

pkg/kubernetes/manager.go
```diff
@@ -216,50 +184,44 @@ func (m *Manager) Derived(ctx context.Context) (*Kubernetes, error) {
...
-		manager: &Manager{
-			clientCmdConfig: clientcmd.NewDefaultClientConfig(clientCmdApiConfig, nil),
-			staticConfig:    m.staticConfig,
-		},
-	}
-	derived.manager.accessControlClientSet, err = NewAccessControlClientset(derived.manager.staticConfig, derivedCfg)
+	derived, err := NewAccessControlClientset(m.staticConfig, clientcmd.NewDefaultClientConfig(clientCmdApiConfig, nil), derivedCfg)
 	if err != nil {
 		if m.staticConfig.RequireOAuth {
 			klog.Errorf("failed to get kubeconfig: %v", err)
```
Cali0707: @manusa is this error message still accurate?

pkg/kubernetes/kubernetes_derived_test.go
```diff
@@ -119,17 +120,140 @@ users:
...
+			pathOptions.LoadingRules.ExplicitPath = badKubeconfigPath
+			badClientCmdConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+				pathOptions.LoadingRules,
+				&clientcmd.ConfigOverrides{})
+			badManager.accessControlClientset.clientCmdConfig = badClientCmdConfig
+
+			ctx := context.WithValue(s.T().Context(), HeaderKey("Authorization"), "Bearer aiTana-julIA")
+			derived, err := badManager.Derived(ctx)
+			s.Require().NoErrorf(err, "expected no error when RequireOAuth=false, got: %v", err)
+			s.Equal(derived.accessControlClientSet, badManager.accessControlClientset, "expected original clientset when RawConfig fails and RequireOAuth=false")
```
Cali0707: nit: use the getter method here
```suggestion
			s.Equal(derived.AccessControlClientset(), badManager.AccessControlClientset(), "expected original clientset when RawConfig fails and RequireOAuth=false")
```

pkg/kubernetes/kubernetes.go
manusa: These methods are the implementation of the [`genericclioptions.RESTClientGetter`](https://pkg.go.dev/k8s.io/cli-runtime/pkg/genericclioptions#RESTClientGetter) (note they basically duplicate the ones already available through the AccessControlClientset).

These were before at the Manager level (here I'm trying to trim off any access to Kubernetes)

It's my understanding that other clients, besides helm, might end up using this interface. That's the main reasoning to having them here.

pkg/kubernetes/manager.go
manusa: Good catch, fixed

pkg/kubernetes/kubernetes_derived_test.go
manusa: Much better, changed. Note that the manager doesn't have a direct accessor to the accessControlClientSet.

pkg/kubernetes/kubernetes.go
Cali0707: cool, I didn't know about that interface - thanks for sharing!

PR #505, opened 2025-11-25
feat: add gevals action to github workflows

Cali0707:
This PR adds the gevals action to run in github workflows. There are a few ways to trigger this:
1. Manually
2. Through a comment on a PR (`/run-gevals`)
3. It will run periodically once per week

I will do future follow ups to allow us to:
1. Specify which model to run
2. Run a matrix of models/agents in the weekly CI
3. Cleanup some of the start/configure server logic

### DIFF
--- a/.github/workflows/gevals.yaml
+++ b/.github/workflows/gevals.yaml
@@ -1,0 +1,155 @@
+name: Gevals MCP Evaluation
+
+on:
+  # Weekly schedule - runs every Monday at 9 AM UTC
+  schedule:
+    - cron: '0 9 * * 1'
+
+  # Manual trigger via PR comments
+  issue_comment:
+    types: [created]
+
+  # Allow manual workflow dispatch for testing
+  workflow_dispatch:
+    inputs:
+      task-filter:
+        description: 'Regular expression to filter tasks (optional)'
+        required: false
+        default: ''
+      verbose:
+        description: 'Enable verbose output'
+        required: false
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
+
+concurrency:
+  # Only run once for latest commit per ref and cancel other (previous) runs.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  GO_VERSION: 1.25
+  KIND_CLUSTER_NAME: mcp-eval-cluster
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  # Check if workflow should run based on trigger
+  check-trigger:
+    name: Check if evaluation should run
+    runs-on: ubuntu-latest
+    if: |
+      github.event_name == 'schedule' ||
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'issue_comment' &&
+       github.event.issue.pull_request &&
+       contains(github.event.comment.body, '/run-gevals'))
+    outputs:
+      should-run: ${{ steps.check.outputs.should-run }}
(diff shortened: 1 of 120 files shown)


Cali0707:
@manusa this also includes all the eval scenarios we ported from kubectl-ai. I realize that makes this PR quite large, so if you prefer I can split this one up

Cali0707:
@manusa @nader-ziada a sample run (which failed, due to not enough tasks succeeding to pass the threshold) can be seen here: https://github.com/Cali0707/kubernetes-mcp-server/actions/runs/19628912129

Cali0707:
Note that this will need a few repository secrets to be set before this can be merged:
- `JUDGE_API_KEY` & `MODEL_KEY`: these are api keys to an openai compatible model
- `JUDGE_BASE_URL` & `MODEL_BASE_URL`: these are the base urls to the models
- `JUDGE_MODEL_NAME` & `MODEL_NAME`: these are the names of the models that we will use for now

As we move to this running a matrix of models and/or having maintainers trigger this with specific models in comments, we will need more secrets and need to figure out naming conventions for those

evals/README.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+Uses the built-in OpenAI agent with model configuration.
+
+## Expected Results
+
+Both examples should produce:
+- ✅ Task passed - pod created successfully
+- ✅ Assertions passed - appropriate tools were called
+- ✅ Verification passed - pod exists and is running
```
nader-ziada, line 108: are there more results captured?

Cali0707: Yes, overtime we are working to expand the top level info exported. For now, this is what we show (along with a full trace of all the MCP messages)

For now, this is all the info shared at a top level

.github/workflows/gevals.yaml
```diff
@@ -0,0 +1,155 @@
...
+  # Check if workflow should run based on trigger
+  check-trigger:
+    name: Check if evaluation should run
+    runs-on: ubuntu-latest
+    if: |
+      github.event_name == 'schedule' ||
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'issue_comment' &&
+       github.event.issue.pull_request &&
+       contains(github.event.comment.body, '/run-gevals'))
```
manusa, line 53: This is the only part I'm worried about.
Since each run incurs a cost, I'm not sure we want to leave this open to any random internet commenter.
Not sure if it's possible to limit this in an easy way to just the project maintainers.

build/gevals.mk
```diff
@@ -0,0 +1,35 @@
+# Gevals evaluation support
```
manusa, line 1: The targets in this file seem more generic than just for the gevals scenario.
Maybe we want to rename the mk file to something else.

.github/workflows/gevals.yaml
```diff
@@ -0,0 +1,155 @@
...
+      pr-number: ${{ steps.check.outputs.pr-number }}
+      pr-ref: ${{ steps.check.outputs.pr-ref }}
+    steps:
+      - name: Check trigger conditions
+        id: check
+        run: |
+          if [[ "${{ github.event_name }}" == "issue_comment" ]]; then
+            # Check if commenter is a maintainer (has write access)
+            PERMISSION=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+              "https://api.github.com/repos/${{ github.repository }}/collaborators/${{ github.event.comment.user.login }}/permission" \
```
Cali0707, line 65: @manusa here is where we check if the person has write access to help address the concern

manusa: oh, I missed that. Nice!

PR #507, opened 2025-11-26
feat(kiali): Possibility of renaming Kiali in the downstream distribution

aljesusg:
Downstream we will add the following override in `pkg/config/config_default_overrides.go`:
```go
ToolsetKialiName: "ossm"
```
This allows the toolset to return `"ossm"` from `GetName()` instead of `"kiali"`.
think this approach works better than passing `internalk8s.Openshift` into `GetName()` the way `GetTools()` does, since it keeps things simpler, requires fewer modifications, and is overall lighter.

Also, each tool already checks whether the environment is **OpenShift** (following the same pattern as the `core` toolset). We expect users to run OpenShift with OSSM, not upstream Istio, so this behavior is consistent with that assumption.

Screenshot From 2025-11-26 11-14-04

cc @manusa WDYT?

### DIFF
--- a/pkg/mcp/testdata/toolsets-kiali-tools.json
+++ b/pkg/mcp/testdata/toolsets-kiali-tools.json
@@ -1,39 +1,6 @@
 [
   {
     "annotations": {
-      "title": "Topology: Mesh, Graph, Health, and Status",
-      "readOnlyHint": true,
-      "destructiveHint": false,
-      "openWorldHint": true
-    },
-    "description": "Returns the topology of a specific namespaces, health, status of the mesh and namespaces. Includes a mesh health summary overview with aggregated counts of healthy, degraded, and failing apps, workloads, and services. Use this for high-level overviews",
-    "inputSchema": {
-      "type": "object",
-      "properties": {
-        "graphType": {
-          "default": "versionedApp",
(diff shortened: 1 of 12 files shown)


aljesusg:
Mmm I need to add some tests to verify that naming is not going to affect to the toml configuration

Ok the other hand I am thinking to use the config to tools naming too.

manusa:
I'm not sure that config is the right place to put this, it also complicates things a little.
Additionally, it provides the "feature" for users to configure the Kiali toolset prefix which is something we might not want to support.

I'm not sure if you want/can to follow a similar pattern like for config but scoped in your toolset:

- `pkg/toolsets/kiali/toolset.go`
- `pkg/toolsets/kiali/toolset_default.go`
- `pkg/toolsets/kiali/toolset_overrides.go`

Also, in relation to #521, all defaults could be moved here.

With regards to downstream.
Ideally, if we manage to create a sufficiently decoupled upstream project, the downstream merge scripts could be created to always accept the upstream changes except for the files that only exist downstream or the xxx_overrides.go (downstream preferred).

/cc @Cali0707 @nader-ziada

aljesusg:
Done! Thanks !

aljesusg:
Thanks ! I'll make changes downstream after sync

PR #511, opened 2025-11-26
feat(kiali): Standardize Certificate Authority Configuration Method

josunect:
- Use a CA file path in addition to the inline certificate.
- Document the standardized pattern for future RH toolsets.

This PR keeps the inline option for back-comp, but I can remove it if it is not required. 

How to manually test it: 

`echo | openssl s_client -showcerts -connect kiali-istio-system.apps-crc.testing:443 2>/dev/null | openssl x509 > /tmp/kiali.crt`

Configuration (Using a kiali https route): 
```
url = "https://kiali-istio-system.apps-crc.testing/"
certificate_authority = "/tmp/kiali.crt"
insecure = false
```

Run the MCP server: 

```
make build
npx @modelcontextprotocol/inspector@latest $(pwd)/kubernetes-mcp-server
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -96,7 +96,6 @@ func withDirPath(path string) ReadConfigOpt {
 }
 
 // WithDirPath returns a ReadConfigOpt that sets the config directory path.
-// Used by tests that need to resolve relative file paths.
 func WithDirPath(path string) ReadConfigOpt {
 	return withDirPath(path)
 }


Cali0707:
> This PR keeps the inline option for back-comp, but I can remove it if it is not required.

@josunect I think it should be safe for you to remove this - we haven't cut a release with kiali toolset yet, so I don't think we need to maintain backwards compatability yet (cc @manusa )

pkg/kiali/config_test.go
```diff
@@ -0,0 +1,130 @@
...
+[toolset_configs.kiali]
+url = "https://kiali.example/"
+certificate_authority = "` + caFileForTOML + `"
+`
+	err := os.WriteFile(configFile, []byte(configContent), 0644)
+	s.Require().NoError(err, "Failed to write config file")
+
+	// Read config - Read() automatically sets the config directory path
+	cfg, err := config.Read(configFile)
+	s.Require().NoError(err, "Failed to read config")
```
manusa: Can't this be replaced with the logic we have in place for other tests e.g.

https://github.com/containers/kubernetes-mcp-server/blob/8d7642698400221d3392def1345f99920a377d67/pkg/kiali/kiali_test.go#L31-L35

josunect: Updated!

aljesusg:
> > This PR keeps the inline option for back-comp, but I can remove it if it is not required.
> 
> @josunect I think it should be safe for you to remove this - we haven't cut a release with kiali toolset yet, so I don't think we need to maintain backwards compatability yet (cc @manusa )

Yes please, The idea is to replace the inline CA by path

manusa:
@aljesusg @josunect I still se the support for inline PEM certificates.
I'm not sure if this is going to be removed in the scope of this PR or in a follow-up.

aljesusg:
> @aljesusg @josunect I still se the support for inline PEM certificates. I'm not sure if this is going to be removed in the scope of this PR or in a follow-up.

The idea is replace it, so yes we are going to remove it in this PR

docs/KIALI.md
```diff
@@ -32,6 +30,6 @@ When the `kiali` toolset is enabled, a Kiali toolset configuration is required v
 - Missing Kiali configuration when `kiali` toolset is enabled → set `[toolset_configs.kiali].url` in the config TOML.
 - Invalid URL → ensure `[toolset_configs.kiali].url` is a valid `http(s)://host` URL.
 - TLS certificate validation:
-  - If `[toolset_configs.kiali].url` uses HTTPS and `[toolset_configs.kiali].insecure` is false, you must set `[toolset_configs.kiali].certificate_authority` with the PEM-encoded certificate(s) used by the Kiali server. This field expects inline PEM content, not a file path. You may concatenate multiple PEM blocks to include an intermediate chain.
+  - If `[toolset_configs.kiali].url` uses HTTPS and `[toolset_configs.kiali].insecure` is false, you must set `[toolset_configs.kiali].certificate_authority` with the path to the CA certificate file. Relative paths are resolved relative to the directory containing the config file. Only file paths are supported; inline PEM content is not allowed.
```
aljesusg:
```suggestion
  - If `[toolset_configs.kiali].url` uses HTTPS and `[toolset_configs.kiali].insecure` is false, you must set `[toolset_configs.kiali].certificate_authority` with the path to the CA certificate file. Relative paths are resolved relative to the directory containing the config file.
```

pkg/config/config.go
```diff
@@ -95,6 +95,13 @@ func withDirPath(path string) ReadConfigOpt {
 	}
 }
 
+// WithDirPath returns a ReadConfigOpt that sets the config directory path.
+// Used by tests that need to resolve relative file paths.
+func WithDirPath(path string) ReadConfigOpt {
+
+	return withDirPath(path)
+}
```
aljesusg, line 101:
```suggestion
func WithDirPath(path string) ReadConfigOpt {
	return withDirPath(path)
}
```

aljesusg (changes requested):
Look good only this 2 minor things, I am going to test it locally

PR #516, opened 2025-11-27
test(mcp): use envtest built-in features to apply CRDs

manusa:
Trying to speed up the test cases that require additional
kubernetes APIs.

Each time a test suite or test case requires an additional
API (like OpenShift Routes or Kiali CRDs), provided by
k8s.io/apiextensions, a fixed 2 seconds delay is introduced.

There's no way to reduce that delay by using envtest features
or APIServer CLI flags.

The delays are introduced by kube-aggregator, apiextensions, and others:

https://github.com/kubernetes/kube-aggregator/blob/a00232cf7e758c2771b3542033bec19e69e6451a/pkg/controllers/openapiv3/controller.go#L34

https://github.com/kubernetes/apiextensions-apiserver/blob/8db5ab628dd026827c1c9677944432db70c065c3/pkg/apiserver/customresource_handler.go#L381-L394

To mitigate this, we're going to declare all CRDs used in the test in
the common_test.go TestMain function.

This will make sure we only have a single delay at the beginning of
the entire test suite instead of per test suite or test case.

The CRDs are NOT served by default.

For each test suite we'll be able to enable them or disable them
by using the EnvTestEnableCRD and EnvTestDisableCRD functions.
These will also ensure that the API server lists the new APIs or not.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,6 @@ require (
 	github.com/google/jsonschema-go v0.3.0
 	github.com/mark3labs/mcp-go v0.43.1
 	github.com/modelcontextprotocol/go-sdk v1.1.0
-	github.com/pkg/errors v0.9.1
 	github.com/spf13/afero v1.15.0
 	github.com/spf13/cobra v1.10.1
 	github.com/spf13/pflag v1.0.10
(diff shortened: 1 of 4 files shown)


manusa:
Merging, I need this as a prereq to complete  #386 test improvements.

PR #518, opened 2025-11-27
chore: switch from gpt-4 to gpt-5 for default evals agent

Cali0707:
I noticed that with the gevals change to builtin agents, we are now declaring the model manually in the eval spec. This updates that model from gpt-4 to gpt-5

### DIFF
--- a/.github/workflows/gevals.yaml
+++ b/.github/workflows/gevals.yaml
@@ -120,11 +120,10 @@ jobs:
           # OpenAI Agent configuration
           MODEL_BASE_URL: ${{ secrets.MODEL_BASE_URL }}
           MODEL_KEY: ${{ secrets.MODEL_KEY }}
-          MODEL_NAME: ${{ secrets.MODEL_NAME }}
           # LLM Judge configuration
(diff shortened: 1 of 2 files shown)


nader-ziada:
can you update the `evals/README.md` as well

Cali0707:
> Looks good, however, I'm not sure how:
> 
> https://github.com/containers/kubernetes-mcp-server/blob/130e42c73b6d296f8b5f94b409d923911d9a7a95/.github/workflows/gevals.yaml#L123
> 
> https://github.com/containers/kubernetes-mcp-server/blob/130e42c73b6d296f8b5f94b409d923911d9a7a95/.github/workflows/gevals.yaml#L127
> 
> relate to this value

I think we should be good to remove them now. I hadn't realized that part of the scope of changes related to the builtin models was to specify the model name through the yaml and not through env vars when I opened this PR (my bad!). Personally, I think it is cleaner now :)

manusa (approved):
LGTM, thx!

manusa:
Removed MODEL_NAME from secrets

PR #519, opened 2025-11-28
feat(test): add utilities for JSONPath-like access to unstructured Kubernetes objects

manusa:
This PR adds test utility functions that provide JSONPath-like field access for `unstructured.Unstructured` objects.

The primary goal is to improve test assertion readability by replacing deeply-nested type assertions with clean, concise path-based field access.

### Readability Improvement

Before:
```golang
s.Equal("quay.io/containerdisks/fedora:latest",
  decodedResult[0].Object["spec"].(map[string]interface{})["template"].(map[string]interface{})["spec"].(map[string]interface{})["volumes"].([]interface{})[0].(map[string]interface{})["containerDisk"].(map[string]interface{})["image"].(string),
  "invalid default image")
```

After:
```golang
s.Equal("quay.io/containerdisks/fedora:latest",
  test.FieldString(vm, "spec.template.spec.volumes[0].containerDisk.image"),
  "invalid default image")
```

### DIFF
--- a/internal/test/unstructured.go
+++ b/internal/test/unstructured.go
@@ -11,6 +11,15 @@ import (
 )
 
 // FieldString retrieves a string field from an unstructured object using JSONPath-like notation.
+// Returns the string value, or empty string if not found or not a string.
+//
+// IMPORTANT: This function cannot distinguish between "field doesn't exist", "field is nil",
+// and "field exists with empty string value". When asserting empty string values (""),
+// you should also verify the field exists using FieldExists:
+//
+//	s.True(test.FieldExists(obj, "spec.emptyField"), "field should exist")
+//	s.Equal("", test.FieldString(obj, "spec.emptyField"), "field should be empty string")
+//
 // Examples:
 //   - "spec.runStrategy"
 //   - "spec.template.spec.volumes[0].containerDisk.image"
@@ -36,7 +45,15 @@ func FieldExists(obj *unstructured.Unstructured, path string) bool {
 }
 
 // FieldInt retrieves an integer field from an unstructured object using JSONPath-like notation.
-// Returns 0 if the field is not found or is not an integer type (int, int64, int32).
+// Returns the integer value (int64), or 0 if not found or not an integer type (int, int64, int32).
+//
+// IMPORTANT: This function cannot distinguish between "field doesn't exist", "field is nil",
+// and "field exists with value 0". When asserting zero values (0), you should also verify
+// the field exists using FieldExists:
+//
+//	s.True(test.FieldExists(obj, "spec.zeroValue"), "field should exist")
+//	s.Equal(int64(0), test.FieldInt(obj, "spec.zeroValue"), "field should be 0")
+//
 // Examples:
 //   - "spec.replicas"
 //   - "spec.ports[0].containerPort"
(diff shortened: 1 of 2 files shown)


internal/test/unstructured.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+			current = slice[segment.index]
+		} else {
+			// Handle map field access
+			m, ok := current.(map[string]interface{})
+			if !ok {
+				return nil, false
+			}
+			val, exists := m[segment.field]
+			if !exists {
```
Cali0707, line 121: @manusa I think you still need a `nil` check on `val` here as well

internal/test/unstructured.go
```diff
@@ -0,0 +1,172 @@
...
+func FieldExists(obj *unstructured.Unstructured, path string) bool {
+	if obj == nil {
+		return false
+	}
+	_, found := Field(obj.Object, path)
+	return found
+}
+
+// FieldInt retrieves an integer field from an unstructured object using JSONPath-like notation.
+// Returns 0 if the field is not found or is not an integer type (int, int64, int32).
```
Cali0707: @manusa I'm not 100% sure this behaviour is ideal - there could be scenarios where I'm testing and expecting to receive a value of 0 - for example status.replicas. How would we be able to distinguish between the value actually being 0 vs. not being set/not being an int? 

Or do you think that's fine because this is for tests? In that case I'm fine with it but let's add a note in the docstring about ensuring that you don't assert expecting an actual value to be 0

internal/test/unstructured.go
```diff
@@ -0,0 +1,172 @@
...
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// FieldString retrieves a string field from an unstructured object using JSONPath-like notation.
+// Examples:
+//   - "spec.runStrategy"
+//   - "spec.template.spec.volumes[0].containerDisk.image"
+//   - "spec.dataVolumeTemplates[0].spec.sourceRef.kind"
```
Cali0707, line 26: Same as my comment about `FieldInt` - it would be nice to be able to tell the difference between not set and set to `""`

manusa: For that purpose you would use `test.FieldExists(vm, "spec.instancetype")` and expect it to be false

https://github.com/marcnuri-forks/kubernetes-mcp-server/blob/8e5f6df7d009ebb7785c12d46faed9e73048be23/pkg/mcp/kubevirt_test.go#L283

internal/test/unstructured.go
manusa: For that purpose you would use `test.FieldExists(vm, "spec.instancetype")` and expect it to be false

https://github.com/marcnuri-forks/kubernetes-mcp-server/blob/8e5f6df7d009ebb7785c12d46faed9e73048be23/pkg/mcp/kubevirt_test.go#L283

manusa: I'll need to revisit for this case for the expecting to be 0

manusa:
I've added the nil-check.
For the other cases you mention, this is not an easy solution.
The goal of the utility is to make the tests more readable.
Any solution I can think of would defeat the purpose of these functions.

For the edge cases you mention, there is both `Field`, `FieldValue` and FieldExists` functions that can be leveraged to perform the extra assertion.
Any of these can be used to perform the extra checks that ensure that a field exists and contains an empty string as opposed to a nil value or a 0 int as opposed to a nil field value.

I've added some godoc to the simplified accessors so that users are aware of the limitation.

Cali0707 (approved):
LGTM, thanks @manusa 

The tests are way more readable now 😄

PR #520, opened 2025-11-28
feat(kubernetes): cluster state and kubeconfig watchers

manusa:
Follows up on #502 to better organize the code and improve its maintenance and sustainability.

Changes:
  - Extract watchers into new pkg/kubernetes/watcher package with a common Watcher interface
  - Move watcher ownership from Manager to Providers (kubeConfigClusterProvider, singleClusterProvider)
  - Simplify Manager by removing watch-related fields and methods (continues the work on #503)
  - MCP server refactor:
    - Creates the provider instance only once (maybe this will need changing depending on the scope of what needs to be done for #493)
    - Exposed a reloadToolsets method that is used as the watcher callback.
  - Extract IsOpenshift() into pkg/openshift package for reuse
  - Add Manager.Invalidate() method for cache invalidation
  - Add environment variable overrides (CLUSTER_STATE_POLL_INTERVAL_MS, CLUSTER_STATE_DEBOUNCE_WINDOW_MS) for testing

### DIFF
--- a/pkg/kubernetes/provider_kubeconfig.go
+++ b/pkg/kubernetes/provider_kubeconfig.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"github.com/containers/kubernetes-mcp-server/pkg/config"
+	"github.com/containers/kubernetes-mcp-server/pkg/kubernetes/watcher"
 	authenticationv1api "k8s.io/api/authentication/v1"
 )
 
@@ -17,8 +18,10 @@ const KubeConfigTargetParameterName = "context"
 // Kubernetes clusters using different contexts from a kubeconfig file.
 // It lazily initializes managers for each context as they are requested.
 type kubeConfigClusterProvider struct {
-	defaultContext string
-	managers       map[string]*Manager
+	defaultContext      string
+	managers            map[string]*Manager
+	kubeconfigWatcher   *watcher.Kubeconfig
+	clusterStateWatcher *watcher.ClusterState
 }
 
 var _ Provider = &kubeConfigClusterProvider{}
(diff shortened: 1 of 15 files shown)


manusa:
For https://github.com/containers/kubernetes-mcp-server/pull/493/commits/f54d30e2f3763daa4c36b1838805fd9b225be649 

IMO we should actually move the SIGHUP logic to the pkg/kubernetes-mcp-server/cmd package where it makes more sense.

nader-ziada:
i like the moving of watcher logic in a separate package, much cleaner. Also the move of open shift check. 👍 

/lgtm

pkg/kubernetes/provider_kubeconfig.go
```diff
@@ -118,14 +123,19 @@ func (p *kubeConfigClusterProvider) GetDefaultTarget() string {
...
 }
 
-func (p *kubeConfigClusterProvider) WatchTargets(onKubeConfigChanged func() error) {
-	m := p.managers[p.defaultContext]
-	m.WatchKubeConfig(onKubeConfigChanged)
-	m.WatchClusterState(DefaultClusterStatePollInterval, DefaultClusterStateDebounceWindow, onKubeConfigChanged)
+func (p *kubeConfigClusterProvider) WatchTargets(reload McpReload) {
+	reloadWithCacheInvalidate := func() error {
+		// Invalidate all cached managers to force reloading on next access
+		for contextName := range p.managers {
```
nader-ziada, line 129: do we need a nil check here if case the manager is nil?

manusa: 👍 good catch

nader-ziada:
thanks for the refactor @manusa, i will rebase the drop-in config and make the suggested changes 

/lgtm
/approve

PR #527, opened 2025-12-01
feat(kiali): Add Kiali with gevals

aljesusg:
Kiali will run Gevals when the task filter is set to 'kiali'


aljesusg:
First approach @Cali0707 
I think it’s better to run Kiali only in certain specific cases, and not with the entire weekly set alongside the rest. WDYT?

evals/tasks/kiali/troubleshooting-log/kiali-troubleshooting-log.yaml
```diff
@@ -0,0 +1,45 @@
...
+                  httpStatus: 500
+                  percentage:
+                    value: 100
+              route:
+              - destination:
+                  host: productpage
+        EOF
+  verify:
+    inline: |-
+      #!/usr/bin/env bash
```
Cali0707: @aljesusg do you maybe want to use the llm judge impl here to check for if the model response contains the correct info? See an example here: https://github.com/containers/kubernetes-mcp-server/blob/77b4bafcf413c2a53f99ce5d69e6a36ec6d1218b/evals/tasks/debug-app-logs/debug-app-logs.yaml#L12-L13

manusa:
> In terms of when to run kiali vs. the whole set of evals I defer to @manusa - what do you think Marc?

I understand that this might be beneficial to run on PRs that affect the Kiali toolset, the filter sounds like a good approach.
However, we should probably document this to avoid forgetting how the PR comments work.

Additionally, we could improve the workflow dispatch to include a drop-down of the features/toolsets to test so at some point we can dispatch an execution for a given toolset evaluation.

aljesusg:
@manusa @Cali0707 WDYT about organizing the tasks into their own subfolders?

Issue #522, opened 2025-11-28
Provider watchers to update underlying manager


manusa:
## Description

In #520 I likely introduced a non-documented behavioral regression.

Before the introduced changes, the `mcp.Server` instance recreated the `kubernetes.Provider` whenever the `reloadKubernetesClusterProvider` callback was called.
This wasn't a straightforward implementation since the callback itself was called from the `kubernetes.Provider` instance.
In the method, the `mcp.Server` teared down the active `kubernetes.Provider` and recreated it.

After the introduced changes, the `mcp.Server` instance exposes a `reloadToolsets` method that is used as a callback by the `kubernetes.Provider` watchers (`WatchTargets(reload McpReload)`).
The `mcp.Server` no longer recreates the `kubernetes.Provider` instance, which is active for the entire lifecycle of the mcp.Server.

This means that if the underlying configurations for the applicable `kubernetes.Manager` instances managed my the `kuberentes.Provider` are changed, the provider itself won't be aware of these.
Same happens in case new contexts/targets are added to the configuration.

PR #531, opened 2025-12-01
test(kubernetes): add WatchTargets test for kubeconfig changes

manusa:
Relates to #522

### DIFF
--- a/pkg/kubernetes/provider_watch_test.go
+++ b/pkg/kubernetes/provider_watch_test.go
@@ -1,0 +1,95 @@
+package kubernetes
+
+import (
+	"os"
+	"reflect"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/containers/kubernetes-mcp-server/internal/test"
+	"github.com/containers/kubernetes-mcp-server/pkg/config"
+	"github.com/stretchr/testify/suite"
+)
+
+type ProviderWatchTargetsTestSuite struct {
+	suite.Suite
+	mockServer *test.MockServer
+}
+
+func (s *ProviderWatchTargetsTestSuite) SetupTest() {
+	s.mockServer = test.NewMockServer()
+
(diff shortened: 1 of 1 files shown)


pkg/kubernetes/provider_watch_test.go
```diff
@@ -0,0 +1,95 @@
...
+// WaitForWatchTargets sets up a WatchTargets callback, executes the provided function, and waits for the callback to be invoked.
+func (s *ProviderWatchTargetsTestSuite) WaitForWatchTargets(timeout time.Duration, provider Provider, fn func()) {
+	callbackInvoked := make(chan struct{})
+	var once sync.Once
+	provider.WatchTargets(func() error {
+		once.Do(func() {
+			close(callbackInvoked)
+		})
+		return nil
+	})
```
nader-ziada, line 40: There is no guarantee here that the watcher has successfully started listening for events. I feel like this might introduce a race?

manusa: This is a utility callback for the tests (not production code).

It's used to wait for the callback to be invoked and then proceed with the assertions.
If the watchers haven't started (or any other issue) the callback will never be invoked and the tests will fail after the provided timeout.


I haven't seen any concurrency issues so far.

manusa: OK, thinking it better, I understand that maybe you're referring that the fn() is invoked without guarantees of watchers having started.

For the `watcher.Kubeconfig` the code is synchronous and shouldn't be an issue, the watch is started by the time the method returns.

For `watcher.ClusterState` the code is not synchronous, but shouldn't be an issue either.

In all cases, the code executed in the enclosed `fn()` is what unblocks the `callbackInvoked` select.
If there's a race condition the test simply won't pass.

manusa:
Closing in favor of #532 which addresses the issue too.

PR #532, opened 2025-12-02
fix(kubernetes)!: Provider updated for each watcher callback

### LINKED ISSUE #522: Provider watchers to update underlying manager

manusa:
Fixes #522 
Replaces #531

The `kubernetes.Provider` watch-related functionality is restarted for each watch notification that affects its state.

### DIFF
--- a/pkg/kubernetes/provider_kubeconfig.go
+++ b/pkg/kubernetes/provider_kubeconfig.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 
 	"github.com/containers/kubernetes-mcp-server/pkg/config"
 	"github.com/containers/kubernetes-mcp-server/pkg/kubernetes/watcher"
@@ -18,6 +19,7 @@ const KubeConfigTargetParameterName = "context"
 // Kubernetes clusters using different contexts from a kubeconfig file.
 // It lazily initializes managers for each context as they are requested.
 type kubeConfigClusterProvider struct {
+	staticConfig        *config.StaticConfig
 	defaultContext      string
 	managers            map[string]*Manager
 	kubeconfigWatcher   *watcher.Kubeconfig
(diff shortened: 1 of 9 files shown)


manusa:
Once this one is merged or the related bug is fixed we should start with the 0.0.55 release which should be stable enough.

pkg/kubernetes/provider_kubeconfig.go
```diff
@@ -35,37 +37,44 @@ func init() {
...
 		rawConfig.CurrentContext: m, // we already initialized a manager for the default context, let's use it
 	}
 
 	for name := range rawConfig.Contexts {
 		if name == rawConfig.CurrentContext {
 			continue // already initialized this, don't want to set it to nil
 		}
-
-		allClusterManagers[name] = nil
+		p.managers[name] = nil
```
nader-ziada, line 69: would these managers have any resources that should be cleaned up

manusa: That's a good question.

We don't provide a `kubernetes.Manager`.Clean method.

The manager itself has a pointer reference to the global `config.StaticConfig` and an instance of `kubernetes.AccessControlClientset`.

The clientset has instances of the discovery client, dynamic client, and kubernetes interface. But AFAICT, none of those clients provide clean up methods.

I might have missed something, but I'd say that there's no clean up needed.

PR #533, opened 2025-12-02
feat(kubevirt): Add VM lifecycle management tools (start, stop, restart)

ksimon1:
## Summary
This PR extends the KubeVirt toolset with VM lifecycle management capabilities, enabling AI agents to start, stop, and restart virtual machines. 

The new vm_lifecycle tool accepts:
- namespace: The namespace of the virtual machine
- name: The name of the virtual machine
- action: The lifecycle action ('start', 'stop', or 'restart')

### `start`

Starts a halted VirtualMachine by updating its `runStrategy` to `Always`.

### `stop`

Stops a running VirtualMachine by updating its `runStrategy` to `Halted`.

### `restart`

Restarts a VirtualMachine by cycling it through a complete stop-start sequence.

### DIFF
--- a/pkg/toolsets/kubevirt/vm/lifecycle/tool.go
+++ b/pkg/toolsets/kubevirt/vm/lifecycle/tool.go
@@ -48,8 +48,8 @@ func Tools() []api.ServerTool {
 				Annotations: api.ToolAnnotations{
 					Title:           "Virtual Machine: Lifecycle",
 					ReadOnlyHint:    ptr.To(false),
-					DestructiveHint: ptr.To(false),
-					IdempotentHint:  ptr.To(true),
+					DestructiveHint: ptr.To(true),
+					IdempotentHint:  ptr.To(false),
 					OpenWorldHint:   ptr.To(false),
 				},
 			},
@@ -137,4 +137,3 @@ func getRequiredString(params api.ToolHandlerParams, key string) (string, error)
 	}
 	return str, nil
 }
-
--- a/pkg/mcp/testdata/toolsets-kubevirt-tools.json
+++ b/pkg/mcp/testdata/toolsets-kubevirt-tools.json
@@ -81,8 +81,7 @@
   {
     "annotations": {
       "title": "Virtual Machine: Lifecycle",
-      "destructiveHint": false,
-      "idempotentHint": true,
+      "destructiveHint": true,
       "openWorldHint": false
     },
     "description": "Manage VirtualMachine lifecycle: start, stop, or restart a VM",


ksimon1:
@manusa, @Cali0707 would you please review this PR?

ksimon1:
@codingben can you please review this PR?

ksimon1:
@Cali0707 for sure I can do that :+1:

codingben:
> Hey @ksimon1 is it possible to consolidate these three tools into one?
> 
> IMO, a tool like `vm_lifecycle_update` or `vm_lifecycle_set` could work well - stop/start/restart could all be options set in the tool call.
> 
> That should help us keep a smaller tool count in this toolset
> 
> WDYT? (also cc @manusa as he has lots of great ideas about tool defs)

It's similar to KubeVirt implementation in [lifecycle.go](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-api/rest/lifecycle.go) from BackEnd perspective.

Why we're not importing `"kubevirt.io/api/core/v1"` to get the running strategy and actually the KubeVirt's client implementation? I'm not fan of dynamic client and unstructured fields.

pkg/toolsets/kubevirt/vm/restart/tool.go
```diff
@@ -0,0 +1,86 @@
...
+	// Format the output
+	marshalledYaml, err := output.MarshalYaml([]*unstructured.Unstructured{vm})
+	if err != nil {
+		return api.NewToolCallResult("", fmt.Errorf("failed to marshal VirtualMachine: %w", err)), nil
+	}
+
+	return api.NewToolCallResult("# VirtualMachine restarted successfully\n"+marshalledYaml, nil), nil
+}
+
+func getRequiredString(params api.ToolHandlerParams, key string) (string, error) {
```
codingben: nit: This is a duplicated function, should be part of API utils or something else.

ksimon1:
@codingben I am not fan of unstructured and dynamic client too, but problem is, that kubevirt.io/client-go has dependency conflicts with packages in this project. That means I would have to update dependencies in this project to fit our kubevirt dependency. @Cali0707 do you think it would be possible?

ksimon1:
@Cali0707, @manusa can you please review new code?

codingben:
@ksimon1 Can you please update the PR description and commit messages?

ksimon1:
@Cali0707, @manusa would you please review this PR?

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
```diff
@@ -0,0 +1,140 @@
...
+							Enum:        []any{"start", "stop", "restart"},
+							Description: "The lifecycle action to perform: 'start' (changes runStrategy to Always), 'stop' (changes runStrategy to Halted), or 'restart' (stops then starts the VM)",
+						},
+					},
+					Required: []string{"namespace", "name", "action"},
+				},
+				Annotations: api.ToolAnnotations{
+					Title:           "Virtual Machine: Lifecycle",
+					ReadOnlyHint:    ptr.To(false),
+					DestructiveHint: ptr.To(false),
```
Cali0707:
```suggestion
					DestructiveHint: ptr.To(true), // stop/restart can disrupt workloads
```
IMO this annotation is wrong, as stopping or restarting a VM can disrupt workloads

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
```diff
@@ -0,0 +1,140 @@
...
+					IdempotentHint:  ptr.To(true),
```
Cali0707:
```suggestion
					IdempotentHint:  ptr.To(false),
```

Here I'm less sure than the `DestructiveHint` annotation - but I think that the restart action is likely not fully idempotent, as restarting a VM multiple times may have side effects on the broader sytem. But, I'm not a kubevirt expert so happy to defer to @ksimon1 and @codingben here

Cali0707 (commented):
Overall this is looking good to me, just a few questions around tool annotations and then I'm happy to merge this

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
ksimon1: These actions should not be destructive, because the VM will be shutted down gracefully - no data should be lost.

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
ksimon1: You are right about this comment. If we consider the broader system, then this action is not completely idempotent, because each launch can cause resource problems for other pods (exceptionally, but it is possible).
However, if we are only talking about the VM pod, then the restart action is idempotent, because every time the action is triggered, the VM is stopped, started and running at the end.
Is this annotation intended for the broader system or only for the resources concerned?

Cali0707: IMO these annotations are all for the broader system. @manusa what are your thoughts here?

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
manusa: The purpose of annotations is mostly (usually) UX-wise.

MCP Hosts/Clients should consider these hints to cater their users with an improved experience.
For example, in case of destructive tools, the MCP host might popup a confirmation dialog asking the users if they are sure to continue with this operation since it might perform irreversible actions.
If the annotation us non-destructive, some agents might opt to just perform the action without user confirmation (human in the loop)

You're the domain experts so I understand that it's your opinion what counts in this subjective matter.

From my outsider perspective, I understand that users might want to be prompted if the LLM is wanting to restart a VM because maybe the VM is running a critical process and by restarting or stopping the machine the LLM will be disrupting or even _destroying_ something the user is interested in.

Like I said, you're the domain experts and I think that your opinion is what counts more. You need to consider the different scenarios and how the different Agents/MCP Hosts might consider these annotations in order to provide a value that will satisfy the users of this toolset.

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
manusa: Mostly what I reflected upon here:
https://github.com/containers/kubernetes-mcp-server/pull/533/files#r2602400034

The annotations should be set based on the understanding of what the consumers (i.e. Agents, MCP Hosts, etc.) will be using them for.

For this annotation, the LLM will usually consider this for retries. Or safely calling the same function/tool multiple times with the same arguments.

From the tests I understand that for start and stop operations the server state is preserved when a call is retried multiple times but the response changes (which should be fine).

For restart, I think that multiple repeated calls might have unexpected side-effects that the user won't expect.

ksimon1:
@Cali0707, @manusa I updated the annotations. Would you please review it again?

Cali0707:
@ksimon1 this looks good from my end now - would you mind running `make test-update-snapshots` so that we can fix the failing unit tests?

manusa (approved):
LGTM, thx!

ksimon1:
@Cali0707 fixed

Cali0707:
Thanks for all your work on this tool @ksimon1 !

manusa:
Regarding the destructive hints, this issue applies https://github.com/containers/kubernetes-mcp-server/issues/568

When the MCP server starts with --disable-destructive, the tool will be disabled. We'll need to work on this in the scope of #568

PR #539, opened 2025-12-03
refactor(api): extract common interface{} to int64 conversion utility

### LINKED ISSUE #537: [REFACTOR] Extract common interface{} to int64 conversion utility
## Description

The codebase has multiple places where `interface{}` values (from JSON-decoded tool arguments) need to be converted to `int64`. This pattern is duplicated across several files, making the code harder to maintain and increasing the risk of subtle bugs (like the one in `nodes.go` where `case int:` is missing the assignment).

## Background

PR #512 introduced a new `parseScaleValue` function in `pkg/toolsets/core/resources.go` to handle this conversion for the `scale` parameter. However, this exact pattern already exists in other locations.

### Known duplications (non-exhaustive, requires analysis):

1. **`pkg/toolsets/core/resources.go:parseScaleValue`** (new in #512)
2. **`pkg/toolsets/core/nodes.go:112`** - `tailLines` parameter (⚠️ has a bug: `case int:` missing assignment!)
3. **`pkg/toolsets/core/pods.go:401`** - `tail` parameter
4. **`pkg/toolsets/kiali/logs.go:89`** - `tail` parameter (converts to string)

> **Note:** This list is non-exhaustive. A thorough analysis of the codebase should be performed to identify all similar patterns before implementing the refactor.

## Proposed Solution

Create a shared utility function in an appropriate package (suggestion: `pkg/api/` or a new `pkg/toolsets/common/` package):

```go
// ParseInt64Value converts an interface{} value (typically from JSON-decoded tool arguments) 
// to int64. Handles float64 (JSON numbers), int, and int64 types.
func ParseInt64Value(value interface{}, paramName string) (int64, error) {
    switch v := value.(type) {
    case float64:
        return int64(v), nil
    case int:
        return int64(v), nil
    case int64:
        return v, nil
    default:
        return 0, fmt.Errorf("failed to parse %s parameter: expected integer, got %T", paramName, value)
    }
}
```

## Acceptance Criteria

- [ ] Analyze the codebase thoroughly to identify ALL instances of this pattern
- [ ] Create a single `ParseInt64Value` utility function in an appropriate location
- [ ] Refactor all existing duplicated code to use the utility
- [ ] Fix the bug in `nodes.go` where `case int:` is missing the assignment
- [ ] Add unit tests for the utility function covering all type cases and error scenarios
- [ ] Consider if similar utilities are needed for other common conversions (e.g., string, bool)

## Additional Notes

This issue was identified during the review of PR #512. The `nodes.go` bug where `case int:` falls through without assigning the value could cause unexpected behavior if an integer (rather than float64) is somehow passed.

manusa:
Fixes #537

Add ParseInt64 utility function in pkg/api/params.go to eliminate duplicated type conversion logic across toolsets. This also fixes a bug in nodes.go where `case int:` was missing the assignment statement.

### DIFF
--- a/pkg/api/params.go
+++ b/pkg/api/params.go
@@ -1,0 +1,28 @@
+package api
+
+import "fmt"
+
+// ErrInvalidInt64Type is returned when a value cannot be converted to int64.
+type ErrInvalidInt64Type struct {
+	Value interface{}
+}
+
+func (e *ErrInvalidInt64Type) Error() string {
+	return fmt.Sprintf("expected integer, got %T", e.Value)
+}
+
+// ParseInt64 converts an interface{} value (typically from JSON-decoded tool arguments)
+// to int64. Handles float64 (JSON numbers), int, and int64 types.
+// Returns the converted value and nil error on success, or 0 and an ErrInvalidInt64Type if the type is unsupported.
+func ParseInt64(value interface{}) (int64, error) {
+	switch v := value.(type) {
+	case float64:
+		return int64(v), nil
+	case int:
+		return int64(v), nil
+	case int64:
+		return v, nil
+	default:
+		return 0, &ErrInvalidInt64Type{Value: value}
+	}
+}
--- a/pkg/api/params_test.go
+++ b/pkg/api/params_test.go
@@ -1,0 +1,105 @@
+package api
+
+import (
+	"testing"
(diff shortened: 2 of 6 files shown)


nader-ziada:
lgtm

PR #550, opened 2025-12-04
test: replace custom WaitForCondition with testify assertions

manusa:


pkg/kubernetes/watcher/cluster_test.go
```diff
@@ -35,11 +41,11 @@ func (s *ClusterStateTestSuite) TearDownTest() {
 
 // waitForWatcherState waits for the watcher to capture initial state
 func (s *ClusterStateTestSuite) waitForWatcherState(watcher *ClusterState) {
-	s.NoError(test.WaitForCondition(watcherStateTimeout, func() bool {
+	s.Eventually(func() bool {
 		watcher.mu.Lock()
 		defer watcher.mu.Unlock()
 		return len(watcher.lastKnownState.apiGroups) > 0
-	}), "timeout waiting for watcher to capture initial state")
+	}, watcherStateTimeout, eventuallyTick, "timeout waiting for watcher to capture initial state")
```
nader-ziada, line 46: what is the timeout value based on? its had multiple different values

manusa: I think I messed with those while trying to fix the windows flakyness, hence the multiple changes.

At the moment it's 100ms which is double than the poll interval set in the test (50ms)

https://github.com/marcnuri-forks/kubernetes-mcp-server/blob/959a45d4a72e3bcf5ed521e927f6a08a21a82e41/pkg/kubernetes/watcher/cluster_test.go#L206-L208

nader-ziada: so you want to keep them at different values for each test?

manusa: Sorry, I misunderstood your question. I see what you mean now, let me change.

manusa: Changed, let me know if the new values make sense.

nader-ziada:
lgtm

PR #551, opened 2025-12-04
test(core): add denied resources tests for resources_scale tool

manusa:


pkg/mcp/resources_test.go
```diff
@@ -723,6 +723,74 @@ func (s *ResourcesSuite) TestResourcesScale() {
...
+		})
+		s.Run("describes denial", func() {
+			msg := deniedByKind.Content[0].(mcp.TextContent).Text
+			s.Contains(msg, "resource not allowed:")
+			expectedMessage := "failed to get/update resource scale:(.+:)? resource not allowed: apps/v1, Kind=Deployment"
+			s.Regexpf(expectedMessage, msg,
+				"expected descriptive error '%s', got %v", expectedMessage, deniedByKind.Content[0].(mcp.TextContent).Text)
+		})
+	})
+	s.Run("resources_scale (denied by group)", func() {
```
Cali0707: @manusa I'm not sure what the difference here with the `(denied by group)` is vs. the `(denied by kind)` test case above - as far as I can tell we are not allowing the StatefulSet kind in the config, so this one is also denied by kind?

manusa: This one was fully implemented by Claude. It probably messed up, I'll review better tomorrow.

manusa: Should be fixed now :)

Cali0707 (approved):
LGTM, thanks @manusa !

PR #524, opened 2025-11-28
feat(config): add drop-in configuration support with dynamic reload

nader-ziada:
follow up to PRs: #493 , #502 and #520 

fixes #548 

Add support for flexible configuration management through both main config files and drop-in directories

- Load config from --config (main file) and --config-dir (drop-in directory)
- Drop-in files processed in alphabetical order for predictable overrides
- Partial configuration support - drop-in files can override specific values
- Dynamic configuration reload via SIGHUP signal
- SIGHUP support requires either --config or --config-dir flag at startup

PR #552, opened 2025-12-04
update readme to fix drop-in config section

nader-ziada:
as a follow up to PR #524, update readme to describe the feature more accurately


Cali0707 (approved):
LGTM, thanks @nader-ziada !

PR #554, opened 2025-12-05
feat(pod_logs): add keyword filtering support

### LINKED ISSUE #350: [RFE] pods_log filter parameter
When looking through logs, sometimes I know what I want to find (e.g., `ERROR`, etc). Having a filter parameter on the tool might be nice. I realize the API doesn't directly provide that, so this would need to be internal to the MCP server.

trytrywei-debug:
fix: #350


trytrywei-debug:
@adietish @giuseppe @lyarwood @bussyjd review this change please

pkg/toolsets/core/pods.go
```diff
@@ -205,8 +205,12 @@ func initPods() []api.ServerTool {
 						Type:        "boolean",
 						Description: "Return previous terminated container logs (Optional)",
 					},
+					"query": {
+						Type:        "string",
+						Description: "filter logs by query",
+					},
 				},
-				Required: []string{"name"},
+				Required: []string{"name", "query"},
```
champly: "query" is required？

pkg/kubernetes/pods.go
```diff
@@ -96,24 +99,28 @@ func (k *Kubernetes) PodsLog(ctx context.Context, namespace, name, container str
...
 
-	req := pods.GetLogs(name, logOptions)
-	res := req.Do(ctx)
-	if res.Error() != nil {
-		return "", res.Error()
+	logData := make([]string, 0)
+	scanner := bufio.NewScanner(stream)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.Contains(line, query) {
```
champly, line 110: query?

pkg/toolsets/core/pods.go
trytrywei-debug: has removed from required

pkg/kubernetes/pods.go
trytrywei-debug: added query parameter to the method's input parameters.

PR #558, opened 2025-12-05
chore(ci): pin gevals version to releases, not main

Cali0707:
This switches to using a specific release, instead of `main` for gevals


nader-ziada:
nice

/lgtm

Issue #226, opened 2025-08-01
[CONFIG] Configurable prompt templates


manusa:
## Description

As a Kubernetes MCP Server administrator, I want to be able to define MCP prompts using the configuration TOML file, so that I can launch a server with custom workflow templates without needing to compile a specific version of the MCP server.

### Background

The [Model Context Protocol (MCP) specification](https://modelcontextprotocol.io/specification/2025-06-18/server/prompts) defines **Prompts** as user-controlled templates that servers expose to clients. Prompts provide structured messages and instructions for interacting with language models, enabling pre-defined workflow templates and guidance for AI assistants.

### Expected Behavior

Administrators can define custom prompts in the TOML configuration file, and these prompts are exposed through the MCP server's `prompts/list` and `prompts/get` endpoints.

**Configuration file example:**

```toml
[[prompts]]
name = "troubleshoot-pod"
title = "Troubleshoot Pod"
description = "Debug a failing or crashed pod"

  [[prompts.arguments]]
  name = "pod_name"
  description = "Name of the pod to troubleshoot"
  required = true

  [[prompts.arguments]]
  name = "namespace"
  description = "Namespace of the pod"
  required = false

  [[prompts.messages]]
  role = "user"
  content = "Help me troubleshoot pod {{pod_name}} in namespace {{namespace}}"

  [[prompts.messages]]
  role = "assistant"
  content = "I'll investigate the pod {{pod_name}} for you."
```

- `prompts/list` returns the prompt metadata (`name`, `title`, `description`, `arguments`)
- `prompts/get` returns the configured messages with `{{argument}}` placeholders substituted with provided values

### Implementation Notes

Since prompt definitions will likely be reused for other purposes (e.g., toolset-defined prompts in future enhancements), the core data structures should be defined in the `pkg/api/` package with appropriate struct tags for JSON/[~~YAML~~/](https://github.com/containers/kubernetes-mcp-server/pull/559/files#r2609742216)TOML serialization and godoc comments referencing the MCP specification.

Only text content type is supported for messages in this implementation.

### Scope

**In scope:**
- Define prompt data structures in `pkg/api/` (`Prompt`, `PromptArgument`, `PromptMessage`, `PromptContent`)
- Add prompt configuration support in `pkg/config/`
- Parse prompt definitions from TOML configuration
- Register and expose configured prompts through the MCP server
- Support `{{argument}}` placeholder substitution in prompt messages
- Validate required arguments when `prompts/get` is called
[...]

PR #510, opened 2025-11-26
feat: add MCP prompts support

nader-ziada:
Fixes: #226 

Add MCP prompts support to enable pre-defined workflow templates and guidance for AI assistants. 

Includes:
- Core prompt types (ServerPrompt, Prompt, PromptArgument, PromptMessage)
- Template argument substitution with {{variable}} syntax
- Required argument validation
- built-in prompts for common Kubernetes workflows (troubleshooting, deployment, scaling, cluster health, networking, resource usage)
- Integration with MCP server to register and serve prompts
- prompts in toml config file

Issue #556, opened 2025-12-05
[API] Toolset-defined prompt templates


manusa:
## Description

As a toolset implementer, I want to be able to define MCP prompts as part of my toolset definition, so that I can provide workflow templates that are bundled with the toolset's functionality.

### Background

The `Toolset` interface in `pkg/api/toolsets.go` currently allows implementers to define tools via `GetTools()`. This issue extends the interface to also support defining prompts via a new `GetPrompts()` method.

This builds on the infrastructure established in #226, which adds MCP prompt support to the configuration module. The `Prompt` struct and argument substitution logic defined there will be reused for toolset-defined prompts.

### Expected Behavior

Toolset implementers can define prompts by implementing the `GetPrompts()` method on their toolset:

```go
type Toolset interface {
    GetName() string
    GetDescription() string
    GetTools(o internalk8s.Openshift) []ServerTool
    // GetPrompts returns the prompts provided by this toolset.
    // Returns nil or empty slice if the toolset doesn't provide any prompts.
    GetPrompts() []Prompt
}
```

Example implementation:

```go
func (t *MyToolset) GetPrompts() []api.Prompt {
    return []api.Prompt{
        {
            Name:        "my-workflow",
            Title:       "My Workflow",
            Description: "A custom workflow for my toolset",
            Arguments: []api.PromptArgument{
                {
                    Name:        "resource_name",
                    Description: "Name of the resource",
                    Required:    true,
                },
            },
            Messages: []api.PromptMessage{
                {
                    Role:    "user",
                    Content: "Help me with {{resource_name}}",
                },
            },
        },
    }
}
```

The MCP server collects prompts from all enabled toolsets and exposes them via `prompts/list` and `prompts/get`. The same argument substitution logic from #226 handles rendering.

### Disabling Toolset Prompts

Administrators can disable all toolset-defined prompts via configuration, using only config-defined prompts:

```toml
disable_toolset_prompts = true
```

### Scope

**In scope:**
- Add `GetPrompts() []Prompt` to the `Toolset` interface
- Update existing toolsets to return `nil` or empty slice from `GetPrompts()`
- Update MCP server to collect and register prompts from enabled toolsets
- Merge toolset prompts with config-defined prompts (config takes precedence for same name)
- Add `disable_toolset_prompts` configuration option

**Out of scope:**
- Built-in/embedded prompts shipped with the server (separate issue)
- System prompts support

### Related

[...]

PR #559, opened 2025-12-08
feat(config): add configurable prompts support

nader-ziada:
This PR replaces: https://github.com/containers/kubernetes-mcp-server/pull/510

Fixes: #226 

Enable administrators to define custom MCP prompts in config.toml using template substitution with {{argument}} syntax.

Features:
- PromptLoader for loading prompts from TOML config
- Core prompt types (ServerPrompt, Prompt, PromptArgument, PromptMessage)
- Template argument substitution with {{variable}} syntax
- Required argument validation
- Integration with MCP server to register and serve config prompts


pkg/config/config.go
```diff
@@ -94,6 +98,28 @@ type GroupVersionKind struct {
...
+	Description string `toml:"description"`
+	Required    bool   `toml:"required"`
+}
+
+// PromptMessageConfig represents a prompt message in TOML
+type PromptMessageConfig struct {
+	Role    string `toml:"role"`
+	Content string `toml:"content"`
+}
+
```
manusa: These structs are exactly the same as those defined in the api package, is there any reasoning to duplicate them?

I see we've created multiple versions of the same data stuctures that could easily be consolidated while preserving their single responsibility.

I believe the pkg/api/prompts.go could contain could define the single struct with all the applicable tags (`json`, `yaml`, `toml`, etc) and be reused everywhere else.

> ### Scope
> - Define prompt data structures in pkg/api/ (Prompt, PromptArgument, PromptMessage, PromptContent)

_[[CONFIG] Configurable prompt templates #226](https://github.com/containers/kubernetes-mcp-server/issues/226)_

pkg/api/prompt_loader.go
```diff
@@ -0,0 +1,136 @@
...
+		return fmt.Errorf("failed to convert config to YAML: %w", err)
+	}
+
+	if err := yaml.Unmarshal(data, &defs); err != nil {
+		return fmt.Errorf("failed to parse prompt config: %w", err)
+	}
+
+	l.definitions = append(l.definitions, defs...)
+	return nil
+}
```
manusa: Ideally all of the config-related should be part of the config package.

By reusing the structs, much of the handling, conversion, and manipulation that's done here and in the mcp package should be completely unnecessary.

pkg/config/config.go
nader-ziada: reworked the types to match the requirements and simplify

pkg/api/prompt_loader.go
nader-ziada: reworked the parsing design to match the requirements and simplify

nader-ziada:
Spec Proposed:
```
type Prompt struct {
    Messages    []PromptMessage  // Static templates
    GetMessages func(arguments map[string]string) []PromptMessage  // Dynamic function
}
```
What Was Actually Implemented:
```
type Prompt struct {
    Templates   []PromptTemplate  // Static templates for TOML parsing
}

type ServerPrompt struct {
    Prompt  Prompt
    Handler PromptHandlerFunc  // Dynamic handler function
}
```

why:
The handler pattern provides separation of concerns by keeping Prompt as just data (serializable to/from TOML) while ServerPrompt adds behaviour through the handler function. This eliminates serialization issues that would require json:"-" tags on function fields or special handling during TOML encoding. Both config and toolset prompts use the same unified PromptHandlerFunc interface, and handlers receive context through PromptHandlerParams 

Config prompts: PromptLoader.createHandler() generates a handler that performs {{argument}} substitution on the Templates field at runtime

(For Later): Toolset prompts: Implementers provide a custom handler function directly in ServerPrompt.Handler

manusa:
Thanks for taking care of this.

I still find that the API is getting far too complicated and complex for something that should be simpler and more intuitive.
This is also revealing a nasty coupling between the config and kubernetes module.

Let me create a PR to your branch with some **proposed** changes so that we can better discuss tradeoffs.

nader-ziada:
@manusa ready for review

pkg/mcp/mcp.go
```diff
@@ -160,9 +162,75 @@ func (s *Server) reloadToolsets() error {
...
+		if _, exists := overrideMap[prompt.Prompt.Name]; !exists {
+			result = append(result, prompt)
+		}
+	}
+
+	// Add all override prompts
+	result = append(result, override...)
+
+	return result
+}
```
manusa, line 232: I believe this is something you'll be using in the follow-up for #556

pkg/api/prompts.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Prompt represents the metadata and content of an MCP prompt.
+// See MCP specification: https://spec.modelcontextprotocol.io/specification/server/prompts/
+type Prompt struct {
+	Name        string           `yaml:"name" json:"name" toml:"name"`
+	Title       string           `yaml:"title,omitempty" json:"title,omitempty" toml:"title,omitempty"`
+	Description string           `yaml:"description,omitempty" json:"description,omitempty" toml:"description,omitempty"`
+	Arguments   []PromptArgument `yaml:"arguments,omitempty" json:"arguments,omitempty" toml:"arguments,omitempty"`
+	Templates   []PromptTemplate `yaml:"messages,omitempty" json:"messages,omitempty" toml:"messages,omitempty"`
```
manusa, line 35: I'm not sure we need the YAML tags or the yaml serialization tests.

Maybe this is a misunderstanding because the issue (#226) explicitly mentions yaml.
I believe I included this because the former PR had the yaml tags.

From my perspective there's no need for yaml since the prompts will be provided by the toml configuration and I don't believe they'll be marshalled to yaml from the MCP layer.

If you don't see any reason to include those, maybe we can simply remove the yaml layer.

pkg/config/config.go
```diff
@@ -74,6 +74,13 @@ type StaticConfig struct {
 	// This map holds raw TOML primitives that will be parsed by registered toolset parsers
 	ToolsetConfigs map[string]toml.Primitive `toml:"toolset_configs,omitempty"`
 
+	// Prompt configuration
+	// Raw TOML primitive for prompt definitions, parsed later
+	// Note: Uses toml:"-" because Primitive can't be encoded, only decoded
+	Prompts         toml.Primitive `toml:"-"`
+	promptsDefined  bool           // Internal: tracks if prompts were defined in config
+	promptsMetadata toml.MetaData  // Internal: metadata for prompts decoding
```
manusa, line 82: I don't quite grasp the limitation here.

I think it would be much clearer if StaticConfig only defined a Prompts []api.Prompt field.

I'll look into this because I'm sure I might be missing something, but the end goal would be that config-wise we didn't have to iterate over the prompts.

manusa: OK; I see we're still having problems with the cyclic dependencies and the kubernetes package. I'll try to properly address this, creating an issue now for tracking.

manusa: https://github.com/containers/kubernetes-mcp-server/issues/577

pkg/api/prompts.go
nader-ziada: i can remove this in a follow up PR

PR #564, opened 2025-12-10
refactor(api): add parameter extraction utilities for tool handlers

manusa:
Follows up on #533

Add RequiredString, OptionalString, and OptionalBool helper functions to reduce code duplication across toolsets. OptionalString and OptionalBool accept default values for cleaner call sites.

- Remove duplicate getRequiredString/getOptionalString/getOptionalBool from kubevirt create and lifecycle tools
- Fix hardcoded enum values in lifecycle tool to use Action constants
- Add comprehensive tests for new parameter utilities

### DIFF
--- a/pkg/api/params.go
+++ b/pkg/api/params.go
@@ -55,9 +55,6 @@ func OptionalString(params ToolHandlerParams, key, defaultVal string) string {
 	if !ok {
 		return defaultVal
 	}
-	if str == "" {
-		return defaultVal
-	}
 	return str
 }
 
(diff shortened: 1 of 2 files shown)


pkg/api/params.go
```diff
@@ -26,3 +26,52 @@ func ParseInt64(value interface{}) (int64, error) {
...
+	if !ok {
+		return defaultVal
+	}
+	str, ok := val.(string)
+	if !ok {
+		return defaultVal
+	}
+	if str == "" {
+		return defaultVal
+	}
```
Cali0707: from the above comment this section seems wrong - wouldn't we only want to default it if it was missing or not a string? If we get here then it was present and a string, just an empty string

manusa: Good catch! Changing.

PR #566, opened 2025-12-10
refactor(config)!: extract config interfaces to pkg/api/config

manusa:
At the moment, the config and kubernetes modules are tightly coupled.

While reviewing the changes in #559, the current state proves to be a problem since workarounds are required to be able to easily unmarshal the configuration.
Ideally, configuration unmarshalling and parsing should be part of the `pkg/config` package, while the api package should focus just on defining the structs and interfaces.

This PR attempts to decouple kubernetes from the config module by providing config API interfaces. `kubernetes->api->config->...`

I'm not proud of how it looks like, but I think that it's a step in the right direction.
Suggestions for further improvements are welcome ;)

---

Decouples the pkg/kubernetes module from pkg/config by introducing interfaces for configuration access. This allows pkg/kubernetes to depend on abstractions rather than the concrete StaticConfig type.

Changes:
- Add AuthProvider, ClusterProvider, DeniedResourcesProvider, and ExtendedProvider interfaces in pkg/api/config
- Move cluster provider strategy constants to pkg/api/config
- Move GroupVersionKind struct to pkg/api/config
- Update pkg/kubernetes to use the new interfaces
- StaticConfig now implements the Config interface

### DIFF
--- a/pkg/api/config/config.go
+++ b/pkg/api/config/config.go
@@ -1,0 +1,53 @@
+package config
+
+const (
+	ClusterProviderKubeConfig = "kubeconfig"
+	ClusterProviderInCluster  = "in-cluster"
+	ClusterProviderDisabled   = "disabled"
+)
+
+type AuthProvider interface {
+	// IsRequireOAuth indicates whether OAuth authentication is required.
+	IsRequireOAuth() bool
+}
+
+type ClusterProvider interface {
+	// GetClusterProviderStrategy returns the cluster provider strategy (if configured).
+	GetClusterProviderStrategy() string
+	// GetKubeConfigPath returns the path to the kubeconfig file (if configured).
+	GetKubeConfigPath() string
+}
+
+// Extended is the interface that all configuration extensions must implement.
+// Each extended config manager registers a factory function to parse its config from TOML primitives
+type Extended interface {
+	// Validate validates the extended configuration.  Returns an error if the configuration is invalid.
+	Validate() error
+}
+
+type ExtendedProvider interface {
+	// GetProviderConfig returns the extended configuration for the given provider strategy.
+	// The boolean return value indicates whether the configuration was found.
+	GetProviderConfig(strategy string) (Extended, bool)
+	// GetToolsetConfig returns the extended configuration for the given toolset name.
+	// The boolean return value indicates whether the configuration was found.
+	GetToolsetConfig(name string) (Extended, bool)
+}
+
+type GroupVersionKind struct {
+	Group   string `json:"group" toml:"group"`
+	Version string `json:"version" toml:"version"`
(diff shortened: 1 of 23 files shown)


pkg/api/config/config.go
```diff
@@ -0,0 +1,8 @@
+package config
+
```
nader-ziada, line 2: I would put all these interfaces in one file instead of multiple small files, maybe even in`pkg/api`

pkg/api/config/config.go
```diff
@@ -0,0 +1,8 @@
...
+type Config interface {
```
nader-ziada: how about calling this interface `Provider`, there are too many things called `Config`

pkg/api/config/resources.go
```diff
@@ -0,0 +1,12 @@
+package config
+
+type GroupVersionKind struct {
```
nader-ziada: I feel like this should be in `pkg/api/types.go`

pkg/api/config/config.go
manusa: Yes, definitely, but api still depends on pkg/kubernetes and that's a dependency that's not going to be easy to remove :(

Many of the methods defined in pkg/api have parameters that depend on this module

pkg/api/config/config.go
manusa: Agreed, I wasn't sure of the name. IMO Provider is not correct either.

How does `Base` sound?

pkg/api/config/resources.go
manusa: Yep, definitely. The reasoning is the same as for creating the separate pkg/api/config package.
So unfortunately, can't be changed either.

pkg/api/config/config.go
manusa: I will move every interface to the same file

pkg/api/config/config.go
nader-ziada: `BaseConfig` ok ?

nader-ziada:
thanks @manusa 

/lgtm

manusa:
@Cali0707 note that this will break downstream again :(

nader-ziada:
> @Cali0707 note that this will break downstream again :(

there is already another breaking change from the [drop-in reload PR](https://github.com/containers/kubernetes-mcp-server/pull/524) that is not synced downstream yet

Cali0707:
Yep I will do another sync later, will aim to do one per week

manusa:
> Yep I will do another sync later, will aim to do one per week

🙌 I'll try to get more involved in the process. I want to check before the end of the week to see if we can better reorganize upstream so that the syncing process could be streamlined even more.

Issue #577, opened 2025-12-11
[API] Decouple pkg/api from pkg/kubernetes to eliminate cyclic dependencies


manusa:
### Background

The current package architecture has evolved organically as the project grew, leading to problematic coupling between core packages. Originally, toolsets directly depended on `pkg/kubernetes`, but as the project matured, the dependency structure became inverted in a way that creates architectural friction.

### Problem Statement

The `pkg/api` package, which should be a foundational package containing type definitions and interfaces, has dependencies on `pkg/kubernetes`:

**Current problematic imports in `pkg/api`:**
- `pkg/api/toolsets.go` imports `pkg/kubernetes` (as `internalk8s`)
- `pkg/api/prompts.go` imports `pkg/kubernetes` (as `internalk8s`)

**Specific coupling points:**

1. **`ToolHandlerParams` struct** (`pkg/api/toolsets.go:67-72`):
   ```go
   type ToolHandlerParams struct {
       context.Context
       *internalk8s.Kubernetes  // ← Direct dependency on concrete type
       ToolCallRequest
       ListOutput output.Output
   }
   ```

2. **`PromptHandlerParams` struct** (`pkg/api/prompts.go:89-93`):
   ```go
   type PromptHandlerParams struct {
       context.Context
       *internalk8s.Kubernetes  // ← Direct dependency on concrete type
       PromptCallRequest
   }
   ```

3. **`Toolset.GetTools()` method** (`pkg/api/toolsets.go:46`):
   ```go
   GetTools(o internalk8s.Openshift) []ServerTool  // ← Uses interface from pkg/kubernetes
   ```

This creates a quasi-circular dependency pattern:
```
pkg/api/config (leaf - only interfaces, no deps)
    ↑
pkg/kubernetes (imports pkg/api/config)
    ↑
pkg/api (imports pkg/kubernetes) ← PROBLEM: api should be a leaf package
    ↑
pkg/toolsets/* (imports pkg/api AND pkg/kubernetes)
```

### Impact

1. **PR #559 (configurable prompts)**: Had to define TOML primitives in `pkg/config` instead of directly deserializing `api.Prompt` structures because of the coupling.

2. **Reduced reusability**: The `pkg/api` package cannot be used independently without pulling in the entire Kubernetes client stack.

3. **Testing complexity**: Testing API types requires setting up Kubernetes dependencies even when not needed.

4. **Architectural violation**: The API/types package should be a stable, dependency-free foundation that other packages build upon.

[...]

PR #578, opened 2025-12-11
refactor(api)!: replace internal Kubernetes references with api package interfaces

manusa:
Initial phases of #577

Sending this for review before going deeper into the rabbit hole and making changes even harder to review (I'll continue to work on the follow ups, but I prefer for you to check if we're going in the right direction).

With this PR we completely decouple the `pkg/api` from the rest of packages.

It's clear now that on the Provider<-Manager<-Kubernetes<-AccessControlClientSet chain (after this PR Provider<-Manager<-api.KubernetesClient<-api.KubernetesClientSet) the KubernetesClient and KubernetesClientSet interfaces and struct implementations are redundant and should be merged.
This will cause even more breaking changes in case there are downstream toolset implementations, but I believe it's worth the trouble since the API will be even cleaner.

### DIFF
--- a/internal/tools/update-readme/main.go
+++ b/internal/tools/update-readme/main.go
@@ -9,8 +9,8 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"github.com/containers/kubernetes-mcp-server/pkg/config"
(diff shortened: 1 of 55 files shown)


nader-ziada:
i'm good with this refactor, thanks @manusa

PR #579, opened 2025-12-11
feat: add support for mcp server instructions

Cali0707:
This PR adds support for mcp server instructions

The current thinking is that they can be set in the config file, and when we see a need for providing any we can add our own default value in the code. For now I didn't have a good default value so I left it empty.

### DIFF
--- a/pkg/mcp/mcp_test.go
+++ b/pkg/mcp/mcp_test.go
@@ -1,11 +1,14 @@
 package mcp
 
 import (
+	"context"
 	"net/http"
+	"net/http/httptest"
 	"sync"
 	"testing"
 
 	"github.com/containers/kubernetes-mcp-server/internal/test"
+	"github.com/mark3labs/mcp-go/client"
 	"github.com/mark3labs/mcp-go/client/transport"
 	"github.com/stretchr/testify/suite"
 )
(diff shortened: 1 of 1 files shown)


nader-ziada:
so currently its empty and there is no way to set it?

Cali0707:
> so currently its empty and there is no way to set it?

yeah with the current code (not with this PR) you can't set it and it is empty.

What this does is adds support for setting it through the config file(s) - if we identify a good default for that we can also set a default value

nader-ziada:
> What this does is adds support for setting it through the config file(s) - if we identify a good default for that we can also set a default value

should we add this when we have a use for it then? or is there a need to add this support

Cali0707:
> should we add this when we have a use for it then? or is there a need to add this support

We have a need for this downstream in the openshift MCP server when running this in an ACM scenario with the ACM Search MCP server enabled as well - I just wasn't sure what value made sense here upstream

manusa:
This PR **does enable** the possibility to set it via config file (although it's undocumented, so it's a preview feature by all means).

At some point we might also want to add the option for toolsets to override/modify this (but we should really think about how to implement since there is the challenge when multiple toolsets converge).

/lgtm

I created a follow-up PR to simplify the test suite and also highlight how the instructions can be actually set (also makes it consistent with other tests such as the denied resources)

PR #581, opened 2025-12-11
feat(api)!: add toolset prompts infrastructure

nader-ziada:
Fixes: #556 

Enable toolset implementers to define MCP prompts programmatically as part of their toolset definition.

- Add GetPrompts() method to Toolset interface
- Collect prompts from all enabled toolsets
- Merge config and toolset prompts (config takes precedence)
- Add disable_toolset_prompts configuration option
- Update all existing toolsets to implement GetPrompts()

### DIFF
--- a/docs/PROMPTS.md
+++ b/docs/PROMPTS.md
@@ -68,7 +68,7 @@ Toolsets can provide built-in prompts by implementing the `GetPrompts()` method.
 ### Implementing Toolset Prompts
 
 ```go
-func (t *MyToolset) GetPrompts(o internalk8s.Openshift) []api.ServerPrompt {
+func (t *MyToolset) GetPrompts() []api.ServerPrompt {
     return []api.ServerPrompt{
         {
             Prompt: api.Prompt{
@@ -102,18 +102,6 @@ func (t *MyToolset) GetPrompts(o internalk8s.Openshift) []api.ServerPrompt {
         },
     }
 }
-```
-
-### Disabling Toolset Prompts
-
-Add `disable_toolset_prompts = true` in your `config.toml` to use only config-defined prompts:
-
-```toml
-disable_toolset_prompts = true
-
-[[prompts]]
-name = "my-custom-prompt"
-# ... rest of config
 ```
 
 ### Prompt Merging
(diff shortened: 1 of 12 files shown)


docs/PROMPTS.md
```diff
@@ -59,4 +59,66 @@ Use `{{argument_name}}` placeholders in message content. The template engine rep
...
+    }
+}
+```
+
+### Disabling Toolset Prompts
+
+Add `disable_toolset_prompts = true` in your `config.toml` to use only config-defined prompts:
+
+```toml
+disable_toolset_prompts = true
```
Cali0707: @nader-ziada will this be a global option, or a per-toolset option?

nader-ziada: global option currently

pkg/api/toolsets.go
```diff
@@ -43,6 +43,9 @@ type Toolset interface {
 	// Will be used to generate documentation and help text.
 	GetDescription() string
 	GetTools(o Openshift) []ServerTool
+	// GetPrompts returns the prompts provided by this toolset.
+	// Returns nil if the toolset doesn't provide any prompts.
+	GetPrompts(o Openshift) []ServerPrompt
```
manusa: The `o Openshift` parameter should be removed.
This is a legacy thing that will also be removed from `GetTools` method eventually.

docs/PROMPTS.md
manusa: I believe I included this in https://github.com/containers/kubernetes-mcp-server/issues/556

> **Disabling Toolset Prompts**
>
> Administrators can disable all toolset-defined prompts via configuration, using only config-defined prompts:

based on what Nader initially implemented in #510 

IMO maybe we SHOULD NOT provide this option at all (prompts don't have any involuntary impact on the LLM behavior or consume context-window tokens)

OR if you consider this of value, provide a consistent implementation with what we have for tools (`EnabledTools` / `DisabledTools`).

nader-ziada: for the sake of simplicity, will remove the flag completely

pkg/api/toolsets.go
nader-ziada: will remove

PR #582, opened 2025-12-12
test(mcp): store initialization results in McpClient for better instruction handling

manusa:
Follows up on #579 to simplify the test


Cali0707 (approved):
LGTM, thanks for cleaning this up @manusa !

Issue #555, opened 2025-12-05
[RELEASE] Publish MCP server to the official Model Context Protocol registry


manusa:
## Description

Register and publish the kubernetes-mcp-server to the official [Model Context Protocol registry](https://github.com/modelcontextprotocol/registry). This will increase discoverability and provide a standardized way for users to find and install the MCP server.

### Background

The MCP Registry is the official registry for Model Context Protocol servers. Publishing to this registry requires:
1. Package metadata updates (npm and Python)
2. A `server.json` manifest file
3. GitHub Actions workflow to automate publishing on releases

### 1. npm Package Metadata Update

Modify `build/node.mk` to add the `mcpName` field to the generated `package.json`:

```json
{
  "mcpName": "io.github.containers/kubernetes-mcp-server"
}
```

This field is required by the MCP Registry to identify the server.

### 2. Python Package Metadata Update

The MCP Registry requires Python packages to include an `mcp-name` metadata field in the package README. This metadata is parsed by the registry to associate the PyPI package with the MCP server entry.

Add the following line to the project's root `README.md`:

```
mcp-name: io.github.containers/kubernetes-mcp-server
```

This line can be placed at the bottom of the README or in a dedicated metadata section. The Python package already uses the root README via `pyproject.toml`:

```toml
readme = {file="README.md", content-type="text/markdown"}
```

Since `build/python.mk` copies the root README during the publish process, no additional changes are needed to the Python build configuration.

### 3. Server Manifest (`server.json`)

Maintain a `server.json` file in the repository root with the following structure:

```json
{
  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json",
  "name": "io.github.containers/kubernetes-mcp-server",
  "description": "A Model Context Protocol (MCP) server for Kubernetes and OpenShift",
  "status": "active",
  "repository": {
    "url": "https://github.com/containers/kubernetes-mcp-server",
    "source": "github"
  },
  "version": "0.0.0",
  "packages": [
    {
      "registryType": "npm",
      "registryBaseUrl": "https://registry.npmjs.org",
      "identifier": "kubernetes-mcp-server",
      "version": "0.0.0",
      "transport": {
        "type": "stdio"
      }
    },
    {
      "registryType": "pypi",
      "registryBaseUrl": "https://pypi.org",
      "identifier": "kubernetes-mcp-server",
      "version": "0.0.0",
      "runtimeHint": "uvx",
      "transport": {
[...]

PR #584, opened 2025-12-12
feat(mcp-registry): add MCP Registry metadata and server manifest

manusa:
Part of #555 (1,2,3)

Add required metadata for publishing to the official Model Context Protocol (MCP) Registry:

- Add mcpName field to npm package.json generation
- Add mcp-name metadata to README.md for Python package
- Add io.modelcontextprotocol.server.name to container image
- Create server.json manifest with npm and PyPI package definitions

### DIFF
--- a/server.json
+++ b/server.json
@@ -1,0 +1,68 @@
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json",
+  "name": "io.github.containers/kubernetes-mcp-server",
+  "description": "A Model Context Protocol (MCP) server for Kubernetes and OpenShift",
+  "status": "active",
+  "repository": {
+    "url": "https://github.com/containers/kubernetes-mcp-server",
+    "source": "github"
+  },
+  "version": "0.0.0",
+  "packages": [
+    {
+      "registryType": "npm",
+      "registryBaseUrl": "https://registry.npmjs.org",
+      "identifier": "kubernetes-mcp-server",
+      "version": "0.0.0",
+      "transport": {
+        "type": "stdio"
+      }
+    },
+    {
+      "registryType": "pypi",
+      "registryBaseUrl": "https://pypi.org",
+      "identifier": "kubernetes-mcp-server",
+      "version": "0.0.0",
+      "runtimeHint": "uvx",
+      "transport": {
+        "type": "stdio"
+      }
+    },
+    {
+      "registryType": "oci",
+      "identifier": "quay.io/containers/kubernetes_mcp_server:0.0.0",
+      "version": "0.0.0",
+      "runtimeHint": "docker",
+      "runtimeArguments": [
+        {
+          "type": "named",
+          "name": "-v",
+          "description": "Volume mount for kubeconfig access (host_path:container_path)",
+          "format": "string",
+          "isRequired": true,
+          "placeholder": "~/.kube/config:/kubeconfig:ro"
+        },
+        {
+          "type": "named",
+          "name": "-e",
+          "description": "Environment variable to set the kubeconfig path inside the container",
+          "format": "string",
+          "isRequired": true,
+          "default": "KUBECONFIG=/kubeconfig"
+        },
+        {
+          "type": "named",
+          "name": "-p",
+          "description": "Port mapping for MCP server HTTP endpoint (host_port:container_port)",
+          "format": "string",
+          "isRequired": true,
+          "default": "8080:8080"
+        }
+      ],
+      "transport": {
+        "type": "streamable-http",
+        "url": "http://localhost:8080/mcp"
+      }
+    }
+  ]
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ COPY ./ ./
 RUN make build
 
 FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
+LABEL io.modelcontextprotocol.server.name="io.github.containers/kubernetes-mcp-server"
 WORKDIR /app
 COPY --from=builder /app/kubernetes-mcp-server /app/kubernetes-mcp-server
 USER 65532:65532
--- a/README.md
+++ b/README.md
@@ -568,3 +568,7 @@ make build
 # Run the Kubernetes MCP server with mcp-inspector
 npx @modelcontextprotocol/inspector@latest $(pwd)/kubernetes-mcp-server
 ```
+
+---
+
+mcp-name: io.github.containers/kubernetes-mcp-server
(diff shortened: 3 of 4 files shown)


server.json
```diff
@@ -0,0 +1,32 @@
...
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json",
+  "name": "io.github.containers/kubernetes-mcp-server",
+  "description": "A Model Context Protocol (MCP) server for Kubernetes and OpenShift",
+  "status": "active",
+  "repository": {
+    "url": "https://github.com/containers/kubernetes-mcp-server",
+    "source": "github"
+  },
+  "version": "0.0.0",
+  "packages": [
```
Cali0707, line 11: Maybe also include the OCI package type here? https://github.com/modelcontextprotocol/registry/blob/main/docs/modelcontextprotocol-io/package-types.mdx#dockeroci-images

Cali0707: Noticing that this will require us to ensure that we label our image correctly: `LABEL io.modelcontextprotocol.server.name="io.github.username/kubernetes-manager-mcp"`

manusa: The problem with this one is that you'll need to add the volume mounts to the local kubeconfig or add extra configuration so that the MCP is functional.

I'm not sure how'd this work in the scope of the registry and the consumer MCP Hosts.

Cali0707: Yeah that's a good point - maybe we can upstream this question to the registry itself? Seems like a reasonable use case

manusa: I added the LABEL to the container file since this PR is mostly to cover the prerequisites to be able to publish the processed `server.json`.

The [issue](https://github.com/containers/kubernetes-mcp-server/issues/555) will remain open until the publishing is complete.

This will also require removing a v1.0.0 of the MCP server that was published by accident: https://github.com/modelcontextprotocol/registry/issues/104

> Yeah that's a good point - maybe we can upstream this question to the registry itself? Seems like a reasonable use case

Maybe that's something to try.
However, this is a particular case for MCP servers such as ours which rely on local files to automatically set up their configuration.

The [other (and less powerful) Kubernetes MCP server](https://github.com/Flux159/mcp-server-kubernetes/) has a hard requirement [on a kubecontext file or passing its content through an environment variable](https://github.com/Flux159/mcp-server-kubernetes/blob/main/ADVANCED_README.md).

In the Docker MCP hub there's a note at the end on how to use the MCP server:
https://hub.docker.com/mcp/server/kubernetes/overview

What I see is that by providing the 3 options, some clients might opt to use the container package (which is the safest) which won't work in most cases.
My concern is that these users might be driven away from the MCP server which essentially defeats the purpose of having it published in a registry.

Cali0707: >  My concern is that these users might be driven away from the MCP server which essentially defeats the purpose of having it published in a registry.

That's a good point. For now let's stick with not using the OCI package then, but ideally we should find a way (perhaps with the upstream registry folks) to specify this config.

server.json
```diff
@@ -0,0 +1,32 @@
...
+      "version": "0.0.0",
+      "transport": {
+        "type": "stdio"
+      }
+    },
+    {
+      "registryType": "pypi",
+      "registryBaseUrl": "https://pypi.org",
+      "identifier": "kubernetes-mcp-server",
+      "version": "0.0.0",
```
Cali0707, line 25: @manusa should these be the same version as our most recent release? or is the idea to replace that version somewhere else?

manusa: This is reflected in the [issue](https://github.com/containers/kubernetes-mcp-server/issues/555): 

> The version fields use a placeholder value (0.0.0) that will be replaced by the workflow during publishing.


The idea is that the pipeline will replace the value with the tagged version (similar to what's done for the wrapper [python](https://github.com/containers/kubernetes-mcp-server/blob/cad863ff2246cacb9f2964f3b8ee35c02127736f/python/pyproject.toml#L7) and npm packages).

This way there's no need to keep updating the version for each release.

server.json
dmartinol: > Noticing that this will require us to ensure that we label our image correctly: `LABEL io.modelcontextprotocol.server.name="io.github.username/kubernetes-manager-mcp"`

Yes, this is required for the upstream repo.
@rdimitrov will this also be part of ToolHive registry validations? (e.g. https://github.com/stacklok/toolhive-registry-server/issues/265)

Cali0707 (approved):
LGTM

Up to you @manusa when you want to merge here

server.json
dmartinol: > The problem with this one is that you'll need to add the volume mounts to the local kubeconfig or add extra configuration so that the MCP is functional.

Are you planning to deploy it locally or in-cluster?
In the latter case, I image it uses the in-cluster config, right?

For those deploying the local server, you could specify the config either as an env variable:
```yaml
          "environmentVariables": [
            {
              "name": "KUBECONFIG",
              "description": "Optional path to kubeconfig file (used with --kubeconfig flag)",
              "format": "string",
              "isSecret": false
            }
```

or using the `--kubeconfig` option:
```yaml
          "packageArguments": [
            {
              "type": "named",
              "name": "--kubeconfig",
              "description": "Optional path to kubeconfig file inside container",
              "format": "filepath",
              "isRequired": false,
              "default": "/root/.kube/config"
            },
            {
              "type": "named",
              "name": "-v",
              "description": "Optional volume mount specification",
              "format": "string",
              "isRequired": false,
              "placeholder": "/host/path:/container/path"
            }
          ],
```
In both cases the `-v` is requested as a `packageArguments`.

> 
> I'm not sure how'd this work in the scope of the registry and the consumer MCP Hosts.

I hope it is clear that the registry only exposes the server metadata, and it's the admin responsibility to deploy it according to the provided instructions (e.g. start a docker container or a K8s Deployment). 
The Anthropic registry is immutable, so you can't add the "remotes" field to the registered server after deployment. Instead, once deployed with ToolHive as an MCPServer instance, it will be automatically discovered by the ToolHive registry and the endpoint published to be consumed:
```yaml
      "remotes": [
        {
          "type": "streamable-http",
          "url": "<deployment>/<namespace>.svc.cluster.local",
          ]
        },
```

manusa: Thanks for the detailed information ❤️ 

> > The problem with this one is that you'll need to add the volume mounts to the local kubeconfig or add extra configuration so that the MCP is functional.
>
> Are you planning to deploy it locally or in-cluster?

Here I'm only considering the local deployment scenario.
For in-cluster, we already provide Helm charts and other docs. In fact, the challenge for in-cluster deployment is mostly about exposing the MCP server in a secure way.

All comments refer to local deployment.

> I hope it is clear that the registry only exposes the server metadata, and it's the admin responsibility to deploy it according to the provided instructions (e.g. start a docker container or a K8s Deployment).

Yes, that's why I have the concern with the container image package. 
The goal is that users can use the registry to discover the MCP server and are able to run it flawlessly.

With the NPM and Python package wrappers, we know for certain that if the user has a kubeconfig file, it will work straight away.
However, for the container image deployment, the volume and kubeconfig CLI flags will need to be set.

I think that if we add the suggested packageArguments, at least users will be aware that they need some extra-config to make things work.

I'm not sure of how clients (MCP Hosts) decide on the packages precedence when setting up an MCP server for a user from the registry. Maybe this is also something to consider registry-wise.

manusa:
> LGTM
> 
> Up to you @manusa when you want to merge here

I believe I've taken care of the OCI package now. I'm still not confident it will work well, but we can remove it later on if needed.
I also updated the issue to reflect the implemented spec.

Please take another look if you can.

manusa:
I'm merging this now, I want to cut a release with prompts before we add the open telemetry and token exchange refactors.

PR #585, opened 2025-12-12
refactor(api)!: consolidate AccessControlClientset into Kubernetes struct

manusa:
Follows up on #578 
Second phase of #577 

After these changes the only thing left will be to clean up the api.KubernetesClient interface to not expose the core methods (ResourcesList, ResourcesGet, NamespacesList, etc.)
This will leave a very clean api.KubernetesClient interface powerful enough to implement any kind of toolset.

---

The AccessControlClientset was an unnecessary layer of indirection between the KubernetesClient interface and the Kubernetes implementation. This change merges the AccessControlClientset directly into the Kubernetes struct, simplifying the architecture and reducing complexity.

- Merge KubernetesClientSet interface into KubernetesClient
- Move all AccessControlClientset methods to Kubernetes struct
- Delete accesscontrol.go and accesscontrol_client_set.go
- Rename NewAccessControlClientset to NewKubernetes
- Update all callers to use Kubernetes directly

BREAKING CHANGE: AccessControlClientset() method removed from KubernetesClient interface. Callers should use the KubernetesClient methods directly.


nader-ziada:
thanks for more simplification @manusa, this makes the code clearer and more straight forward

PR #434, opened 2025-11-03
prompt(core): add cluster health check

ropatil010:
Hi Team, 

Can you PTAL on this. 

PR about:
  Implements comprehensive cluster health check tool that examines:
  - Cluster operators (OpenShift)
  - Nodes (readiness, schedulability, resource pressure)
  - Pods (failures, crash loops, image pull errors, high restarts)
  - Workload controllers (Deployments, StatefulSets, DaemonSets)
  - Storage (PVC status)
  - Recent warning events

  Features:
  - Text and JSON output formats
  - Verbose mode for detailed diagnostics
  - Configurable event checking
  - Clear severity indicators (critical/warning/healthy)

PR #591, opened 2025-12-17
test(prompts): split notification tests to assert specific list changes

manusa:
- enhance WaitForNotification to accept specific method for notifications

### DIFF
--- a/pkg/mcp/mcp_watch_test.go
+++ b/pkg/mcp/mcp_watch_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/BurntSushi/toml"
 	"github.com/containers/kubernetes-mcp-server/internal/test"
 	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/stretchr/testify/suite"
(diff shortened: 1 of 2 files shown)


pkg/mcp/mcp_watch_test.go
```diff
@@ -40,13 +56,22 @@ func (s *WatchKubeConfigSuite) TestNotifiesToolsChange() {
...
 	s.NotNil(notification, "WatchKubeConfig did not notify")
-	s.True(
-		notification.Method == "notifications/tools/list_changed" || notification.Method == "notifications/prompts/list_changed",
-		"WatchKubeConfig did not notify tools or prompts change, got: %s", notification.Method,
-	)
+	s.Equal("notifications/tools/list_changed", notification.Method, "WatchKubeConfig did not notify tools change")
+}
+
+func (s *WatchKubeConfigSuite) TestNotifiesPromptsChange() {
+	s.T().Skip("Enable when prompt unmarshalling is fixed") // TODO: Fix prompt unmarshalling issue
```
nader-ziada, line 66: if there is an issue with the prompt unmarshalling, can we fix it as part of the same PR here?

manusa: I will cover that later.
The idea is to simplify the Config unmarshalling, but I haven't analyzed what that would require.
The main purpose of these changes is to unlock #434 which has some changes in this test and shouldn't.

PR #604, opened 2025-12-19
feat: add token exchange provider support

Cali0707:
This PR adds a tokenexchange package as well as the ability for providers to provide target specific token exchange config.

As a follow up, after adding more tests to verify existing behaviour, I plan on removing the `sts.go`/`sts_test.go` files and move to just using our RFC6893 token exchanger. The STS files currently use the google external account package to do a token exchange, which does not meet all of our requirements for token exchange (e.g. the keycloak v1 flow). I hope that we can consolidate all this into a single unified impl, and unify the config as well after we stabilize these changes a bit.

### DIFF
--- a/pkg/http/authorization.go
+++ b/pkg/http/authorization.go
@@ -9,7 +9,6 @@ import (
 	"github.com/coreos/go-oidc/v3/oidc"
 	"github.com/go-jose/go-jose/v4"
 	"github.com/go-jose/go-jose/v4/jwt"
-	"golang.org/x/oauth2"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/strings/slices"
 
@@ -50,16 +49,7 @@ func write401(w http.ResponseWriter, wwwAuthenticateHeader, errorType, message s
 //	         - The token is then validated against the OIDC Provider.
 //
 //	         see TestAuthorizationOidcToken
-//
-//	    2.3. OIDC Token Exchange (oidcProvider is not nil, StsClientId and StsAudience are set):
-//	         - The token is validated offline for basic sanity checks (audience and expiration).
-//	         - If OAuthAudience is set, the token is validated against the audience.
-//	         - The token is then validated against the OIDC Provider.
-//	         - If the token is valid, an external account token exchange is performed using
-//	           the OIDC Provider to obtain a new token with the specified audience and scopes.
-//
-//	         see TestAuthorizationOidcTokenExchange
-func AuthorizationMiddleware(staticConfig *config.StaticConfig, oidcProvider *oidc.Provider, httpClient *http.Client) func(http.Handler) http.Handler {
+func AuthorizationMiddleware(staticConfig *config.StaticConfig, oidcProvider *oidc.Provider) func(http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			if r.URL.Path == healthEndpoint || slices.Contains(WellKnownEndpoints, r.URL.EscapedPath()) {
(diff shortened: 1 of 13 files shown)


pkg/http/authorization.go
```diff
@@ -50,16 +49,7 @@ func write401(w http.ResponseWriter, wwwAuthenticateHeader, errorType, message s
...
-//	    2.3. OIDC Token Exchange (oidcProvider is not nil, StsClientId and StsAudience are set):
-//	         - The token is validated offline for basic sanity checks (audience and expiration).
-//	         - If OAuthAudience is set, the token is validated against the audience.
-//	         - The token is then validated against the OIDC Provider.
-//	         - If the token is valid, an external account token exchange is performed using
-//	           the OIDC Provider to obtain a new token with the specified audience and scopes.
-//
-//	         see TestAuthorizationOidcTokenExchange
-func AuthorizationMiddleware(staticConfig *config.StaticConfig, oidcProvider *oidc.Provider, httpClient *http.Client) func(http.Handler) http.Handler {
+func AuthorizationMiddleware(staticConfig *config.StaticConfig, oidcProvider *oidc.Provider) func(http.Handler) http.Handler {
```
manusa, line 52: One of the things I don't like about the new approach is that now all the authorization bits are scattered around and the authorization flow is less clear.

However, I assume that this is how it should be. As I understand there are 2 parts:
- HTTP authentication/authorization: JWT validation (offline), OIDC validation (online)
- Token exchange: applies to the Kubernetes layer and for the Derived client provider.

Cali0707: Yes that is the idea

pkg/kubernetes/sts_test.go
manusa, line 1: The tests defined here are supposed to be black-boxed, they should serve to verify the new implementation in the tokenexchange module.

pkg/kubernetes/token_exchange.go
```diff
@@ -0,0 +1,81 @@
...
+	"strings"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/config"
+	"github.com/containers/kubernetes-mcp-server/pkg/tokenexchange"
+	"github.com/coreos/go-oidc/v3/oidc"
+	"golang.org/x/oauth2"
+	"k8s.io/klog/v2"
+)
+
+func ExchangeTokenInContext(
```
manusa, line 15: I haven't looked into this, but it seems like having this in the MCP layer has not much sense.

As I see it the HTTP layer sets the Bearer token header that gets propagated to the MCP layer through the context.

Then in the MCP layer we perform the token exchange by invoking provider methods to replace the Bearer token header with the exchanged token.

Then we call GetDerivedKubernetes on the provider to retrieve the applicable Kubernets client-set.

IMO at this point, the MCP layer could just call GetDerived and don't have to deal with auth-related tasks that have nothing to do with the MCP domain.

Cali0707: the reason why I added this into the MCP layer is that the token exchange relies on MCP tool call params (in particular the target for the tool). This means that we need to parse the MCP tool call request to extract the target before we can know how to do the token exchange.

If you prefer, we can move this back to the http layer, this will just require us manually parsing the tool call requests (at least the parameter) in the HTTP layer

manusa (commented):
Overall looks good, thanks!

As mentioned in a comment, It'd be nicer (IMO) if the mcp layer didn't get polluted with auth-related variables (oidcProvider, httpClient) or token exchange orchestration.

The tokenexchange module looks fine. It'd be good if we had some tests to verify it works as expected. I believe that the `sts_test.go` tests should verify that the compatible exchanger works.

The original `SecurityTokenService` struct was already designed with something like what you did in mind, it's currently using google externalaccount, but the body of the function was to be replaced with our own implementation at some point (it's mostly following what's defined in the `TokenExchanger` interface).

pkg/kubernetes/token_exchange.go
manusa: No, that's worse. 
```golang
cluster := toolCallRequest.GetString(s.p.GetTargetParameterName(), s.p.GetDefaultTarget())
ctx = kubernetes.ExchangeTokenInContext(ctx, s.configuration.StaticConfig, s.oidcProvider, s.httpClient, s.p, cluster)
k, err := s.p.GetDerivedKubernetes(ctx, cluster)
```
Is this the part you mean?
GetDerivedKubernetes is already receiving the cluster

matzew:
LGTM. 
Perhaps we can figure a better layering in a follow up PR. 

Perhaps same for refactoring the testing on token excange, as suggested as well?

/cc @Cali0707

PR #588, opened 2025-12-15
refactor!: remove kubernetes API server token validation

Cali0707:
Resolves #587 

This PR removes the token validation done against the Kubernetes API server, as the assumption is that the subsequent request for an actual operation will surface the error.

PR #614, opened 2025-12-23
refactor(test): use metav1 types for mock server discovery handlers

manusa:
While reviewing https://github.com/openshift/openshift-mcp-server/pull/87/changes
I recalled that we already had some work underway to make the mock server `DiscoveryClientHandler` extensible and more reusable.

This was used by the `TokenReviewHandler` that was removed in #588

With the current changers, users/developers should be able to define additional kubernetes API resources easily.

---

#### refactor(test): reuse DiscoveryClientHandler in InOpenShiftHandler

InOpenShiftHandler now embeds DiscoveryClientHandler instead of
duplicating the /api and /apis endpoint handling logic. Added a
constructor function to initialize the OpenShift-specific API groups.

#### refactor(test): use metav1 types for mock server discovery handlers

- Add NewDiscoveryClientHandler constructor with default resources
- Add AddAPIResourceList method for dynamic resource additions
- Auto-generate /api, /apis, and /apis/<group>/<version> endpoints
- Simplify NewInOpenShiftHandler to a constructor with variadic params
- Update all test files to use the new declarative API

### DIFF
--- a/internal/test/mock_server.go
+++ b/internal/test/mock_server.go
@@ -3,7 +3,6 @@ package test
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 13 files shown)


Cali0707 (approved):
LGTM

This is way clearer when reading the tests, and easier to extend where needed. Thanks @manusa !

PR #616, opened 2025-12-23
chore(ci): restrict release workflows to version tags only

manusa:
Change tag filter from '*' to 'v*' so release workflows only trigger for semantic version tags (e.g., v1.0.0) and ignore other tags.


Cali0707 (approved):
LGTM

PR #617, opened 2025-12-23
test(kiali): contract tests

josunect:
Automated Consistency Tests: Verify MCP Tool Definitions against Kiali API Routes

How do we know that APIs used by the consumer (kiali mcp) still exist and return what it expects?

- Kiali mcp declares what is using
- It is expected to run these tests by Kiali to verify those expectations in CI

Contract tests keep coupling low because they only describe: 

- endpoints
- request/responses
- status codes

TODO: 

- [x] Kiali CI: Use these tests: https://github.com/kiali/kiali/pull/9018 (It will be merged after this PR)
- [x] A test that validates that all the routes declared in `endpoints.go` are tested

### DIFF
--- a/pkg/kiali/tests/backend/api_backend_test.go
+++ b/pkg/kiali/tests/backend/api_backend_test.go
@@ -1,5 +1,5 @@
-//go:build kiali-contract
-// +build kiali-contract
+//go:build kiali_contract
+// +build kiali_contract
 
 package backend
 


aljesusg:
cc @Cali0707 Can we merge this test validation?

josunect:
> This seems reasonable. However, I fail to see why the backend tests are maintained here if they are executed from a different repo.

That's because they are testing the endpoints implemented by the other repo (Maintained by the same team). So, if any endpoint is modified, the tests will fail, and the way to fix it would be: 

- Or update the endpoints used by Kiali MCP accordingly (And the tests)
- Or version the Kiali endpoints to keep compatibility and don't break the Kiali MCP

pkg/kiali/tests/backend/api_backend_test.go
```diff
@@ -0,0 +1,683 @@
+//go:build contract
```
Cali0707: nit: can we make this something like kiali-contract ? Just to be a bit more explicit about _which_ contract we are testing

PR #624, opened 2026-01-07
fix: strip http forwarded headers from requests to kc

Cali0707:
While testing the server deployed in an openshift cluster, the wellknown proxy to keycloak was not working correctly, as all the various forwarded headers led keycloak to set the base url to the url to the MCP server, not the keycloak server.

This PR just strips those forwarded headers before passing requests to the keycloak server.

### DIFF
--- a/pkg/http/wellknown.go
+++ b/pkg/http/wellknown.go
@@ -57,6 +57,13 @@ func (w WellKnown) ServeHTTP(writer http.ResponseWriter, request *http.Request) 
 		return
 	}
 	for key, values := range request.Header {
+		// Skip headers that would cause Keycloak to use the proxy's hostname in URLs
+		lowerKey := strings.ToLower(key)
+		if lowerKey == "host" ||
+			strings.HasPrefix(lowerKey, "x-forwarded-") ||
+			lowerKey == "forwarded" {
+			continue
+		}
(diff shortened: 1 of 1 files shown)


manusa:
I need to properly check this. I recall that the header forwarding was added precisely for a proxy-related issue.
Do you have some environment to try or a way to reproduce the issue?

Cali0707:
>  Do you have some environment to try or a way to reproduce the issue?

To reproduce, you can install keycloak + the MCP server in a OpenShift cluster (behind routes), and then try to authenticate

I was using the make targets in https://github.com/openshift/openshift-mcp-server/pull/80

manusa:
> To reproduce, you can install keycloak + the MCP server in a OpenShift cluster (behind routes), and then try to authenticate


OK, I've checked the history.
This was added for the MCP inspector and the CORS issue #406

Was not present in the early implementation when everything was tested with the OpenShift setup.

I assume that the problem is that the URLs from the Well Known document now point to the MCP URL instead of that of the exposed Keycloak endpoints.

Cali0707:
> I assume that the problem is that the URLs from the Well Known document now point to the MCP URL instead of that of the exposed Keycloak endpoints.

Yes, that is exactly the issue this is trying to resolve

PR #629, opened 2026-01-08
feat(mcp): implement logging capability with error handling and secret redaction

nader-ziada:
Add MCP logging support to enable debug information flow to clients while maintaining security.

- Enable MCP logging capability in server initialization
- Add generic K8s error handler with user-friendly messages for all error types
- Implement logging with dedicated "mcp" logger for client-facing logs
- Add automatic sanitization for sensitive data (tokens, keys, passwords, cloud credentials)
- Integrate error handling in all core toolsets (pods, resources, nodes, events, namespaces)
- Add authentication/authorization failure logging with secure messages

### DIFF
--- a/pkg/mcplog/log.go
+++ b/pkg/mcplog/log.go
@@ -1,0 +1,167 @@
+package mcplog
+
+import (
+	"context"
+	"regexp"
+
+	"github.com/go-logr/logr"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"k8s.io/klog/v2"
+)
+
+// ContextKey is a type for context keys to avoid collisions
+type ContextKey string
+
+// MCPSessionContextKey is the context key for storing MCP ServerSession
+const MCPSessionContextKey = ContextKey("mcp_session")
+
+// Level represents MCP log severity levels per RFC 5424 syslog specification.
+// https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/logging#log-levels
+type Level int
+
+// Log levels from least to most severe, per MCP specification.
+const (
+	// LevelDebug is for detailed debugging information.
+	LevelDebug Level = iota
+	// LevelInfo is for general informational messages.
+	LevelInfo
+	// LevelNotice is for normal but significant events.
+	LevelNotice
+	// LevelWarning is for warning conditions.
+	LevelWarning
+	// LevelError is for error conditions.
+	LevelError
+	// LevelCritical is for critical conditions.
+	LevelCritical
+	// LevelAlert is for conditions requiring immediate action.
+	LevelAlert
+	// LevelEmergency is for system unusable conditions.
+	LevelEmergency
+)
+
+// levelStrings maps Level values to their MCP protocol string representation.
+var levelStrings = [...]string{
+	LevelDebug:     "debug",
+	LevelInfo:      "info",
+	LevelNotice:    "notice",
+	LevelWarning:   "warning",
+	LevelError:     "error",
+	LevelCritical:  "critical",
+	LevelAlert:     "alert",
+	LevelEmergency: "emergency",
+}
+
+// String returns the MCP protocol string representation of the level.
+func (l Level) String() string {
+	if l >= 0 && int(l) < len(levelStrings) {
+		return levelStrings[l]
+	}
+	return "debug"
+}
+
+var (
+	// mcpLogger is a dedicated named logger for MCP client-facing logs
+	// This provides complete separation from server logs
+	// issue for the sdk to implement this https://github.com/modelcontextprotocol/go-sdk/issues/748
+	mcpLogger logr.Logger = klog.NewKlogr().WithName("mcp")
+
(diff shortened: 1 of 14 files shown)


pkg/mcplog/log.go
```diff
@@ -0,0 +1,122 @@
...
+		// Database connection strings
+		regexp.MustCompile(`(postgres://[^:]+:)([^@]+)(@)`),
+		regexp.MustCompile(`(mysql://[^:]+:)([^@]+)(@)`),
+		regexp.MustCompile(`(mongodb(\+srv)?://[^:]+:)([^@]+)(@)`),
+	}
+)
+
+func sanitizeMessage(msg string) string {
+	// JSON/YAML field patterns (indices 0-6) - preserve field name
+	for i := 0; i < 7 && i < len(sensitivePatterns); i++ {
```
matzew, line 113: I guess there hard coded checks seem a bit funky ? :sweat_smile:

matzew:
I noticed that for instance on `helm.go` (e.g. in the `helmList`) there is no usage of that 

Would this be a follow up?

matzew:
Overall 
LGTM

nader-ziada:
> I noticed that for instance on `helm.go` (e.g. in the `helmList`) there is no usage of that
> 
> Would this be a follow up?

I probably missed that one, will fix

pkg/mcplog/log.go
nader-ziada: yeah, I feel the same, but couldn't think of something better, will take another pass at it

Cali0707: Would this not be one of the sanitizations that the MCP gateway would handle? (Although I get that that depends on a specific deployment topology, so maybe not the best fix for here)

pkg/mcplog/log.go
```diff
@@ -0,0 +1,122 @@
...
+		msg = sensitivePatterns[i].ReplaceAllString(msg, `[REDACTED]`)
+	}
+
+	return msg
+}
+
+// SendMCPLog sends a log notification to the MCP client and server logs.
+// Uses dedicated "mcp" named logger. Message is automatically sanitized.
+// Level: "debug", "info", "notice", "warning", "error", "critical", "alert", "emergency"
+func SendMCPLog(ctx context.Context, level, message string) {
```
manusa: Maybe we can use a specific type and some constants to make avoid errors

```
type Level int


const (
  Error Level = iota
  Critical
  Alert
  Emergency
)
```

For the levels, maybe reproduce whatever the protocol defines: https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/logging#log-levels

pkg/mcplog/log.go
manusa: The spec states that the server should take care of this: https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/logging#implementation-considerations

However, I believe this should be part of the SDK itself, in addition to our own implementation, we might want to open an issue for this.

It's my understanding that most servers will deal with the same kind of logic to remove sensitive information.

pkg/mcplog/log.go
```diff
@@ -0,0 +1,122 @@
...
+
+// MCPSessionContextKey is the context key for storing MCP ServerSession
+const MCPSessionContextKey = ContextKey("mcp_session")
+
+var (
+	// mcpLogger is a dedicated named logger for MCP client-facing logs
+	// This provides complete separation from server logs
+	mcpLogger logr.Logger = klog.NewKlogr().WithName("mcp")
+
+	// Patterns for redacting sensitive data from industry-standard secret detection tools
```
manusa: This comment states that these patterns are from industry-standard secret detection tools. It'd be better if we could include references for maintenance and future updates

pkg/toolsets/core/events.go
```diff
@@ -42,6 +43,7 @@ func eventsList(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
 	}
 	eventMap, err := kubernetes.NewCore(params).EventsList(params, namespace.(string))
 	if err != nil {
+		mcplog.HandleK8sError(params.Context, err, "events listing")
```
manusa, line 46: This pattern seems brittle.
Implementors need to take care of handling the error which might be eventually forgotten.
As a follow up we should try to find a (decoupled) way of integrating the logging directly in the Kubernetes package or some other alternative that would not require the extra-handling for each tool function

pkg/mcplog/log.go
nader-ziada: created an issue https://github.com/modelcontextprotocol/go-sdk/issues/748

manusa: awesome, thx!

pkg/mcplog/log.go
nader-ziada: its not coming from one place, I kind of looked in multiple places plus did some research, so don't really have one reference to link here. will remove the comment so its not misleading, but will mention the sdk issue as a future replacement

matzew:
LGTM!

PR #630, opened 2026-01-09
fix(http): add CORS headers for well-known endpoints

### LINKED ISSUE #625: Investigate Well Known middleware header forwarding
Currently in https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/http/wellknown.go we forward all headers (except for the "forwarded" headers) to the IdP. This was initially done to make CORS work in the OAuth flow.

We should investigate specifically which headers need to be forwarded in this flow, and only forward those.

manusa:
Fixes #625
Relates to #406
Relates to #624

- Stop propagating client headers to authorization server requests
- Add explicit CORS headers (Access-Control-Allow-Origin, Methods, Headers)
- Preserve backend response header propagation with CORS override
- Migrate well-known tests to testify/suite in dedicated test file
- Export kubeconfig to _output during kind cluster creation

### DIFF
--- a/pkg/http/wellknown.go
+++ b/pkg/http/wellknown.go
@@ -87,9 +87,13 @@ func (w WellKnown) ServeHTTP(writer http.ResponseWriter, request *http.Request) 
 	}
 	writer.Header().Set("Content-Type", "application/json")
 	writer.Header().Set("Content-Length", fmt.Sprintf("%d", len(body)))
+	withCORSHeaders(writer)
+	writer.WriteHeader(resp.StatusCode)
+	_, _ = writer.Write(body)
+}
+
+func withCORSHeaders(writer http.ResponseWriter) {
 	writer.Header().Set("Access-Control-Allow-Origin", "*")
 	writer.Header().Set("Access-Control-Allow-Methods", "GET, OPTIONS")
 	writer.Header().Set("Access-Control-Allow-Headers", "Content-Type, Authorization")
-	writer.WriteHeader(resp.StatusCode)
-	_, _ = writer.Write(body)
 }


pkg/http/wellknown.go
```diff
@@ -97,7 +85,11 @@ func (w WellKnown) ServeHTTP(writer http.ResponseWriter, request *http.Request)
 			writer.Header().Add(key, value)
 		}
 	}
+	writer.Header().Set("Content-Type", "application/json")
 	writer.Header().Set("Content-Length", fmt.Sprintf("%d", len(body)))
+	writer.Header().Set("Access-Control-Allow-Origin", "*")
```
matzew, line 96: nit: I guess we could encapsulate those CORS headers into some helper, like `func writeCORSHeaders` ?

pkg/http/http_test.go
```diff
@@ -337,121 +337,6 @@ func TestWellKnownReverseProxy(t *testing.T) {
 	})
 }
 
-func TestWellKnownHeaderPropagation(t *testing.T) {
```
matzew, line 340: nice to move out to dedicated file!

pkg/http/wellknown.go
manusa: I personally don't like the helper method approach in this case. It's a function with side-effects and doesn't really simplify things (in this case we're replacing 3 quite semantic lines with 6 lines with an ad-hoc method that isn't reused elsewhere).

Nonetheless, I added a subsequent commit with these changes ;)

pkg/http/http_test.go
manusa: I'll follow up with further refactor to move the entire suite to testify. Some tech debt pending for weeks now.

matzew:
Great!

LGTM

Cali0707:
@manusa when I tried running with this code I saw CORS failed errors from the inspector with firefox (and could not get the OAuth flow to work):

Screenshot 2026-01-09 at 08 47 12

manusa:
hmm, didn't try this one with Firefox, let me give it a spin. Was working fine from Chrome.

manusa:
> @manusa when I tried running with this code I saw CORS failed errors from the inspector with firefox (and could not get the OAuth flow to work):
> 
> Screenshot 2026-01-09 at 08 47 12

Was this working in firefox before? I notice that Firefox is performing an HTTP**S** connection as opposed to Chrome.

Cali0707:
> Was this working in firefox before? I notice that Firefox is performing an HTTP**S** connection as opposed to Chrome.

@manusa it seems to not be working in firefox with code built from `main` either -  however back when we tested this before (when we did the initial CORS fix) it did work there. Probably some combo of newer firefox version/newer MCP inspector version...

manusa:
> > Was this working in firefox before? I notice that Firefox is performing an HTTP**S** connection as opposed to Chrome.
> 
> @manusa it seems to not be working in firefox with code built from `main` either - however back when we tested this before (when we did the initial CORS fix) it did work there. Probably some combo of newer firefox version/newer MCP inspector version...

OK, I'm having a complicated day today.

I had wrong settings in Firefox and also tried to perform the client registration from the inspector (which won't work).


I'm using Firefox 140.6.0esr (macOS).

The flow works correctly with some nuances:
- You need to first go into the keycloak HTTPS web interface to add the self-signed certificate exception
- You need to set the Client ID to `mcp-client` otherwise, client registration is attempted and won't work (same CORS issue, should not be a problem with conventional MCP clients/hosts)
  Client Registration also fails in Chrome.

If the settings are fine, everything should work:

image

Cali0707:
@manusa for setting that exception in the UI - let's see if we can add that into the keycloak setup in a separate PR

manusa:
> @manusa for setting that exception in the UI - let's see if we can add that into the keycloak setup in a separate PR

I didn't explain this correctly.

It's not something you add to keycloak.

It's an exception you add to Firefox to acknowledge the security risk of accessing a TLS protected site with a self-signed certificate (not present in the local CA list):

image

`Accept the Risk and Continue`

image

PR #631, opened 2026-01-12
build: Add Podman support for local development

lyarwood:
Enhance the local development environment to fully support Podman as an alternative to Docker, with improved container engine detection and fixes for Kind cluster networking.

Container Engine Detection:
- Verify that docker or podman are actually running before selecting them by checking `docker info` / `podman info`
- Prevents errors when a container engine is installed but not active
- Consistently use CONTAINER_ENGINE variable throughout makefiles

Keycloak Connectivity:
- Add curl --resolve flag to bypass DNS and connect directly to localhost
- Ensures reliable connectivity to Keycloak running in the Kind cluster
- Prevents DNS-related connection failures

Kind Cluster Management:
- Set KIND_EXPERIMENTAL_PROVIDER inline when using podman on Linux
- Simplify cluster creation and deletion logic
- Consistently use CONTAINER_ENGINE for all container operations

Ingress Controller Fixes:
- Enable hostNetwork mode for nginx ingress controller
- Required for proper port binding with Podman in Kind clusters
- Limit nginx worker processes to 2 to avoid pthread resource exhaustion
- Remove redundant hostPort bindings (not needed with hostNetwork)
- Set dnsPolicy to ClusterFirstWithHostNet for proper name resolution

These changes enable seamless use of Podman for local Kubernetes development with Kind, Keycloak, and the MCP server.

### DIFF
--- a/build/keycloak.mk
+++ b/build/keycloak.mk
@@ -1,8 +1,21 @@
 # Keycloak IdP for development and testing
 
+# Detect container engine (docker or podman) - prefer the one that's actually running
+CONTAINER_ENGINE ?= $(shell \
+	if command -v docker >/dev/null 2>&1 && docker info >/dev/null 2>&1; then \
+		echo docker; \
+	elif command -v podman >/dev/null 2>&1 && podman info >/dev/null 2>&1; then \
+		echo podman; \
+	else \
+		command -v docker 2>/dev/null || command -v podman 2>/dev/null; \
+	fi)
+
 KEYCLOAK_NAMESPACE = keycloak
 KEYCLOAK_ADMIN_USER = admin
 KEYCLOAK_ADMIN_PASSWORD = admin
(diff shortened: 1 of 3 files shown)


build/keycloak.mk
```diff
@@ -1,8 +1,21 @@
 # Keycloak IdP for development and testing
 
+# Detect container engine (docker or podman) - prefer the one that's actually running
+CONTAINER_ENGINE ?= $(shell \
```
matzew, line 4: we have some check like this here: https://github.com/containers/kubernetes-mcp-server/blob/main/build/kind.mk#L19 

`Podman` is what most of the engineers are using

lyarwood: ACK apologies, I ended up getting claude all confused attempting to get keycloak to work with gevals and forgot to drop all of this. I'll close this out and move things over to the main KubeVirt evals PR.

PR #635, opened 2026-01-12
feat(config_toolset): add targets_list tool

Cali0707:
I recently realized that after #348 we never added a way to support a more generic `configuration_targets_list` tool. The kubeconfig one is there as there were extra user requirements we had for that specific case (associating context and server url). However, for our other multi cluster impls (e.g. ACM downstream), we only really need to list the names.

I'm not the happiest with the way this code was organized, but I didn't think adding these tools as a part of the provider interface made sense either.

IMO this change is needed for the multi cluster providers to be useable when there are > 5 targets (enum goes away)

### DIFF
--- a/pkg/mcp/tool_mutator_test.go
+++ b/pkg/mcp/tool_mutator_test.go
@@ -345,3 +345,50 @@ func (s *TargetParameterToolMutatorSuite) TestNonClusterAwareTool() {
 func TestTargetParameterToolMutator(t *testing.T) {
 	suite.Run(t, new(TargetParameterToolMutatorSuite))
 }
+
+type TargetListToolMutatorSuite struct {
+	suite.Suite
+}
+
+func (s *TargetListToolMutatorSuite) TestMutatesTargetsListTool() {
+	tool := createTestTool(TargetsListToolName)
+	tm := WithTargetListTool("default-cluster", "cluster", []string{"cluster-1", "cluster-2", "cluster-3"})
+	result := tm(tool)
+
+	s.Run("renames tool based on target parameter", func() {
+		s.Equal("cluster_list", result.Tool.Name)
+	})
+	s.Run("updates description", func() {
+		s.Contains(result.Tool.Description, "cluster")
+		s.Contains(result.Tool.Description, "List all available")
+	})
+	s.Run("updates title annotation", func() {
+		s.Equal("Cluster List", result.Tool.Annotations.Title)
+	})
+	s.Run("sets handler", func() {
+		s.NotNil(result.Handler)
+	})
+}
+
+func (s *TargetListToolMutatorSuite) TestDoesNotMutateOtherTools() {
+	tool := createTestTool("some-other-tool")
+	tm := WithTargetListTool("default", "cluster", []string{"cluster-1", "cluster-2"})
+	result := tm(tool)
+
+	s.Equal("some-other-tool", result.Tool.Name, "tool name should remain unchanged")
+}
+
+func (s *TargetListToolMutatorSuite) TestHandlerWithEmptyTargets() {
+	tool := createTestTool(TargetsListToolName)
+	tm := WithTargetListTool("default", "cluster", []string{})
+	result := tm(tool)
+
+	s.Require().NotNil(result.Handler)
+	callResult, err := result.Handler(api.ToolHandlerParams{})
+	s.NoError(err)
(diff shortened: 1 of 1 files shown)


pkg/mcp/tool_mutator.go
```diff
@@ -62,3 +76,60 @@ func createTargetProperty(defaultCluster, targetName string, targets []string) *
...
+		tool.Handler = createTargetListHandler(targets, targetParameterName, defaultTarget)
+
+		return tool
+	}
+}
+
+func createTargetListHandler(targets []string, targetParameterName, defaultTarget string) api.ToolHandlerFunc {
+	return func(_ api.ToolHandlerParams) (*api.ToolCallResult, error) {
+		if len(targets) == 0 {
+			return api.NewToolCallResult(fmt.Sprintf("No %ss available", targetParameterName), nil), nil
```
matzew, line 104: `ss` ?

Cali0707: This should format to "no clusters available" instead of "no cluster available"

Cali0707:
Running this code with the ACM provider I can see the following output:
```
Available clusters (2 total, default: local-cluster):                                                                                         
                                                                                                                                                      
Format: [*] CLUSTER_NAME                                                                                                                         
 (* indicates the default cluster used in tools if cluster is not set)                                                                           
                                                                                                                                                      
Clusters:                                                                                                                                        
---------                                                                                                                                        
  cmurray-managed                                                                                                                                
* local-cluster                                                                                                                                  
---------                                                                                                                                        
                                                                                                                                                      
To use a specific cluster with any tool, set the 'cluster' parameter in the tool call arguments
```

pkg/toolsets/config/configuration.go
```diff
@@ -32,6 +32,30 @@ func initConfiguration() []api.ServerTool {
...
+					ReadOnlyHint:    ptr.To(true),
+					DestructiveHint: ptr.To(false),
+					IdempotentHint:  ptr.To(true),
+					OpenWorldHint:   ptr.To(false),
+				},
+			},
+			ClusterAware:       ptr.To(false),
+			TargetListProvider: ptr.To(true),
+			Handler:            nil,
+		},
```
manusa, line 58: Is the tool always listed?

Cali0707: It follows similar filtering logic to the `configuration_contexts_list`. That one was a special case as there was a user request to format the output with a mapping from context name to cluster API server url.

The idea here is that for any provider that is multi cluster (which is not the special case from a user request), we provide a targets list tool of some kind. Otherwise the enum going away makes it so the model doesn't know what it can target

I feel like we could probably find a more elegant way to handle the kubeconfig context case, but I didn't have any good ideas while working on this of how to do that

manusa: I was asking more in the context of the tool being present if the function was hidden in case the enum was present or if there was just a single target.

I will check better.

Cali0707: With the original user request, they wanted the list tool to always be available when there was more than one target. 

My first pass with the multi cluster only enabled the tool when we got rid of the enum

Happy to implement either way

pkg/mcp/tool_mutator.go
matzew: ah...

pkg/mcp/tool_mutator.go
```diff
@@ -62,3 +76,60 @@ func createTargetProperty(defaultCluster, targetName string, targets []string) *
 
 	return baseSchema
 }
+
+// WithTargetListTool mutates the generic "targets_list" tool to have the correct name,
+// description, and handler based on the provider's target parameter name.
+// For example, with ACM provider (targetParameterName="cluster"), it becomes "cluster_list".
+func WithTargetListTool(defaultTarget, targetParameterName string, targets []string) ToolMutator {
```
matzew, line 83: Would you mind adding a test to the `tool_mutator_test` file? 

In `mcp.go` it gets called like: 

```go
	mutator := ComposeMutators(
		WithTargetParameter(s.p.GetDefaultTarget(), s.p.GetTargetParameterName(), targets),
		WithTargetListTool(s.p.GetDefaultTarget(), s.p.GetTargetParameterName(), targets),
...
```

but I see only `WithTargetParameter` test in the `_test.go` file for it

matzew:
LGTM

PR #636, opened 2026-01-12
feat(helm): add support for specifying RBAC

Cali0707:
As in some deployments the MCP server itself needs RBAC (beyond using the user's role when executing tool calls), this provides support in the helm chart for setting that RBAC

### DIFF
--- a/charts/kubernetes-mcp-server/templates/rbac.yaml
+++ b/charts/kubernetes-mcp-server/templates/rbac.yaml
@@ -1,0 +1,107 @@
+{{/*
+Generic RBAC resources template.
+This allows downstream forks and users to add custom RBAC resources without
+modifying upstream templates. Define your RBAC needs in values.yaml under:
+  - rbac.extraClusterRoles
+  - rbac.extraClusterRoleBindings
+  - rbac.extraRoles
+  - rbac.extraRoleBindings
+*/}}
+
+{{- range .Values.rbac.extraClusterRoles }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "kubernetes-mcp-server.fullname" $ }}-{{ .name }}
+  labels:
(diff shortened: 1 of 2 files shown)


manusa:
## Post-Merge Review Comments

Thanks for this contribution! The RBAC support is a valuable addition to the Helm chart. Here are some suggestions for potential follow-up improvements:

### Suggestions (Should Consider)

1. **Name length validation concern** (`templates/rbac.yaml:17`)
   - The fullname template truncates at 63 characters. When appending `-<name>` for RBAC resources (e.g., `{{ include "kubernetes-mcp-server.fullname" $ }}-{{ .name }}`), the total could exceed 63 characters if users provide long names.
   - Suggestion: Consider using `trunc` on the combined name or documenting the name length limitation.

2. **Missing `rbac.create` toggle** - Many Helm charts include an `rbac.create: true/false` top-level toggle to enable/disable all RBAC resource creation. While empty arrays achieve the same effect, an explicit toggle would be more consistent with common Helm patterns.

3. **ClusterRoleBinding roleRef name always prefixes** (`rbac.yaml:50`)
   - The roleRef name is always prefixed with the fullname: `{{ include "kubernetes-mcp-server.fullname" $ }}-{{ .roleRef.name }}`
   - This means users cannot bind to existing cluster roles (like `view`, `edit`, `admin`). Consider adding a `roleRef.external: true` option or similar to allow referencing external roles without prefixing.

### Minor Comments (Nice to Have)

1. **Consistent annotation handling** - The `serviceaccount.yaml` uses `tpl (toYaml .)` for annotations, but `rbac.yaml` uses just `toYaml .`. While this is fine, using `tpl` would allow template expressions in annotations if users need that flexibility.

2. **Documentation** - The values.yaml comments are well-documented. It might be helpful to also update the chart README.md with examples of common RBAC configurations.

---

*🤖 Generated with [Claude Code](https://claude.ai/code)*

PR #639, opened 2026-01-12
feat!: add toolset versioning

Cali0707:
This PR adds toolset versioning and versioning guidelines to the server.

A couple key points for reviewers:
1. Based on the current guidelines, no toolsets are "stable". The best is "core" which is "beta". As such, the default version has been set to "beta"
2. The helm toolset has no evals, so it is considered "alpha". As such, it is not included by default in tool lists (**this is a breaking change**)


docs/TOOLSET_VERSIONING.md
```diff
@@ -0,0 +1,53 @@
...
+## How Toolsets and Tools are Versioned
+
+All tools/prompts and toolsets are versioned as one of "alpha", "beta", or "ga"/"stable". Each toolset has a default version
+for the toolset, however individual tools/prompts may have their own versions. For example, a toolset as a whole may be in beta,
+however a newly added tool in that toolset may only be in alpha.
+
+The general idea for these versions is:
+- "alpha": the toolset is not guaranteed to work well
+- "beta": the toolset is not guaranteed to work well, but we are evaluating how well it works
+- "stable": the toolset works well, and we are evaluating how well it works to avoid regressions
```
matzew, line 15: `stable`: this reads (b/c of evaluating) not that stable. that sentence reads a bit vague.

docs/TOOLSET_VERSIONING.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Alpha
+
+All tools/prompts/toolsets begin in "alpha". If you are contributing a new tool/prompt/toolset, this is the version to set.
+There are no minimum requirements for something to be considered alpha, apart from the code getting merged.
+
+### Beta
+
+For a tool/prompt/toolset to enter into "beta", we require that there are eval scenarios. For a toolset to enter "beta", there must be scenarios
+excercising all of the tools and prompts in the toolset. For individual tools and prompts to enter "beta", we only require an eval scenario
```
matzew, line 29: perhaps would be good to eventually point to concrete example for that. But generally I like the requirement of having some sort of eval (w/ our toolkit)

docs/TOOLSET_VERSIONING.md
```diff
@@ -0,0 +1,53 @@
...
+for the specific tool or prompt.
+
+**Note**: for beta we do not require that all the eval scenarios are passing - we just require that they exist.
+
+### GA/Stable
+
+For a tool/prompt/toolset to enter into "stable", we require that 95% or more of the eval scenarios are passing. There is the same requirements as "beta" in terms of the number of evaluation scenarios.
```
matzew, line 36: 95% comes from?

docs/TOOLSET_VERSIONING.md
```diff
@@ -0,0 +1,53 @@
...
+
+You can also enable specific minimum versions for specific toolsets using the "toolset:version" syntax when enabling the toolset.
+For example, if you want to allow all the "alpha" tools in the "core" toolset, you could set `toolsets = [ "core:alpha" ]`, and this would
+enable all alpha+ tools in the core toolset.
+
+See a full config example below:
+```toml
+default_toolset_version = "beta"
+
+toolsets = [ "core", "config", "helm:alpha" ]
```
matzew, line 52: I like this schema

pkg/config/config.go
```diff
@@ -86,6 +86,9 @@ type StaticConfig struct {
 	// This can be used to provide specific instructions on how the client should use the server
 	ServerInstructions string `toml:"server_instructions,omitempty"`
 
+	// Which toolset version to enable (any tools/toolsets below this will be disabled)
+	DefaultToolsetVersion api.Version `toml:"default_toolset_version"`
```
matzew, line 90: So, at "server level", we say `stable` - hence no `beta` (for instance) enabled. 

what about explicit enablement - given a "global default" ? 

E.g. something like `toolsets = [ "core", "config", "helm:alpha" ]` would than "win", right?

pkg/api/toolsets.go
```diff
@@ -46,6 +76,9 @@ type Toolset interface {
 	// GetPrompts returns the prompts provided by this toolset.
 	// Returns nil if the toolset doesn't provide any prompts.
 	GetPrompts() []ServerPrompt
+	// GetVersion returns the version of the toolset.
+	// This version can be overridden by specific tools/prompts (e.g. a toolset may be beta, but have an alpha tool).
+	GetVersion() Version
```
matzew, line 81: for downstreaming impls we would than just set those?

pkg/mcp/testdata/toolsets-kiali-tools.json
```diff
@@ -1,292 +1 @@
...
-      },
-      "required": [
-        "namespace",
-        "workload"
-      ]
-    },
-    "name": "kiali_workload_logs"
-  }
-]
+[]
```
matzew: remove?

docs/TOOLSET_VERSIONING.md
Cali0707: I think @mrunalp was mentioning this as a threshold

pkg/config/config.go
Cali0707: IMO there are two parts to configuring everything:
1. Which toolsets you want. In my mind, this doesn't necessarily align with how mature the toolsets are, but rather with which domains you want to interact
2. What level of maturity of tools you want to use

So, if you set "core", "config", and "helm:alpha" in the current setup, what would happen is:
1. The core and config toolsets would both pick up the default version of "stable". As they are both in "beta" currently, no tools would be selected
2. The helm toolset would use the overridden "alpha" version, and since it is in alpha, all of it's tools would be available

I wasn't 100% convinced that the way I wrote it is the most intuititive, I just want to capture somewhere that there are those two key separate ideas in the config (which toolsets/domains, which versions you are okay with). My main thought is that enabling "stable" or "beta" should not enable all the toolsets with that version

pkg/mcp/testdata/toolsets-kiali-tools.json
Cali0707: Oh I think this is a mistake - because now by default this toolset is not enabled, there are no tools so the json snapshot update removed these.

Will switch it so these are enabled in the tests

manusa:
Given what we discussed internally yesterday and this proposal, I'm not that sure that _versioning_ is exactly what we want.
Also, I don't think this fits exactly the **upstream** purposes.

This is how I see it:
- I don't think that 3 levels (alpha, beta, stable) are really necessary but I might be missing something (e.g. I don't see a need to differentiate between beta and alpha). (See next point)
- Instead of providing _versioning_, I believe that for the downstream story, it should be something more like _certified_ or _supported_
  This should cover the productization story where the MCP server is provided with a given set of tools that are supported for customers.
 In case _certified_ misses something, then I guess _version_ is fine. 
- Upstream (this repo), SHOULD provide the required infrastructure to declare a toolset as certified/versioned.
  However, I think it SHOULD NOT take on the responsibility of declaring a toolset certified (or GA), since it doesn't really matter for the upstream world and its licensing (`"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND`).
- For maintenance reasons and also to facilitate the creation of the _certified_ infrastructure upstream and being able to leverage it downstream, I believe that a slice of certified toolsets should be declared somewhere (maybe the toolsets package) upstream with the `_overrides.go` pattern to be able to redefine it downstream.
  This is considering that entire toolsets should be certified as opposed to the current approach which provide granularity for individual tools or prompts (I don't really see the case for that, but I might be missing some points here too).
- I like the fact that certified toolsets (or a given version threshold in case we go that route) can be enforced while starting the server.
 But this should really account for their intersections and edge cases (enabled toolsets, enabled/disabled tools, readonly, etc.) Maybe warnings or exceptions should be printed with informative messages (e.g. a server is started with settings that effectively provide no toolsets)

matzew:
> for the downstream story, it should be something more like certified or supported

I think for that (downstream) **certified/supported** versus **NOT certified** is good. 

For upstream (here) I do see the reasoning behind the idea of coming in at a more flexible stage (e.g. alpha, beta, ga), allowing to define a set of robustness there. E.g. in some CNCF projects we had similar stage of "feature promotion". Showing how robust a (new) feature is. Which still than would fit with the term on the license.  

@manusa I guess you are more for a lower (or different) bar of adding toolsets, like "if it passes X % of eval it can come in" ?  Should all be enabled by default? (Which of course can be changed on the downstream too)

Cali0707:
From discussions today the idea (as I understand it) is:
1. Toolsets will either be "certified" or not (possibly under some other name, like "passing evals")
2. We will have some slice of the certified toolsets
3. There will be an eval requirement to make a toolset certified - if the toolset falls below the bar, we remove it from certified
4. There will be some flag (default false) to enable only certified toolsets
5. We will warn if a toolset is disabled by the flag

The idea then for new toolsets is that they will not be certified by default, until they get evals. For existing toolsets we can add a `toolsetname_experimental` toolset for adding/making large changes to existing tools

docs/TOOLSET_VERSIONING.md
nader-ziada, line 29: type `excercising` -> `exercising`

pkg/api/toolsets.go
```diff
@@ -3,16 +3,46 @@ package api
...
+
+func (v *Version) UnmarshalText(text []byte) error {
+	var tmp Version
+	switch strings.ToLower(string(text)) {
+	case "alpha":
+		tmp = VersionAlpha
+	case "beta":
+		tmp = VersionBeta
+	case "ga", "", "stable":
+		tmp = VersionGA
```
nader-ziada, line 30: minor issue: 
The empty string "" is grouped with "ga" and "stable", so if a user has a config like:

 ` default_toolset_version = "" `

it silently becomes VersionGA instead of the default of `VersionBeta`

PR #641, opened 2026-01-13
feat(kubernetes): add field selector support for filtering pods and resources

### LINKED ISSUE #621: [RFE] Add fieldSelector parameter to pod and resource listing tools
When trying to find crashloop, unhealthy, or other types of pods right now, the MCP server can only list _all_ the pods, and then requires something else (probably the LLM) to parse them.

This is extremely problematic for "are there any crashlooping pods in my cluster?" because this requires the ENTIRE pod list to be passed to the LLM, often exploding the context window.

Adding an optional "status" paremeter to the pod retrieval tools will fix this problem.

manusa:
Fixes #621

Will probably need to add some evals to see if this is useful or not based on what the user originally reported.

### DIFF
--- a/pkg/mcp/pods_test.go
+++ b/pkg/mcp/pods_test.go
@@ -697,6 +697,120 @@ func (s *PodsSuite) TestPodsListWithLabelSelector() {
 	})
 }
 
+func (s *PodsSuite) TestPodsListWithFieldSelector() {
+	s.InitMcpClient()
+	kc := kubernetes.NewForConfigOrDie(envTestRestConfig)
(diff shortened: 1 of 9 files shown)


Cali0707 (approved):
LGTM

Agree on evals for the scenario - up to you @manusa if you want to add it now, or hold off and we can open an issue to follow up with evals

manusa:
I'll play a little bit more with this tomorrow morning, see if I can finally make gevals work.

If it doesn't, I'll create a follow-up issue and merge this one.

nader-ziada:
LGTM

manusa:
Merging now, will work on an eval later.

Issue #602, opened 2025-12-19
Add configurable for running server in stateless mode


alexeykazakov:
Currently, the MCP server's stateless mode is hardcoded to false in the HTTP handler: https://github.com/containers/kubernetes-mcp-server/blob/974c79250ee3b4e45b2708c3b26e5a356ca2f36f/pkg/mcp/mcp.go#L269C3-L269C19, which means the server always operates in stateful mode. It's needed for sending notifications to clients about tool/prompt changes. This creates challenges for running multiple replicas of the server. Sticky sessions has to be used.

A new configuration would allow to run the server in stateless mode if dynamic tool/prompt change notifications are not required.

PR #646, opened 2026-01-14
test(auth): restore token exchange integration tests

manusa:
After the changes in:
- #604 
- #602 

The token exchange functionality was extracted to a separate package and its invocation moved to the mcp layer.

These changes also removed some tests that asserted our current authorization stack for token exchange.

While restoring the tests I noticed a change in the behavior that we should probably address

In the previous state, whenever the token exchange failed, the server returned a failure (through the HTTP layer).
Now, these failures are silently omitted, while the kubernetes layer will interact with the target server using the original token. I understand that eventually this will end up in an error for the user, but we might want to pick it up earlier.

There was also an open discussion about moving the exchange logic to the MCP layer:
https://github.com/containers/kubernetes-mcp-server/pull/604/changes#r2645800581

I believe this can be handled from the DerivedClient logic directly (isolating/decoupling the mcp layer from any auth-related stuff)  but will probably mean some breaking changes.

### DIFF
--- a/pkg/http/authorization_mcp_test.go
+++ b/pkg/http/authorization_mcp_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/coreos/go-oidc/v3/oidc/oidctest"
 	"github.com/mark3labs/mcp-go/client"
 	"github.com/mark3labs/mcp-go/client/transport"
+	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/stretchr/testify/suite"
 	"k8s.io/klog/v2"
 	"k8s.io/klog/v2/textlogger"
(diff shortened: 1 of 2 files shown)


Cali0707:
> I believe this can be handled from the DerivedClient

@manusa do you mean in the `GetDerivedKubernetes` method in the providers?

My only concern is I don't think providers should be reimplementing the actual logic of the token exchange, they should only be responsible for handling any target specific config relating to the token exchange. 

Maybe we can handle this in a way where we get the derived kubernetes and handle the token exchange, without forcing this responsibility into the providers?

manusa:
> My only concern is I don't think providers should be reimplementing the actual logic of the token exchange, they should only be responsible for handling any target specific config relating to the token exchange.
> 
> Maybe we can handle this in a way where we get the derived kubernetes and handle the token exchange, without forcing this responsibility into the providers?

Yes, as it is it's not possible, hence the breaking changes. We'll have to look into it, I believe there are more things that could be extracted away from the providers

Cali0707:
>  believe there are more things that could be extracted away from the providers

+1 from me there 😄

PR #653, opened 2026-01-15
feat(kubevirt): add troubleshoot action to vm_lifecycle tool

ksimon1:
Add a new "troubleshoot" action to the vm_lifecycle tool that generates a step-by-step troubleshooting guide for diagnosing VirtualMachine issues.

The troubleshoot action:
- Renders a diagnostic plan for VMs
- Guides the AI through checking VM, VMI, DataVolumes, PVCs, pods, and events
- Includes a summary template for reporting findings

This helps AI assistants systematically diagnose VM issues by providing structured instructions on which MCP tools to use and how to interpret the results.
Original credit goes to @lyarwood 

Code was assisted by Cursor AI

Signed-off-by: Karel Simon <ksimon@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -664,6 +664,16 @@ In case multi-cluster support is enabled (default) and you have access to multip
 
 </details>
 
+<details>
+
+<summary>kubevirt</summary>
+
+- **vm-troubleshoot** - Generate a step-by-step troubleshooting guide for diagnosing VirtualMachine issues
+  - `namespace` (`string`) **(required)** - The namespace of the VirtualMachine to troubleshoot
+  - `name` (`string`) **(required)** - The name of the VirtualMachine to troubleshoot
+
+</details>
+
 
 <!-- AVAILABLE-TOOLSETS-PROMPTS-END -->
 
--- a/evals/tasks/kubevirt/troubleshoot-vm/task.yaml
+++ b/evals/tasks/kubevirt/troubleshoot-vm/task.yaml
@@ -1,0 +1,133 @@
+kind: Task
+metadata:
+  name: "troubleshoot-vm"
+  difficulty: hard
+  description: "Use the vm-troubleshoot prompt to diagnose and fix VirtualMachine issues"
+steps:
+  setup:
+    inline: |-
+      #!/usr/bin/env bash
+      NS="${EVAL_NAMESPACE:-vm-test}"
+      kubectl delete namespace "$NS" --ignore-not-found
+      kubectl create namespace "$NS"
+
+      # Create a VM that references a missing Secret for cloud-init
+      # The agent should identify the missing Secret and create it to fix the VM
+      cat <<EOF | kubectl apply -f -
+      apiVersion: kubevirt.io/v1
+      kind: VirtualMachine
+      metadata:
+        name: broken-vm
+        namespace: $NS
+        labels:
+          app: broken-vm
+      spec:
+        runStrategy: Always
+        template:
+          spec:
+            domain:
+              devices:
+                disks:
+                - name: containerdisk
+                  disk:
+                    bus: virtio
+                - name: cloudinit
+                  disk:
+                    bus: virtio
+              resources:
+                requests:
+                  memory: 2Gi
+            terminationGracePeriodSeconds: 0
+            volumes:
+            - name: containerdisk
+              containerDisk:
+                image: quay.io/containerdisks/fedora:latest
+            - name: cloudinit
+              cloudInitNoCloud:
+                secretRef:
+                  name: vm-cloud-init
+      EOF
+
+      # Wait for VM to be created
+      kubectl wait --for=create vm/broken-vm -n "$NS" --timeout=10s
+      echo "VM created with missing Secret reference - waiting for failure state"
+      
+      # Give some time for the VM to attempt to start and fail
+      sleep 10
+  verify:
+    inline: |-
+      #!/usr/bin/env bash
+      NS="${EVAL_NAMESPACE:-vm-test}"
+
+      echo "=== Verification: Checking if agent fixed the VM ==="
+      
+      # Verify that the VM still exists
+      if ! kubectl get virtualmachine broken-vm -n "$NS" > /dev/null 2>&1; then
+          echo "✗ VirtualMachine broken-vm no longer exists"
+          exit 1
+      fi
+      echo "✓ VirtualMachine broken-vm exists"
+
+      # Check if the Secret was created by the agent
+      if kubectl get secret vm-cloud-init -n "$NS" > /dev/null 2>&1; then
+          echo "✓ Secret vm-cloud-init was created"
+      else
+          echo "✗ Secret vm-cloud-init was not created - agent did not fix the issue"
+          exit 1
+      fi
+
+      # Wait for VM to become ready after the fix (with timeout)
+      echo "Waiting for VM to become ready after fix..."
+      READY=false
+      for i in {1..30}; do
+          VM_READY=$(kubectl get virtualmachine broken-vm -n "$NS" -o jsonpath='{.status.ready}' 2>/dev/null || echo "false")
+          if [[ "$VM_READY" == "true" ]]; then
+              READY=true
+              break
+          fi
+          sleep 5
+      done
+
+      if [[ "$READY" == "true" ]]; then
+          echo "✓ VM is now ready - fix was successful!"
+      else
+          VM_STATUS=$(kubectl get virtualmachine broken-vm -n "$NS" -o jsonpath='{.status.printableStatus}' 2>/dev/null || echo "Unknown")
+          echo "⚠ VM is not ready yet (status: $VM_STATUS) - fix may need more time or was incomplete"
+          # Don't fail here as the VM may still be starting up
+      fi
+
+      # Check if virt-launcher pod exists and is running
+      LAUNCHER_POD=$(kubectl get pods -n "$NS" -l kubevirt.io=virt-launcher,vm.kubevirt.io/name=broken-vm -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || true)
+      if [[ -n "$LAUNCHER_POD" ]]; then
+          POD_PHASE=$(kubectl get pod "$LAUNCHER_POD" -n "$NS" -o jsonpath='{.status.phase}' 2>/dev/null || true)
+          echo "✓ virt-launcher pod exists (phase: $POD_PHASE)"
+      else
+          echo "ℹ No virt-launcher pod found yet"
+      fi
+
+      echo ""
+      echo "=== Troubleshooting and Fix Eval Complete ==="
+      echo "The agent should have:"
+      echo "  1. Used the vm-troubleshoot prompt with namespace=$NS and name=broken-vm"
+      echo "  2. Identified the root cause (missing Secret vm-cloud-init)"
+      echo "  3. Created the missing Secret to fix the VM"
+      echo "  4. Reported the action taken and result"
+      echo ""
+      
+      exit 0
+  cleanup:
+    inline: |-
+      #!/usr/bin/env bash
+      NS="${EVAL_NAMESPACE:-vm-test}"
+      kubectl delete virtualmachine broken-vm -n "$NS" --ignore-not-found
+      kubectl delete secret vm-cloud-init -n "$NS" --ignore-not-found
+      kubectl delete namespace "$NS" --ignore-not-found
+  prompt:
+    inline: |-
+      There is a VirtualMachine named "broken-vm" in the ${EVAL_NAMESPACE:-vm-test} namespace that is not working correctly.
+      
+      Please use the vm-troubleshoot prompt to diagnose the issue with this VirtualMachine.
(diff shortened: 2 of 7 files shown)


ksimon1:
@lyarwood, @Cali0707, @manusa can you please review this PR?

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
```diff
@@ -75,14 +87,13 @@ func lifecycle(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
 		return api.NewToolCallResult("", err), nil
 	}
 
-	dynamicClient := params.DynamicClient()
```
codingben, line 78: What is the reason for removing it from here and putting the `dynamicClient` in each switch-case?

Cali0707:
> IIRC we don't have a way of measuring this clearly through gevals yet 

This is on the roadmap, but not implemented yet (sorry 😞 )

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
```diff
@@ -106,14 +118,32 @@ func lifecycle(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
...
 
 	case ActionRestart:
+		dynamicClient := params.DynamicClient()
 		vm, err = kubevirt.RestartVM(params.Context, dynamicClient, namespace, name)
 		if err != nil {
 			return api.NewToolCallResult("", err), nil
 		}
 		message = "# VirtualMachine restarted successfully\n"
 
+	case ActionTroubleshoot:
```
Cali0707: IMO this is more of a MCP prompt than a tool. We have added support for toolsets to define their own prompts (see https://github.com/containers/kubernetes-mcp-server/blob/main/docs/PROMPTS.md#toolset-prompts)

lyarwood: TIL looks like this dropped while I was out, thanks! 

@ksimon1 are you able to rework this to use a prompt? Would be good to rebase and land this in the next 2 weeks with some basic eval coverage.

pkg/toolsets/kubevirt/vm_troubleshoot.go
```diff
@@ -0,0 +1,113 @@
...
+
+	if namespace == "" {
+		return nil, fmt.Errorf("namespace argument is required")
+	}
+	if name == "" {
+		return nil, fmt.Errorf("name argument is required")
+	}
+
+	// Build the troubleshooting guide message
+	guideText := fmt.Sprintf(`# VirtualMachine Troubleshooting Guide
```
ksimon1, line 73: @Cali0707 Can be the prompt in this format as a string, or do you want to have written it in code?

Cali0707: @ksimon1 I think it can be either a string or a function - depends on what you think works best. I just think we should use the MCP "prompt" API for this :)

README.md
```diff
@@ -613,10 +613,15 @@ In case multi-cluster support is enabled (default) and you have access to multip
...
-  - `action` (`string`) **(required)** - The lifecycle action to perform: 'start' (changes runStrategy to Always), 'stop' (changes runStrategy to Halted), or 'restart' (stops then starts the VM)
+  - `action` (`string`) **(required)** - The lifecycle action to perform: 'start' (changes runStrategy to Always), 'stop' (changes runStrategy to Halted), 'restart' (stops then starts the VM)
   - `name` (`string`) **(required)** - The name of the virtual machine
   - `namespace` (`string`) **(required)** - The namespace of the virtual machine
 
+### Prompts
+- **vm-troubleshoot** - Generate a step-by-step troubleshoot guide for diagnosing VirtualMachine issues
+  - `namespace` (`string`) **(required)** - The namespace of the VirtualMachine
+  - `name` (`string`) **(required)** - The name of the VirtualMachine
+
```
manusa: This is auto-generated, it will be erased once we update the readme docs.

If you want to change the tool definition, you need to do that here:

https://github.com/containers/kubernetes-mcp-server/blob/045255849865072b25535e499a52be804cf49c88/pkg/toolsets/kubevirt/vm/lifecycle/tool.go#L42

and run `make update-readme-tools`
 
We don't support the prompt definitions yet, [I'll create an issue for that and work on it](https://github.com/containers/kubernetes-mcp-server/issues/698).

ksimon1: ok, I will keep it as it is and once the #698 will be implemented, the README will be updated automatically

ksimon1:
@lyarwood, @Cali0707, @manusa can you please review this PR?

README.md
ksimon1: The prompt description is now generated.

pkg/toolsets/kubevirt/vm_troubleshoot.go
```diff
@@ -0,0 +1,113 @@
...
+
+## VM: %s (namespace: %s)
+
+Follow these steps to diagnose issues with the VirtualMachine:
+
+## Step 1: Check VirtualMachine Status
+Use resources_get with apiVersion=kubevirt.io/v1, kind=VirtualMachine, namespace=%s, name=%s
```
manusa: Some ideas that float into my head that might or might not apply:

The prompt could probably benefit from some dynamic content injection (as defined in [Claude Skills](https://code.claude.com/docs/en/skills#inject-dynamic-context)).

The idea here would be to execute the relevant queries ourselves and injecting the desired result, instead of delegating the task to the model and avoiding the extra round-trips and overhead. (I believe this is a pattern we're already following with the cluster-health-check prompt).

I would use this dynamic content injection at least for those requests that we know beforehand that will have to be performed by the model.

Regarding token usage, I understand the initial toll of this pattern should be higher, but it should compensate the extra overhead and roundtrips the model+agent would need to complete the actual troubleshooting task.


Thoughts?

ksimon1: IIUC you mean to prepopulate the prompt text with basic information like vm/vmi manifest, pod description, ...?

ksimon1: @manusa would you please review the new code, which injects the content to the prompt?

evals/tasks/kubevirt/troubleshoot-vm/task.yaml
```diff
@@ -0,0 +1,105 @@
...
+      exit 0
+  cleanup:
+    inline: |-
+      #!/usr/bin/env bash
+      NS="${EVAL_NAMESPACE:-vm-test}"
+      kubectl delete virtualmachine broken-vm -n "$NS" --ignore-not-found
+      kubectl delete namespace "$NS" --ignore-not-found
+  prompt:
+    inline: |-
+      There is a VirtualMachine named "broken-vm" in the ${EVAL_NAMESPACE:-vm-test} namespace that is not working correctly.
```
lyarwood, line 127: Have you tried running this with mcpchecker? IIRC it no longer supports bash substitutions, something we can address in the project but it will lead to this being passed directly to the agent/model in it's current form.

pkg/toolsets/kubevirt/vm_troubleshoot.go
```diff
@@ -0,0 +1,445 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	"github.com/containers/kubernetes-mcp-server/pkg/kubernetes"
+	"github.com/containers/kubernetes-mcp-server/pkg/output"
+)
+
+// GVRs for KubeVirt and related resources
+var (
```
lyarwood: supernit - we should share these between tools and maybe even toolsets.

pkg/toolsets/kubevirt/vm_troubleshoot.go
```diff
@@ -0,0 +1,445 @@
...
+---
+
+## Step 1: VirtualMachine Status
+
+Check the VirtualMachine status:
+- status.printableStatus (should be "Running")
+- status.ready (should be true)
+- status.conditions for any errors
+
+%s
```
lyarwood, line 88: Do you need to pass the entire VM here if we are only checking status?

pkg/toolsets/kubevirt/vm_troubleshoot.go
```diff
@@ -0,0 +1,445 @@
...
+
+---
+
+## Step 2: VirtualMachineInstance Status
+
+Check the VirtualMachineInstance status:
+- status.phase (should be "Running")
+- status.conditions for "Ready" condition
+
+%s
```
lyarwood: As above.

pkg/toolsets/kubevirt/vm_troubleshoot.go
```diff
@@ -0,0 +1,445 @@
...
+- status.phase (should be "Bound")
+- Ensure storage is properly provisioned
+
+%s
+
+---
+
+## Step 5: Other Volume Types
+
+Review the VM spec (from Step 1) for other volume types that may cause issues:
```
lyarwood: Provide specific parts of the spec here instead?

evals/tasks/kubevirt/troubleshoot-vm/task.yaml
```diff
@@ -0,0 +1,105 @@
...
+      
+      Please use the vm-troubleshoot prompt to diagnose the issue with this VirtualMachine.
+      Follow the troubleshooting guide and report your findings, including the root cause and recommended action.
```
lyarwood: Do you not want to fix the VM and assert that behaviour above instead?

evals/tasks/kubevirt/troubleshoot-vm/task.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+      echo ""
+      echo "=== Troubleshooting Eval Complete ==="
+      echo "The agent should have:"
+      echo "  1. Used the vm-troubleshoot prompt with namespace=$NS and name=broken-vm"
+      echo "  2. Followed the troubleshooting steps to gather information"
+      echo "  3. Identified the root cause (container image pull failure)"
+      echo ""
+      
+      exit 0
```
lyarwood, line 117: As below, I think asking the agent to troubleshoot and fix would be cleaner to assert instead of just asserting that the VM remains broken.

evals/tasks/kubevirt/troubleshoot-vm/task.yaml
```diff
@@ -0,0 +1,105 @@
...
+              echo "⚠ virt-launcher pod status: $POD_STATUS"
+          fi
+      else
+          echo "ℹ No virt-launcher pod found (VM may not have started yet)"
+      fi
+
+      echo ""
+      echo "=== Troubleshooting Eval Complete ==="
+      echo "The agent should have:"
+      echo "  1. Used the vm-troubleshoot prompt with namespace=$NS and name=broken-vm"
```
lyarwood, line 111: This is also missing from the `Task` API at the moment IMHO, we can define this in `Evals` but I also think each `Task` should be able to assert that tools and/or prompts are called.

Cali0707: @lyarwood +1 here - we have an open discussion trying to figure out how we want to solve this: https://github.com/mcpchecker/mcpchecker/discussions/126

Interested in hearing if you have any thoughts 😄

evals/tasks/kubevirt/troubleshoot-vm/task.yaml
ksimon1: The mcpchecker passed. Since this substitution is in all kubevirt's eval tasks, I would update it in different PR in all tasks.

ksimon1:
@Cali0707, @manusa can you please review this PR?

pkg/toolsets/kubevirt/vm_troubleshoot.go
manusa: > IIUC you mean to prepopulate the prompt text with basic information like vm/vmi manifest, pod description, ...?

Yes, exactly

> @manusa would you please review the new code, which injects the content to the prompt?

Sorry, last Friday I was completely focused on the code mode feature and demo. Let me check this now.

manusa: I'm not sure how the compiled prompt looks like, but you're definitely doing what I was suggesting.
This should prevent the LLM from doing multiple roundtrips which we know beforehand it's going to try because we were instructing it to do so.

lyarwood:
> Evals look fine to me, thanks @ksimon1 !
> 
> Only thing to note is that we are hoping to move onto the new task format and off of using bash scripts as much as possible. So in the future, we will need to rework this to e.g. leverage the [kubernetes extension](https://github.com/mcpchecker/kubernetes-extension)

ACK thanks for the reminder, something for a follow up.

I can create an issue if you want to us to track this?

Cali0707:
> I can create an issue if you want to us to track this?

That would be helpful!

ksimon1:
So can we merge this PR?

manusa:
> So can we merge this PR?

Merged 🚀, thx!

PR #657, opened 2026-01-16
feat(kcp): Add kcp provider

mjudeikis:
This adds tooling to treat kcp (kcp.io) workspaces as clusters.

This PR is primarily to measure temperature and gauge community interest; if so, what changes are needed (design by PR). 
Let me know if this is something we could merge :) 

Testing in real life:

```
# start kcp 
kcp start 
# repeat a few times
kubectl ws create foo --enter # repeat few times
# mcp works on relative path so move to start in workspace tree
k ws use :root

# run mcp with kcp toolset
npx @modelcontextprotocol/inspector@latest \
    ./kubernetes-mcp-server \
    --kubeconfig=.kcp/admin.kubeconfig \
    --toolsets=core,helm,kcp \
    --log-level=3                                      
```
connect with STDIO and test

and see some results.
Screenshot 2026-01-16 at 10 06 34

Screenshot 2026-01-16 at 10 06 13

### DIFF
--- a/pkg/kubernetes/provider_kcp.go
+++ b/pkg/kubernetes/provider_kcp.go
@@ -1,0 +1,283 @@
+package kubernetes
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"reflect"
+	"sort"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	kcppkg "github.com/containers/kubernetes-mcp-server/pkg/kcp"
+	"github.com/containers/kubernetes-mcp-server/pkg/kubernetes/watcher"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/klog/v2"
+)
+
+// kcpTargetParameterName is the parameter name used to specify
+// the workspace when using the kcp cluster provider strategy.
+const kcpTargetParameterName = "workspace"
+
+// kcpClusterProvider implements Provider for managing multiple
+// kcp workspaces as separate cluster targets.
+// It discovers workspaces via the kcp tenancy API and creates
+// managers for each workspace on-demand.
+type kcpClusterProvider struct {
+	config              api.BaseConfig
+	baseServerURL       string
+	restConfig          *rest.Config
+	clientCmdConfig     clientcmd.ClientConfig
+	defaultWorkspace    string
+	managers            map[string]*Manager
+	workspaceWatcher    *watcher.Workspace
+	clusterStateWatcher *watcher.ClusterState
+}
+
+var _ Provider = &kcpClusterProvider{}
+
+func init() {
+	RegisterProvider(api.ClusterProviderKcp, newKcpClusterProvider)
+}
+
+// newKcpClusterProvider creates a provider that manages multiple kcp workspaces.
+// Each workspace is treated as a separate cluster target.
+func newKcpClusterProvider(cfg api.BaseConfig) (Provider, error) {
+	ret := &kcpClusterProvider{config: cfg}
+	if err := ret.reset(); err != nil {
+		return nil, err
+	}
+	return ret, nil
+}
+
+func (p *kcpClusterProvider) reset() error {
+	// Load kubeconfig
+	pathOptions := clientcmd.NewDefaultPathOptions()
+	if p.config.GetKubeConfigPath() != "" {
+		pathOptions.LoadingRules.ExplicitPath = p.config.GetKubeConfigPath()
+	}
+
+	p.clientCmdConfig = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		pathOptions.LoadingRules,
+		&clientcmd.ConfigOverrides{})
+
+	rawConfig, err := p.clientCmdConfig.RawConfig()
+	if err != nil {
+		return fmt.Errorf("failed to load kubeconfig: %w", err)
+	}
+
+	// Get current context
+	currentContext := rawConfig.Contexts[rawConfig.CurrentContext]
+	if currentContext == nil {
+		return errors.New("no current context in kubeconfig")
+	}
+
+	currentCluster := rawConfig.Clusters[currentContext.Cluster]
+	if currentCluster == nil {
+		return errors.New("current context's cluster not found in kubeconfig")
+	}
+
+	// Parse kcp server URL to extract base URL and workspace
+	p.baseServerURL, p.defaultWorkspace = kcppkg.ParseServerURL(currentCluster.Server)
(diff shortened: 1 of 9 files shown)


pkg/kubernetes/provider.go
```diff
@@ -60,6 +62,9 @@ func resolveStrategy(cfg api.BaseConfig) string {
 	}
 
 	if cfg.GetKubeConfigPath() != "" {
+		if isKcpKubeconfig(cfg) {
```
Cali0707: IMO the KCP provider (instead of kubeconfig) should be opt-in (not something that may happen without the user explicitly setting the strategy)

pkg/kubernetes/provider_kcp.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+// constructWorkspaceURL builds a full kcp server URL for a workspace.
+func constructWorkspaceURL(baseURL, workspace string) string {
+	return fmt.Sprintf("%s/clusters/%s", baseURL, workspace)
+}
+
+// discoverWorkspaces queries the kcp tenancy API to discover available workspaces.
+// It recursively discovers nested workspaces in the workspace hierarchy.
+func (p *kcpClusterProvider) discoverWorkspaces(baseManager *Manager) ([]string, error) {
```
Cali0707: this logic seems quite similar to the tool in the toolset - maybe we can create some kind of kcp package that will contain the shared logic?

pkg/toolsets/kcp/workspaces.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/containers/kubernetes-mcp-server/pkg/kubernetes"
+	"github.com/containers/kubernetes-mcp-server/pkg/output"
+)
+
+func initWorkspaceTools() []api.ServerTool {
+	return []api.ServerTool{
+		{
+			Tool: api.Tool{
+				Name:        "kcp_workspaces_list",
+				Description: "List all available kcp workspaces in the cluster",
```
Cali0707: Since there is already a tool which will list all the targets for a provider if num targets >1, it's not clear to me what the value of this tool is. Is this just useful if someone is using the kcp toolset but not the kcp provider?

mjudeikis: This is very specific to kcp. First - recursive list for nested workspaces, second - having dedicated tool makes AI agent to understand this better without systemPromt magic. But agree, one could do this with a resource list but prompt engineering gets complicated.

mjudeikis: Same argument like list pods... Its arbitrary list, but somehow we have a dedicated tool to make ai less unpredictability

pkg/kubernetes/provider_kcp.go
mjudeikis: Will try to explore this bus wondering how much of this would be pre-mature optimization

pkg/kubernetes/provider.go
mjudeikis: Make sense. This was very much exploration

pkg/toolsets/kcp/workspaces.go
mjudeikis: + kcp tool without kcp provider is kinda useless. KCP is more of a distribution of K8S, than a tool on top of K8S. Hence, I don't see the tool being used without a provider set.

pkg/kubernetes/provider_kcp.go
mjudeikis: After re-reading this, it makes sense. Will change

mjudeikis:
@Cali0707  I did the following:
1. Created a dedicated package for KCP. I suspect more things will come so better do it early as per your suggestion it makes logic cleaner
2. Added `cluster-provider` logic. I could imagine somebody adding like `cluster-api` as value :)
3. Added minimal docs

pkg/kubernetes/provider_kcp.go
```diff
@@ -0,0 +1,283 @@
...
+	"github.com/containers/kubernetes-mcp-server/pkg/kubernetes/watcher"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/klog/v2"
+)
+
+// kcpTargetParameterName is the parameter name used to specify
+// the workspace when using the kcp cluster provider strategy.
+const kcpTargetParameterName = "cluster"
```
Cali0707: @mjudeikis I defer to your expertise here, but I am wondering do you think naming this "workspace" would be more intuitive for agents/users interacting with agents if they know they are using kcp?

pkg/kubernetes-mcp-server/cmd/root.go
```diff
@@ -75,6 +82,7 @@ const (
 	flagServerUrl            = "server-url"
 	flagCertificateAuthority = "certificate-authority"
 	flagDisableMultiCluster  = "disable-multi-cluster"
+	flagClusterProvider      = "cluster-provider"
```
Cali0707: This is more of a personal preference, but IMO this should be part of a separate PR as the change isn't super related to enabling the kcp provider/toolset

mjudeikis: will take this out and move to separate PR

pkg/kubernetes/provider_kcp.go
mjudeikis: Update:

I totaly missread this. yes :D

pkg/kubernetes-mcp-server/cmd/root.go
mjudeikis: https://github.com/containers/kubernetes-mcp-server/pull/667

matzew:
@mjudeikis are you adding a bit of documentation for `kcp provider` ?

mjudeikis:
@matzew, the plan was to add it as a follow-up as we go. We will need more methods/tools, and some for now, it's just adding the first code. 
If you want I can add it now, but with just list and describe it is not very usefull (yet) 

Plan is to start shipping this mcp together with https://platform-mesh.io/ (its kcp based) so we can have better AI journey :)

matzew:
I am ok with a separate PR for docs - I also see that you add a new `cluster-provider` flag on a different PR.

I think for now, let's keep it as is: users can enable it when they want. Makes sense?

LGTM

mjudeikis:
> I am ok with a separate PR for docs - I also see that you add a new `cluster-provider` flag on a different PR.
> 
> I think for now, let's keep it as is: users can enable it when they want. Makes sense?
> 
> LGTM

Very much. I would not want this to be ever enabled by default. In general clusters concepts in this mcp implementation is very interesting :) Imaginging cluster-api, vcluster, karamda, etc becoming one opf the providers too soon

PR #659, opened 2026-01-17
chore(docs): add GitHub Copilot configuration for local MCP server

itobey:
This PR inserts a new "GitHub Copilot" subsection in README.md. The subsection documents how to configure the IntelliJ GitHub Copilot plugin to use a local Docker-hosted MCP server and includes a ready-to-use mcp.json snippet pointing Copilot at http://localhost:8080/mcp with SSE headers. It took me a while to figure out the correct way to configure it, so it might as well be in the official docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,27 @@ extensions:
 
 ```
 
+### GitHub Copilot
+
+This configuration is for the IntelliJ GitHub Copilot plugin using a local Docker container. Add the following configuration to the `mcp.json` point Copilot to the local MCP server:
+
+```json
+{
+  "servers": {
+    "kubernetes-mcp": {
+      "type": "http",
+      "url": "http://localhost:8080/mcp",
+      "requestInit": {
+        "method": "GET",
+        "headers": {
+          "Accept": "text/event-stream"
+        }
+      }
+    }
+  }
+}
+```
+
 ## 🎥 Demos <a id="demos"></a>
 
 ### Diagnosing and automatically fixing an OpenShift Deployment


README.md
```diff
@@ -138,6 +138,27 @@ extensions:
...
+
+This configuration is for the IntelliJ GitHub Copilot plugin using a local Docker container. Add the following configuration to the `mcp.json` point Copilot to the local MCP server:
+
+```json
+{
+  "servers": {
+    "kubernetes-mcp": {
+      "type": "http",
+      "url": "http://localhost:8080/mcp",
+      "requestInit": {
```
matzew, line 151: Not used copilot before - is that section really needed? 

some search showed this was used in context of (auth) tokens

README.md
```diff
@@ -138,6 +138,27 @@ extensions:
...
+        "method": "GET",
+        "headers": {
+          "Accept": "text/event-stream"
+        }
+      }
+    }
+  }
+}
```
manusa, line 159:
```suggestion
{
    "servers": {
        "kubernetes-mcp-server": {
            "command": "npx",
            "args": ["-y", "kubernetes-mcp-server@latest"]
        }
    }
}
```

We need to keep it consistent

README.md
manusa: This should be the configuration to keep it consistent with the other sections: https://github.com/containers/kubernetes-mcp-server/pull/659/changes#r2703697058

itobey:
I didn't find any documentation on how to use the containerized version. The current documentation only shows the local call with npx. So that was the main reason for this pull request. I think the way Copilot mcp works is quite similar to other offerings. If you do not want the http way of calling the mcp documented I think this PR does not need to be merged.

manusa:
I just read about the Docker container note in the PR description.
You can actually do this by using a docker/podman command while mounting your .kube/config file to the container and setting the KUBECONFIG environment variable.
This should be a more maintainable way of doing things since the container could be ephemeral and delete itself once you shutdown IntelliJ.

itobey:
I'm afraid I do not understand your comment. My issue has been that I was not aware of how to connect through http/sse to the mcp server (e.g. using Copilot in IntelliJ). I did only find the local `mcp.json` specifications for directly connecting a running process with `npx`. That is why I thought others may benefit of this `mcp.json`. Some people may prefer to have a central container running somewhere (or even on their local machine for that matter - not taking security in account here), so that might be of help.

manusa:
X Y problem. As I understand, you want to connect using HTTP/SSE to be able to use a the MCP server that's exposed from a **local (Docker) container**, right?

If this IS the problem, then you can actually run the server with the guidelines I stated in the last comment (they need elaboration though).
This would be reusable for any other user who wants to deploy the MCP server locally in a secure, sandboxed way.

If this IS NOT the problem, I'm afraid that your use-case is too specific to document the configuration in the getting started section.
Main reason is that users would need to first deploy somehow the MCP server and this is not explained.
The configuration you show is really straightforward and common to any other MCP server that can be used through SSE or streamable HTTP mode. With the exception of the `requestInit` config that I'm not really sure why it's necessary.

itobey:
Well, I can clear that up now. I was not explicitly talking about localhost, but the SSE / HTTP way of connecting to the mcp. I was under the impression that for the kubernetes-mcp-server to work using SSE this additional configuration was necessary. As it turns out: my IntelliJ / Copilot Plugin bugged out on me and this part was in fact not necessary - the regular MCP config is sufficient, `requestInit` has somehow been mandatory when I first set it up, but it does not seem to be the case anymore. So I will close this PR, since this does not add any additional value.

PR #590, opened 2025-12-17
feat(observability): add OpenTelemetry tracing and metrics support

nader-ziada:
Add observability capabilities through OpenTelemetry, providing distributed tracing and metrics for MCP operations and HTTP requests.

- Automatic tracing of all MCP tool calls via middleware
- HTTP request/response tracing in server mode
- New telemetry package with configurable OTLP tracing via environment variables
- Metrics collection for tool execution statistics and HTTP request counters
- Middleware-based tracing that follows MCP semantic conventions
- Session-based span context propagation between tool calls
- Enable via OTEL_EXPORTER_OTLP_ENDPOINT environment variable
- Custom sampling rates via OTEL_TRACES_SAMPLER environment variables

PR #670, opened 2026-01-21
feat(metrics): add Prometheus exporter and monitoring resources

nader-ziada:
Follow-up to Adding OpenTelemetry PR: #590 

Add a Prometheus HTTP endpoint (/metrics) to expose OpenTelemetry metrics in Prometheus format. Also add Helm chart templates for ServiceMonitor and PrometheusRule to enable integration with Prometheus operators.

Changes:
 - Add prometheus exporter to OtelStatsCollector
 - Expose /metrics endpoint in HTTP server
 - Add ServiceMonitor template for Prometheus scraping
 - Add PrometheusRule template with recording rules for Telemeter
 - Inject POD_NAMESPACE env var for namespace-level metric aggregation

### DIFF
--- a/docs/OTEL.md
+++ b/docs/OTEL.md
@@ -417,6 +417,49 @@ The stats endpoint is useful for:
 
 **Note**: The `/stats` endpoint is only available in HTTP mode. In STDIO mode, use OTLP export for metrics.
 
+## Metrics Endpoint
+
+When running in HTTP mode, the server exposes a `/metrics` endpoint for Prometheus scraping:
+
+```bash
+curl http://localhost:8080/metrics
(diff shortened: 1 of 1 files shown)


pkg/http/http.go
```diff
@@ -21,6 +21,7 @@ import (
 const (
 	healthEndpoint     = "/healthz"
 	statsEndpoint      = "/stats"
+	metricsEndpoint    = "/metrics"
```
matzew, line 24: Do we want to also document this available endpoint?

nader-ziada: added to otel.md

PR #684, opened 2026-01-22
fix(ci): gevals permission check captures SHA to checkout

Cali0707:
This PR cleans up the gevals action in a few ways:
1. Capture the SHA which a reviewer checked with they triggered the gevals action (and use that SHA to checkout the code)
2. Separate posting of comments to a separate workflow so that the workflow running the user code does not have write permissions

### DIFF
--- a/.github/workflows/gevals-report.yaml
+++ b/.github/workflows/gevals-report.yaml
@@ -1,0 +1,63 @@
+name: Gevals MCP Evaluation - Report
+
+on:
+  workflow_run:
+    workflows: ["Gevals MCP Evaluation"]
+    types: [completed]
+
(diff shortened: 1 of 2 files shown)


nader-ziada:
looks great @Cali0707 

LGTM

manusa:
I haven't checked them, but there are some failures in the new pipeline https://github.com/containers/kubernetes-mcp-server/actions/workflows/gevals-report.yaml

@Cali0707 if you could please take a look whenever you have time 🙏

Cali0707:
Hey @manusa thanks for the heads up - I think I've found the root cause and added a fix in #693

PR #689, opened 2026-01-23
fix(kubernetes): use lazy evaluation for restMapper in AccessControlRoundTripper

### LINKED ISSUE #688: Nil pointer dereference in AccessControlRoundTripper
Before we start, thanks for the good work on the project!

## Issue
Starting from v0.0.55, the kubernetes-mcp-server crashes with a nil pointer dereference panic during startup when the AccessControlRoundTripper attempts to use an uninitialized restMapper.

## Error
```
npx -y kubernetes-mcp-server@0.0.55 --kubeconfig ~/.kube/k8s-mcp-server-kubeconfig.yaml --port 8080
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x105629c74]

goroutine 1 [running]:
github.com/containers/kubernetes-mcp-server/pkg/kubernetes.(*AccessControlRoundTripper).RoundTrip(0x1400012ecf0, 0x140002e6640)
        pkg/kubernetes/accesscontrol_round_tripper.go:26 +0x64
```

Same kubeconfig file works fine for 0.0.54:
```
npx -y kubernetes-mcp-server@0.0.54 --kubeconfig ~/.kube/k8s-mcp-server-kubeconfig.yaml --port 8080
E0123 18:15:39.547939   80691 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: the server has asked for the client to provide credentials" logger="UnhandledError"
E0123 18:15:40.065036   80691 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: the server has asked for the client to provide credentials" logger="UnhandledError"
I0123 18:15:40.068521   80691 http.go:57] "Streaming and SSE HTTP servers starting on port 8080 and paths /mcp, /sse, /message"
^CI0123 18:15:58.548835   80691 http.go:65] "Received signal interrupt, initiating graceful shutdown"
I0123 18:15:58.548904   80691 http.go:77] "Shutting down HTTP server gracefully..."
I0123 18:15:58.549151   80691 http.go:83] "HTTP server shutdown complete"
Received SIGINT, terminating child process...
```

## Possible root cause:

Since #585, the [restConfig.Wrap() function](https://github.com/containers/kubernetes-mcp-server/blob/353772b59be56a9c11e00abc58d2cfcd59e4c8ef/pkg/kubernetes/kubernetes.go#L61) is called before restMapper is [initialized](https://github.com/containers/kubernetes-mcp-server/blob/353772b59be56a9c11e00abc58d2cfcd59e4c8ef/pkg/kubernetes/kubernetes.go#L73)  in pkg/kubernetes/kubernetes.go, leading to the nil pointer issue

manusa:
Fixes #688

This fixes a nil pointer dereference panic in AccessControlRoundTripper as reported in https://github.com/containers/kubernetes-mcp-server/issues/688

The bug occurred because rest.Config.Wrap() was called before restMapper was initialized. When discovery.NewDiscoveryClientForConfig() triggered the wrapper, k.restMapper was still nil, causing panics when resource API calls went through the discovery client's transport.

The fix changes restMapper from a direct value to a provider function that returns the restMapper at request time (lazy evaluation). This ensures the wrapper always gets the current restMapper value, even if it was nil when the wrapper was created.

If the provider returns nil (during discovery client initialization), the request passes through without access control checks.

### DIFF
--- a/pkg/kubernetes/accesscontrol_round_tripper.go
+++ b/pkg/kubernetes/accesscontrol_round_tripper.go
@@ -28,9 +28,7 @@ func (rt *AccessControlRoundTripper) RoundTrip(req *http.Request) (*http.Respons
 	// was created before restMapper was set (fixes issue #688)
 	restMapper := rt.restMapperProvider()
 	if restMapper == nil {
-		// restMapper not yet initialized, pass through
-		// This can happen during discovery client initialization
-		return rt.delegate.RoundTrip(req)
+		return nil, fmt.Errorf("failed to make request: AccessControlRoundTripper restMapper not initialized")
 	}
 
 	gvk, err := restMapper.KindFor(gvr)
--- a/pkg/kubernetes/accesscontrol_round_tripper_test.go
+++ b/pkg/kubernetes/accesscontrol_round_tripper_test.go
@@ -77,8 +77,8 @@ func (s *AccessControlRoundTripperTestSuite) TestRoundTripForNonAPIResources() {
 
 func (s *AccessControlRoundTripperTestSuite) TestRoundTripWithNilRestMapper() {
 	// This test covers the nil restMapper branch (issue #688 fix).
-	// When restMapperProvider returns nil, requests should pass through
-	// to the delegate without panicking.
+	// When restMapperProvider returns nil, requests should fail with an error
+	// to prevent bypassing access control.
 
 	delegateCalled := false
 	mockDelegate := &mockRoundTripper{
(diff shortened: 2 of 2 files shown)


pkg/kubernetes/accesscontrol_round_tripper.go
```diff
@@ -23,7 +23,17 @@ func (rt *AccessControlRoundTripper) RoundTrip(req *http.Request) (*http.Respons
...
 	}
 
-	gvk, err := rt.restMapper.KindFor(gvr)
+	// Get restMapper at request time (lazy evaluation)
+	// This ensures we get the initialized restMapper even if the wrapper
+	// was created before restMapper was set (fixes issue #688)
+	restMapper := rt.restMapperProvider()
+	if restMapper == nil {
+		// restMapper not yet initialized, pass through
+		// This can happen during discovery client initialization
```
Cali0707: @manusa I think this _may_ be incorrect - as far as I can tell the discovery requests would go direct to the delegate already (in lines 20-24)

So as far as I can tell, this would be a legitimate resource request, and sending it to the delegate would be bypassing our access control rules

manusa: this is more a defensive programming branch than anything else.
Technically, it shouldn't be possible to happen (it was completely impossible for me to reproduce the issue unless I crafted a test that modified the configurations with a hack).
I can change this to fail the request instead.

manusa: This was a good catch, I addressed the issue and made it more defensive. I feel the roundtripper is safer now.

matzew:
LGTM

Issue #568, opened 2025-12-10
Support granular ReadOnly permissions


aljesusg:
# What?

The current implementation of the MCP server creates a binary distinction between "Read-Only" and "Write" tools. When the server is started with the --read-only flag, any tool that is not explicitly marked with ReadOnlyHint: true is effectively disabled or hidden from the client.

# Example

Currently, the manage_istio_config tool in the Kiali toolset has the ReadOnlyHint explicitly set to false. Ref: [pkg/toolsets/kiali/tools/manage_istio_config.go#L56](https://www.google.com/search?q=%5Bhttps://github.com/containers/kubernetes-mcp-server/blob/main/pkg/toolsets/kiali/tools/manage_istio_config.go%23L56%5D(https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/toolsets/kiali/tools/manage_istio_config.go%23L56))

This is technically correct because the tool supports destructive actions like create, patch, and delete.

## The Problem
When the MCP server is running in Read-Only mode, the entire manage_istio_config tool becomes unavailable/unusable.

However, this tool also handles safe list and get operations. This means that in a restricted environment (read-only), users lose the ability to view Istio configurations entirely via this tool, which is a significant limitation.

## Proposed Solution
We should find a way to allow this tool to function in Read-Only mode for specific actions, rather than splitting it into two separate tools (e.g., get_config vs modify_config).

Goal: Keep the tool interface unified but restrict operations dynamically.

## Suggestions:

Dynamic Permission Check: Allow the tool to be invoked in Read-Only mode, but strictly validate the action argument.

If action is list or get -> Allow.

If action is create, patch, or delete -> Deny (return a clear error message citing read-only mode).

Metadata Enhancement: (Longer term) Investigate if the server implementation can support finer-grained hints, where a tool is partially available based on input parameters.

## Current Workaround
The only current workaround would be splitting the tool into two, which degrades the user experience and adds maintenance overhead. We should avoid this if possible.

manusa:
Same applies to `DestructiveHint` and `--disable-destructive`

Cali0707:
@manusa @aljesusg love this idea - thinking about the proposal a bit more, maybe we can:
1. add a new hint to _our_ api layer like `PartialReadOnly` so that when the `--read-only` flag is set those tools are still exposed (maybe even with the `ReadOnly` annotation set) while the actual `ReadOnly` annotation can still normally be set to false (as it isn't fully a read only tool)
[...]

PR #692, opened 2026-01-23
feat(kiali): split istio config read tool

aljesusg:
- Split Kiali Istio config management into read-only and write tools so list/get remain available in read-only mode.
- Added a dedicated read-only tool and registered it in the Kiali toolset.
- Updated tool metadata snapshot to reflect the new tool and revised write-tool schema.

Related with https://github.com/containers/kubernetes-mcp-server/issues/568

cc @manusa

### DIFF
--- a/pkg/toolsets/kiali/tools/manage_istio_config.go
+++ b/pkg/toolsets/kiali/tools/manage_istio_config.go
@@ -117,11 +117,6 @@ func validateIstioConfigInput(action, namespace, group, version, kind, name, jso
 				return fmt.Errorf("json_data is required for action %q", action)
 			}
 		}
-		if action == "get" {
-			if name == "" {
-				return fmt.Errorf("name is required for action %q", action)
-			}
-		}
 		if action == "delete" {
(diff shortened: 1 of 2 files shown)


pkg/toolsets/kiali/tools/manage_istio_config.go
```diff
@@ -83,29 +83,26 @@ func istioConfigHandler(params api.ToolHandlerParams) (*api.ToolCallResult, erro
...
-			if group == "" {
-				return fmt.Errorf("group is required for action %q", action)
-			}
-			if version == "" {
-				return fmt.Errorf("version is required for action %q", action)
-			}
-			if kind == "" {
-				return fmt.Errorf("kind is required for action %q", action)
-			}
+	case "create", "patch", "delete":
```
matzew, line 94: in line 120 there is still the validation for the `get` - which got removed from this `switch`

matzew:
perhaps a bit of enhancing tests (on separate PR?) would be good.

PR #695, opened 2026-01-23
cleanup(tokenexchange): exchange logic contained in kubernetes provider layer

Cali0707:
This PR is a follow up on #604 

In that PR we added more comprehensive token exchange support, however ran into issues with the token exchange code ending up split across many places. This PR aims to simplify this flow.

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -8,7 +8,6 @@ import (
 	"slices"
 	"time"
 
-	"github.com/coreos/go-oidc/v3/oidc"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
@@ -70,7 +69,7 @@ type Server struct {
 	metrics        *metrics.Metrics // Metrics collection system
 }
 
-func NewServer(configuration Configuration, oidcProvider *oidc.Provider, httpClient *http.Client) (*Server, error) {
+func NewServer(configuration Configuration, targetProvider internalk8s.Provider) (*Server, error) {
 	s := &Server{
 		configuration: &configuration,
 		server: mcp.NewServer(
@@ -89,6 +88,7 @@ func NewServer(configuration Configuration, oidcProvider *oidc.Provider, httpCli
 				},
 				Instructions: configuration.ServerInstructions,
 			}),
+		p: targetProvider,
 	}
 
 	// Initialize metrics system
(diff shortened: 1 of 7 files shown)


matzew:
LGTM

Perhaps we could add some test for new `tokenExchangingProvider` ?

manusa:
This is great, I did an initial pass to do sth similar last week but most attempts introduced far too many breaking changes.
Let me review this properly, but by the initial look of the mcp package it's going in the right direction ❤️

pkg/mcp/mcp.go
```diff
@@ -116,7 +112,7 @@ func NewServer(configuration Configuration, oidcProvider *oidc.Provider, httpCli
 	if configuration.RequireOAuth && false { // TODO: Disabled scope auth validation for now
 		s.server.AddReceivingMiddleware(toolScopedAuthorizationMiddleware)
 	}
-	s.p, err = internalk8s.NewProvider(s.configuration.StaticConfig)
+	s.p, err = internalk8s.NewProvider(s.configuration.StaticConfig, internalk8s.WithTokenExchange(oidcProvider, httpClient))
```
manusa: At some point we can even go further and pass in the provided instance instead of the oidcProvider and httpClient

Cali0707: @manusa would you mind expanding a bit on what you mean by "provided instance"? Not sure I follow...

Cali0707: Happy to add this to the PR, just not sure what it is you are asking for

manusa:
```
internalk8s.NewProvider(s.configuration.StaticConfig, internalk8s.WithTokenExchange(oidcProvider, httpClient))
```

This is instantiated in the block calling `mcp.NewServer` instead of within the NewServer function.

(i.e. right before https://github.com/containers/kubernetes-mcp-server/blob/dd5724af55ef85c3b38f45e69247bf5f12f24ab7/pkg/kubernetes-mcp-server/cmd/root.go#L361)

The function signature for NewServer is changed to `func NewServer(configuration Configuration, provider kubernetes.Provider)`, which receives a Kubernetes provider instead of the OIDC provider and http client that are strictly related to auth

manusa: My initial explanation was poor but also contained a typo, meant provider not provided :)

Cali0707: Ah okay, that makes sense - will update here in a bit. Thx for clarifying!

Cali0707:
@manusa I think this should be good now :)

Issue #642, opened 2026-01-14
[EVALS] Kiali tasks run during standard evaluations despite filtering infrastructure


manusa:
## Description

The gevals workflow (`gevals.yaml`) includes infrastructure intended to filter out Kiali tasks when Istio/Kiali is not installed. However, this filtering mechanism is not working as intended, causing Kiali tasks to run and fail during standard evaluation runs, which negatively impacts the overall success rate.

**This is also a problem when running gevals locally** - there is currently no easy or documented way to exclude Kiali tasks from a local evaluation run.

### Current Workflow Logic

The workflow at `.github/workflows/gevals.yaml` has the following logic:

1. [**Lines 82-87**](https://github.com/containers/kubernetes-mcp-server/blob/f6a3e4fbab9d7cff72a426129c8c34ffb8c9b290/.github/workflows/gevals.yaml#L82-L87): Sets `kiali-run=true` ONLY if the `task-filter` input contains the word "kiali"
2. [**Lines 109-111**](https://github.com/containers/kubernetes-mcp-server/blob/f6a3e4fbab9d7cff72a426129c8c34ffb8c9b290/.github/workflows/gevals.yaml#L109-L111): Istio/Kiali infrastructure is ONLY installed when `kiali-run=true`
3. [**Lines 113-116**](https://github.com/containers/kubernetes-mcp-server/blob/f6a3e4fbab9d7cff72a426129c8c34ffb8c9b290/.github/workflows/gevals.yaml#L113-L116): The MCP server enables the kiali toolset ONLY when `kiali-run=true`
4. [**Line 124**](https://github.com/containers/kubernetes-mcp-server/blob/f6a3e4fbab9d7cff72a426129c8c34ffb8c9b290/.github/workflows/gevals.yaml#L124): The `task-filter` parameter defaults to empty string `''`

### The Problem

When `task-filter` is empty (the default for scheduled runs and local executions):
- **All tasks run** because the glob pattern in `eval.yaml` (`../tasks/*/*/*.yaml`) matches both `kubernetes/` and `kiali/` task directories
- **Istio/Kiali is NOT installed** because `task-filter` doesn't contain "kiali"
- **Result**: All ~19 Kiali tasks run but fail because the required infrastructure isn't present

### Why Regex Filtering Won't Work

The gevals tool filters tasks using the `--run` flag, which matches against `taskSpec.Metadata.Name` (the task name in the YAML file), NOT the file path.

[...]

PR #696, opened 2026-01-26
feat(evals): update mcpchecker to use labels

josunect:
Update mcpchecker workflow to use labels to filter different tasks. 

Requires: https://github.com/genmcp/gevals/issues/111 

Fixes https://github.com/containers/kubernetes-mcp-server/issues/642

### DIFF
--- a/.github/workflows/mcpchecker.yaml
+++ b/.github/workflows/mcpchecker.yaml
@@ -122,16 +122,7 @@ jobs:
             esac
           fi
 
-          # Backward compatible behavior:
-          # - If suite is explicitly provided, it wins.
-          # - Otherwise, if task-filter mentions "kiali", treat this as a kiali run.
-          # - Otherwise default to kubernetes-only.
-          SUITE="kubernetes"
-          if [[ -n "$SUITE_INPUT" ]]; then
-            SUITE="$SUITE_INPUT"
-          elif [[ "$TASK_FILTER" =~ kiali ]]; then
-            SUITE="kiali"
-          fi
+          SUITE="${SUITE_INPUT:-$SUITE}"
 
           # Select label-selector and infrastructure based on suite
           # All suites use the same eval.yaml file; suite controls label-selector + infra.


josunect:
Waiting for the new release of mcpchecker with labels support

Cali0707:
@josunect here is the release: https://github.com/mcpchecker/mcpchecker/releases/tag/v0.0.4

josunect:
@Cali0707 @manusa this should be ready for review

manusa (approved):
LGTM, thx.
At some point we should improve the pipeline to support kubevirt too.

.github/workflows/mcpchecker.yaml
```diff
@@ -90,13 +100,53 @@ jobs:
...
+              echo "kiali-run=true" >> $GITHUB_OUTPUT
+              ;;
+            all)
+              echo "label-selector=" >> $GITHUB_OUTPUT  # No filter: run all taskSets
+              echo "kiali-run=true" >> $GITHUB_OUTPUT
+              ;;
+            *)
+              echo "label-selector=suite=kubernetes" >> $GITHUB_OUTPUT
+              echo "kiali-run=false" >> $GITHUB_OUTPUT
+              ;;
```
Cali0707: IMO we don't need to keep the backwards compatible behaviour, especially since it is complicating this already complex github action. Since this is only triggerable by maintainers, I think it is reasonable to expect that it is called correctly with the new command

evals/tasks/kiali/show-topology/show-topology.yaml
```diff
@@ -1,5 +1,8 @@
 kind: Task
 metadata:
+  labels:
+    suite: kiali
+    requires: istio
```
Cali0707: what are we using these `requires` labels for?

josunect: At the moment they are not used, it was an example of classification. But, I will remove it.

.github/workflows/mcpchecker.yaml
josunect: Ok, so I will remove it as well

josunect:
@Cali0707 conflicts solved!

nader-ziada:
@Cali0707 can this be merged?

PR #697, opened 2026-01-26
chore(ci): rename gevals to mcpchecker

Cali0707:
The gevals project was recently renamed and migrated to mcpchecker. This PR migrates our CI over to the new version & release of the project.

### DIFF
--- a/.github/workflows/mcpchecker.yaml
+++ b/.github/workflows/mcpchecker.yaml
@@ -134,7 +134,7 @@ jobs:
         uses: mcpchecker/mcpchecker/.github/actions/mcpchecker-action@v0.0.3
         with:
           eval-config: 'evals/openai-agent/eval.yaml'
-          mcpchecker-version: 'v0.0.3'
+          mcpchecker-version: 'latest'
           task-filter: ${{ github.event.inputs.task-filter || '' }}
           output-format: 'json'
           verbose: ${{ github.event.inputs.verbose || 'false' }}


.github/workflows/mcpchecker.yaml
manusa: I believe we should also update the trigger command to `/run-mcpchecker`
```suggestion
       contains(github.event.comment.body, '/run-mcpchecker'))
```

.github/workflows/mcpchecker.yaml
```diff
@@ -129,17 +129,17 @@ jobs:
...
-      - name: Run gevals evaluation
-        id: gevals
-        uses: genmcp/gevals/.github/actions/gevals-action@v0.0.2
+      - name: Run mcpchecker evaluation
+        id: mcpchecker
+        uses: mcpchecker/mcpchecker/.github/actions/mcpchecker-action@v0.0.3
         with:
           eval-config: 'evals/openai-agent/eval.yaml'
-          gevals-version: 'latest'
+          mcpchecker-version: 'v0.0.3'
```
manusa: If we're going to actively work on mcpchecker in paraller, maybe leave this as latest to avoid the releasing overhead.

Cali0707: Sure, I was just thinking that this way we are less likely to introduce a breaking change. What do you prefer?

manusa: I don't mind, I pointed this out in case we are going to add additional functionality that will be leveraged here. If we go with versioned tags, for each iteration we'll need to cut a mcpchecker release and update the pipeline.
If we expect to add features to mcpcheckers (e.g. result parsing and tooling), then we might want to keep this as latest.
If we don't, or the changes are going to be minimal, then let's keep the release version reference.

Cali0707: good point - switched back to latest

manusa: Just to be clear, once we have things stable, we need to reference the version or the hash to harden the pipeline security 😄🔒

Cali0707: In theory I also fixed the bug where the version tag for the action was not being parsed and used for the mcpchecker version which is installed. So then we can pick up dependabot updates instead (if we just remove this param).

Cali0707: Current logic is:
1. If mcpchecker-version is set, use that
2. Fallback to action tag
3. Fallback to latest

PR #701, opened 2026-01-27
fix(ci): remove version field from OCI package for MCP registry compliance

manusa:
Part of #555

The MCP registry now requires OCI packages to not have a separate 'version' field - the version should only be in the 'identifier' (e.g., image:1.0.0).

https://github.com/containers/kubernetes-mcp-server/actions/runs/21391968827/job/61580817303


matzew:
> OCI packages must not have 'version' field 

Interesting, but I guess less redundancy anyways :sweat_smile:

PR #703, opened 2026-01-27
feat(kubevirt): VirtualMachineSnapshot Support

Ruclo:
In this PR, I expanded the kubevirt toolset with additional tools allowing users to create virtual machine snapshots, restore snapshots and manage the snapshot feature gate. The second commit contains tests which cover this functionality.

### DIFF
--- a/pkg/kubevirt/snapshot_test.go
+++ b/pkg/kubevirt/snapshot_test.go
@@ -1,0 +1,602 @@
+package kubevirt
+
+import (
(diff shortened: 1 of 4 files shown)


lyarwood:
Thanks for looking into this @Ruclo! As discussed privately, could you look at writing some basic evals for these use cases first in this PR and then use the results to understand what if anything else is required in terms of tooling and prompts here?

PR #712, opened 2026-01-29
refactor(mcp): make mcp registration/deregistration consistent

Cali0707:
While exploring adding MCP resources/resource templates to the repo I found that there was a lot of duplicated logic around prompt/tool registration/removal.

This PR refactors that logic to make it more maintainable and easier to introduce resources in the future

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -129,11 +129,14 @@ func (s *Server) reloadToolsets() error {
 		return err
 	}
 
+	// TODO: No option to perform a full replacement of tools.
+	// s.server.SetTools(tools...)
+
 	// Collect applicable items
 	applicableTools := s.collectApplicableTools(targets)
 	applicablePrompts := s.collectApplicablePrompts()
 
-	// Reload tools
+	// Reload tools, and track the newly enabled tools so that we can diff on reload to figure out which to remove (if any)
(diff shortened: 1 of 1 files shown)


pkg/mcp/mcp.go
```diff
@@ -126,96 +129,124 @@ func (s *Server) reloadToolsets() error {
...
-	// s.server.SetTools(m3labsServerTools...)
-
-	// Track previously enabled tools
-	previousTools := s.enabledTools
+	// Start new watch
+	s.p.WatchTargets(s.reloadToolsets)
+	return nil
+}
 
-	// Build new list of applicable tools
```
manusa, line 145: These comments are very important.
It basically states the main reasoning for the overhead and why we're storing a snapshot of the tools (and now prompts, resources).

For example, I did a PR review with Opus and it didn't understand why this was being done, even reported it as something to fix (performance improvement).

manusa (approved):
Thanks for taking care of this.

It looks good, but the documentation and reasoning for why we keep a snapshot of the tools and why we deregister and reregister them again is very important and should be kept.

If this needs further clarification, let me know and I'll elaborate a little bit more.

Cali0707:
Added back some comments, lmk if those makes sense @manusa 

Sorry it took so long, forgot about this PR 😅

PR #713, opened 2026-01-29
[DNM] Spike: add resources

Cali0707:
This PR explores how to add resources to the MCP server

*DO NOT MERGE*

I want to explore how different clients use resources, with this branch as an experiment, to see how we should best implement these

### DIFF
--- a/pkg/toolsets/core/resource_templates.go
+++ b/pkg/toolsets/core/resource_templates.go
@@ -1,0 +1,128 @@
+package core
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	apiextensionsv1spec "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func initResourceTemplates() []api.ServerResourceTemplate {
+	return []api.ServerResourceTemplate{
+		crdOpenAPISpecResourceTemplate(),
+	}
+}
+
+// crdOpenAPISpecResourceTemplate returns a resource template that provides the OpenAPI spec for a CRD.
+// The URI format is: k8s://crds/{name}/openapi
+// Where {name} is the full CRD name (e.g., "virtualmachines.kubevirt.io")
+func crdOpenAPISpecResourceTemplate() api.ServerResourceTemplate {
+	return api.ServerResourceTemplate{
+		ResourceTemplate: api.ResourceTemplate{
+			Name:        "crd-openapi-spec",
+			Title:       "CRD OpenAPI Specification",
+			Description: "Returns the OpenAPI v3 schema for a Custom Resource Definition (CRD). The schema describes the structure and validation rules for custom resources of this type. Use this to figure out how to structure resource tool calls when you are unsure of the schema",
+			URITemplate: "k8s://crds/{name}/openapi",
+			MIMEType:    "application/json",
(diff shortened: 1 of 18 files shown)


Cali0707:
It seems like, at a minimum, we probably want to only use static resources and not a template: https://github.com/anthropics/claude-code/issues/4110

This is _probably_ fine since I believe most clients do not load each resource into the agent's context

manusa:
> It seems like, at a minimum, we probably want to only use static resources and not a template: [anthropics/claude-code#4110](https://github.com/anthropics/claude-code/issues/4110)
> 
> This is _probably_ fine since I believe most clients do not load each resource into the agent's context

In general, I never attempted to add resources to the server because this are mostly used at the expense of the user. At least, a couple of months ago, users of VSCode, claude, etc. had to specifically reference a Resource as part of their prompt.

I think it was part of the Stacklock Kubernetes MCP server that they were actually exposing all Kubernetes resources as discoverable MCP resources. Something pretty neat, but that it just doesn't work because the agents/MCPHosts aren't designed that way.

I'm fine with adding the API support for resources, but I wouldn't go much far that that for the moment. This might be a required feature for some toolset implementers and we should definitely add support for it. But other than that, I wouldn't work on our toolsets to add resource discovery until this proves to be something valuable in any of the mainstream coding agents.

pkg/toolsets/core/resource_templates.go
```diff
@@ -0,0 +1,128 @@
...
+func initResourceTemplates() []api.ServerResourceTemplate {
+	return []api.ServerResourceTemplate{
+		crdOpenAPISpecResourceTemplate(),
+	}
+}
+
+// crdOpenAPISpecResourceTemplate returns a resource template that provides the OpenAPI spec for a CRD.
+// The URI format is: k8s://crds/{name}/openapi
+// Where {name} is the full CRD name (e.g., "virtualmachines.kubevirt.io")
+func crdOpenAPISpecResourceTemplate() api.ServerResourceTemplate {
```
matzew, line 23: In theory I get this via `resources_get` somewhat, but here we have a little more convenient usage. Instead of going via `CustomResourceDefinition` tool for `_get`

matzew:
> I'm fine with adding the API support for resources, but I wouldn't go much far that that for the moment. This might be a required feature for some toolset implementers and we should definitely add support for it.

I like adding the api enhancements that are done in the first commit - I also don't mind that crd template. Or would that be already too much, @manusa ?

matzew:
> we probably want to only use static resources

@Cali0707 like included/embedded documentation?

manusa:
> I like adding the api enhancements that are done in the first commit - I also don't mind that crd template. Or would that be already too much, @manusa ?

My reply was mostly generic on the PR itself, didn't check the actual code.
However, it's my understanding that in this case the user will still need to know the name of the CRD, etc. to be able to make use of it in a prompting session. I'm not sure of the value that this will provide.

The same outcome can probably be achieved by https://github.com/containers/kubernetes-mcp-server/issues/664

But haven't checked this out, so maybe I'm just wrong with my current assumptions.

matzew:
> However, it's my understanding that in this case the user will still need to know the name of the CRD, etc. to be able to make use of it in a prompting session. 

yeah, I had similar thought: https://github.com/containers/kubernetes-mcp-server/pull/713#discussion_r2745676562 

Didn't know about https://github.com/containers/kubernetes-mcp-server/issues/664 - but some tool like that might be handy, agreed

manusa:
Closing this out since it was a DNM spike and most of it has already landed.

The resources/templates infrastructure went in via #1091, got hardened in #1114, and was made transactional in #1130. All toolsets already implement the new GetResources / GetResourceTemplates interface.

The bits this spike surfaced that we didn't pick up (annotations, title field, params-struct handler signature) are now tracked in #1157. The CRD OpenAPI template idea is covered by #664.

Thanks for the exploration @Cali0707!

Issue #717, opened 2026-02-02
Some tools still missing `properties` field in InputSchema causing OpenAI API errors


hpcnt-leo-s:
### Description
  Following up on #340, some tools still have `InputSchema` definitions without the `properties` field,
   causing errors when used with OpenAI API through MCP proxy (e.g., Windmill).
  **Error Message:**

  ExecutionErr: execution error:
  Internal: API error: HTTP status client error (400 Bad Request) for url
  (https://api.openai.com/v1/responses) - 
```
{
    "error": {
      "message": "Invalid schema for function 'mcp_k8s_configuration_contexts_list': In context=(), object schema missing properties.",
      "type": "invalid_request_error",
      "param": "tools[19].parameters",
      "code": "invalid_function_parameters"
    }
}
```

  ### Root Cause
  The following tools define `InputSchema` with only `Type: "object"` but without `Properties`:
  1. **`configuration_contexts_list`** in `pkg/toolsets/config/configuration.go` (line 20-22):

  InputSchema: &jsonschema.Schema{
      Type: "object",
  },

  2. **`targets_list`** in `pkg/toolsets/config/configuration.go` (line 44-46):

  InputSchema: &jsonschema.Schema{
      Type: "object",
  },

  3. **`kcp_workspaces_list`** in `pkg/toolsets/kcp/workspaces.go` (line 20-22):

  InputSchema: &jsonschema.Schema{
      Type: "object",
  },

  While `{"type":"object"}` is a valid JSON Schema, OpenAI's function calling API requires the
  `properties` field to be present (even if empty).
  ### Expected Behavior
  These tools should include an empty `Properties` field:

  InputSchema: &jsonschema.Schema{
      Type:       "object",
      Properties: map[string]*jsonschema.Schema{},
  },

  ### Environment
  - kubernetes-mcp-server: latest (commit ceb360b)
  - Client: Windmill MCP Proxy → OpenAI API
  ### Related Issues
  - #340 (closed but fix was incomplete)
  - #339 (PR that fixed namespace tools only)
  - https://github.com/containers/kubernetes-mcp-server/pull/606

PR #718, opened 2026-02-02
fix(mcp): ensure InputSchema.Properties is initialized for all tools

manusa:
Initialize InputSchema.Properties to an empty map in ServerToolToGoSdkTool
when converting tools to MCP SDK format. This ensures OpenAI API
compatibility, which requires the properties field to be present even
when empty.

Previously, only cluster-aware tools had Properties initialized via
the WithTargetParameter mutator. Non-cluster-aware tools like
configuration_contexts_list and kcp_workspaces_list were missing
this field, causing OpenAI API errors.

Fixes: #717

### DIFF
--- a/pkg/mcp/gosdk.go
+++ b/pkg/mcp/gosdk.go
@@ -7,11 +7,18 @@ import (
 	"fmt"
 
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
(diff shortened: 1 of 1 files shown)


matzew:
Interesting issue! :sweat_smile:

PR #734, opened 2026-02-03
feat(config): add evals for 'config' toolset

matzew:
Adding initial evals for the `config` toolset


evals/tasks/config/list-contexts/cleanup.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+# No cleanup needed - we didn't create any resources
+echo "No cleanup required for config inspection task"
```
matzew: I guess these could be skipped and just use the `verify` .

matzew:
Also might be good to use the `extensions` for this, see https://github.com/mcpchecker/kubernetes-extension/pull/15

evals/tasks/config/list-contexts/cleanup.sh
Cali0707: Yeah I think we are okay to not have cleanup

Cali0707:
> Also might be good to use the `extensions` for this

@matzew I made a new release here for you: https://github.com/mcpchecker/kubernetes-extension/releases/tag/v0.0.3

matzew:
@Cali0707 changed to use the new release!

evals/tasks/config/list-contexts/list-contexts.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+  setup:
+    # No setup needed - working with existing kubeconfig
+
+  verify:
+    - kubernetes.listContexts:
+        # This operation lists all contexts and returns:
+        # - current: name of current context
+        # - count: number of contexts found
+        # Will fail if no contexts exist
```
Cali0707, line 18: @matzew whats the point of this? I guess we maybe need support in mcpchecker to reference previous outputs within the LLM judge or something to compare the answer?

Cali0707:
> it's backwards compatible.

This is the answer 😄

PR #739, opened 2026-02-04
feat(http): implement tls serving certs for server

mjudeikis:
Add tls service certs, allowing to provide TLS certificates to the server itself.

This allows usages:
* serviceType: Loadbalancer + external certs
* certmagic as side container

In general simplified configuration when neither ingress or gateway is available

### DIFF
--- a/charts/kubernetes-mcp-server/README.md
+++ b/charts/kubernetes-mcp-server/README.md
@@ -76,6 +76,7 @@ Each container accepts any valid Kubernetes container field including `image`, `
 | configFilePath | string | `"/etc/kubernetes-mcp-server/config.toml"` |  |
 | defaultPodSecurityContext | object | `{"seccompProfile":{"type":"RuntimeDefault"}}` | Default Security Context for the Pod when one is not provided |
 | defaultSecurityContext | object | `{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"runAsNonRoot":true}` | Default Security Context for the Container when one is not provided |
+| extraArgs | list | `[]` | Note: For TLS configuration, use the tls section above instead of extraArgs. |
 | extraContainers | list | `[]` | Each container is defined as a complete container spec. |
 | extraVolumeMounts | list | `[]` | Additional volumeMounts on the output Deployment definition. |
 | extraVolumes | list | `[]` | Additional volumes on the output Deployment definition. |
@@ -122,14 +123,21 @@ Each container accepts any valid Kubernetes container field including `image`, `
 | replicaCount | int | `1` | This will set the replicaset count more information can be found here: https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/ |
 | resources | object | `{"limits":{"cpu":"100m","memory":"128Mi"},"requests":{"cpu":"100m","memory":"128Mi"}}` | Resource requests and limits for the container. |
 | securityContext | object | `{}` | Define the Security Context for the Container |
-| service | object | `{"port":8080,"targetPort":"http","type":"ClusterIP"}` | This is for setting up a service more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/ |
+| service | object | `{"annotations":{},"port":8080,"targetPort":"http","type":"ClusterIP"}` | This is for setting up a service more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/ |
+| service.annotations | object | `{}` | Annotations to add to the service |
 | service.port | int | `8080` | This sets the ports more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports |
-| service.targetPort | string | `"http"` | Target port for the service. Useful when deploying with an proxy sidecar. Set this to the sidecar's port to route traffic through the proxy before reaching the main container. |
+| service.targetPort | string | `"http"` | Target port for the service. Useful when deploying with an proxy sidecar or exposing a different port. Set this to the sidecar's port to route traffic through the proxy before reaching the main container. |
 | service.type | string | `"ClusterIP"` | This sets the service type more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types |
 | serviceAccount | object | `{"annotations":{},"create":true,"name":""}` | This section builds out the service account more information can be found here: https://kubernetes.io/docs/concepts/security/service-accounts/ |
 | serviceAccount.annotations | object | `{}` | Annotations to add to the service account |
 | serviceAccount.create | bool | `true` | Specifies whether a service account should be created |
 | serviceAccount.name | string | `""` | If not set and create is true, a name is generated using the fullname template |
+| tls | object | `{"certFile":"tls.crt","enabled":false,"keyFile":"tls.key","mountPath":"/etc/tls","secretName":""}` | This is the recommended way to enable TLS instead of using extraArgs. |
+| tls.certFile | string | `"tls.crt"` | Name of the certificate file within the secret (default: tls.crt) |
+| tls.enabled | bool | `false` | Enable TLS for the MCP server |
+| tls.keyFile | string | `"tls.key"` | Name of the key file within the secret (default: tls.key) |
+| tls.mountPath | string | `"/etc/tls"` | Path where the TLS secret will be mounted inside the container |
+| tls.secretName | string | `""` | The secret should be of type kubernetes.io/tls with tls.crt and tls.key. |
 | tolerations | list | `[]` |  |
 
 ## Updating the README
--- a/charts/kubernetes-mcp-server/values.yaml
+++ b/charts/kubernetes-mcp-server/values.yaml
@@ -118,8 +118,11 @@ service:
   type: ClusterIP
   # -- This sets the ports more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports
   port: 8080
-  # -- Target port for the service. Useful when deploying with an proxy sidecar. Set this to the sidecar's port to route traffic through the proxy before reaching the main container.
+  # -- This sets the target port on the pod. Defaults to "http" (the container port name).
+  # -- Target port for the service. Useful when deploying with an proxy sidecar or exposing a different port. Set this to the sidecar's port to route traffic through the proxy before reaching the main container.
   targetPort: http
+  # -- Annotations to add to the service
+  annotations: {}
 
 # -- This block is for setting up the ingress for more information can be found here: https://kubernetes.io/docs/concepts/services-networking/ingress/
 ingress:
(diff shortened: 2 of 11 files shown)


mjudeikis:
Still, a few other things are needed. Not straight forward as I expected :D

charts/kubernetes-mcp-server/README.md
```diff
@@ -1,8 +1,6 @@
 # kubernetes-mcp-server
```
mjudeikis, line 1: This file was regenerated using helm-docs. looks like it was not updated for some time.

pkg/http/http.go
```diff
@@ -87,9 +105,14 @@ func Serve(ctx context.Context, mcpServer *mcp.Server, staticConfig *config.Stat
 	// Wrap with metrics middleware
 	instrumentedHandler := metricsMiddleware(wrappedMux, mcpServer)
 
+	// Use a custom error logger to filter out noisy TLS handshake errors
+	// from load balancer health checks
+	errorLog := log.New(&tlsErrorFilterWriter{underlying: os.Stderr}, "", 0)
```
mjudeikis: aws is at fault for this code :D

Dockerfile
```diff
@@ -10,6 +10,6 @@ LABEL io.modelcontextprotocol.server.name="io.github.containers/kubernetes-mcp-s
 WORKDIR /app
 COPY --from=builder /app/kubernetes-mcp-server /app/kubernetes-mcp-server
 USER 65532:65532
-ENTRYPOINT ["/app/kubernetes-mcp-server", "--port", "8080"]
```
mjudeikis: There is a default port in Helm, but if one runs this in a container, they would need to provide the port manually. 
We can set it in `config_default.go`, but it will change how stdin mode behaves, as `--port` controls the mode of operations.

Dockerfile
```diff
@@ -10,6 +10,6 @@ LABEL io.modelcontextprotocol.server.name="io.github.containers/kubernetes-mcp-s
...
+ENTRYPOINT ["/app/kubernetes-mcp-server"]
```
Cali0707, line 13: IMO let's move this change to a separate PR, as it is a breaking change

mjudeikis: https://github.com/containers/kubernetes-mcp-server/pull/752

mjudeikis:
@manusa ready. tests look like flakes, and I don't have access to rerun

charts/kubernetes-mcp-server/README.md
```diff
@@ -92,6 +87,24 @@ Each container accepts any valid Kubernetes container field including `image`, `
...
+| metrics.serviceMonitor | object | `{"annotations":{},"enabled":false,"interval":"","labels":{},"metricRelabelings":[],"relabelings":[],"scheme":"","scrapeTimeout":"","tlsConfig":{}}` | ServiceMonitor configuration for Prometheus Operator monitoring |
+| metrics.serviceMonitor.annotations | object | `{}` | Annotations for the ServiceMonitor |
+| metrics.serviceMonitor.enabled | bool | `false` | Enable ServiceMonitor for Prometheus scraping |
+| metrics.serviceMonitor.interval | string | `""` | Scrape interval (e.g., "30s", "1m") |
+| metrics.serviceMonitor.labels | object | `{}` | Additional labels for the ServiceMonitor (useful for prometheus-operator serviceMonitorSelector) |
+| metrics.serviceMonitor.metricRelabelings | list | `[]` | Metric relabeling rules |
+| metrics.serviceMonitor.relabelings | list | `[]` | Relabeling rules for metrics |
+| metrics.serviceMonitor.scheme | string | `""` | Scheme to use for scraping (http or https) |
+| metrics.serviceMonitor.scrapeTimeout | string | `""` | Scrape timeout (e.g., "10s") |
+| metrics.serviceMonitor.tlsConfig | object | `{}` | TLS configuration for scraping |
```
manusa, line 108: I'm not sure if these were added to document missing values from other features, but it'd be best if we could add them as a separate PR/commit to keep things under the same scope.

pkg/config/config.go
```diff
@@ -69,6 +69,13 @@ type StaticConfig struct {
 	StsScopes            []string `toml:"sts_scopes,omitempty"`
 	CertificateAuthority string   `toml:"certificate_authority,omitempty"`
 	ServerURL            string   `toml:"server_url,omitempty"`
+
+	// TLS configuration for the HTTP server
+	// TLSCert is the path to the TLS certificate file for HTTPS
+	TLSCert string `toml:"tls_cert,omitempty"`
+	// TLSKey is the path to the TLS private key file for HTTPS
+	TLSKey string `toml:"tls_key,omitempty"`
```
manusa, line 77: Maybe add a test case (subtest) to [`TestReadConfigValid`](https://github.com/containers/kubernetes-mcp-server/blob/fac9426a4e602a427747fd9e6f6a4a53242d940f/pkg/config/config_test.go#L70-L207) to ensure deserialization for the fields work and that the fields are not removed eventually.

pkg/http/http.go
```diff
@@ -18,6 +21,21 @@ import (
...
+
+func (w *tlsErrorFilterWriter) Write(p []byte) (n int, err error) {
+	msg := string(p)
+	// Filter out TLS handshake EOF errors - these are typically from
+	// load balancer health checks that just do TCP connects
+	if strings.Contains(msg, "TLS handshake error") && strings.Contains(msg, "EOF") {
+		return len(p), nil // Silently discard
+	}
+	return w.underlying.Write(p)
+}
```
manusa, line 40: Is this really necessary? I'm worried that actually valid messages are silently  discarded.
Maybe we can log them with a higher log level instead? users can then configure logging to filter them.

pkg/http/http.go
```diff
@@ -87,9 +105,14 @@ func Serve(ctx context.Context, mcpServer *mcp.Server, staticConfig *config.Stat
...
+	// Use a custom error logger to filter out noisy TLS handshake errors
+	// from load balancer health checks
+	errorLog := log.New(&tlsErrorFilterWriter{underlying: os.Stderr}, "", 0)
+
 	httpServer := &http.Server{
-		Addr:    ":" + staticConfig.Port,
-		Handler: instrumentedHandler,
+		Addr:     ":" + staticConfig.Port,
+		Handler:  instrumentedHandler,
+		ErrorLog: errorLog,
```
manusa: Maybe this ErrorLog can be set conditionally within the if-else clause next?

pkg/kubernetes-mcp-server/cmd/root.go
```diff
@@ -296,6 +308,22 @@ func (m *MCPServerOptions) Validate() error {
...
+	if tlsCert != "" {
+		if _, err := os.Stat(tlsCert); err != nil {
+			return fmt.Errorf("tls-cert must be a valid file path: %w", err)
+		}
+	}
+	if tlsKey != "" {
+		if _, err := os.Stat(tlsKey); err != nil {
+			return fmt.Errorf("tls-key must be a valid file path: %w", err)
+		}
+	}
```
manusa, line 329: I think that tests for these validation errors can be easily created in root_test.go, we have similar tests for other validations, just follow the pattern.

charts/kubernetes-mcp-server/README.md
```diff
@@ -82,6 +76,7 @@ Each container accepts any valid Kubernetes container field including `image`, `
 | configFilePath | string | `"/etc/kubernetes-mcp-server/config.toml"` |  |
 | defaultPodSecurityContext | object | `{"seccompProfile":{"type":"RuntimeDefault"}}` | Default Security Context for the Pod when one is not provided |
 | defaultSecurityContext | object | `{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"runAsNonRoot":true}` | Default Security Context for the Container when one is not provided |
+| extraArgs | list | `[]` | Useful for passing TLS keys or other configuration options. |
```
manusa: Note that the new settings could also be provided as part of the configmap.yaml

charts/kubernetes-mcp-server/README.md
mjudeikis: will do. might be easier change.

pkg/http/http.go
mjudeikis: I can remove, but this generated a lot of noise in aws (they have tcp as probes). Would be nice to have it to kill the errors.

I can either remove it and live with it or make it higher log level?

charts/kubernetes-mcp-server/README.md
mjudeikis: adding ci job to test dirtiness and check https://github.com/containers/kubernetes-mcp-server/pull/798

manusa: great improvement, thx! shall I merge that one first?

pkg/config/config.go
mjudeikis: done

pkg/http/http.go
mjudeikis: I added that it's initiated only when TLS certs are used. This is when traffic will get served directly, and these errors might pop up. And only at V4

mjudeikis:
@manusa hope this is what you expected in your comments :)

mjudeikis:
@mjudeikis-bot follow this pr and let me know if my attention here is needed.

charts/kubernetes-mcp-server/values.yaml
```diff
@@ -166,6 +169,12 @@ extraVolumeMounts: []
 #   mountPath: "/etc/foo"
 #   readOnly: true
 
+# -- Extra arguments to pass to the kubernetes-mcp-server command line.
+# -- Useful for passing TLS keys or other configuration options. It can also be configured using ConfigMap.
+extraArgs: []
```
manusa, line 200: When TLS is enabled via `extraArgs`, the default `livenessProbe` and `readinessProbe` in `values.yaml` use `httpGet` with `port: http`. These HTTP probes will fail against an HTTPS endpoint since the server won't respond to plain HTTP.

Since TLS is configured through opaque `extraArgs`, the chart has no way to auto-set `scheme: HTTPS` on the probes. Users would need to manually override probes, volumes, and volume mounts — which is error-prone.

**Minimal fix**: Document the required probe override in the comments here (e.g., show that users must also set `livenessProbe.httpGet.scheme: HTTPS`).

**Better approach**: Add a first-class `tls` section to the chart values instead of relying on `extraArgs` for TLS:

```yaml
tls:
  enabled: false
  secretName: ""
  mountPath: /etc/tls
```

When `tls.enabled` is true, the chart template would automatically:
- Inject `--tls-cert` and `--tls-key` args
- Mount the TLS secret as a volume
- Set `scheme: HTTPS` on health probes

This is the standard pattern used by most Helm charts (cert-manager webhook, ingress-nginx, etc.) and provides a much better UX — users just set `tls.enabled: true` and `tls.secretName: my-cert`.

The `extraArgs` value is still useful as a generic escape hatch for other flags.

pkg/config/config.go
```diff
@@ -69,6 +69,13 @@ type StaticConfig struct {
 	StsScopes            []string `toml:"sts_scopes,omitempty"`
 	CertificateAuthority string   `toml:"certificate_authority,omitempty"`
 	ServerURL            string   `toml:"server_url,omitempty"`
+
+	// TLS configuration for the HTTP server
+	// TLSCert is the path to the TLS certificate file for HTTPS
+	TLSCert string `toml:"tls_cert,omitempty"`
```
manusa, line 75: These new TOML options should also be documented in `docs/configuration.md`. Other similar fields like `certificate_authority` and `server_url` are already documented in the configuration reference table there.

pkg/kubernetes-mcp-server/cmd/root.go
```diff
@@ -296,6 +308,22 @@ func (m *MCPServerOptions) Validate() error {
 			return fmt.Errorf("certificate-authority must be a valid file path: %w", err)
 		}
 	}
+	// Validate TLS configuration
```
manusa, line 311: Nit: TLS is only meaningful in HTTP mode (when `port` is set). Consider also validating that `port` is configured when TLS cert/key are provided:

```go
if tlsCert != "" && m.StaticConfig.Port == "" {
    return fmt.Errorf("--tls-cert and --tls-key require --port to be set (TLS is only supported in HTTP mode)")
}
```

This would give users a clear early error instead of silently ignoring the TLS config in STDIO mode.

manusa (changes requested):
Hi, thx for taking care of the previous comments.
I reported a few more minor things 😅 otherwise, I think we should be ready to go 🚀

Issue #743, opened 2026-02-05
helm chart: tpl wrong context in with blocks causes "expected chartutil.Values; got []interface {}"


shyoon-devops:
# Helm chart: `tpl` wrong context in `with` blocks causes "expected chartutil.Values; got []interface {}"

## Summary

Installing the chart with values that set `imagePullSecrets` (or any nested value used inside `{{- with .Values.xxx }}` + `tpl`) fails with:

```
Error: template: kubernetes-mcp-server/templates/deployment.yaml:28:27: executing "kubernetes-mcp-server/templates/deployment.yaml" at <.>: wrong type for value; expected chartutil.Values; got []interface {}
```

With `imagePullSecrets: []`, a similar error appears later (e.g. line 53) for `livenessProbe` because the default values also trigger the same bug.

## Cause

In `charts/kubernetes-mcp-server/templates/deployment.yaml`, every `{{- with .Values.xxx }}` block uses:

```yaml
{{- tpl (toYaml .) . | nindent N }}
```

Inside a `with` block, `.` is the current value (e.g. the array for `imagePullSecrets`, or the probe object for `livenessProbe`). The `tpl` function expects its second argument to be the root context (`chartutil.Values`) so that template expressions in the string can resolve. Passing `.` (array or nested object) causes Helm to error with "expected chartutil.Values; got []interface {}".

## Fix

Use the root context `$` for `tpl` instead of `.`:

- **Before:** `{{- tpl (toYaml .) . | nindent N }}`
- **After:** `{{- tpl (toYaml .) $ | nindent N }}`

Apply this change everywhere in `deployment.yaml` where `tpl (toYaml .) .` appears inside a `with` block (podAnnotations, podLabels, imagePullSecrets, livenessProbe, readinessProbe, resources, extraVolumeMounts, extraContainers, extraVolumes, nodeSelector, affinity, tolerations). The lines that use `tpl` with `.Values.xxx` directly (e.g. `default .Values.defaultPodSecurityContext .Values.podSecurityContext`) are already outside `with`, so `.` is correct there; only the `with`-block usages need to become `$`.

## Environment

- Chart: `oci://ghcr.io/containers/charts/kubernetes-mcp-server` version `0.1.0`
- Helm: 3.x

## Workaround

[...]

PR #745, opened 2026-02-05
chore(ci): add helm chart validation to CI pipeline

manusa:
Add GitHub Actions workflow to validate Helm chart on PRs/pushes affecting chart files. Enhance helm-validate target to test with both default values and tpl-exercising values to catch template context bugs.

Refs #743
Should go in before #744 to verify the fix


manusa:
The new pipeline should fail until we also implement the fix in #744, that PR should rebase once we merge this one.

PR #746, opened 2026-02-05
feat(helm): Adding evals for helm toolset

matzew:


Cali0707 (approved):
LGTM, thanks @matzew - as follow up (as we add support), let's get rid of the scripts and use the newer format

matzew:
yes, that is the plan

PR #748, opened 2026-02-06
fix(helm): use namespace label in recording rules for multi-tenant RBAC

nader-ziada:
Update PrometheusRule recording rules to use the 'namespace' label
(provided by ServiceMonitor discovery) instead of 'k8s_namespace_name'.
This enables namespace-scoped metric queries for multi-tenant RBAC.

### DIFF
--- a/charts/kubernetes-mcp-server/templates/prometheusrule.yaml
+++ b/charts/kubernetes-mcp-server/templates/prometheusrule.yaml
@@ -47,20 +47,19 @@ spec:
         - record: cluster:k8s_mcp_server_info:count
           expr: count(k8s_mcp_server_info)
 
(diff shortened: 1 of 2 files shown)


nader-ziada:
hold while checking if Prometheus would automatically add the namespace label when it discovers the services/pods from the Kubernetes API

nader-ziada:
hold cancel, 
removed the addition of the namespace label as the ServiceMonitor will automatically populate the namespace label from `__meta_kubernetes_namespace`, which represents the metadata.namespace of the scraped pod and  we don't need to duplicate what Prometheus already does.

jan--f:
/lgtm
Minor nit: maybe update the PR description for future readers?

nader-ziada:
> Minor nit: maybe update the PR description for future readers?

updated PR desc and commit msg

PR #757, opened 2026-02-10
feat: Allow Toolset-specific instructions to be added to server instructions

saswatamcode:
This commit adds a new method GetToolsetInstructions to Toolset interface, to allow Toolsets to specify their own custom instructions (LLM-friendly text describing how tools within a particular toolset can be used in tandem).

This is then bubbled up to MCP server initialize response and appended to the provided server prompt. Also added a config option to disable this behavior.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -86,6 +86,10 @@ type StaticConfig struct {
 	// This can be used to provide specific instructions on how the client should use the server
 	ServerInstructions string `toml:"server_instructions,omitempty"`
 
+	// DisableToolsetInstructions indicates whether toolset instructions are to be excluded
+	// from being provided by the MCP server to the MCP client in the initialize response.
+	DisableToolsetInstructions bool `toml:"disable_toolset_instructions,omitempty"`
+
 	// Telemetry contains OpenTelemetry configuration options.
 	// These can also be configured via OTEL_* environment variables.
 	Telemetry TelemetryConfig `toml:"telemetry,omitempty"`
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -71,9 +71,9 @@ type Server struct {
 }
 
 func NewServer(configuration Configuration, targetProvider internalk8s.Provider) (*Server, error) {
-	instructions := buildServerInstructions(configuration.ServerInstructions, configuration.Toolsets())
-	if configuration.DisableToolsetInstructions {
-		instructions = configuration.ServerInstructions
+	instructions := configuration.ServerInstructions
+	if !configuration.DisableToolsetInstructions {
+		instructions = buildServerInstructions(configuration.ServerInstructions, configuration.Toolsets())
 	}
 
 	s := &Server{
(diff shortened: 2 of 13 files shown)


pkg/mcp/mcp.go
```diff
@@ -118,6 +124,26 @@ func NewServer(configuration Configuration, targetProvider internalk8s.Provider)
...
+func buildServerInstructions(serverInstructions string, toolsets []api.Toolset) string {
+	var instructions []string
+
+	if serverInstructions != "" {
+		instructions = append(instructions, serverInstructions)
+	}
+
+	for _, toolset := range toolsets {
+		if toolsetInstructions := toolset.GetToolsetInstructions(); toolsetInstructions != "" {
+			instructions = append(instructions, toolsetInstructions)
```
andreasgerstmayr: I wonder if there should be a separation between the toolsets instructions? Something like
```
# toolset1_name
toolset1_instructions

# toolset2_name
toolset2_instructions
```
or similar?

saswatamcode: Was assuming we'd leave it to the toolsets themselves

pkg/mcp/mcp.go
```diff
@@ -70,6 +71,11 @@ type Server struct {
 }
 
 func NewServer(configuration Configuration, targetProvider internalk8s.Provider) (*Server, error) {
+	instructions := buildServerInstructions(configuration.ServerInstructions, configuration.Toolsets())
+	if configuration.DisableToolsetInstructions {
```
matzew: why bother with the `buildServerInstructions` in case it is disabled?

pkg/config/config.go
```diff
@@ -86,6 +86,10 @@ type StaticConfig struct {
 	// This can be used to provide specific instructions on how the client should use the server
 	ServerInstructions string `toml:"server_instructions,omitempty"`
 
+	// DisableToolsetInstructions indicates whether toolset instructions are to be excluded
+	// from being provided by the MCP server to the MCP client in the initialize response.
+	DisableToolsetInstructions bool `toml:"disable_toolset_instructions,omitempty"`
```
matzew, line 91: do we want this to be always on? (e.g. disabled:false) ?

saswatamcode: My thinking was that the toolsets in this repo would be "trusted", and reviewed, so the instructions could safely always be passed in, and optionally be disabled if the user would like to use something custom

pkg/mcp/mcp.go
saswatamcode: Addressed

pkg/mcp/mcp.go
Cali0707: IMO it would probably be better to handle the separation logic in one place, so that we can keep all of it consistent between toolsets

saswatamcode: I'm separating by adding a newline here, essentially.
However, if we want to centralize the header-based separation in one place, I can add it in

saswatamcode: Added here https://github.com/containers/kubernetes-mcp-server/pull/757/commits/a8a5807d9abecaeb2c093d1582034e90a82faaea

Issue #758, opened 2026-02-10
JSON schema conversion error in llama.cpp


ThomasBuchinger:
I've been debugging a problem with kubernetes-mcp-sever and llama.cpp. The Short version is, that the `pattern` field in JSON-Schema must include the start/end anchors (`^regex$`) or llama.cpp will throw an error.

The error occurs on parsing the request. Therefore just having kubernetes-mcp-server enabled causes all requests to the LLM to fail. I think the easiest solution is to just add start/end anchors to all `patterns`, since there does not seem to be a way to disable this behavior in llama.cpp (also affects Ollama and LMStudio).

## Problem Description

The MCP server itself works fine. The Agent can list the tools and include the list of MCP-Tools in the request to the LLM. However the Request fails with the following error message:

```
POST /v1/chat/completions 127.0.0.1 400
{"error":{"code":400,"message":"JSON schema conversion failed:\nPattern must start with '^' and end with '$'","type":"invalid_request_error"}}
```

llama.cpp internally converts the JSON schema provided by MCP into a grammar, that it checks the output against. ([Code in llama.cpp](https://github.com/ggml-org/llama.cpp/blob/292f6908cdc6abb5c38581e34fa141973e5aba82/common/json-schema-to-grammar.cpp#L346)), that conversion fails if the regex doesn't have the start/end anchors and as far as I can tell, there is not option to disable this.

## Solution

I've tested, that adding `^$` to every `pattern` using curl and confirmed, that this will fix the issue. I also checked the existing patterns and it should be fine to add the anchors to them.

I can look into creating a PR for this sometime this week

## Workarounds

If you run a Proxy (Like Envoy-AI-Gateways `MCPRoute` or LiteLLM) that can filter tools from the MCP, you can disable the affected tools

**Affected Tools (in the default tools)**
- nodes_top
- pods_list
- pods_list_in_namespace
- pods_top
- resources_list

manusa:
Relates to https://github.com/containers/kubernetes-mcp-server/pull/277

PR #759, opened 2026-02-10
fix(mcp): kubernetes-mcp-server is incompatible with llama.cpp (JSON-Schema)

ThomasBuchinger:
fix: https://github.com/containers/kubernetes-mcp-server/issues/758

I changed the Regex from the original `([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]` to 2 different Regex for LabelSelector and FieldSelector.
* LabelSelector: I am just checking if the characters are valid, because the API server will validate the selector anyway (hopefully) and I didn't think a Mega-Regex-To-Rule-Them-All was a good choice here
* FieldSelector: This one checks: KEY (OP_OR_COMMA  + VALUE). since FieldSelectors are more contraint, I opted to check at least a little bit of syntax

I didn't know if I should put the Regex into their own file or not. But creating a seperate file for 2 lines seemed a bit much and toolsets.go seemed the best place to add it

Let me know if you want any changes. 

Supersedes closes https://github.com/containers/kubernetes-mcp-server/pull/277

### DIFF
--- a/pkg/toolsets/core/pods.go
+++ b/pkg/toolsets/core/pods.go
@@ -31,7 +31,7 @@ func initPods() []api.ServerTool {
 					"fieldSelector": {
 						Type:        "string",
 						Description: "Optional Kubernetes field selector to filter pods by field values (e.g. 'status.phase=Running', 'spec.nodeName=node1'). Supported fields: metadata.name, metadata.namespace, spec.nodeName, spec.restartPolicy, spec.schedulerName, spec.serviceAccountName, status.phase (Pending/Running/Succeeded/Failed/Unknown), status.podIP, status.nominatedNodeName. Note: CrashLoopBackOff is a container state, not a pod phase, so it cannot be filtered directly. See https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/",
-						Pattern:     REGEX_LABELSELECTOR_VALID_CHARS,
+						Pattern:     REGEX_FIELDSELECTOR,
 					},
 				},
 			},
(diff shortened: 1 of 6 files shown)


pkg/toolsets/core/pods.go
```diff
@@ -26,12 +26,12 @@ func initPods() []api.ServerTool {
...
 						Type:        "string",
 						Description: "Optional Kubernetes label selector (e.g. 'app=myapp,env=prod' or 'app in (myapp,yourapp)'), use this option when you want to filter the pods by label",
-						Pattern:     "([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]",
+						Pattern:     REGEX_LABELSELECTOR_VALID_CHARS,
 					},
 					"fieldSelector": {
 						Type:        "string",
 						Description: "Optional Kubernetes field selector to filter pods by field values (e.g. 'status.phase=Running', 'spec.nodeName=node1'). Supported fields: metadata.name, metadata.namespace, spec.nodeName, spec.restartPolicy, spec.schedulerName, spec.serviceAccountName, status.phase (Pending/Running/Succeeded/Failed/Unknown), status.podIP, status.nominatedNodeName. Note: CrashLoopBackOff is a container state, not a pod phase, so it cannot be filtered directly. See https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/",
-						Pattern:     "([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]",
+						Pattern:     REGEX_LABELSELECTOR_VALID_CHARS,
```
manusa:
```suggestion
						Pattern:     REGEX_FIELDSELECTOR,
```

I think this is a typo

ThomasBuchinger: yep :)

PR #760, opened 2026-02-10
feat(http): propagate user agent to api server

Cali0707:
Currently, when an agent or user interacts with the MCP server, the API requests made to the API server are not correlated with that client by the `User-Agent` header. This PR adds support for identifying the client by:
1. propagating the user agent if it is present
2. creating a user agent using the mcp session initialization info (client name + version)

In both cases, we add the kubernetes mcp server and version into the user agent, to disambiguate it from requests coming directly from the client.

### DIFF
--- a/pkg/mcp/mcp_test.go
+++ b/pkg/mcp/mcp_test.go
@@ -1,7 +1,9 @@
 package mcp
 
 import (
+	"fmt"
 	"net/http"
+	"runtime"
 	"sync"
 	"testing"
 
@@ -166,7 +168,7 @@ func (s *UserAgentPropagationSuite) TestPropagatesExplicitUserAgentToKubeAPI() {
 	s.InitMcpClient(transport.WithHTTPHeaders(map[string]string{
 		"User-Agent": "custom-mcp-client/2.0",
 	}))
-	_, _ = s.CallTool("pods_list", map[string]interface{}{})
+	_, _ = s.CallTool("pods_list", map[string]any{})
 
 	s.pathHeadersMux.Lock()
 	podsHeaders := s.pathHeaders["/api/v1/namespaces/default/pods"]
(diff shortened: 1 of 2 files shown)


pkg/kubernetes/useragent_roundtripper.go
```diff
@@ -0,0 +1,25 @@
...
+	if userAgent == nil {
+		return u.delegate.RoundTrip(req)
+	}
+
+	userAgentHeader, ok := userAgent.(string)
+	if !ok || userAgentHeader == "" {
+		return u.delegate.RoundTrip(req)
+	}
+
+	req.Header.Set(string(UserAgentHeader), userAgentHeader)
```
nader-ziada, line 25: According to the http.RoundTripper documentation: "RoundTrip should not modify the request, except for consuming and closing the Request's Body."

i believe you are supposed to clone it to avoid a race condition on the shared Header map.

pkg/mcp/mcp_test.go
```diff
@@ -126,3 +127,101 @@ func (s *ServerInstructionsSuite) TestServerInstructionsFromConfiguration() {
...
+	}))
+	_, _ = s.CallTool("pods_list", map[string]interface{}{})
+
+	s.pathHeadersMux.Lock()
+	podsHeaders := s.pathHeaders["/api/v1/namespaces/default/pods"]
+	s.pathHeadersMux.Unlock()
+
+	s.Require().NotNil(podsHeaders, "No requests were made to /api/v1/namespaces/default/pods")
+	s.Run("DynamicClient propagates User-Agent with server prefix to Kube API", func() {
+		s.Equal("kubernetes-mcp-server/0.0.0 custom-mcp-client/2.0", podsHeaders.Get("User-Agent"))
```
matzew: would we want to have the platform too? (e.g. see https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/kubernetes/manager_test.go#L53)

nader-ziada:
LGTM, thanks @Cali0707 

will wait for other folks to take a look

PR #764, opened 2026-02-10
feat(validation): add pre-execution validation layer

nader-ziada:
Add validation middleware that catches errors before they reach the Kubernetes API.

Changes:
 - Resource validation: catches typos in resource types (GVK)
 - Schema validation: validates manifests against OpenAPI schema
 - RBAC validation: pre-checks permissions using SelfSubjectAccessReview
 - All validators are enabled by default and can be configured via environment variables (MCP_VALIDATION_*) or TOML config.

Closes #775


pkg/validation/middleware.go
```diff
@@ -0,0 +1,337 @@
...
+			// Skip validation if we couldn't extract enough info
+			if validationReq == nil {
+				return next(ctx, method, req)
+			}
+
+			// Run validation
+			result := validator.Validate(ctx, validationReq)
+			if !result.Valid {
+				errorMessage := FormatErrors(result.Errors)
+				klog.V(4).Infof("Validation failed for tool %s: %s", params.Name, errorMessage)
```
matzew: I like the logs

```
I0211 08:15:07.026316       1 rbac_validator.go:100] "RBAC pre-check: denied list on /events in : "
I0211 08:15:07.026332       1 middleware.go:108] "Validation failed for tool events_list: Validation Error [PERMISSION_DENIED]: Cannot list events (cluster-scoped)":
```

pkg/validation/rbac_validator.go
```diff
@@ -0,0 +1,117 @@
...
+			req.Namespace,
+		))
+	}
+
+	return NewValidResult()
+}
+
+// canI performs a SelfSubjectAccessReview to check if the current identity
+// has permission to perform the specified operation.
+func (v *RBACValidator) canI(ctx context.Context, req *ValidationRequest) (bool, error) {
```
matzew: @Cali0707 just wondering (un releated) some sort of validation of "larger cluster" access would be nice for multi-cluster/acm use-case. I think we currently list all "managed clusters", w/o some sort of "validation"

pkg/validation/middleware.go
```diff
@@ -0,0 +1,337 @@
...
+
+			// Skip validation if we couldn't extract enough info
+			if validationReq == nil {
+				return next(ctx, method, req)
+			}
+
+			// Run validation
+			result := validator.Validate(ctx, validationReq)
+			if !result.Valid {
+				errorMessage := FormatErrors(result.Errors)
```
matzew: would it make sense to add some metrics for failed validations?

pkg/mcp/mcp.go
```diff
@@ -109,6 +110,16 @@ func NewServer(configuration Configuration, targetProvider internalk8s.Provider)
 	s.server.AddReceivingMiddleware(authHeaderPropagationMiddleware)
 	s.server.AddReceivingMiddleware(toolCallLoggingMiddleware)
 	s.server.AddReceivingMiddleware(s.metricsMiddleware())
+
+	// Add validation middleware if enabled
+	if configuration.Validation.IsEnabled() {
```
manusa: To keep things organized, we should add this to `middleware.go` or a dedicated middleware file.

IMO, from a domain perspective logic, it doesn't make sense either to have a pkg/validation/middleware.go file which even references the go-sdk MCP module.
Validation pkg should be focused on providing the described feature and mcp pkg should be focused on plumbing it.
Ideally pkg/validation should be completely decoupled from `github.com/modelcontextprotocol/go-sdk`.

pkg/mcp/mcp.go
```diff
@@ -280,6 +291,27 @@ func (s *Server) GetMetrics() *metrics.Metrics {
...
+	ctx := context.Background()
+
+	// Get the default target's kubernetes client
+	k8s, err := s.p.GetDerivedKubernetes(ctx, s.p.GetDefaultTarget())
+	if err != nil {
+		klog.Warning("Validation enabled but kubernetes client unavailable, validation disabled: ", err)
+		return nil
+	}
+
+	factory := validation.NewValidatorFactory(k8s.DiscoveryClient(), k8s, k8s.RESTMapper())
```
manusa: I believe this prevents the functionality from workin with multiple clusters. Since we're not using the derived client per request, the validation assumptions will probably be based on the wrong cluster target.

manusa (commented):
Thx for putting this together.
I gave it an initial look and added a few comments.

pkg/validation/rbac_validator.go
manusa: Note that we already have a canIUser method defined here:

https://github.com/containers/kubernetes-mcp-server/blob/dbc1bd0e7ff8e991fc8176d07bb7c055bc5c78ee/pkg/kubernetes/resources.go#L232-L248

pkg/validation/middleware.go
nader-ziada: good idea, will add metrics for validation

pkg/validation/rbac_validator.go
nader-ziada: both methods are unexposed (private methods) and serve different purposes in different contexts
I prefer to have them in different packages with clear separation of concerns.

but let me know if you prefer to combine them somehow and I can do that

manusa: Just noting in the context of the downstream issue that relates to this fix.
I'm not sure that if in that context, we might just want to put this in the AccessControlRoundTripper. 
Here, we're already checking for denied resources in the configuration (for example)

pkg/mcp/mcp.go
nader-ziada: refactored to seperate concerns, moved the middleware func(s) to the mcp package and created a factory in the validation package

pkg/mcp/mcp.go
nader-ziada: fixed to follow the tools patterns of getting the target client

pkg/validation/rbac_validator.go
nader-ziada: my understanding is that AccessControlRoundTripper checks configured denials (checks against a static deny list that the server operator configures. This is separate from Kubernetes RBAC) while the RBACValidator checks actual RBAC (what the cluster says about the current user's permissions)

Are you saying we don't need the RBACValidator since we will get the AccessControlRoundTripper check anyways? but the error message and logging would be different?

manusa: No, I was saying that maybe (if we're just tackling the downstream issue -it's unclear to me the scope of it since it's not properly described-) we can just add RBAC validation to the roundrtipper (simple solution: KISS, YAGNI)

In any case, I'll review everything else in the issue.

What I'd also suggest is to make this opt-in for a couple of vesions at least upstream, since it might have side-effects we haven't accounted for.

nader-ziada:
made a refactor of how this works, sorry to the reviewers, to make work from the kubernetes pkg instead of the mcp pkg, now more dynamic and flexible. also disabled by default.

swghosh (approved):

pkg/config/validation_config.go
```diff
@@ -0,0 +1,28 @@
...
+// When enabled, validates tool calls before execution
+type ValidationConfig struct {
+	// Defaults to false.
+	Enabled *bool `toml:"enabled,omitempty"`
+}
+
+// IsEnabled returns true if validation is enabled.
+// Environment variable MCP_VALIDATION_ENABLED takes precedence over config.
+// Defaults to false.
+func (c *ValidationConfig) IsEnabled() bool {
```
manusa: I don't recall we support config overrides through environment variables at the moment, we do support those through CLI flags.

Is there any reason why we need to be able to override this through an environment variable?

pkg/config/validation_config.go
```diff
@@ -0,0 +1,28 @@
+package config
+
+import (
+	"os"
+	"strconv"
+)
+
+// ValidationConfig contains pre-execution validation configuration.
+// When enabled, validates tool calls before execution
+type ValidationConfig struct {
```
manusa: Are we planning to add more config options to validation?
If so, I understand the nesting of the config into a different struct.
If not, why is the nesting needed?

pkg/kubernetes/validation_round_tripper.go
```diff
@@ -0,0 +1,236 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	authv1client "k8s.io/client-go/kubernetes/typed/authorization/v1"
+	"k8s.io/klog/v2"
+)
+
+// ValidationRoundTripper intercepts HTTP requests to validate them before they reach the K8s API.
+// It extends the AccessControlRoundTripper pattern with additional validator chain execution.
+type ValidationRoundTripper struct {
```
manusa: This seems to be a substitute to the AccessControlRoundTripper, but the former is still retained (which is confusing).

I'm not sure if we prefer to make the composable and compose the when wiring together in kubernetes.go.
Or merging both into this one.
Whatever works better, but we should trim and clean up the redundant bits.

pkg/mcp/events_test.go
```diff
@@ -151,17 +150,12 @@ func (s *EventsSuite) TestEventsListForbidden() {
...
-			s.Contains(toolResult.Content[0].(mcp.TextContent).Text, "forbidden",
-				"error message should indicate forbidden")
-		})
-		s.Run("sends log notification", func() {
-			logNotification := capture.RequireLogNotification(s.T(), 2*time.Second)
-			s.Equal("error", logNotification.Level, "forbidden errors should log at error level")
-			s.Contains(logNotification.Data, "Permission denied", "log message should indicate permission denied")
+			msg := toolResult.Content[0].(mcp.TextContent).Text
+			s.Truef(strings.Contains(msg, "forbidden") || strings.Contains(msg, "PERMISSION_DENIED"),
+				"error message should indicate forbidden or permission denied, got: %s", msg)
```
manusa: Ideally, these tests should not require any changes, so we ensure that there are no behavioral changes after the access control refactor (this applies to all tests in this package that have been changed with similar assertions).

pkg/mcp/pods_test.go
```diff
@@ -66,6 +66,10 @@ func (s *PodsSuite) TestPodsListInAllNamespaces() {
 }
 
 func (s *PodsSuite) TestPodsListInAllNamespacesUnauthorized() {
+	// Disable validation for this test - we're testing the tool's fallback behavior
+	// when cluster-wide list fails but namespace-scoped list succeeds
+	validationDisabled := false
+	s.Cfg.Validation.Enabled = &validationDisabled
```
manusa: I think this is legacy from the initial approach

README.md
```diff
@@ -431,10 +431,10 @@ The following sets of tools are available (toolsets marked with ✓ in the Defau
 
 | Toolset  | Description                                                                                                                                                          | Default |
 |----------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
+| kiali    | Most common tools for managing Kiali, check the [Kiali documentation](https://github.com/containers/kubernetes-mcp-server/blob/main/docs/KIALI.md) for more details. |         |
```
manusa, line 314: I'm not sure why this changed.

pkg/kubernetes/validation_round_tripper.go
nader-ziada: merged the validation login into the AccessControlRoundTripper

pkg/config/validation_config.go
nader-ziada: i originally had different flags for each validator, but decided its too much complexity, will remove struct

pkg/config/validation_config.go
nader-ziada: removed

README.md
nader-ziada: cleaned up

pkg/mcp/pods_test.go
nader-ziada: cleaned up

pkg/mcp/events_test.go
```diff
@@ -3,7 +3,6 @@ package mcp
 import (
```
manusa, line 3: The changes in this file weren't reverted but can be reverted (checked original tests pass locally)

pkg/mcp/namespaces_test.go
```diff
@@ -3,8 +3,8 @@ package mcp
 import (
```
manusa, line 3: The changes in this file weren't reverted but can be reverted (checked original tests pass locally)

pkg/mcp/pods_test.go
```diff
@@ -194,17 +194,12 @@ func (s *PodsSuite) TestPodsListForbidden() {
 	_ = client.RbacV1().ClusterRoles().Delete(s.T().Context(), "allow-all", metav1.DeleteOptions{})
```
manusa, line 194: The changes in this file weren't reverted but can be reverted (checked original tests pass locally)

pkg/mcp/resources_test.go
manusa, line 1: The changes in this file weren't reverted but can be reverted (checked original tests pass locally)

pkg/api/config.go
```diff
@@ -53,6 +53,11 @@ type StsConfigProvider interface {
 	GetStsScopes() []string
 }
 
+// ValidationEnabledProvider provides access to validation enabled setting.
+type ValidationEnabledProvider interface {
+	IsValidationEnabled() bool
+}
+
 type BaseConfig interface {
```
manusa, line 61: We can add ValidationEnabledProvider to this interface so that we don't have to do the type check in `pkg/kubernetes/kubernetes.go`

pkg/kubernetes/kubernetes.go
```diff
@@ -50,24 +50,30 @@ type Kubernetes struct {
...
 		restConfig:      rest.CopyConfig(restConfig),
 	}
 	if k.restConfig.UserAgent == "" {
 		k.restConfig.UserAgent = rest.DefaultKubernetesUserAgent()
 	}
+
+	var validationEnabled bool
+	if vep, ok := baseConfig.(api.ValidationEnabledProvider); ok {
+		validationEnabled = vep.IsValidationEnabled()
+	}
```
manusa: See https://github.com/containers/kubernetes-mcp-server/pull/764/changes#r2828148113

pkg/kubernetes/accesscontrol_round_tripper.go
```diff
@@ -28,17 +67,55 @@ func (rt *AccessControlRoundTripper) RoundTrip(req *http.Request) (*http.Respons
...
 	restMapper := rt.restMapperProvider()
 	if restMapper == nil {
-		return nil, fmt.Errorf("failed to make request: AccessControlRoundTripper restMapper not initialized")
+		return nil, fmt.Errorf("failed to make request: restMapper not initialized")
 	}
 
 	gvk, err := restMapper.KindFor(gvr)
 	if err != nil {
-		return nil, fmt.Errorf("failed to make request: AccessControlRoundTripper failed to get kind for gvr %v: %w", gvr, err)
+		return nil, fmt.Errorf("failed to make request: failed to get kind for gvr %v: %w", gvr, err)
```
manusa: The `ResourceValidator` is actually redundant to what we do here (which is a prereq to ensure that only resources for which the config grants access are enabled).

The easiest solution in this case is to just remove the resource_validator and if necessary improve the error message here by doing the IsNoMatchError

pkg/kubernetes/schema_validator.go
```diff
@@ -0,0 +1,130 @@
...
+	return "schema"
+}
+
+func (v *SchemaValidator) Validate(ctx context.Context, req *api.HTTPValidationRequest) error {
+	if req.GVK == nil || len(req.Body) == 0 {
+		return nil
+	}
+
+	// Only validate for create/update operations
+	if req.Verb != "create" && req.Verb != "update" && req.Verb != "patch" {
```
manusa: Patch can cause false positives in case of partial bodes(JSON Merge Patch, Strategic Merge Patch, or JSON Patch).
For simplicity, maybe we can just exclude patch operations from this validator too.

pkg/kubernetes/accesscontrol_round_tripper.go
```diff
@@ -28,17 +67,55 @@ func (rt *AccessControlRoundTripper) RoundTrip(req *http.Request) (*http.Respons
 	// was created before restMapper was set (fixes issue #688)
 	restMapper := rt.restMapperProvider()
 	if restMapper == nil {
-		return nil, fmt.Errorf("failed to make request: AccessControlRoundTripper restMapper not initialized")
+		return nil, fmt.Errorf("failed to make request: restMapper not initialized")
```
manusa: Since this is AccessControlRoundTripper (again), this change shouldn't be necessary, the original error provides better context for debugging.

pkg/api/validation.go
```diff
@@ -0,0 +1,123 @@
...
+// NewInvalidManifestError creates an error for malformed manifests.
+func NewInvalidManifestError(reason string) *ValidationError {
+	return &ValidationError{
+		Code:    ErrorCodeInvalidManifest,
+		Message: fmt.Sprintf("Invalid resource manifest: %s", reason),
+	}
+}
+
+// FormatValidationErrors formats multiple validation errors into a single string.
+func FormatValidationErrors(errors []*ValidationError) string {
```
manusa: Unused function, can be removed.

pkg/api/validation.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return &ValidationError{
+		Code:    ErrorCodePermissionDenied,
+		Message: msg,
+	}
+}
+
+// NewInvalidManifestError creates an error for malformed manifests.
+func NewInvalidManifestError(reason string) *ValidationError {
```
manusa: Unused function, can be removed

pkg/api/validation.go
```diff
@@ -0,0 +1,123 @@
...
+// NewResourceNotFoundError creates an error for unknown resource types.
+func NewResourceNotFoundError(apiVersion, kind string) *ValidationError {
+	return &ValidationError{
+		Code:    ErrorCodeResourceNotFound,
+		Message: fmt.Sprintf("Resource %s/%s does not exist in the cluster", apiVersion, kind),
+	}
+}
+
+// NewInvalidFieldError creates an error for invalid fields.
+func NewInvalidFieldError(field, resourceKind string) *ValidationError {
```
manusa: Unused function, can be removed

pkg/api/validation.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// ValidationErrorCode categorizes validation failures.
+type ValidationErrorCode string
+
+const (
+	ErrorCodeResourceNotFound ValidationErrorCode = "RESOURCE_NOT_FOUND"
+	ErrorCodeInvalidField     ValidationErrorCode = "INVALID_FIELD"
+	ErrorCodePermissionDenied ValidationErrorCode = "PERMISSION_DENIED"
+	ErrorCodeInvalidManifest  ValidationErrorCode = "INVALID_MANIFEST"
```
manusa: Can be potentially removed after removing the unused function

pkg/api/config.go
nader-ziada: moved it

pkg/kubernetes/accesscontrol_round_tripper.go
nader-ziada: removed resource validator as suggested

nader-ziada:
@manusa thanks for the review, I addressed the comments and mostly applied your suggestions to simplify the code.

PR #773, opened 2026-02-11
refactor(http): improve user agent handling and test infrastructure

manusa:
- Fix trailing space in user agent when client info is unavailable
- Simplify UserAgentRoundTripper type assertion
- Add McpClientOption interface for extensible test client configuration
- Simplify BaseMcpSuite.InitMcpClient to use new options API
- Add test coverage for empty client info edge case


matzew (approved):
LGTM

PR #775, opened 2026-02-11
refactor(kubernetes): export k.CanIUse func

swghosh:
Certain downstream implementations ([eg.](https://github.com/openshift/openshift-mcp-server/pull/51/changes#diff-b2f101ddb4fbeac7dc05117140c00da705dbdd190d70aabf04d009d197666420R246-R277)) would want to call `k.CanIUse` in other packages outside `pkg/kubernetes`. CanIUse performs generic SubjectAccessReview calls to check if current user is authorized with certain resources, this is useful beyond the immediate package.

Also, added a fake Kube client that allows mocking the CanIUse func in other dependent packages.

xref: https://github.com/openshift/openshift-mcp-server/pull/51#discussion_r2794887577

### DIFF
--- a/pkg/kubernetes/fakeclient/sarc_fakeclient.go
+++ b/pkg/kubernetes/fakeclient/sarc_fakeclient.go
@@ -1,0 +1,198 @@
+package fakeclient
+
+import (
+	"context"
+
+	authv1 "k8s.io/api/authorization/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	fakedynamic "k8s.io/client-go/dynamic/fake"
+	"k8s.io/client-go/kubernetes"
+	authorizationv1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
+	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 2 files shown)


pkg/kubernetes/resources.go
```diff
@@ -229,7 +229,9 @@ func (c *Core) supportsGroupVersion(groupVersion string) bool {
 	return true
 }
 
-func (c *Core) canIUse(ctx context.Context, gvr *schema.GroupVersionResource, namespace, verb string) bool {
+// CanIUse checks if the current user has permission to perform a specific verb on a resource.
+// It uses SelfSubjectAccessReview to determine if the operation is allowed.
+func (c *Core) CanIUse(ctx context.Context, gvr *schema.GroupVersionResource, namespace, verb string) bool {
```
matzew, line 234: currently this file just exposes "crud" operations. 

Not sure if for a public func we might want to move it else? 
Before was private, which is different story

swghosh: any suggestions for a new package / existing package to place this func instea?

manusa:
This has a slight overlap with https://github.com/containers/kubernetes-mcp-server/pull/764

Not sure if we can extract this somewhere so that it can be reused by both, or if we can reuse what's added in #764 once we merge it (cc @nader-ziada)

swghosh:
> This has a slight overlap with https://github.com/containers/kubernetes-mcp-server/pull/764

https://github.com/containers/kubernetes-mcp-server/pull/764/changes#diff-6a980b448ac8e6b02bc47f25c026c06bb7d31298e327a7bb0e27782446ad1380R15 does similar.

/close

nader-ziada:
> /close

@swghosh  were you trying to close the PR? i don't think this command works in this repo

manusa:
I added the closes ref to the other PR. This PR will be closed automatically when we merge the other one

PR #779, opened 2026-02-12
feat(helm-charts): make Service targetPort Configurable

dkhachyan:
This PR makes the targetPort field configurable in the Helm chart's Service definition, allowing users to customize which port the Service forwards traffic to. This is particularly useful when deploying kubernetes-mcp-server with an authentication proxy sidecar container.

### DIFF
--- a/charts/kubernetes-mcp-server/values.yaml
+++ b/charts/kubernetes-mcp-server/values.yaml
@@ -118,6 +118,7 @@ service:
   type: ClusterIP
   # -- This sets the ports more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports
   port: 8080
+  # -- Target port for the service. Useful when deploying with an proxy sidecar. Set this to the sidecar's port to route traffic through the proxy before reaching the main container.
   targetPort: http
 
 # -- This block is for setting up the ingress for more information can be found here: https://kubernetes.io/docs/concepts/services-networking/ingress/


charts/kubernetes-mcp-server/values.yaml
```diff
@@ -118,6 +118,7 @@ service:
   type: ClusterIP
   # -- This sets the ports more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports
   port: 8080
+  targetPort: http
```
manusa, line 122: The port field in values.yaml has a descriptive comment (`# -- This sets the ports...`), but the new `targetPort` field lacks one. Consider adding a comment explaining the purpose.

dkhachyan: done

manusa: Thx, you need to signoff the commits to that the DCO check passes

dkhachyan: done

dkhachyan:
@manusa, are you planning to release a new version of the chart?

manusa:
> @manusa, are you planning to release a new version of the chart?

will be released along with the next version of the MCP server.

In the meantime, you should be able to use the dev version of the chart (https://github.com/containers/kubernetes-mcp-server/pkgs/container/charts%2Fkubernetes-mcp-server/681457054?tag=0.1.0-dev.a215678) and verify that everything works as expected (the dev version is exactly what the 0.1.0 version will be if we don't change anything else).

dkhachyan:
thx!

PR #735, opened 2026-02-04
feat(evals): add kubevirt tasks for vmsnapshot and vmrestore

Ruclo:
This PR adds 2 tasks into the kubevirt evals, testing the scenario of a snapshot creation and restoring from a snapshot into a vm.

PR #785, opened 2026-02-17
Kubevirt snapshot tooling

r2dedios:
Adds VM snapshot management to KubeVirt toolset (create, list, delete, restore).

### DIFF
--- a/README.md
+++ b/README.md
@@ -657,6 +657,24 @@ In case multi-cluster support is enabled (default) and you have access to multip
   - `name` (`string`) **(required)** - The name of the virtual machine
   - `namespace` (`string`) **(required)** - The namespace of the virtual machine
 
+- **vm_snapshot_create** - Create a snapshot of a VirtualMachine for backup or cloning purposes
+  - `namespace` (`string`) **(required)** - The namespace of the virtual machine
+  - `snapshot_name` (`string`) - Optional name for the snapshot. If not provided, a name will be generated
+  - `vm_name` (`string`) **(required)** - The name of the virtual machine to snapshot
+
+- **vm_snapshot_delete** - Delete a VirtualMachine snapshot
+  - `namespace` (`string`) **(required)** - The namespace of the snapshot
+  - `snapshot_name` (`string`) **(required)** - The name of the snapshot to delete
+
+- **vm_snapshot_list** - List VirtualMachine snapshots for a specific VM or all snapshots in a namespace
+  - `namespace` (`string`) **(required)** - The namespace to list snapshots from
+  - `vm_name` (`string`) - Optional name of the virtual machine to filter snapshots. If not provided, lists all snapshots in the namespace
+
+- **vm_snapshot_restore** - Restore a VirtualMachine to a previous snapshot state
+  - `namespace` (`string`) **(required)** - The namespace of the virtual machine
+  - `snapshot_name` (`string`) **(required)** - The name of the snapshot to restore from
+  - `vm_name` (`string`) **(required)** - The name of the virtual machine to restore
+
 </details>
 
 <details>
--- a/pkg/kubevirt/gvr.go
+++ b/pkg/kubevirt/gvr.go
@@ -79,6 +79,30 @@ var (
 	}
 )
 
+// Snapshot resources
+var (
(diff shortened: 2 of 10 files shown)


lyarwood:
Thanks for the PR @r2dedios, can I ask what the justification is for this additional tooling?

For context @Ruclo recently landed https://github.com/containers/kubernetes-mcp-server/pull/735 that introduced evals around creating VirtualMachineSnapshots and VirtualMachineRestores with the existing generic CRD support provided by the core toolset.

If possible could you introduce evals showing the need for any additional tooling *before* introducing the additional tooling in follow up commits / PRs.

r2dedios:
Hi @lyarwood 

My team is working on [agentic packs](https://github.com/RHEcosystemAppEng/agentic-collections) connecting multiple AI models with MCP server to increase operational capabilities. We are providing AI Agent Skills for openshift  virtualization administrators, and we thought these tools could be useful. If this PR is accepted we will define the skills for using them properly

Being honest, I didn't notice the PR #735, but if I understood it correctly, it doesn't include specialized tooling for snapshots, but it demonstrates the generic `resources_create_or_update` tool can create Snapshots CRs correctly. My original idea was to contribute with specialized tooling for this topic, like the "vm-create" tool already available. 

Do you have any policy regarding this? Use always generic tooling or create specialized tools?

lyarwood:
> Hi @lyarwood
> 
> My team is working on [agentic packs](https://github.com/RHEcosystemAppEng/agentic-collections) connecting multiple AI models with MCP server to increase operational capabilities. We are providing AI Agent Skills for openshift virtualization administrators, and we thought these tools could be useful. If this PR is accepted we will define the skills for using them properly

Interesting @r2dedios, I had looked into this upstream for KubeVirt but ended up creating a more dev focused ai-helpers set of skills.

I'd like to revisit this if there's a downstream need for such skills. We do have plans to extract the kubevirt toolset from this project and place it under the kubevirt org FWIW.

> Being honest, I didn't notice the PR #735, but if I understood it correctly, it doesn't include specialized tooling for snapshots, but it demonstrates the generic `resources_create_or_update` tool can create Snapshots CRs correctly. My original idea was to contribute with specialized tooling for this topic, like the "vm-create" tool already available.
> 
> Do you have any policy regarding this? Use always generic tooling or create specialized tools?

We (the kubevirt toolset) don't have any defined policies in place yet no but I think this project as a whole is asking folks to first demonstrate a need for specific tooling through the use of evals to avoid saturating callers with tool definitions.

@manusa @Cali0707 @matzew keep me honest here please.

r2dedios:
I thought the specific snapshots tooling could be interesting in order to have a simpler and more specialized tool for this use case. As far as I understand, is the same scenario with vm-create tool.. Specific vs Generic tools

I'm pending on your feedback, thanks!

lyarwood:
> I thought the specific snapshots tooling could be interesting in order to have a simpler and more specialized tool for this use case. As far as I understand, is the same scenario with vm-create tool.. Specific vs Generic tools

The vm-create tool itself isn't required to create a VirtualMachine object, it just encodes lots of additional best practices that we found through the use of evals were often missed without the tool even when using frontier models.

Again, we really don't want to saturate the agent and model with a huge number of tool definitions if we can help it. So for now if the generic tools pass the use cases we provide in our evals then there's no need for more specialised tooling.

r2dedios:
Ok, you're the owners, so you decide :+1:

lyarwood:
> Ok, you're the owners, so you decide 👍

I'm not formally an owner of the toolset so we should let others chime in here as my reasoning might be out of date given the introduction of the MCP proxy and code execution support etc.

manusa:
> @manusa @Cali0707 @matzew keep me honest here please.

In general, for generic resource operations we want the model to reuse the generic resources tools.
The main reasoning for this is that providing many tools will overwhelm most models and eventually defeat the purpose of specialization.
Most frontier models are capable of using the generic resources to perform most tasks.

The main way to deal with this is through evals.
If you have common prompts that you know your users will be sending to their coding agents to perform a given task, then you can introduce that as an eval task/test and see if the MCP server in the current state makes the LLM capable of achieving the expected purpose.

r2dedios:
Ok, that makes sense. I will work on the skills with the generic tooling. Thanks everyone!

PR #792, opened 2026-02-19
feat(logging): centralize K8s error logging in MCP tool handler

manusa:
Move `HandleK8sError` calls from individual tool handlers to the central `ServerToolToGoSdkTool` dispatch point in `gosdk.go`. This eliminates duplicated logging across ~20 call sites and ensures consistent error categorization for all tools automatically.

Add `errors.As` guard in `HandleK8sError` fallback to only log actual K8s API errors, silently ignoring non-K8s errors (e.g. access control denials). Extract README logging section into `docs/logging.md` for consistency with new docs organization.

Add unit tests for `HandleK8sError` covering all K8s error types, wrapped error chain traversal, and non-K8s error filtering. Add integration tests verifying non-K8s errors produce no log notifications (pods, helm) and K8s forbidden errors from Helm produce correct notifications. Add `RequireNoLogNotification` test helper.

### DIFF
--- a/pkg/mcplog/k8s.go
+++ b/pkg/mcplog/k8s.go
@@ -7,39 +7,49 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
-// HandleK8sError sends appropriate MCP log messages based on Kubernetes API error types.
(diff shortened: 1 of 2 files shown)


nader-ziada:
LGTM, great simplification

PR #795, opened 2026-02-23
fix(toolsets): sort Toolsets() output for deterministic ordering (#791)

### LINKED ISSUE #791: Toolsets() returns toolsets in non-deterministic init() order
## Problem

`Toolsets()` in `pkg/toolsets/toolsets.go` returns the registered toolsets slice in `init()` registration order, which depends on the Go package dependency graph. This means the order can change when imports are added or modified, leading to non-alphabetical ordering in the README (e.g., kiali appearing at the top instead of in its alphabetical position).

`ToolsetNames()` sorts its output, but `Toolsets()` does not. The `update-readme` tool and other consumers iterate `Toolsets()` directly.

## Fix

Sort the slice in `Toolsets()` by toolset name before returning, similar to how `ToolsetNames()` already sorts:

```go
func Toolsets() []api.Toolset {
	slices.SortFunc(toolsets, func(a, b api.Toolset) int {
		return strings.Compare(a.GetName(), b.GetName())
	})
	return toolsets
}
```

## Context

Observed after changes in #764 altered the dependency graph, causing kiali's `init()` to run before other toolsets.

manusa:
Toolsets() returned toolsets in non-deterministic init() registration order, causing the README toolset table to change when the Go dependency graph shifted. Sort by name before returning, matching ToolsetNames().

Fixes #791

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,12 +251,12 @@ The following sets of tools are available (toolsets marked with ✓ in the Defau
 
 | Toolset  | Description                                                                                                                                                          | Default |
 |----------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
-| kiali    | Most common tools for managing Kiali, check the [Kiali documentation](https://github.com/containers/kubernetes-mcp-server/blob/main/docs/KIALI.md) for more details. |         |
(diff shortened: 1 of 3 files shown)


nader-ziada:
LGTM

thanks for fixing this @manusa the README changes were annoying

PR #796, opened 2026-02-23
feat(ci): add kubevirt toolset support to mcpchecker workflow

lyarwood:
Extend the mcpchecker evaluation workflow to support testing the kubevirt toolset alongside kubernetes and kiali. Adds Kind cluster KubeVirt installation, suite selection via workflow_dispatch and PR comments (/run-mcpchecker kubevirt), and kubevirt taskSet entries in eval configs.

### DIFF
--- a/.github/workflows/mcpchecker.yaml
+++ b/.github/workflows/mcpchecker.yaml
@@ -9,16 +9,24 @@ on:
   issue_comment:
     types: [created]
 
+  # Automatically run kubevirt evals when kubevirt-related files change
+  pull_request:
+    paths:
+      - 'pkg/toolsets/kubevirt/**'
+      - 'evals/tasks/kubevirt/**'
+      - 'build/kubevirt.mk'
+
   # Allow manual workflow dispatch for testing
   workflow_dispatch:
     inputs:
       suite:
-        description: 'Which task suite to run (kubernetes, kiali, or all)'
+        description: 'Which task suite to run (kubernetes, kubevirt, kiali, or all)'
         required: false
         default: 'kubernetes'
         type: choice
         options:
           - kubernetes
+          - kubevirt
           - kiali
           - all
       task-filter:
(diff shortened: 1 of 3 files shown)


lyarwood:
Ah looks like we can't test this pre-merge?

https://github.com/containers/kubernetes-mcp-server/actions/runs/22303542468

@matzew @Cali0707 @manusa - any thoughts on landing something like this for the kubevirt toolset?

manusa:
> @matzew @Cali0707 @manusa - any thoughts on landing something like this for the kubevirt toolset?

I'm not sure if you're asking about your changes or about implementing something like what your changes do.

Ever since adding the kiali eval tests, I've been wanting to improve the evals pipeline (we've already made some progress there).
Adding additional toolsets is cumbersome and error-prone.
So IMO, anything we can do to improve the pipeline is welcome. (Not sure if this is what you're asking here)

lyarwood:
> > @matzew @Cali0707 @manusa - any thoughts on landing something like this for the kubevirt toolset?
> 
> I'm not sure if you're asking about your changes or about implementing something like what your changes do.
> 
> Ever since adding the kiali eval tests, I've been wanting to improve the evals pipeline (we've already made some progress there). Adding additional toolsets is cumbersome and error-prone. So IMO, anything we can do to improve the pipeline is welcome. (Not sure if this is what you're asking here)

@manusa Yes sorry I was asking if extending the pipeline to also cover kubevirt was something we wanted to do as I had heard about plans to move to prow?

manusa:
OK, I'm not sure about the state on that (migration to prow), but until then it makes sense to have the improved GH pipeline IMO.

Cali0707:
@lyarwood I think the plan (last that I heard of it) was for this to stay in github actions in this repo, and use prow in the openshift-mcp-server downstream fork

Cali0707 (approved):
This looks fine to me

Now that we have multiple different toolsets that may be run in combo, we should look at if there is a way (ideally in mcpchecker) to simplify this action

lyarwood:
@Cali0707 Thanks, is there anyway of testing this pre-merge? AFAICT the workflow/action doesn't run at the moment.

PR #802, opened 2026-02-23
fix(core): resources_create_or_update forces SSA

### LINKED ISSUE #799: Force Server Side Apply in `resources_create_or_update`
In our evals we are seeing failures on the `resources_create_or_update` tool due to SSA conflicts (the fields are generally managed by kubectl).

We should force these for now, and add elicitation support to confirm this eventually.

In evals we see agents working around this by deleting and then recreating the resource, which is not ideal

Cali0707:
Fixes #799


nader-ziada:
LGTM

PR #803, opened 2026-02-23
fix(core)!: resources_create_or_update removes status updates

### LINKED ISSUE #800: Disallow status updates in `resources_create_or_update`
We should block any updates to the status in the `resources_create_or_update` tool.

We should figure out whether we want to:
1. Silently drop the status update
2. Drop the status and tell the agent we dropped it (unclear how to do this while keeping everything serializable/deserializable - especially with structured outputs)
3. Return an error if status is being updated, make the model retry

IMO 1 or 2 is probably best, as 3 would reduce agent perf on smaller models that struggle more with iterative tool calling, and increase token counts on any model

Cali0707:
Fixes #800 

Note: this is technically a breaking change, but I really hope no one is currently relying on this behaviour 😅

### DIFF
--- a/pkg/mcp/resources_test.go
+++ b/pkg/mcp/resources_test.go
@@ -533,6 +533,37 @@ func (s *ResourcesSuite) TestResourcesCreateOrUpdate() {
 			s.Equalf("true", annotations["updated"], "custom resource not updated")
 		})
 	})
+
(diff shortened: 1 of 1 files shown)


matzew (approved):
LGTM

PR #804, opened 2026-02-23
feat(api): support elicitation from tool calls

### LINKED ISSUE #801: Initial support for elicitation in the toolset API
In the toolset API we should allow for the server to send elicitation requests to the user.

As far as I can tell, elicitations can happen at any point in the MCP session. To keep things simple, let's initially support this within existing prompt/tool flows. In the future if needed, we can consider how to send an elicitation request outside of those flows

Cali0707:
This PR adds a first pass at adding elicitation support, so that toolset authors can use elicitation flows within their tools.

Note that I have not used this in any tools yet, just working to land support first here.

Fixes #801

### DIFF
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -108,12 +108,30 @@ type ToolHandlerParams struct {
 
 type ToolHandlerFunc func(params ToolHandlerParams) (*ToolCallResult, error)
 
+// Elicitor provides a mechanism for tools and prompts to request additional information
+// from the user during execution via the MCP elicitation protocol.
+// The elicitation request is forwarded to the MCP client, which presents a form to the user.
+// See MCP specification: https://modelcontextprotocol.io/specification/2025-11-25/client/elicitation
 type Elicitor interface {
 	Elicit(ctx context.Context, message string, requestedSchema *jsonschema.Schema) (*ElicitResult, error)
 }
 
+// ElicitAction constants define the possible user responses to an elicitation request.
+// See MCP specification: https://modelcontextprotocol.io/specification/2025-11-25/client/elicitation
+const (
+	// ElicitActionAccept indicates the user submitted the form with content.
+	ElicitActionAccept = "accept"
+	// ElicitActionDecline indicates the user explicitly declined the request.
+	ElicitActionDecline = "decline"
+	// ElicitActionCancel indicates the user dismissed the form without making a choice.
+	ElicitActionCancel = "cancel"
+)
+
+// ElicitResult represents the user's response to an elicitation request.
 type ElicitResult struct {
-	Action  string
+	// Action is one of ElicitActionAccept, ElicitActionDecline, or ElicitActionCancel.
+	Action string
+	// Content contains the submitted form data. Only populated when Action is ElicitActionAccept.
 	Content map[string]any
 }
 
--- a/pkg/mcp/elicit.go
+++ b/pkg/mcp/elicit.go
@@ -2,13 +2,19 @@ package mcp
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"github.com/containers/kubernetes-mcp-server/pkg/mcplog"
 	"github.com/google/jsonschema-go/jsonschema"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
+
+// ErrElicitationNotSupported is returned when the MCP client does not support elicitation.
+// Tool authors can check for this error using errors.Is() to implement fallback behavior.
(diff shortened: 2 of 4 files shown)


pkg/api/toolsets.go
```diff
@@ -103,10 +103,20 @@ type ToolHandlerParams struct {
...
+	Elicitor
 }
 
 type ToolHandlerFunc func(params ToolHandlerParams) (*ToolCallResult, error)
 
+type Elicitor interface {
+	Elicit(ctx context.Context, message string, requestedSchema *jsonschema.Schema) (*ElicitResult, error)
+}
+
+type ElicitResult struct {
```
matzew, line 131: should we document this a bit? Like others, e.g. `Tool` ?

pkg/mcp/elicit.go
```diff
@@ -0,0 +1,30 @@
...
+var _ api.Elicitor = &sessionElicitor{}
+
+func (s *sessionElicitor) Elicit(ctx context.Context, message string, requestedSchema *jsonschema.Schema) (*api.ElicitResult, error) {
+	session, ok := ctx.Value(mcplog.MCPSessionContextKey).(*mcp.ServerSession)
+	if !ok || session == nil {
+		return nil, fmt.Errorf("no MCP session found in context")
+	}
+
+	result, err := session.Elicit(ctx, &mcp.ElicitParams{Message: message, RequestedSchema: requestedSchema})
+	// TODO: check if the error is because the client does not support elicitation, possibly return default value
```
manusa: The TODO should be more actionable for tool authors. When `session.Elicit()` fails because the client doesn't support elicitation, the go-sdk returns a JSON-RPC error with code `-32602` and message `"client does not support elicitation"`. Tool authors need to know how to distinguish this from other errors so they can implement fallback behavior.

Consider either:
- Documenting the expected error type/code here so callers know what to check
- Wrapping it in a sentinel error (e.g., `ErrElicitationNotSupported`) that callers can use with `errors.Is()`

This becomes important as soon as actual tools start using elicitation.

pkg/api/toolsets.go
```diff
@@ -103,10 +103,20 @@ type ToolHandlerParams struct {
...
+	Action  string
+	Content map[string]any
```
manusa, line 135: `ElicitResult.Action` should document the valid values. Per the MCP spec, the action is one of `"accept"`, `"decline"`, or `"cancel"`. Without documentation or constants, every consumer of this interface must know the magic strings.

Consider adding a comment here, or defining constants:
```go
const (
    ElicitActionAccept  = "accept"
    ElicitActionDecline = "decline"
    ElicitActionCancel  = "cancel"
)
```

manusa (changes requested):
## Pull Request Review: #804

### Summary

Adds foundational MCP elicitation support: an `Elicitor` interface in `pkg/api/`, a `sessionElicitor` implementation in `pkg/mcp/elicit.go` that delegates to the go-sdk `ServerSession.Elicit()`, and wiring into both `ToolHandlerParams` and `PromptHandlerParams`. No tools use elicitation yet — this is plumbing only.

The design is clean and follows the existing patterns in the codebase (embedding interfaces in params structs, concrete implementations in `pkg/mcp/`).

### Review Verdict

REQUEST_CHANGES

### Findings

#### Critical Issues (Must Fix)

1. **Rebase on main required** — The latest main (`66106e7c`) switched the test infrastructure from `mark3labs/mcp-go` to the official `modelcontextprotocol/go-sdk` client. The go-sdk client supports `ElicitationHandler` in `ClientOptions`, which enables real end-to-end elicitation integration tests. Please rebase on main.

2. **Missing tests** — After rebasing, the go-sdk test client can be extended with a `WithElicitationHandler(...)` option (following the existing `WithHTTPHeaders`, `WithClientInfo` patterns in `internal/test/mcp.go`). This enables behavior-based integration tests that exercise the actual elicitation protocol flow:

   - **`TestElicitationAccepted`** — Client accepts with content, tool receives `action=accept` and provided data
   - **`TestElicitationDeclined`** — Client declines, tool receives `action=decline`
   - **`TestElicitationWithUnsupportedClient`** — Client has no handler, tool gets an error from `Elicitor.Elicit()` and handles it gracefully (no crash, returns fallback)

   These test the observable end-to-end behavior (server → client → response → server) with no mocks of the elicitor itself.

#### Suggestions (Should Consider)

See inline comments for:
- `ElicitResult.Action` should document valid values or define constants
- The TODO in `elicit.go` should be more actionable for tool authors

### Security Concerns

None. The elicitation mechanism delegates to the MCP SDK's session handling, and the `jsonschema.Schema` parameter constrains client responses.

### Overall Assessment

The interface design and wiring are solid. The main gap is the rebase needed to align with the current test infrastructure and enable proper integration tests.

PR #805, opened 2026-02-24
test(core): add test for resources_create_or_update SSA force

manusa:
## Summary
- Adds an MCP-layer test for `resources_create_or_update` to verify that Server-Side Apply uses `Force: true`
- The test creates a ConfigMap with a different field manager, then confirms `resources_create_or_update` can update fields owned by that other manager without conflict errors

Follows up on #802 (fix) and #799 (issue).

## Test plan
- [x] Test passes with `Force: true` in place
- [x] Test fails without `Force: true` with `Apply failed with 1 conflict: conflict with "other-manager": .data.key`

### DIFF
--- a/pkg/mcp/resources_test.go
+++ b/pkg/mcp/resources_test.go
@@ -535,6 +535,48 @@ func (s *ResourcesSuite) TestResourcesCreateOrUpdate() {
 	})
 }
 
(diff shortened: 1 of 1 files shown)


matzew (approved):
LGTM

PR #809, opened 2026-02-24
fix(test): On test suites always assume a certain default configuration to be executed

matzew:
Following up on https://github.com/containers/kubernetes-mcp-server/pull/738

For now test suites should set configuration for toolsets' intended behavior (or default configuration). Just that downstream repos can disable some options (or toolsets) does not mean those should not be tested. Users can always re-enable any config.

As discussed in the other PR a larger refactor for a more modular approach is the ultimate goal

### DIFF
--- a/pkg/mcp/common_test.go
+++ b/pkg/mcp/common_test.go
@@ -192,6 +192,9 @@ type BaseMcpSuite struct {
 
 func (s *BaseMcpSuite) SetupTest() {
 	s.Cfg = config.Default()
+	// Reset to upstream defaults for testing (ignore downstream build-time overrides)
+	s.Cfg.ReadOnly = false
+	s.Cfg.Toolsets = []string{"core", "config", "helm"}
 	s.Cfg.ListOutput = "yaml"
 	s.Cfg.KubeConfig = filepath.Join(s.T().TempDir(), "config")
 	s.Require().NoError(os.WriteFile(s.Cfg.KubeConfig, envTest.KubeConfig, 0600), "Expected to write kubeconfig")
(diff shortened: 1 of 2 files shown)


pkg/mcp/common_test.go
```diff
@@ -192,6 +192,9 @@ type BaseMcpSuite struct {
 
 func (s *BaseMcpSuite) SetupTest() {
 	s.Cfg = config.Default()
+	// Reset to upstream defaults for testing (ignore downstream build-time overrides)
+	s.Cfg.ReadOnly = false
+	s.Cfg.Toolsets = []string{"core", "config", "helm"}
```
Cali0707, line 197: IMO we should think long term about if there is a better fix, since I would prefer for the tests to pick up on the "Defaults" we define in this repo, rather than syncing that between multiple places.

Maybe we can make the `config.Default()` method different from something like `config.DefaultNoOverrides()` ?

matzew: yeah, some more modular approach, less coupled

manusa: We need something different to this, having a hardcoded list here will eventually diverge from whatever the server defaults to.

matzew: yeah - I run into this now on a downstream repo... I will spend some time on finding something here

nader-ziada: Maybe we can do something like a BaseDefaults:
```
func baseDefaults() *StaticConfig {
    return &StaticConfig{
        ReadOnly: false,
        Toolsets: []string{"core", "config", "helm"},
    }
}
```
 - Defaults are defined in one place (no divergence risk)
 - BaseDefault() returns upstream defaults 
 - Downstream can override by setting exported variables in their fork's init() when needed

matzew: yeah, looking into this - along the lines of what has been also discussed on the other PR

matzew:
/close
replaced by https://github.com/containers/kubernetes-mcp-server/pull/817

Issue #722, opened 2026-02-02
[Feature Request] Add Efficient Resource Listing Tools to Prevent Response Truncation


ajaysingh-sc30:
## Executive Summary

The current `resources_list` tool returns complete YAML manifests for all matching resources, causing response truncation when listing resources in production environments. This feature request proposes seven new tools and enhancements to enable efficient resource discovery without truncation.

## Proposed Tools Summary

| Tool Name | Problem Solved | Priority | Response Size |
|-----------|---------------|----------|---------------|
| `resources_list_summary` | Returns compact metadata instead of full YAML specs | ⭐ Critical | 90% reduction |
| `resources_count` | Quick resource counting without data transfer | ⭐ Critical | <1KB |
| `resources_list` (enhanced) | Add `outputFormat` parameter for flexible output | ⭐ Critical | Configurable |
| `resources_labels_list` | Discover available labels for efficient filtering | Medium | ~5KB |
| `resources_list` (pagination) | Handle large result sets with continuation tokens | Medium | Configurable |
| `resources_get_batch` | Fetch multiple specific resources efficiently | Low | 50% reduction |
| `resources_list` (field projection) | Extract specific fields using JSONPath | Low | 70% reduction |

## Problem Statement

Current `resources_list` tool limitations in production environments:

- **Response Truncation**: Exceeds ~60KB limit when listing 50+ resources (e.g., deployments)
- **Context Loss**: AI agents cannot see resources beyond truncation point
- **Token Inefficiency**: Full YAML specifications consume excessive tokens for simple queries
- **Limited Discovery**: No efficient method to enumerate resources before fetching details

### Impact Example

```bash
# Query: List all deployments in namespace
resources_list(apiVersion="apps/v1", kind="Deployment", namespace="production")

# Current behavior:
- Returns: 50+ complete YAML manifests (~100KB total)
- Result: Response truncated at ~60KB (only 23/50 deployments visible)
- Missing: auth-service, api-gateway, data-processor, and 27 other deployments

# Desired behavior:
resources_list_summary(apiVersion="apps/v1", kind="Deployment", namespace="production")
- Returns: Compact metadata for all 50 deployments (~10KB total)
- Result: All resources visible, no truncation
```

### Tool 1: `resources_list_summary`

**Purpose**: Return essential resource metadata without full YAML specifications.

**Problem Solved**: Eliminates truncation by reducing response size by 90%.

**Tool Definition:**
```yaml
name: resources_list_summary
[...]

Issue #608, opened 2025-12-22
n8n: The model returned the following errors: Input is too long for requested model.


marcofranssen:
I'm continuously hitting following problem when trying to use this MCP in combination with n8n and Claude Sonnet 4.5 or basically any other model.

## Problem

```
The model returned the following errors: Input is too long for requested model.
```

I think this is mainly caused by the large textual output of the MCP tools. It would probably be great if the tools offered in this MCP can have more parameters where filters and output formats can be controlled to have more targeted results.

## Steps to reproduce

1. Configure a AI agent in a n8n workflow with following details.

   **System message**

   ```
   You are a k8s expert and you will use the tool K8s MCP to get information from the cluster. You will collect data in the most compact form possible to keep the prompt size as small as possible.
   ```

   **Prompt (user message)**

   ```
   Give me an overview of failing pods. For each failing pod you will collect conditions, events, logs to determine the failure.
   ```

2. Add Claude Sonnet 4.5 as a model (or basically any other AI will face same problem).
3. Add simple memory to the AI agent can remember context
4. Run the workflow

As long there are no failing pods the workflow will succeed, but as soon there are failing pods, the context will grow out of bounds.

barth12:
It is exactly the same what we are observing. MCP output is huge.

manusa:
We've recently added more filtering options to the MCP server (#641 / #621)

In addition, I've also created this PoC to allow models to evaluate code https://github.com/containers/kubernetes-mcp-server/pull/711 (you can check the linked demo and spec for more information). If you find this interesting, please comment in the pull request, it'd be great if you could try it and provide feedback (especially how it behaves with your models and clusters)

manusa:
Closing — the underlying concerns are tracked in more specific issues:
- #824 — list table output hardcoded to wide/show-labels
- #877 — output_query parameter for reducing response size
- #876 — statusOnly parameter for resources_get

PR #813, opened 2026-02-25
feat(core): add output format parameter to resources_list

gregtuc:
Gives the resources_list tool an optional output parameter with two modes: table (which is compact like results from kubectl) and yaml (full resource manifests). Takes precedence over the server-wide list_output config, falling back to it if not provided.

Previously, output format was only configurable server-wide via the list_output config setting. MCP clients had no way to control verbosity per request. Listing nodes or CRDs in yaml mode can easily return tens of thousands of tokens, whereas table mode returns a few dozen lines.

### DIFF
--- a/pkg/mcp/resources_test.go
+++ b/pkg/mcp/resources_test.go
@@ -301,6 +301,50 @@ func (s *ResourcesSuite) TestResourcesListAsTable() {
 	})
 }
 
+func (s *ResourcesSuite) TestResourcesListOutputParameter() {
+	// Server default is yaml (set in BaseMcpSuite.SetupTest)
+	s.InitMcpClient()
+	s.Run("output=table overrides server yaml config", func() {
+		result, err := s.CallTool("resources_list", map[string]interface{}{
+			"apiVersion": "v1",
+			"kind":       "Namespace",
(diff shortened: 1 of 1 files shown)


manusa:
The default output is table, unless configured otherwise.

I'm not sure why is this needed in a per function-call basis. If the LLM need the entire YAML resource it can fetch it by doing a subsequent request to the resources_get tool.

Can you provide more details or elaborate on why this might be useful. Adding more parameters to the tools make them more complicated for some LLMs and consume more tokens from the context window.

gregtuc:
Sorry @manusa, reading more deeply I was confused about what was causing the large response my MCP keeps getting compared to what the equivalent calls get with kubectl. Just as an example, on one of our clusters running `kubectl get nodes` returns 4.5k characters, whereas the equivalent call with the MCP returns 88k characters (despite using the default table mode) - uses up a lot of context window during sessions.

The actual issue I'm hitting is that the table printer in `pkg/output/output.go` hardcodes `Wide: true` and `ShowLabels: true`. Different issue, and I see there's already three issues (#608, #688, #722) about it. I'll close out this PR, thanks for the review.

PR #817, opened 2026-02-26
refactor(config): extract BaseDefault() and decouple tests from downstream overrides

matzew:
Introduce `BaseDefault` returning upstream-only defaults, separate from `Default` which merges downstream build-time overrides on top. 

Remove `HasDefaultOverrides` and all skip guards that relied on it.

Replace hardcoded default values in test assertions with `config.Default()` and use `BaseDefault()` in `BaseMcpSuite` so feature tests (Helm, Core, etc.) run regardless of downstream overrides. ToolsetsSuite retains `Default()` to verify effective default behavior.

Downstream only needs to regenerate snapshots (`UPDATE_TOOLSETS_JSON=1`) when overrides change.

### DIFF
--- a/pkg/config/config_default.go
+++ b/pkg/config/config_default.go
@@ -6,26 +6,24 @@ import (
 	"github.com/BurntSushi/toml"
 )
 
-func Default() *StaticConfig {
-	defaultConfig := StaticConfig{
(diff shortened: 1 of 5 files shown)


nader-ziada:
LGTM

PR #818, opened 2026-02-26
fix(test): improve test accuracy and naming after BaseDefault refactor

manusa:
Follows up on #817 to address some minor nitpicks

Add missing LogLevel assertion to TestBaseDefaultValues, fix stale assertion message in TestReadConfigValidPreservesDefaultsForMissingFields, and rename dynamic default-checking subtests to "matches default config" since they no longer assert hardcoded values.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -53,6 +53,9 @@ func (s *ConfigSuite) TestBaseDefaultValues() {
 	s.Run("Stateless is false", func() {
 		s.False(base.Stateless)
 	})
+	s.Run("LogLevel is 0", func() {
+		s.Equal(0, base.LogLevel)
+	})
 }
 
 func (s *ConfigSuite) TestReadConfigMissingFile() {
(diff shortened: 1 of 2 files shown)


pkg/config/config_test.go
```diff
@@ -262,7 +265,7 @@ func (s *ConfigSuite) TestReadConfigValidPreservesDefaultsForMissingFields() {
 		s.Equalf(0, config.LogLevel, "Expected LogLevel to be 0, got %d", config.LogLevel)
 	})
 	s.Run("port parsed correctly", func() {
-		s.Equalf("1337", config.Port, "Expected Port to be 9999, got %s", config.Port)
+		s.Equalf("1337", config.Port, "Expected Port to be 1337, got %s", config.Port)
```
matzew, line 268: 😅

manusa: that was a previous mistake I might have introduced TBH 😅 I realized when I reviewed your changes

PR #825, opened 2026-02-26
fix(test): TestReloadUpdatesToolsets was reloading a different server than the one under test

matzew:
The `TestReloadUpdatesToolsets` accidentally created two independent server instances. The reload was applied to one, while the MCP client assertions ran against the other. The test passed coincidentally because the toolset being added was already present in the default configuration.

Remove the redundant server setup and reload `s.mcpServer` directly — the server `InitMcpClient()` creates and the MCP client is connected to.

I noticed this when looking at when removing `helm` from default cfg

### DIFF
--- a/pkg/mcp/mcp_reload_test.go
+++ b/pkg/mcp/mcp_reload_test.go
@@ -180,14 +180,6 @@ func (s *ConfigReloadSuite) TestMultipleReloads() {
 }
 
 func (s *ConfigReloadSuite) TestReloadUpdatesToolsets() {
-	provider, err := kubernetes.NewProvider(s.Cfg)
-	s.Require().NoError(err)
-	server, err := NewServer(Configuration{
-		StaticConfig: s.Cfg,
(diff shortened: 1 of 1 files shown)


manusa:
Good catch!

manusa:
I asked Claude Code to analyze the rest of the codebase for tests with the same dual-server flaw — creating a Server via `NewServer()` while also calling `InitMcpClient()`, which spins up a second one. Here's the summary:

- `mcp_reload_test.go`: the other 4 tests calling `NewServer()` only check internal state, never use MCP client methods — safe
- `mcp_test.go` (`UserAgentPropagationSuite`): two tests create servers but assign directly to `s.mcpServer` — clean
- `toolsets_test.go`: custom `InitMcpClient()` wraps `NewServer()` internally — clean
- All remaining 18 test files: use `InitMcpClient()` exclusively — no risk

This PR was the only instance of this bug in the codebase.

PR #826, opened 2026-02-26
fix(config): remove helm toolset from default enabled toolsets

### LINKED ISSUE #819: [Breaking Change]: Do not enable helm by default
As the number of toolsets grows, we should be more picky about what is enabled by default. IMO not all users of this server will benefit from the helm toolset, so we can disable it by default.

Since it is easy to enable, this should not be hard to turn back on for users who do use it, while also helping non-helm users by having less tokens from the MCP server and less chances of the agent doing the wrong thing

matzew:
Fixes #819 

Helm is an optional tool that not every user needs. Removing it from the default set reduces the initial context size and avoids exposing destructive operations (install, uninstall) unless explicitly opted in.


matzew:
/hold
Should be merged/rebased after https://github.com/containers/kubernetes-mcp-server/pull/825 lands

matzew:
@nader-ziada @Cali0707 is now ready and rebased

manusa:
As discussed internally, please don't merge until we cut a new release later today.

PR #827, opened 2026-02-26
fix(ci): remove pr target from mcpchecker action as secrets won't be included properly

Cali0707:
As a security measure, github does not include secrets from the base repo into PRs triggered from a fork.

In general this is probably a good idea and not something we want to work around, in particular as the secrets here are API credentials and can run up a large bill.

This PR removes the automatic run of the evals on a PR, meaning we will have to manually trigger it

### DIFF
--- a/.github/workflows/mcpchecker.yaml
+++ b/.github/workflows/mcpchecker.yaml
@@ -9,13 +9,6 @@ on:
   issue_comment:
     types: [created]
 
-  # Automatically run kubevirt evals when kubevirt-related files change
-  pull_request:
-    paths:
-      - 'pkg/toolsets/kubevirt/**'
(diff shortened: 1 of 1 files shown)


Cali0707:
FYI @lyarwood I think this is why you weren't seeing the eval results you expected to see...

manusa:
PR branch workflow runs should never rely on repo secrets (there's a workaround for this, but it SHOULD NEVER be used).

As a follow up, I think that what we need to do is improve the workflow_dispatch so that  it runs the action from the main branch but executes the evals (and evaluated code) from the PR branch (an extra-level of carefulness should be required for those authorized to trigger the workflow in order to ensure that no secrets would be leaked by a maliciously crafted PR)

PR #828, opened 2026-02-26
fix(ci): improve mcpchecker workflow reliability

nader-ziada:
the cp of the results file had an incorrect file name

### DIFF
--- a/.github/workflows/mcpchecker.yaml
+++ b/.github/workflows/mcpchecker.yaml
@@ -260,7 +260,7 @@ jobs:
 
       - name: Upload PR context
         if: always() && needs.check-trigger.outputs.is-pr == 'true'
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: eval-context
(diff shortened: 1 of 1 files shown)


.github/workflows/mcpchecker.yaml
```diff
@@ -283,7 +283,7 @@ jobs:
           ref: main
 
       - name: Download mcpchecker results
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v6
```
Cali0707: @nader-ziada is there a reason for downgrading this?

nader-ziada: to match the upload one `uses: actions/upload-artifact@v6`

should i upgrade both?

Cali0707: yeah let's upgrade both 😄

nader-ziada: done

Cali0707 (approved):
LGTM, thanks @nader-ziada !

PR #835, opened 2026-02-27
fix(evals): add missing suite labels to kubevirt tasks

lyarwood:
/cc @ksimon1 
/cc @Cali0707 

5 kubevirt eval tasks (clone-vm, create-vm-with-vlan, snapshot-vm, restore-vm, troubleshoot-vm) were missing the suite: kubevirt label, causing them to be silently skipped when running the mcpchecker workflow with --label-selector suite=kubevirt.

### DIFF
--- a/evals/tasks/kubevirt/clone-vm/task.yaml
+++ b/evals/tasks/kubevirt/clone-vm/task.yaml
@@ -1,6 +1,9 @@
 kind: Task
 apiVersion: mcpchecker/v1alpha2
 metadata:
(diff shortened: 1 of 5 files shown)


Cali0707:
Looks like the evals are running again, thanks for fixing @lyarwood !

lyarwood:
> Looks like the evals are running again, thanks for fixing @lyarwood !

Yeah not very well unfortunately, I'll follow up with some fixes.

PR #836, opened 2026-02-27
refactor: expose cluster provider strategy to toolsets

Cali0707:
When we initially made the cluster provider strategy, the thinking was that everything could be handled transparently from the toolsets. While this has been true for the most part, as some toolsets are now adding functionality that calls to non kube API server APIs (e.g. a prometheus instance in the cluster), this somewhat breaks down.

In these scenarios, there is sometimes a need for the toolset to know what the strategy was so that it can determine if it needs/should use extra config (e.g. TLS config if not in the cluster).

This attempts to expose that config to toolsets where they can optionally consume it as needed

### DIFF
--- a/internal/tools/update-readme/main.go
+++ b/internal/tools/update-readme/main.go
@@ -21,13 +21,17 @@ import (
 	_ "github.com/containers/kubernetes-mcp-server/pkg/toolsets/kubevirt"
 )
 
-type OpenShift struct{}
+type ToolsetEnv struct{}
 
-func (o *OpenShift) IsOpenShift(_ context.Context) bool {
+func (o *ToolsetEnv) IsOpenShift(_ context.Context) bool {
 	return true
 }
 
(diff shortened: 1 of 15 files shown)


manusa:
Can we hold on this one until I send a proposed alternative via Middleware, that OpenShift struct was something I wanted to get rid of, so making it bigger won't help 😅

Cali0707:
@manusa for sure, happy to look into different ways of approaching this

Cali0707:
Marking as draft so it doesn't get accidentally merged early

iNecas:
What's the status of this PR? If I understand it correctly, this blocks https://github.com/openshift/openshift-mcp-server/pull/124

manusa:
I believe this was superseded by #916. Should we close it?

manusa:
Closing as I think this one is no longer necessary #916

Issue #838, opened 2026-02-27
KubeVirt eval failures


lyarwood:
/cc @ksimon1 
/cc @codingben 
/cc @Ruclo 

While these evals are now running we are seeing failures that need to be addressed:

https://github.com/containers/kubernetes-mcp-server/pull/835#issuecomment-3973698738

```
$ mcpchecker summary ~/Downloads/mcpchecker-results/mcpchecker-openai-agent-kubernetes-test-out.json
=== Evaluation Summary ===

  ✓ clone-vm (assertions: 3/3)
  ✓ create-basic-vm (assertions: 3/3)
  ✓ create-ubuntu-vm (assertions: 3/3)
  ✓ create-vm-with-instancetype (assertions: 3/3)
  ✓ create-vm-with-size (assertions: 3/3)
  ✓ create-vm-with-vlan (assertions: 3/3)
  ✗ delete-vm
      failed to setup task: setup[0] failed: script execution failed: exit status 1
output: namespace/vm-test created
virtualmachine.kubevirt.io/deleted-vm created
virtualmachine.kubevirt.io/deleted-vm condition met
error: timed out waiting for the condition on virtualmachines/deleted-vm

  ✗ pause-vm (assertions: 3/3)
      verification failed: verify[0] failed: script execution failed: exit status 1
output: error: timed out waiting for the condition on virtualmachines/paused-vm

  ✗ restore-vm (assertions: 1/3)
      verification failed: verify[0] failed: script execution failed: exit status 1
output: ERROR: VirtualMachineRestore 'test-restore' not found

      - ToolsUsed: Required tool not called: server=kubernetes, tool=, pattern=.*
      - MinToolCalls: Too few tool calls: expected >= 1, got 0
  ✗ snapshot-vm (assertions: 1/3)
      verification failed: verify[0] failed: script execution failed: exit status 1
output: ERROR: VirtualMachineSnapshot 'test-snapshot' not found

      - ToolsUsed: Required tool not called: server=kubernetes, tool=, pattern=.*
      - MinToolCalls: Too few tool calls: expected >= 1, got 0
  ✗ troubleshoot-vm (assertions: 3/3)
      verification failed: verify[0] failed: script execution failed: exit status 1
output: === Verification: Checking if agent fixed the VM ===
✓ VirtualMachine broken-vm exists
✗ Secret vm-cloud-init was not created - agent did not fix the issue

  ✗ update-vm-resources (assertions: 1/3)
[...]

PR #841, opened 2026-02-27
fix(ci): include default toolsets when enabling extra toolsets in mcpchecker

lyarwood:
The TOOLSETS env var was only passing the extra toolsets (kubevirt,
kiali) without including the default toolsets (core, config), causing
the MCP server to run without core tools during evaluations.

Fixes: https://github.com/containers/kubernetes-mcp-server/issues/838


lyarwood:
/cc @Cali0707 
/cc @manusa 

Would you mind running the kubevirt evals against this?

ksimon1:
/close
replaced by #959

PR #843, opened 2026-02-27
refactor(kubernetes)!: provider.GetTargets is optionally user scoped now

Cali0707:
To use the user scoping, provider authors can use the context supplied into the call, and it will have the user auth info.

Note: this is a first pass at trying to handle this in a way where each provider can choose the semantics of how they want to do the auth checks. Open to other approaches

I dropped the enum for cluster names from the tool defs as in the past we have not wanted to create any user specific versions of the MCP server (require since tools are registered globally for each instance), and we are no longer guaranteeing that the list is the same for every user

Actual implementations for the providers to use this should be follow ups as I am not sure what semantics make sense for each provider

### DIFF
--- a/pkg/mcp/tool_mutator.go
+++ b/pkg/mcp/tool_mutator.go
@@ -28,7 +28,7 @@ const TargetsListToolName = "targets_list"
 // WithTargetParameter adds a target selection parameter to the tool's input schema if the tool is cluster-aware
 func WithTargetParameter(defaultCluster, targetParameterName string, isMultiCluster bool) ToolMutator {
 	return func(tool api.ServerTool) api.ServerTool {
-		if !tool.IsClusterAware() || !isMultiCluster {
+		if !tool.IsClusterAware() {
 			return tool
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/mcp/tool_mutator_test.go
```diff
@@ -66,67 +69,71 @@ func createTestToolWithExistingProperties(name string) api.ServerTool {
...
+func (m *mockProvider) IsMultiCluster() bool                                            { return len(m.targets) > 1 }
+func (m *mockProvider) GetTargets(_ context.Context) ([]string, error)                  { return m.targets, m.err }
+func (m *mockProvider) GetDerivedKubernetes(_ context.Context, _ string) (*internalk8s.Kubernetes, error) {
+	return nil, nil
+}
+func (m *mockProvider) GetDefaultTarget() string            { return "" }
+func (m *mockProvider) GetTargetParameterName() string      { return "" }
+func (m *mockProvider) WatchTargets(_ internalk8s.McpReload) {}
+func (m *mockProvider) Close()                              {}
+
```
manusa: To avoid having to create this struct, it might be easier to have a minimal interface near `func WithTargetListTool` definition similar to:

```
type targetLister interface {
  GetTargets(ctx context.Context) ([]string, error)
}
```

This would also reduce the coupling with the kubernetes package

pkg/kubernetes/provider.go
```diff
@@ -19,6 +19,7 @@ type Provider interface {
 	// For the kubecontext case, a user might be targeting both an OpenShift flavored cluster and a vanilla Kubernetes cluster.
 	// See: https://github.com/containers/kubernetes-mcp-server/pull/372#discussion_r2421592315
 	api.Openshift
+	IsMultiCluster() bool
```
manusa: nit: since the rest of the interface defines `target`, maybe we should name this `IsMultiTarget` instead

Cali0707:
Hey @manusa sorry this one took me so long to get back to - should have your comments addressed now

PR #844, opened 2026-02-27
fix(mcp): init request w/o client info does not panic

### LINKED ISSUE #842: 0.0.58 panic (SIGSEGV) in MCP middleware during initialize/tools flow
Hi maintainers,

Homebrew core bump PR:
- https://github.com/Homebrew/homebrew-core/pull/269781

CI failure log (Linux arm64):
- https://github.com/Homebrew/homebrew-core/actions/runs/22496536198/job/65172938539?pr=269781

The build/install steps pass, but the runtime test crashes with:
- `panic: runtime error: invalid memory address or nil pointer dereference`
- `SIGSEGV`
- stack includes `pkg/mcp/middleware.go` in `getMcpReqUserAgent` / middleware chain.

Observed around MCP request handling after initialize/tools flow (from Homebrew test harness), with a command like:

```bash
(echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26"}}'
echo '{"jsonrpc":"2.0","id":2,"method":"tools/list"}'
sleep 1) | kubernetes-mcp-server --kubeconfig ./kubeconfig 2>&1
```

Then panic occurs before clean completion.

Could you check whether this is a known regression in `0.0.58`?

Thanks.

Cali0707:
Fixes #842 

It looks like when init request were sent without client info our user agent middleware (added in #760 ) would panic trying to access the client info.

IMO this case is rare since the spec seems to [mandate that client info is sent](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#initialization), but putting a nil check here is probably fine.

Not sure if we should return something other than "" in this case though

### DIFF
--- a/pkg/mcp/middleware.go
+++ b/pkg/mcp/middleware.go
@@ -204,7 +204,7 @@ func getMcpReqUserAgent(req mcp.Request) string {
 		return ""
 	}
 	initParams := session.InitializeParams()
-	if initParams == nil || (initParams.ClientInfo.Name == "" && initParams.ClientInfo.Version == "") {
+	if initParams == nil || initParams.ClientInfo == nil || (initParams.ClientInfo.Name == "" && initParams.ClientInfo.Version == "") {
 		return ""
 	}
 	return fmt.Sprintf("%s/%s", initParams.ClientInfo.Name, initParams.ClientInfo.Version)


Cali0707:
@chenrui333 does this fix the panic you were seeing?

manusa:
Thx for fixing this so quickly 🙌 

> IMO this case is rare since the spec seems to [mandate that client info is sent](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#initialization), but putting a nil check here is probably fine.
> 
> Not sure if we should return something other than "" in this case though

Most MCP Hosts -> Clients should take care of this automatically.
I've even tried to add a test case for this scenario with go-sdk and seemed quite an impossible task to do.

Since [v0.0.58](https://github.com/containers/kubernetes-mcp-server/releases/tag/v0.0.58) was just released, we'll definitely see other reports if this requires further attention.

PR #847, opened 2026-02-27
Fix tool schema compatibility with OpenAI models

edcdavid:
OpenAI models require the "properties" field to be present in the JSON schema for tool input, even when a tool takes no parameters. Add an empty properties map to the configuration_contexts_list and targets_list tool schemas to ensure compatibility.

Assisted by: Cursor

### DIFF
--- a/pkg/toolsets/config/configuration.go
+++ b/pkg/toolsets/config/configuration.go
@@ -18,7 +18,8 @@ func initConfiguration() []api.ServerTool {
 				Name:        "configuration_contexts_list",
 				Description: "List all available context names and associated server urls from the kubeconfig file",
 				InputSchema: &jsonschema.Schema{
-					Type: "object",
+					Type:       "object",
+					Properties: make(map[string]*jsonschema.Schema),
 				},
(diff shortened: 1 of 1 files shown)


manusa (commented):
Hi @edcdavid, thx for getting involved.
This is not a proper fix.
The toolset definition gets transformed in a latter stage at the MCP domain layer, so whatever you change here, doesn't mean that it will be sent to the MCP client and host.

This issue (#717) was already fixed, first in this repository, and later in the go-sdk directly.

We have specific tests to ensure that this is the case:

https://github.com/containers/kubernetes-mcp-server/blob/b371f492bb57389617cae6e53d423be5ce821257/pkg/mcp/toolsets_test.go#L195-L225

Are you using the latest version of the server.

If you are, and are still facing the issue, try to provide a reproducer first.

edcdavid:
Sorry for the delay, and thanks for considering the PR. I was able to retest this using the latest main and seems to work well. I used a branch that had not merged the fix yet. Closing the PR.

PR #848, opened 2026-02-28
test(mcp): add regression test for nil ClientInfo panic

manusa:
Adds a regression test for the nil `ClientInfo` panic fixed in #844 (originally reported in #842).

The existing `TestFallsBackToServerPrefixWhenNoClientInfo` test uses `WithEmptyClientInfo()` which sets `ClientInfo` to an empty (but non-nil) `*Implementation{}`. This doesn't cover the actual panic scenario where `ClientInfo` is `nil` — which happens when a client omits `clientInfo` from the initialize request entirely.

Since the go-sdk client [always sets `clientInfo`](https://github.com/modelcontextprotocol/go-sdk/blob/v1.3.0/mcp/client.go#L241) and [panics if the implementation is nil](https://github.com/modelcontextprotocol/go-sdk/blob/v1.3.0/mcp/client.go#L46-L48), we can't reproduce this through the standard test client. Instead, the new test sends raw HTTP JSON-RPC requests with an initialize payload that omits `clientInfo`, simulating a non-compliant client. This approach is inspired by the [`CallToolRaw` pattern](https://github.com/manusa/podman-mcp-server/blob/b6b0cc1/internal/test/mcp.go#L141) used in [podman-mcp-server](https://github.com/manusa/podman-mcp-server) for testing raw JSON-RPC interactions.

The [MCP spec mandates](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#initialization) that `clientInfo` is sent during initialization, but some clients don't follow this, so we should handle it gracefully.

### DIFF
--- a/pkg/mcp/mcp_test.go
+++ b/pkg/mcp/mcp_test.go
@@ -2,7 +2,9 @@ package mcp
 
 import (
 	"fmt"
+	"io"
 	"net/http"
+	"net/http/httptest"
 	"runtime"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


manusa:
## Note: overlap with #859

PR #859 introduces `test.McpRawPost` — a reusable helper for sending raw JSON-RPC requests to MCP HTTP endpoints, bypassing the go-sdk client's automatic normalization. This is the same pattern used inline in this PR.

If #859 merges first, this PR can be rebased and simplified by replacing the inline HTTP request construction with `test.McpRawPost` calls:

```go
// Raw initialize (no clientInfo)
initResp := test.McpRawPost(s.T(), httpServer.URL+"/mcp", "",
    `{"jsonrpc":"2.0","id":0,"method":"initialize","params":{"protocolVersion":"2025-03-26"}}`)
defer initResp.Body.Close()
_, _ = io.ReadAll(initResp.Body)
sessionID := initResp.Header.Get("Mcp-Session-Id")

// Raw tool call
toolResp := test.McpRawPost(s.T(), httpServer.URL+"/mcp", sessionID,
    `{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"pods_list","arguments":{}}}`)
defer toolResp.Body.Close()
```

This eliminates the duplicated HTTP boilerplate (Content-Type, Accept headers, error handling) while keeping the custom `httptest.Server` setup this test needs for User-Agent stripping.

pkg/mcp/mcp_test.go
```diff
@@ -262,6 +265,64 @@ func (s *UserAgentPropagationSuite) TestFallsBackToServerPrefixWhenNoClientInfo(
...
+	handler := s.mcpServer.ServeHTTP()
+	strippedHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		r.Header.Del("User-Agent")
+		handler.ServeHTTP(w, r)
+	})
+	httpServer := httptest.NewServer(strippedHandler)
+	defer httpServer.Close()
+
+	// Send raw initialize request without clientInfo (non-compliant client)
+	initBody := `{"jsonrpc":"2.0","id":0,"method":"initialize","params":{"protocolVersion":"2025-03-26"}}`
```
Cali0707: I like this @manusa ! I couldn't get the SDK to send a init request without the clientInfo, but didn't think of this as another way to get the test to run...

manusa: I had to do a few tests like this in podman-mcp-server to complete the behavioral coverage for edge cases. Thought we could reuse here too. Funny that I ended up needing this for two separate tasks in almost the same day 😅

PR #850, opened 2026-03-02
fix(kubernetes): creating shared http client and close connections on the client

### LINKED ISSUE #830: Derived() creates new Kubernetes client per request without pooling; causes OOM under concurrency
The `Derived()` method in `pkg/kubernetes/manager.go` creates a new Kubernetes client for every request when a Bearer token is present. There is no pooling or caching. Under concurrent load, this leads to unbounded memory growth and OOMKilled pods.

## Steps to reproduce
1. Deploy the MCP server in-cluster with `REQUIRE_OAUTH=true` and Bearer token auth.
2. Run a load test (e.g. k6) with 50+ concurrent virtual users, each sending MCP tool calls with the same Bearer token.
3. Observe pod memory growth and eventual OOMKilled (exit code 137).

## Expected behavior

- Clients keyed by Bearer token (or token hash) should be reused from a bounded pool/cache.
- Memory usage should plateau as the number of unique tokens is typically small (e.g. one per user session).

## Actual behavior

- A new client is created for every request.
- Memory grows with concurrency until OOM.
- In our tests: error-free until ~50 VUs per pod with 8Gi memory; OOM beyond that.

## Proposed solution

Add a connection pool (or cache) keyed by Bearer token (or a hash of it for security):

- **Key:** Bearer token (or SHA256 hash to avoid storing raw tokens)
- **Value:** `*Kubernetes` (or a wrapper with ref-count / last-used)
- **Eviction:** LRU or TTL-based eviction when cache exceeds a configurable max size
- **Thread safety:** Use `sync.Map` or a mutex-protected map with proper cleanup on eviction

rajeeshckr:
Fixes #830

## Problem
`Manager.Derived()` creates a new Kubernetes client for every request carrying a Bearer token. Each client internally created 4 independent HTTP transports (discovery, clientset, dynamic, metrics) via `NewForConfig()`. None of these transports were ever closed after the request completed, leaving TCP sockets, TLS state, read/write buffers, and goroutines to accumulate. Under concurrent load this caused unbounded memory growth and eventual OOMKilled pods.

### Shared HTTP transport
`NewKubernetes` now creates a single shared `*http.Client` via `rest.HTTPClientFor()` and passes it to all sub-clients using `NewForConfigAndClient()`. This reduces per-client resource usage by ~4x and provides a single handle for cleanup.
This follows the pattern recommended by [Kubernetes SIG](https://github.com/kubernetes/kubernetes/issues/118703).

### Implement `utilnet.RoundTripperWrapper` on custom round trippers
Both `UserAgentRoundTripper` and `AccessControlRoundTripper` now implement `WrappedRoundTripper() http.RoundTripper`, which is the interface `utilnet.CloseIdleConnectionsFor` uses to recursively unwrap the transport chain and reach the base `http.Transport`. Without this, the close call would be a no-op since the outermost transport doesn't implement `closeIdler`.

### Context-based cleanup for derived clients
A new unexported `close()` method calls `utilnet.CloseIdleConnectionsFor(k.httpClient.Transport)` to release all HTTP transport resources (TCP sockets, TLS sessions, buffers, goroutines). This method is registered via `context.AfterFunc(ctx, derived.close)` in `Manager.Derived()`, so derived clients are automatically cleaned up when the request context finishes. The base client is unaffected since its context is never cancelled.

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -7,6 +7,7 @@ import (
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
@@ -109,7 +110,7 @@ func (k *Kubernetes) close() {
 	if k == nil || k.httpClient == nil {
 		return
 	}
-	k.httpClient.CloseIdleConnections()
+	utilnet.CloseIdleConnectionsFor(k.httpClient.Transport)
 }
 
 func (k *Kubernetes) RESTConfig() *rest.Config {
--- a/pkg/kubernetes/accesscontrol_round_tripper.go
+++ b/pkg/kubernetes/accesscontrol_round_tripper.go
@@ -54,6 +54,10 @@ func NewAccessControlRoundTripper(cfg AccessControlRoundTripperConfig) *AccessCo
 	return rt
 }
 
+func (rt *AccessControlRoundTripper) WrappedRoundTripper() http.RoundTripper {
+	return rt.delegate
+}
+
 func (rt *AccessControlRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	gvr, ok := parseURLToGVR(req.URL.Path)
 	// Not an API resource request, just pass through
--- a/pkg/kubernetes/kubernetes_close_test.go
+++ b/pkg/kubernetes/kubernetes_close_test.go
@@ -66,7 +66,6 @@ func (s *DerivedClientCleanupSuite) activeConnCount() int {
 func (s *DerivedClientCleanupSuite) TestClosesIdleConnectionsWhenContextIsCancelled() {
 	// https://github.com/containers/kubernetes-mcp-server/issues/830
 	// https://github.com/containers/kubernetes-mcp-server/pull/850
-	s.T().Skip("Pending fix: derived client must close idle connections when its context is cancelled")
 	baseConns := s.activeConnCount()
 
 	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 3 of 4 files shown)


pkg/kubernetes/kubernetes.go
```diff
@@ -74,27 +76,42 @@ func NewKubernetes(baseConfig api.BaseConfig, clientCmdConfig clientcmd.ClientCo
...
 
+// Close releases HTTP transport resources held by this client.
+// Only acts on derived (per-request) clients to avoid disrupting the base client.
+// Safe to call on any Kubernetes instance including nil.
+func (k *Kubernetes) Close() {
+	if k == nil || !k.isDerived || k.httpClient == nil {
+		return
+	}
+	k.httpClient.CloseIdleConnections()
+}
```
Cali0707: I think we can simplify this (and in particular remove the `derived` field) by registering this call with `context.AfterFunc(ctx, derived.close)`

The mcp go sdk has a per-request context which would be passed through here, so that should work correctly

manusa:
Thx for taking care of this, besides what Calum said, you should also [signoff](https://github.com/containers/kubernetes-mcp-server/pull/850/checks?check_run_id=65335838449) your commit.

pkg/kubernetes/kubernetes.go
rajeeshckr: Done, that looks way better

manusa:
## Review findings & acceptance tests

I've submitted #879 with skipped acceptance tests that verify derived client connections are actually cleaned up when the context is cancelled. These tests track TCP connections via `httptest.Server` `ConnState` and assert they close after `cancel()`.

### Key finding: `httpClient.CloseIdleConnections()` is a no-op

During testing, I found that the current implementation doesn't actually close connections. Here's why:

1. **`http.Client.CloseIdleConnections()`** checks if `Transport` implements `closeIdler`. The outermost transport in the chain is client-go's `userAgentRoundTripper`, which does **not** implement it. So the call is a no-op.

2. **`utilnet.CloseIdleConnectionsFor()`** exists in `k8s.io/apimachinery` and recursively unwraps via `RoundTripperWrapper` to reach the base `http.Transport`. However, the project's custom round trippers (`UserAgentRoundTripper` and `AccessControlRoundTripper`) don't implement `utilnet.RoundTripperWrapper`, so the unwrap chain breaks before reaching the base transport.

### What needs to change

For the fix to actually work, this PR needs to:

1. **Use `utilnet.CloseIdleConnectionsFor(k.httpClient.Transport)`** instead of `k.httpClient.CloseIdleConnections()` in the `close()` method.

2. **Implement `utilnet.RoundTripperWrapper`** on both custom round trippers:

```go
// In UserAgentRoundTripper
func (u *UserAgentRoundTripper) WrappedRoundTripper() http.RoundTripper {
    return u.delegate
}

// In AccessControlRoundTripper
func (rt *AccessControlRoundTripper) WrappedRoundTripper() http.RoundTripper {
    return rt.delegate
}
```

### Suggested workflow

1. Merge #879 first (skipped tests, no functional changes)
2. Rebase this PR on top
3. Apply the fixes above
4. Remove the `s.T().Skip(...)` lines from the tests
5. All tests should pass

rajeeshckr:
@manusa  this is ✅

thanks for adding the test and finding the issue!

Cali0707:
Thanks for catching and resolving this issue @rajeeshckr !

PR #862, opened 2026-03-02
feat(mcp): add URL-mode elicitation support

manusa:
Extend the `Elicitor` interface to support URL-mode elicitation. Refactor `Elicit()` to accept an `*ElicitParams` struct (instead of separate `message`/`schema` arguments) with fields for both form mode (`Message`, `RequestedSchema`) and URL mode (`URL`, `ElicitationID`).

Fix error detection for mode-specific unsupported elicitation errors by matching all three go-sdk error variants.

URL-mode elicitation was made available in go-sdk v1.4.0, bumped in #854.

Follow-up to #804.

### DIFF
--- a/pkg/mcp/elicit.go
+++ b/pkg/mcp/elicit.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"github.com/containers/kubernetes-mcp-server/pkg/mcplog"
-	"github.com/google/jsonschema-go/jsonschema"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/mcp/elicit.go
```diff
@@ -20,15 +19,24 @@ type sessionElicitor struct{}
...
+		URL:             params.URL,
+		ElicitationID:   params.ElicitationID,
+	})
 	if err != nil {
-		if strings.Contains(err.Error(), "does not support elicitation") {
+		// The go-sdk does not export a typed error for unsupported elicitation.
+		// This string check mirrors the go-sdk's own test approach (mcp/mcp_test.go).
+		// The go-sdk returns three variants: "client does not support elicitation",
+		// "client does not support "form" elicitation", and "client does not support "url" elicitation".
+		if strings.Contains(err.Error(), "does not support") && strings.Contains(err.Error(), "elicitation") {
```
Cali0707, line 39: 🤦 ... thanks for catching this @manusa !

manusa: I'm not 100% sure, but I think this changed from 1.3 to 1.4 (so when you created the original PR, probably it was just fine)

PR #881, opened 2026-03-05
chore(formatting) fix incorrect formatting 

matzew:
Since `make build` runs formatting rules, now after each build code is locally changed, due to  https://github.com/containers/kubernetes-mcp-server/pull/861 removing field


matzew:
@manusa I guess worth to eventually add a check for these kind of things?

manusa:
> @manusa I guess worth to eventually add a check for these kind of things?

Yes, we do!

https://github.com/containers/kubernetes-mcp-server/pull/882

PR #884, opened 2026-03-05
fix(kubevirt): fix vm_lifecycle pause/unpause blocked by REST mapper

lyarwood:
## Summary
- Adds pause and unpause actions to the `vm_lifecycle` tool using KubeVirt's subresource API
- Fixes `AccessControlRoundTripper` blocking requests to API groups not in the REST mapper (e.g. `subresources.kubevirt.io`), which prevented pause/unpause from working

### DIFF
--- a/pkg/kubernetes/accesscontrol_round_tripper.go
+++ b/pkg/kubernetes/accesscontrol_round_tripper.go
@@ -72,10 +72,11 @@ func (rt *AccessControlRoundTripper) RoundTrip(req *http.Request) (*http.Respons
 	gvk, err := restMapper.KindFor(gvr)
 	if err != nil {
 		if meta.IsNoMatchError(err) {
-			return nil, &api.ValidationError{
-				Code:    api.ErrorCodeResourceNotFound,
-				Message: fmt.Sprintf("Resource %s does not exist in the cluster", api.FormatResourceName(&gvr)),
-			}
+			// Some API groups (e.g. subresources.kubevirt.io) serve valid
+			// endpoints that are not discoverable via the REST mapper.
+			// Let the API server decide whether the resource exists.
+			klog.V(4).Infof("Resource %s not found in REST mapper, passing through to API server", api.FormatResourceName(&gvr))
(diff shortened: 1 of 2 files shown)


ksimon1:
@lyarwood can you please regenerate README?

pkg/kubernetes/accesscontrol_round_tripper.go
```diff
@@ -89,10 +72,11 @@ func (rt *AccessControlRoundTripper) RoundTrip(req *http.Request) (*http.Respons
...
 		if meta.IsNoMatchError(err) {
-			return nil, &api.ValidationError{
-				Code:    api.ErrorCodeResourceNotFound,
-				Message: fmt.Sprintf("Resource %s does not exist in the cluster", api.FormatResourceName(&gvr)),
-			}
+			// Some API groups (e.g. subresources.kubevirt.io) serve valid
+			// endpoints that are not discoverable via the REST mapper.
+			// Let the API server decide whether the resource exists.
+			klog.V(4).Infof("Resource %s not found in REST mapper, passing through to API server", api.FormatResourceName(&gvr))
+			return rt.delegate.RoundTrip(req)
```
Cali0707, line 79: @lyarwood I get why you need this change for kubevirt, but the original behavior was actually a requirement we were given.

I believe the `AccessControlRoundTripperConfig` already contains a `DiscoveryProvider` which should contain all the API groups (including subresources). Would that work for your use case?

If not we can discuss moving forwards with the proposed change, I would just prefer to find a way to keep the existing behaviour for actual invalid groups

manusa: AccessControlRoundTripper SHOULD NOT be made.
If certain API groups need exceptions, we'll need to find a proper way to handle those via the toolset API, and make it clear and explicit that certain routes are being allowed regardless of the security enforcements.

lyarwood:
/close

https://github.com/containers/kubernetes-mcp-server/issues/1061 will track fixing this and reintroducing the pause tool.

PR #892, opened 2026-03-06
feat(tekton): add Tekton toolset with pipeline and task management

ksimon1:
## Summary

This PR introduces a new `tekton` toolset to `kubernetes-mcp-server`, exposing domain-specific Tekton operations that go beyond what the generic `core` toolset's `resources_*` tools provide. This functionality is required for new Kubevirt functionality.

### Toolset (`pkg/toolsets/tekton`)

5 tools are exposed:

| Tool | Description |
|---|---|
| `tekton_pipeline_start` | Start a Pipeline by creating a PipelineRun |
| `tekton_pipelinerun_restart` | Restart a PipelineRun with the same spec |
| `tekton_task_start` | Start a Task by creating a TaskRun |
| `tekton_taskrun_restart` | Restart a TaskRun with the same spec |
| `tekton_taskrun_logs` | Fetch logs from a TaskRun's underlying pod |

The `tekton_pipeline_start` and `tekton_task_start` tools accept an optional `params` argument — a JSON object where values can be **strings**, **arrays of strings**, or **objects (map of string to string)**, matching all three Tekton `ParamType` variants.

Generic CRUD operations (list, get, create, update, delete) for Tekton resources remain available via the `core` toolset's `resources_*` tools using `apiVersion: tekton.dev/v1`.

### Eval tasks (`evals/tasks/tekton`)

9 declarative eval tasks validated against a live OpenShift cluster (**9/9 passed, 27/27 assertions**):

- `create-tekton-pipeline`, `create-tekton-task`
- `get-tekton-pipeline`, `list-tekton-pipelines`, `list-tekton-pipelineruns`
- `delete-tekton-pipelinerun`, `restart-tekton-pipelinerun`
- `start-tekton-pipeline`, `start-tekton-task`

Setup scripts handle a known OpenShift Pipelines quirk: the Tekton Results operator adds `results.tekton.dev/taskrun` finalizers to TaskRuns, which can block namespace termination between tasks. The scripts actively remove these finalizers to ensure clean teardown.

The `tekton` suite is registered in both `claude-code` and `openai-agent` eval configurations.

## Test plan

- [x] Unit tests: `go test ./pkg/toolsets/tekton/...` — all pass
- [x] Snapshot tests: `go test ./pkg/mcp/...` — all pass
- [x] Eval tests: 9/9 tasks passed on a live OpenShift cluster with OpenShift Pipelines installed


README.md
```diff
@@ -34,6 +34,11 @@ A powerful and flexible Kubernetes [Model Context Protocol (MCP)](https://blog.m
   - **Install** a Helm chart in the current or provided namespace.
   - **List** Helm releases in all namespaces or in a specific namespace.
   - **Uninstall** a Helm release in the current or provided namespace.
+- **🔧 Tekton**: Tekton-specific operations that complement generic Kubernetes resource management.
+  - **Pipeline**: Start a Tekton Pipeline by creating a PipelineRun.
+  - **PipelineRun**: Restart a PipelineRun with the same spec, or bulk-delete PipelineRuns with selectors.
```
waveywaves: This PR does not have bulk-delete implemented, maybe a missed commit ?

pkg/toolsets/tekton/taskrun.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return api.NewToolCallResult("", fmt.Errorf("failed to get TaskRun %s/%s: %w", namespace, name, err)), nil
+	}
+
+	if tr.Status.PodName == "" {
+		return api.NewToolCallResult(fmt.Sprintf("TaskRun '%s' in namespace '%s' has not started a pod yet", name, namespace), nil), nil
+	}
+
+	logs := ""
+	for _, step := range tr.Status.Steps {
```
waveywaves, line 177: we should get logs for sidecars as well

README.md
```diff
@@ -515,6 +520,36 @@ In case multi-cluster support is enabled (default) and you have access to multip
 
 </details>
 
+<details>
+
+<summary>tekton</summary>
+
+> **Note**: Generic CRUD operations (list, get, create, update, delete) for Tekton resources (Pipeline, PipelineRun, Task, TaskRun) are available via the `core` toolset's `resources_list`, `resources_get`, `resources_create_or_update`, and `resources_delete` tools using `apiVersion: tekton.dev/v1`.
```
waveywaves: https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/toolsets/core/resources.go#L115

don't think resources_delete supports bulk deletion using label selectors

README.md
ksimon1: Many actions can be done with general tools which are already implemented in this MCP server. If you think we will need some special delete function, then we can add it :+1:

Cali0707: Maybe we can start here with an eval task to see if it is needed?

pkg/toolsets/tekton/tekton_test.go
```diff
@@ -0,0 +1,126 @@
+package tekton_test
+
+import (
+	"testing"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/toolsets/tekton"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
```
matzew, line 8: test should be grouped via `testify/suite`, like:

```go
type TektonSuite struct {
	suite.Suite
}
```

README.md
ksimon1: @Cali0707 I created new PR https://github.com/containers/kubernetes-mcp-server/pull/917

pkg/toolsets/tekton/taskrun.go
```diff
@@ -0,0 +1,164 @@
...
+		req := params.KubernetesClient.CoreV1().Pods(namespace).GetLogs(tr.Status.PodName, &corev1.PodLogOptions{
+			Container: step.Container,
+		})
+		stream, err := req.Stream(params.Context)
```
vdemeester: `io.ReadAll` reads without bound. A chatty step could return hundreds of MB, hitting LLM context limits or OOMing the server.

Suggest capping per step:
```go
const maxStepLogBytes = 1 << 20 // 1 MB per step
bytes, err := io.ReadAll(io.LimitReader(stream, maxStepLogBytes))
```

pkg/toolsets/tekton/taskrun.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return api.NewToolCallResult("", fmt.Errorf("failed to create Tekton client: %w", err)), nil
+	}
+
+	tr, err := tektonClient.TektonV1().TaskRuns(namespace).Get(params.Context, name, metav1.GetOptions{})
+	if err != nil {
+		return api.NewToolCallResult("", fmt.Errorf("failed to get TaskRun %s/%s: %w", namespace, name, err)), nil
+	}
+
+	if tr.Status.PodName == "" {
```
vdemeester, line 171: String concatenation with `+=` in a loop is O(n²) in allocations. For multi-step tasks this adds up. Use `strings.Builder` instead:
```go
var sb strings.Builder
// ...
if len(bytes) > 0 {
    fmt.Fprintf(&sb, "[step: %s]
%s
", step.Name, string(bytes))
}
// ...
return api.NewToolCallResult(sb.String(), nil), nil
```

pkg/toolsets/tekton/client.go
```diff
@@ -0,0 +1,11 @@
+package tekton
+
+import (
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	versioned "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
```
vdemeester: The alias `versioned` is identical to the package name — it's redundant. Drop it:
```go
import (
    "github.com/containers/kubernetes-mcp-server/pkg/api"
    "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
)
```

pkg/toolsets/tekton/client.go
```diff
@@ -0,0 +1,11 @@
...
+)
+
+// newTektonClient creates a Tekton clientset from the REST config in the provided KubernetesClient.
+func newTektonClient(k api.KubernetesClient) (versioned.Interface, error) {
+	return versioned.NewForConfig(k.RESTConfig())
```
vdemeester: A new `versioned.Interface` is constructed on every tool call. Other toolsets (e.g. kubevirt) use `params.DynamicClient()` which is pre-initialised per session.

`versioned.NewForConfig` is cheap (just builds an HTTP client struct, no network round-trip), but for consistency you could cache it lazily in the `Toolset` struct:
```go
type Toolset struct {
    once   sync.Once
    client versioned.Interface
    err    error
}

func (t *Toolset) tektonClient(k api.KubernetesClient) (versioned.Interface, error) {
    t.once.Do(func() { t.client, t.err = versioned.NewForConfig(k.RESTConfig()) })
    return t.client, t.err
}
```

pkg/toolsets/tekton/taskrun.go
vdemeester: I didn't look at how it is done directly in the kubernetes mcp server though, but we could re-use code/way of doing.

manusa:
#917 was merged, I think we need to rebase this one now.

pkg/toolsets/tekton/taskrun.go
ksimon1: Fixed

pkg/toolsets/tekton/client.go
ksimon1: fixed

ksimon1:
@manusa the PR is rebased

pkg/toolsets/tekton/tekton_test.go
manusa, line 1: This test is mostly redundant with toolset_tests.go which check this with a snapshot of the tool schema json

pkg/toolsets/tekton/client.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
+)
+
+var (
+	tektonClientOnce sync.Once
+	tektonClient     versioned.Interface
+	tektonClientErr  error
+)
+
+func getOrCreateTektonClient(k api.KubernetesClient) (versioned.Interface, error) {
```
manusa: The `sync.Once` singleton pattern here has a few concerns:

1. **No recovery from initialization failure**: If `versioned.NewForConfig` fails once (transient network issue, misconfigured REST config), every subsequent call returns the same cached error forever — the `sync.Once` will never re-execute.

2. **Diverges from codebase conventions**: No other toolset uses this pattern. The core toolset creates `kubernetes.NewCore(params)` per call, Kiali creates `kialiclient.NewKiali(params, params.RESTConfig())` per call. This should follow the same approach.

3. **Untestable**: The package-level singleton can't be reset or replaced with a fake client for integration tests.

Consider creating the client per-call (or caching it on the `Toolset` struct with a mutex if performance is a concern).

pkg/toolsets/tekton/taskrun.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	return api.NewToolCallResult(sb.String(), nil), nil
+}
+
+func collectContainerLogs(params api.ToolHandlerParams, sb *strings.Builder, podName, namespace, kind, name, container string) {
+	req := params.CoreV1().Pods(namespace).GetLogs(podName, &corev1.PodLogOptions{
+		Container: container,
+	})
+	stream, err := req.Stream(params.Context)
```
manusa, line 196: The log stream should be closed with `defer` immediately after the error check:

```go
stream, err := req.Stream(params.Context)
if err != nil {
    fmt.Fprintf(sb, "[%s: %s] error retrieving logs: %v\n", kind, name, err)
    return
}
defer stream.Close()
```

Currently `stream.Close()` is called manually after `io.ReadAll` (line 167), but if `ReadAll` panics the stream leaks. Using `defer` is the standard Go idiom and is safer.

pkg/toolsets/tekton/taskrun.go
```diff
@@ -0,0 +1,175 @@
...
+
+	created, err := tektonClient.TektonV1().TaskRuns(namespace).Create(params.Context, newTR, metav1.CreateOptions{})
+	if err != nil {
+		return api.NewToolCallResult("", fmt.Errorf("failed to create restart TaskRun for %s/%s: %w", namespace, name, err)), nil
+	}
+
+	return api.NewToolCallResult(fmt.Sprintf("TaskRun '%s' restarted as '%s' in namespace '%s'", name, created.Name, namespace), nil), nil
+}
+
+func getTaskRunLogs(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
```
manusa, line 139: In other log retrieval tools we provide a parameter for the llm to limit (tail) the number of log lines to retrieve:

https://github.com/containers/kubernetes-mcp-server/blob/7cfe708ec00563a654eb386ea8496880a6091534/pkg/toolsets/core/pods.go#L208-L213

this is essential to enable the llm to reduce the context window usage.

waveywaves:
@ksimon1 can you check why the build is failing on this one ?

pkg/toolsets/tekton/tekton_test.go
ksimon1: Tests were removed.

pkg/toolsets/tekton/client.go
ksimon1: fixed

pkg/toolsets/tekton/taskrun.go
ksimon1: fixed

ksimon1:
@manusa, @Cali0707, @waveywaves, @vdemeester  can you please review this PR?

manusa (approved):
LGTM, thx!

ksimon1:
@nader-ziada it should be fixed now

nader-ziada:
> @nader-ziada it should be fixed now

thanks @ksimon1 

LGTM

PR #910, opened 2026-03-10
chore(test): Increase timeout on namespaces test

### LINKED ISSUE #909: TestNamespacesListForbidden flaky on CI
On downstream CI I noticed:

```
  --- FAIL: TestNamespaces/TestNamespacesListForbidden (2.08s)
      --- FAIL: TestNamespaces/TestNamespacesListForbidden/namespaces_list_(forbidden) (2.01s)
          --- FAIL: .../returns_error (0.00s)
          --- FAIL: .../sends_log_notification (2.01s)
```

The itself test uses a 2-second timeout, sometimes this than fails. Idea is to enhance that timeout

matzew:
Fixes #909 

Increase test to 5 seconds to avoid some flake, when CI is under load.

### DIFF
--- a/pkg/mcp/namespaces_test.go
+++ b/pkg/mcp/namespaces_test.go
@@ -101,7 +101,7 @@ func (s *NamespacesSuite) TestNamespacesListForbidden() {
 				"error message should indicate forbidden")
 		})
 		s.Run("sends log notification", func() {
-			logNotification := capture.RequireLogNotification(s.T(), 2*time.Second)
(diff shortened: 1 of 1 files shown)


matzew:
That's a good point! 

I noticed this only on the TestNamespace (downstream) CI. I have seen other flakes, and I am looking into that.

I guess we could update all others to be also 5 seconds - but than perhaps we could also embed some "default" directly into the `func` ? :man_shrugging: 

I think for now just this might be just fine

manusa:
> I think for now just this might be just fine

Let's merge and see how it pans out.

PR #913, opened 2026-03-11
cleanup(kubevirt): Remove snapshot patch in snapshot-vm and restore-vm tasks

### LINKED ISSUE #806: [kubevirt] Redundant feature gate patch in snapshot-vm and restore-vm evaluation tasks
The patch in the setup of the snapshot-vm and restore-vm tasks which enables the Snapshot feature gate in Kubevirt CR should be removed to improve clarity as this feature gate is already enabled during the setup of the local environment. This is already omitted in the clone-vm task so I think similar approach should be utilized in this situation.
Currently there is #756 which makes use of  kubernetes extension of mcpchecker in the tasks.

Ruclo:
To increase clarity, I removed the snapshot feature gate patch in these tasks. 
The snapshot feature gate is now enabled by default.

Fixes #806

### DIFF
--- a/evals/tasks/kubevirt/restore-vm/task.yaml
+++ b/evals/tasks/kubevirt/restore-vm/task.yaml
@@ -11,15 +11,6 @@ spec:
     - extension: kubernetes
       as: k8s
   setup:
-    # Enable Snapshot feature gate - requires kubectl patch (no extension op available)
-    - script:
-        inline: |-
-          #!/usr/bin/env bash
-          set -e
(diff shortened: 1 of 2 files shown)


Ruclo:
@Cali0707 Any idea what might be wrong with the CI?

manusa:
> @Cali0707 Any idea what might be wrong with the CI?

we have some flaky tests, ci doesn't run pipelines specific to these changes automatically

PR #916, opened 2026-03-12
refactor(api): expose BaseConfig to tool and prompt handlers

manusa:
## Summary

PR #836 identified the need to expose the cluster provider strategy to toolsets. As some toolsets now call non-kube-API-server endpoints (e.g. Prometheus), they need to know the strategy to determine TLS and authentication configuration.

The approach in #836 expanded the `GetTools()` parameter, but this is evaluated at server startup — if the strategy changes via drop-in configuration support, toolsets would still see the stale value. Additionally, `GetTools()` should eventually take no parameters.

This PR takes an alternative approach: embed `BaseConfig` (instead of `ExtendedConfigProvider`) in `ToolHandlerParams` and `PromptHandlerParams`. Since `BaseConfig` already composes `ClusterProvider` (which provides `GetClusterProviderStrategy()`) and `ExtendedConfigProvider`, no new interface is needed. Toolsets can access the strategy at execution time — always reflecting the current configuration, even after a drop-in config reload.

Downstream consumers (e.g. [openshift/openshift-mcp-server#124](https://github.com/openshift/openshift-mcp-server/pull/124)) can now call `params.GetClusterProviderStrategy()` in their tool handlers to determine TLS/auth behavior, following the same pattern Kiali already uses for its configuration.

### Changes

- **`pkg/api/toolsets.go`** / **`pkg/api/prompts.go`** — Embed `BaseConfig` instead of `ExtendedConfigProvider` in handler params
- **`pkg/mcp/gosdk.go`** / **`pkg/mcp/prompts_gosdk.go`** — Update field assignment
- **`pkg/mcp/mcp_config_provider_test.go`** — Behavioral tests verifying strategy access from tool/prompt handlers and config reload

Refs #836

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -26,12 +26,7 @@ type ExtendedConfig interface {
 	Validate() error
 }
 
-// ConfigProvider provides read-only access to configuration from tool and prompt handlers.
-// This interface is embedded in ToolHandlerParams and PromptHandlerParams, allowing
-// toolsets to access configuration at execution time (not just at initialization).
-type ConfigProvider interface {
-	// GetClusterProviderStrategy returns the cluster provider strategy (e.g. "kubeconfig", "in-cluster").
-	GetClusterProviderStrategy() string
+type ExtendedConfigProvider interface {
 	// GetProviderConfig returns the extended configuration for the given provider strategy.
 	// The boolean return value indicates whether the configuration was found.
 	GetProviderConfig(strategy string) (ExtendedConfig, bool)
@@ -67,7 +62,7 @@ type BaseConfig interface {
 	AuthProvider
 	ClusterProvider
 	DeniedResourcesProvider
-	ConfigProvider
+	ExtendedConfigProvider
 	StsConfigProvider
 	ValidationEnabledProvider
 }
--- a/pkg/api/prompts.go
+++ b/pkg/api/prompts.go
@@ -84,7 +84,7 @@ func NewPromptCallResult(description string, messages []PromptMessage, err error
 // PromptHandlerParams contains the parameters passed to a prompt handler
 type PromptHandlerParams struct {
 	context.Context
-	ConfigProvider
+	BaseConfig
 	KubernetesClient
 	PromptCallRequest
 	Elicitor
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -99,7 +99,7 @@ func NewToolCallResultStructured(structured any, err error) *ToolCallResult {
 
 type ToolHandlerParams struct {
 	context.Context
-	ConfigProvider
+	BaseConfig
 	KubernetesClient
 	ToolCallRequest
 	ListOutput output.Output
--- a/pkg/kiali/kiali.go
+++ b/pkg/kiali/kiali.go
@@ -24,7 +24,7 @@ type Kiali struct {
 }
 
 // NewKiali creates a new Kiali instance
-func NewKiali(configProvider api.ConfigProvider, kubernetes *rest.Config) *Kiali {
+func NewKiali(configProvider api.ExtendedConfigProvider, kubernetes *rest.Config) *Kiali {
 	kiali := &Kiali{bearerToken: kubernetes.BearerToken}
 	if cfg, ok := configProvider.GetToolsetConfig("kiali"); ok {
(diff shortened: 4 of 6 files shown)


pkg/api/config.go
```diff
@@ -26,7 +26,12 @@ type ExtendedConfig interface {
...
-type ExtendedConfigProvider interface {
+// ConfigProvider provides read-only access to configuration from tool and prompt handlers.
+// This interface is embedded in ToolHandlerParams and PromptHandlerParams, allowing
+// toolsets to access configuration at execution time (not just at initialization).
+type ConfigProvider interface {
+	// GetClusterProviderStrategy returns the cluster provider strategy (e.g. "kubeconfig", "in-cluster").
+	GetClusterProviderStrategy() string
 	// GetProviderConfig returns the extended configuration for the given provider strategy.
 	// The boolean return value indicates whether the configuration was found.
 	GetProviderConfig(strategy string) (ExtendedConfig, bool)
```
matzew, line 32: we wanna change this to just Config? Or wanna keep the `ExtendedConfig` here and below?

pkg/api/config.go
```diff
@@ -26,7 +26,12 @@ type ExtendedConfig interface {
 	Validate() error
 }
 
-type ExtendedConfigProvider interface {
+// ConfigProvider provides read-only access to configuration from tool and prompt handlers.
+// This interface is embedded in ToolHandlerParams and PromptHandlerParams, allowing
+// toolsets to access configuration at execution time (not just at initialization).
+type ConfigProvider interface {
```
matzew: Rename to better communicate that this is the build-in config provider, rather than something that was intentially extending it?

pkg/api/config.go
matzew: Ah, I see that just actually retunrs the extended config - nervemind

matzew:
LGTM

pkg/api/config.go
manusa: ExtendedConfig is something else altogether.
The idea is that toolset implementers can provide custom Config entries in the toml file (hence the extended).

You can check how Kiali deals with this:
https://github.com/containers/kubernetes-mcp-server/blob/4f0ed3e59898cf856203dd5b83bd21076507d1a0/pkg/kiali/config.go#L17-L22

By having an extended config, users can the provide specific values to configure the toolset:
```toml
[toolset_configs.kiali]
url = "https://kiali.example/"
certificate_authority = "ca.crt"
```

And then consumed through:
https://github.com/containers/kubernetes-mcp-server/blob/b93fadefe78e11c5f4ce9a27abdcf949d75f050c/pkg/kiali/kiali.go#L29-L35

This basically makes the config extensible by toolsets in a decoupled way.

pkg/api/config.go
manusa: I'm not sure I'm following, do you mean to rename to `BuiltinConfigProvider`? 
Basically the api interface here is just exposing the Config values (those we've decided to make public) instead of completely exposing the StaticConfig.
Any Config value that needs to be accessed by toolsets would need to be explicitly exposed here.

pkg/api/config.go
```diff
@@ -26,7 +26,12 @@ type ExtendedConfig interface {
...
+	// GetClusterProviderStrategy returns the cluster provider strategy (e.g. "kubeconfig", "in-cluster").
+	GetClusterProviderStrategy() string
```
Cali0707: @manusa is there any concern about this method being duplicated between this interface and the `ClusterProvider` interface which also goes into `BaseConfig` ?

This all compiles fine and is concretely backed by a single func, so i guess maybe it is fine?

manusa: Yes, there is :) I had that flagged.

Another option is to revert the changes to the api package, and actually expose the api.BaseConfig interface instead of `api.ConfigProvider`.

I'll add this as a second commit and let you folks decide what's best (@matzew @Cali0707)

matzew:
LGTM

I think renaming of `ExtendedConfigProvider` to `BaseConfig` is fine. Should the PR title cover that too?

manusa:
> LGTM
> 
> I think renaming of `ExtendedConfigProvider` to `BaseConfig` is fine. Should the PR title cover that too?

🤦 forgot to update the PR title and description. Thanks for catching that.

Technically it's not renaming but exposing BaseConfig as part of the tools and prompt handler params.

Issue #877, opened 2026-03-03
Explore output_query parameter across tools


Cali0707:
In eval runs, we saw experimentally that adding an output_query param allowed the model to reduce token usage by about 30% across our eval suite. The key idea behind this param is the model could use a jq query to filter the data down to what it needs.

This was done with https://github.com/Cali0707/mcp-token-experiments, and leveraged #812 

However another approach we should consider here is to use yq against the yaml formatted text results, instead of forcing the structured outputs.

Either way, since there are multiple issues about adding parameters to reduce the output size in particular ways, since this is a general and experimentally backed approach I think we should add it (behind a config flag).

manusa:
We should expand on the context of `StaticConifg.ListOutput`.
Maybe create additional options, improve `pkg/output` to cover additional options, etc.

This is also something that's needed in the scope of #753, where the output for each took needs processing if the feature is enabled.

We could also look into the intersection of output filtering and code mode (#711). To reduce the number of output tokens, tools could accept some JavaScript (or any other scripting language that can easily be integrated) to post-process the output before receiving it.

pixelsoccupied:
thanks @Cali0707 for pointing me to #877 and #812 and saw @manusa's comment about the broader output processing vision #753, #711.

My core need is simpler i think? when diagnosing resources, I only need .status — not the full resource with managedFields, spec, and annotations bloating the context.

PR #812 adds MCP structuredContent (not to be confused with LLM structured output...you have something called PruneForStructuredOutput and also description said structured output)...but the server already has the data as Go maps before YAML serialization so filtering can happen server-side without needing StructuredContent on the response.

Also, the pruning unconditionally removes last-applied-configuration and managedFields — the former can be useful for debugging drift or understanding what was previously applied to a CR. Did i miss something there?

A simpler first step: add a filter parameter (e.g., jsonpath) to tools like resources_get. Apply it on the existing Go map, return only the filtered YAML. Similar to kubectl get -o jsonpath='{.status}' but done server-side. This can land now and be refactored into the broader pkg/output layer later.

Happy to open a PR if this direction works.

pixelsoccupied:
In the spirit of "reduce context bloat" 

Another concrete case....I have 9 ACM policies but only care about `NonCompliant` (as opposed to `Compliant`) ones. There's no field selector or label for status.compliant, so `resources_list` returns all 9 with full output into the LLM context.

With kubectl jsonpath this is straightforward:

```
[...]

PR #918, opened 2026-03-12
feat(resources): add gotemplate parameter for resources_list and resources_get

edcdavid:
Adds an optional gotemplate parameter (same syntax as oc get -o go-template) that lets callers extract specific fields server-side instead of returning full YAML.

Why: LLMs currently receive entire resource YAML even when they only need one field. This wastes context window tokens and slows down responses. With gotemplate, the LLM declares exactly what it needs and the server returns only that — easily 10-100x fewer tokens for targeted queries.

Benefits:

Token efficiency — {{.status.phase}} returns a single value instead of hundreds of YAML lines
CLI parity — uses the same Go template syntax as oc get -o go-template, so templates are portable between CLI and MCP workflows
Server-side projection — iteration, conditionals, and field extraction run on the server, not in the LLM
Backward compatible — parameter is optional; omitting it preserves existing full-YAML behavior

Examples:

resources_list(kind="Pod", gotemplate="{{range .items}}{{.metadata.name}}\n{{end}}")
resources_get(kind="Application", name="my-app", gotemplate="{{.spec.source.repoURL}}")

Assisted by: Cursor

### DIFF
--- a/pkg/mcp/resources_test.go
+++ b/pkg/mcp/resources_test.go
@@ -1029,6 +1029,113 @@ func (s *ResourcesSuite) TestResourcesScaleDenied() {
 	})
 }
 
+func (s *ResourcesSuite) TestResourcesListGoTemplate() {
+	s.InitMcpClient()
+	s.Run("resources_list with gotemplate extracts names", func() {
+		result, err := s.CallTool("resources_list", map[string]interface{}{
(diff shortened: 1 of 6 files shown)


manusa:
https://github.com/containers/kubernetes-mcp-server/issues/877

PR #925, opened 2026-03-16
fix(kubernetes): avoid startup deadlock when the Kubernetes API is served behind a path prefix

zhiyu0729:
## Summary

This PR fixes a startup deadlock when `kubernetes-mcp-server` talks to a Kubernetes API through a gateway or proxy that exposes the API under a path prefix.

Sanitized example:

- Base URL: `https://<gateway-host>/<proxy-prefix>`
- Discovery requests:
  - `GET /<proxy-prefix>/api`
  - `GET /<proxy-prefix>/apis`
  - `GET /<proxy-prefix>/openapi/v2`

Before this change, `AccessControlRoundTripper` assumed request paths started directly at the Kubernetes API root (`/api` or `/apis`). When the configured API server URL included an extra path prefix, discovery requests could be misinterpreted as normal resource requests. That could cause the access-control layer to call `restMapper.KindFor(...)` while discovery / RESTMapper initialization was already in progress, leading to recursive discovery and a deadlock during startup.

In practice, this could prevent the server from becoming ready before it even handled the first MCP request.

## Problem

In environments where the Kubernetes API is exposed behind a path-prefixed gateway, startup can trigger early discovery calls as part of:

- toolset loading
- cluster capability detection
- OpenShift detection
- RESTMapper initialization

Because the round-tripper parsed the full HTTP path without accounting for the configured host prefix, requests such as `/<proxy-prefix>/api` were not treated as plain discovery endpoints. Instead, they could fall into the access-control path and try to resolve a resource kind via the RESTMapper.

That is unsafe during discovery bootstrap, because `DeferredDiscoveryRESTMapper` may re-enter the cached discovery client while discovery state is still being populated, which can deadlock.

## Root Cause

The issue comes from a mismatch between:

- the configured API server host, which may include a path prefix
- the request parsing logic in `AccessControlRoundTripper`, which assumed root-relative Kubernetes paths

More specifically:

1. `rest.Config.Host` may be something like `https://<gateway-host>/<proxy-prefix>`
2. client-go then sends discovery requests against paths like `/<proxy-prefix>/api`
3. `AccessControlRoundTripper` parsed `req.URL.Path` as-is
4. the path was interpreted using Kubernetes root-path rules
5. the request could incorrectly enter the RESTMapper / access-control flow
6. `restMapper.KindFor(...)` triggered discovery again
7. discovery re-entered the cached client during initialization and the process deadlocked

## What This PR Changes

This PR makes path handling prefix-aware.

### Changes included

- Extract the path prefix from `rest.Config.Host` when constructing the Kubernetes client
- Pass that prefix into `AccessControlRoundTripper`
- Strip the configured prefix from outbound request paths before:
  - GVR parsing
  - namespace/name extraction
  - HTTP verb derivation
  - validation request path generation

After stripping the prefix, the round-tripper works with the Kubernetes-relative path instead of the proxy-relative path.

## Why This Fix Works

With this change:

- `/<proxy-prefix>/api` is normalized to `/api`
- `/<proxy-prefix>/apis` is normalized to `/apis`
- `/<proxy-prefix>/api/v1/namespaces/default/pods` is normalized to `/api/v1/namespaces/default/pods`

That restores the assumptions already used by the parser and the validators, without changing behavior for standard Kubernetes API server URLs.

Discovery requests remain discovery requests.
Normal resource requests remain normal resource requests.
The access-control layer no longer misclassifies prefixed discovery traffic.

### DIFF
--- a/pkg/kubernetes/accesscontrol_round_tripper_test.go
+++ b/pkg/kubernetes/accesscontrol_round_tripper_test.go
@@ -315,6 +315,23 @@ func (s *AccessControlRoundTripperTestSuite) TestRoundTripForDeniedAPIResources(
 			s.False(delegateCalled)
 			s.Contains(err.Error(), "resource not allowed")
 		})
+
+		s.Run("List pods behind an API path prefix is denied", func() {
+			delegateCalled = false
+			originalAPIPathPrefix := rt.apiPathPrefix
+			defer func() {
+				rt.apiPathPrefix = originalAPIPathPrefix
+			}()
+			rt.apiPathPrefix = "/api/v1/kube/clusters/test-cluster"
+
+			req := httptest.NewRequest("GET", "/api/v1/kube/clusters/test-cluster/api/v1/namespaces/default/pods", nil)
+			resp, err := rt.RoundTrip(req)
+			s.Error(err)
+			s.Nil(resp)
+			s.False(delegateCalled, "Expected delegate not to be called for denied resource behind a path prefix")
+			s.Contains(err.Error(), "resource not allowed")
+			s.Contains(err.Error(), "Pod")
+		})
 	})
 
 	s.Run("Entire group/version is denied", func() {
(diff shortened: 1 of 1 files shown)


pkg/kubernetes/accesscontrol_round_tripper_test.go
```diff
@@ -220,6 +260,16 @@ func (s *AccessControlRoundTripperTestSuite) TestRoundTripForAllowedAPIResources
 		s.NotNil(resp)
 		s.True(delegateCalled, "Expected delegate to be called for namespaced deployments list")
 	})
+
+	s.Run("List pods in namespace is allowed when the API server has a path prefix", func() {
+		delegateCalled = false
+		rt.apiPathPrefix = "/api/v1/kube/clusters/test-cluster"
```
Cali0707, line 270: maybe we can defer something to restore the api path prefix here, as this may cause other tests added to this suite later to fail (modifying the shared roundtripper between test cases)

pkg/kubernetes/kubernetes.go
```diff
@@ -63,11 +64,17 @@ func NewKubernetes(baseConfig api.BaseConfig, clientCmdConfig clientcmd.ClientCo
 		k.restConfig.UserAgent = rest.DefaultKubernetesUserAgent()
 	}
 
+	apiPathPrefix := ""
+	if hostURL, err := url.Parse(k.restConfig.Host); err == nil {
```
Cali0707: If err is non nil, let's log a warning to help users debug what's going wrong

zhiyu0729:
Thanks! I addressed both comments in the latest update.

pkg/kubernetes/accesscontrol_round_tripper.go
```diff
@@ -189,6 +193,22 @@ func parseURLToGVR(path string) (gvr schema.GroupVersionResource, ok bool) {
...
 }
 
+func stripAPIPathPrefix(path, prefix string) string {
+	if prefix == "" || prefix == "/" {
+		return path
+	}
+
+	if path == prefix {
+		return "/"
+	}
```
matzew, line 203: do we want a simple unit test for these cases?
as far as I see the `TestRoundTripForPrefixedDiscoveryRequests` covers the third branch.

matzew:
Do we want to add a test for denied resources behind a path prefix?

zhiyu0729:
Added coverage for the denied-resource case behind an API path prefix.

Issue #926, opened 2026-03-17
Limit Kiali API response body size to prevent unbounded memory consumption


matzew:
The Kiali HTTP client in `pkg/kiali/kiali.go` reads response bodies using `io.ReadAll(resp.Body)` without any size limit. A misbehaving or compromised Kiali server could return an arbitrarily large response, causing the MCP server process to exhaust available memory.

PR #927, opened 2026-03-17
fix(kiali): limit response body size to prevent unbounded memory consumption

matzew:
Wrap Kiali API response reads with io.LimitReader to cap at 512 KiB, preventing a misbehaving or compromised upstream server from exhausting MCP server memory via arbitrarily large responses.

Closes https://github.com/containers/kubernetes-mcp-server/issues/926

### DIFF
--- a/pkg/kiali/kiali.go
+++ b/pkg/kiali/kiali.go
@@ -137,6 +137,11 @@ func (k *Kiali) authorizationHeader() string {
 	return "Bearer " + token
 }
 
+// maxResponseBodySize is the maximum number of bytes read from a Kiali API
+// response. Responses exceeding this limit are truncated to prevent unbounded
+// memory consumption from a misbehaving or compromised upstream server.
+const maxResponseBodySize = 512 << 10 // 512 KiB
+
 // executeRequest executes an HTTP request (optionally with a body) and handles common error scenarios.
 func (k *Kiali) executeRequest(ctx context.Context, method, endpoint, contentType string, body io.Reader) (string, error) {
 	if method == "" {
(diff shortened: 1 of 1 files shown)


pkg/kiali/kiali.go
```diff
@@ -137,6 +137,11 @@ func (k *Kiali) authorizationHeader() string {
 	return "Bearer " + token
 }
 
+// maxResponseBodySize is the maximum number of bytes read from a Kiali API
+// response. Responses exceeding this limit are truncated to prevent unbounded
+// memory consumption from a misbehaving or compromised upstream server.
+const maxResponseBodySize = 512 << 10 // 512 KiB
```
matzew, line 143: @aljesusg not sure if that is the best (gu)estimated size 😅

aljesusg:
We are working on a new approach to reduce the output. To avoid breaking the MCP, we will implement these changes two weeks after the Kiali release

matzew:
@aljesusg Is up to you, if we should close this now, and wait for the new version. Or we merge this now, and the proper fix lands than when kiali release is done? 

The size was something I am not sure, but open to suggestions.

CC @manusa

aljesusg:
Initial results indicate an 80–85% reduction in size. We are scheduled for the next release on the 27th, and I plan to submit the PR then.

I’m currently unable to verify if this threshold covers the new response. We can either merge now and verify post-release, or wait until the new outputs are available

matzew:
@manusa how about we merge this after you did cut the relase?

manusa:
Merging as is for now, since we just release [v0.0.59](https://github.com/containers/kubernetes-mcp-server/releases/tag/v0.0.59), even if this breaks the kiali toolset, users won't be affected until the next release.

matzew:
Thx

PR #929, opened 2026-03-18
fix(ci): add v prefix to OCI identifier in MCP registry publish

manusa:
The `release-image` workflow tags container images with the full git ref name
(e.g., `v0.0.59`), but the MCP registry workflow was stripping the `v` prefix
from the OCI identifier, producing a tag (`0.0.59`) that doesn't exist in the
container registry.

This caused the MCP registry publish to fail with:
> OCI image 'ghcr.io/containers/kubernetes-mcp-server:0.0.59' does not exist in the registry

The fix prepends `v` to the version when updating the OCI identifier in `server.json`.

### DIFF
--- a/.github/workflows/release-mcp-registry.yaml
+++ b/.github/workflows/release-mcp-registry.yaml
@@ -53,7 +53,7 @@ jobs:
           VERSION: ${{ steps.version.outputs.version }}
         run: |
           jq --arg v "$VERSION" \
-            '.version = $v | .packages[] |= if .registryType == "oci" then .identifier = (.identifier | sub(":[^:]+$"; ":" + $v)) else .version = $v end' \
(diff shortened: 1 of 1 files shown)


matzew:
@manusa means we will just "repush" or redo the 0.0.59 release?

manusa:
> @manusa means we will just "repush" or redo the 0.0.59 release?

I've been working on the workflow for months now.
No need to repush, I did a manual trigger after the fix and it finally worked 🎉 https://github.com/containers/kubernetes-mcp-server/actions/runs/23250704781

Now I need to find some way to deal with the invalid 1.0.0 version that the registry features and that's not valid: https://registry.modelcontextprotocol.io/v0/servers?search=kubernetes&limit=100

But that will be a different issue 😓

PR #931, opened 2026-03-18
feat(elicitation): add confirmation rules system

Cali0707:
This PR adds a set of confirmation rules that can be added to the server config to allow admins to decide when/how HITL confirmation is presented to users of the MCP server


docs/configuration.md
```diff
@@ -483,6 +484,70 @@ For detailed information about the validation flow, error codes, and behavior, s
...
 ```
 
+### Confirmation Rules
+
+Prompt users for confirmation before dangerous actions. Rules operate at two levels:
+
+- **Tool-level** — matches on tool name, input arguments, or `DestructiveHint` annotation. Fires once before the tool handler runs.
+- **Kube-level** — matches on Kubernetes API verb, kind, group, version, or namespace. Fires per API call during handler execution.
+
+When a client doesn't support elicitation, the `fallback` determines behavior: `"allow"` proceeds silently, `"deny"` blocks the action. The global default is set via `confirmation_fallback` (defaults to `"allow"`). Per-rule `fallback` overrides the global default.
```
manusa: I think this is aligned with one of the initially discussed concerns about clients (MCP Hosts) that don't support  elicitation. 👍

docs/configuration.md
```diff
@@ -483,6 +484,70 @@ For detailed information about the validation flow, error codes, and behavior, s
...
+
+If multiple rules match at the same level, their messages are merged into a single prompt, and the most restrictive fallback wins (`"deny"` beats `"allow"`).
```
manusa: Do we really need granular fallback?
This might end up being confusing, and might backfire with some edge cases.
I mean, the rule fallback seems nice, but I'm not sure the UX complication and increased allow/deny logic is a good tradeoff.

docs/configuration.md
```diff
@@ -483,6 +484,70 @@ For detailed information about the validation flow, error codes, and behavior, s
...
+| `input` | map | tool | Key-value pairs that must match tool arguments |
+| `verb` | string | kube | Kubernetes verb (`"get"`, `"delete"`, `"list"`, etc.) |
+| `kind` | string | kube | Resource kind (`"Secret"`, `"Deployment"`, etc.) |
+| `group` | string | kube | API group (`"apps"`, `""` for core, etc.) |
+| `version` | string | kube | API version (`"v1"`, `"v1beta1"`, etc.) |
+| `namespace` | string | both | Namespace to match |
+| `message` | string | both | Message shown in the confirmation prompt |
+| `fallback` | string | both | Per-rule fallback override: `"allow"` or `"deny"` |
+
+A rule must not mix tool-level fields (`tool`, `destructive`) with kube-level fields (`verb`, `kind`, `group`, `version`).
```
manusa: Haven't checked yet, but this should probably be validated at server startup

docs/configuration.md
```diff
@@ -483,6 +484,70 @@ For detailed information about the validation flow, error codes, and behavior, s
...
+When a client doesn't support elicitation, the `fallback` determines behavior: `"allow"` proceeds silently, `"deny"` blocks the action. The global default is set via `confirmation_fallback` (defaults to `"allow"`). Per-rule `fallback` overrides the global default.
+
+If multiple rules match at the same level, their messages are merged into a single prompt, and the most restrictive fallback wins (`"deny"` beats `"allow"`).
+
+| Field | Type | Level | Description |
+|-------|------|-------|-------------|
+| `confirmation_fallback` | string | global | Default fallback: `"allow"` or `"deny"` (default: `"allow"`) |
+| `tool` | string | tool | Tool name to match (e.g. `"helm_uninstall"`) |
+| `destructive` | boolean | tool | Match tools with `DestructiveHint` annotation |
+| `input` | map | tool | Key-value pairs that must match tool arguments |
```
manusa: I'm not sure if we're making this far too flexible.
Just like with fallback, the number of permutations and combinations can make the resolution logic brittle.
Also very unclear for users.

IMO we should start with a simple feature (e.g. tool name matching or even destructive might be just enough for most use-cases, while input key pairs might be way too much at the moment)

pkg/api/confirmation.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// Validate checks that the rule is well-formed.
+// A rule must not mix tool-level fields (tool, destructive) with kube-level fields (verb, kind, group, version).
+// The namespace field is allowed alongside tool-level fields because it also appears in tool input matching.
+func (r *ConfirmationRule) Validate() error {
+	if r.IsToolLevel() && r.IsKubeLevel() {
+		return fmt.Errorf("confirmation rule mixes tool-level fields (tool, destructive) with kube-level fields (verb, kind, group, version): %q", r.Message)
+	}
+	return nil
```
manusa: In line with what's stated about UX, 
What if `!r.IsToolLevel() && !r.IsKubeLevel()`? 
For example a user that inputs as Confirmation rule a an input key-value pair assuming that this is applicable to any tool

pkg/config/config.go
```diff
@@ -308,6 +317,17 @@ func ReadToml(configData []byte, opts ...ReadConfigOpt) (*StaticConfig, error) {
 		return nil, err
 	}
 
+	var ruleErrors []error
+	for i := range config.ConfirmationRules {
```
manusa, line 324: I think this answers one of my previous questions:
https://github.com/containers/kubernetes-mcp-server/pull/931/changes#r2982520304

Marking it as resolved (once I submit the review).

pkg/kubernetes/accesscontrol_round_tripper.go
```diff
@@ -100,9 +105,8 @@ func (rt *AccessControlRoundTripper) RoundTrip(req *http.Request) (*http.Respons
 		return nil, fmt.Errorf("resource not allowed: %s", gvk.String())
 	}
 
-	// Skip validators if disabled or if this is SelfSubjectAccessReview (used by RBAC validator)
-	skipValidation := !rt.validationEnabled || (gvr.Group == "authorization.k8s.io" && gvr.Resource == "selfsubjectaccessreviews")
-	if skipValidation {
+	// Skip validators for SelfSubjectAccessReview to avoid recursion from RBAC validator
+	if gvr.Group == "authorization.k8s.io" && gvr.Resource == "selfsubjectaccessreviews" {
```
manusa, line 109: I can't seem to find where the `!rt.validationEnabled` logic has been moved. I think there was a test covering it so it should be fine, but still I don't see where this is addressed now.

pkg/mcp/confirmation_test.go
```diff
@@ -0,0 +1,187 @@
...
+	s.InitMcpClient(test.WithElicitationHandler(
+		func(_ context.Context, _ *mcp.ElicitRequest) (*mcp.ElicitResult, error) {
+			return &mcp.ElicitResult{Action: "decline"}, nil
+		},
+	))
+	result, err := s.CallTool("pods_list", map[string]any{})
+	s.Run("returns error content", func() {
+		s.NoError(err)
+		s.Require().NotNil(result)
+		s.True(result.IsError)
```
manusa, line 75: we should assert the elicitation specific error too.

manusa: This should be applicable to all error assertions

pkg/mcp/confirmation_test.go
```diff
@@ -0,0 +1,187 @@
...
+		s.Require().NotNil(result)
+		s.False(result.IsError)
+	})
+}
+
+func (s *ConfirmationRulesSuite) TestToolRuleWithInputMatch() {
+	s.Cfg.ConfirmationFallback = "deny"
+	s.Cfg.ConfirmationRules = []api.ConfirmationRule{
+		{Tool: "pods_list", Input: map[string]any{"namespace": "kube-system"}, Message: "Listing kube-system pods."},
+	}
```
manusa: It'd be nice to use regular TOML syntax (like in other tests).
It's cheap and it helps understand how the real-world config would look like.

pkg/mcp/gosdk.go
```diff
@@ -39,6 +40,14 @@ func ServerToolToGoSdkTool(s *Server, tool api.ServerTool) (*mcp.Tool, mcp.ToolH
...
 			return nil, fmt.Errorf("%v for tool %s", err, tool.Tool.Name)
 		}
+		// Check confirmation rules before executing the tool
+		if confirmErr := confirmation.CheckToolRules(
+			ctx, s.configuration, &sessionElicitor{},
+			tool.Tool.Name, toolCallRequest.GetArguments(), tool.Tool.Annotations.DestructiveHint,
+		); confirmErr != nil {
+			return NewTextResult("", confirmErr), nil
+		}
+
```
manusa, line 50: I'm not sure if it's feasible to move this into a middleware (or maybe a mutator) so that the conversion function just deals with conversion and doesn't get cluttered with other concerns

pkg/config/config.go
```diff
@@ -102,6 +104,13 @@ type StaticConfig struct {
 	// Defaults to false.
 	ValidationEnabled bool `toml:"validation_enabled,omitempty"`
 
+	// ConfirmationFallback is the global default fallback behavior when a client
+	// does not support elicitation. Valid values are "deny" and "allow".
+	// Defaults to "deny".
```
manusa:
```suggestion
	// Defaults to "allow".
```

Maybe we should just remove this line since the defaults are defined elsewhere and can be changed downstream

pkg/api/confirmation.go
manusa: We should also validate that if `IsToolLevel`, only tool-specific configs are created.
Same for `IsKubeLevel`, only kube-level-specific configs should be handled.

Imagine a user that sets the following config thinking that Namespace can apply:
```toml
[[confirmation_rules]]
tool = "resources_delete"
namespace = "kube-system"
message = "Dangerous!"
```

docs/configuration.md
```diff
@@ -483,6 +484,70 @@ For detailed information about the validation flow, error codes, and behavior, s
...
+| `verb` | string | kube | Kubernetes verb (`"get"`, `"delete"`, `"list"`, etc.) |
+| `kind` | string | kube | Resource kind (`"Secret"`, `"Deployment"`, etc.) |
+| `group` | string | kube | API group (`"apps"`, `""` for core, etc.) |
+| `version` | string | kube | API version (`"v1"`, `"v1beta1"`, etc.) |
+| `namespace` | string | both | Namespace to match |
```
manusa: I'm not sure (by reading the code) that `both` is respected.

docs/configuration.md
```diff
@@ -483,6 +484,70 @@ For detailed information about the validation flow, error codes, and behavior, s
...
+- **Tool-level** — matches on tool name, input arguments, or `DestructiveHint` annotation. Fires once before the tool handler runs.
+- **Kube-level** — matches on Kubernetes API verb, kind, group, version, or namespace. Fires per API call during handler execution.
+
+When a client doesn't support elicitation, the `fallback` determines behavior: `"allow"` proceeds silently, `"deny"` blocks the action. The global default is set via `confirmation_fallback` (defaults to `"allow"`). Per-rule `fallback` overrides the global default.
+
+If multiple rules match at the same level, their messages are merged into a single prompt, and the most restrictive fallback wins (`"deny"` beats `"allow"`).
+
+| Field | Type | Level | Description |
+|-------|------|-------|-------------|
+| `confirmation_fallback` | string | global | Default fallback: `"allow"` or `"deny"` (default: `"allow"`) |
```
manusa, line 504: Maybe we want to add a warning in the logs in case we detect that a client doesn't support elicitations and allow is set as fallback.

docs/configuration.md
Cali0707: This is a good point, I think for now we can remove this and if a user needs these fallbacks and requests it in the future we can re-add them

docs/configuration.md
Cali0707: yeah that makes sense

docs/configuration.md
Cali0707: I think for now I will remove the input key pairs, also since LLMs can probably find ways to work around those.

manusa:
@Cali0707 kind reminder. (this was almost ready to merge, just want to confirm you're aware this was missing a few minor things)

pkg/kubernetes/accesscontrol_round_tripper.go
```diff
@@ -53,14 +53,19 @@ func NewAccessControlRoundTripper(cfg AccessControlRoundTripperConfig) *AccessCo
...
+	// Schema/RBAC validators run first so the user isn't prompted for
+	// confirmation on an operation that would fail anyway.
 	if cfg.ValidationEnabled {
-		rt.validators = CreateValidators(ValidatorProviders{
+		rt.validators = append(rt.validators, CreateValidators(ValidatorProviders{
 			Discovery:  cfg.DiscoveryProvider,
 			AuthClient: cfg.AuthClientProvider,
-		})
+		})...)
+	}
```
Cali0707, line 65: @manusa the rt.validationEnabled got moved here - basically we just don't append the validators if the validation is not enabled, so there is no need for the redundant check at runtime - iterating over the validators is a noop (unless there is also a confirmation rule that applies)

pkg/mcp/gosdk.go
Cali0707: It is possible, but I am not sure if it is worth it - it's not clear to me how we could get the annotations (in particular the destructive annotation) into the middleware without building a map in server.go and then keeping that updated with tool changes. IMO that is more complex/messy than just keeping this here for now.

Maybe we can revisit in the future if there are better ideas on how to decouple this?

pkg/api/confirmation.go
Cali0707: added a validation check for that, thx!

pkg/kubernetes/accesscontrol_round_tripper.go
Cali0707: https://github.com/containers/kubernetes-mcp-server/pull/931#discussion_r3009685547

PR #942, opened 2026-03-20
refactor(mcp): clean up multi-target naming and tests after PR #843

manusa:
## Summary

Follow-up cleanup to #843 addressing naming inconsistencies, missing documentation,
and redundant tests left over from the enum removal.

- Rename `isMultiCluster` parameter to `isMultiTarget` for consistency with the
  `Provider.IsMultiTarget()` interface method
- Add godoc to `IsMultiTarget()` clarifying its relationship to `GetTargets` and
  intended implementation contract
- Remove `toolsets-full-tools-multicluster-enum.json` testdata and
  `TestDefaultToolsetsToolsInMultiClusterEnum` (identical to the multicluster test
  after enum removal in #843)
- Fix `TestClusterAwareTool` to use mutator return value instead of relying on
  shared pointer side effects
- Remove redundant `TestClusterAwareToolMultipleClusters` (identical to
  `TestClusterAwareTool` after enum removal)


pkg/mcp/tool_filter.go
```diff
@@ -20,12 +20,12 @@ func CompositeFilter(filters ...ToolFilter) ToolFilter {
 	}
 }
 
-func ShouldIncludeTargetListTool(targetName string, isMultiCluster bool) ToolFilter {
+func ShouldIncludeTargetListTool(targetName string, isMultiTarget bool) ToolFilter {
```
matzew, line 23: 👍

PR #949, opened 2026-03-21
feat: add make target to install claude-agent-acp

### LINKED ISSUE #947: Add/update make target to install deps for mcpchecker agents
Currently `make mcpchecker` only installs mcpchecker itself. However for some flows, users will also need the `claude-agent-acp` to be installed (if they want to use that agent).

We should support installing that as well

Deepak8858:
This PR updates the `mcpchecker` target to also install the `claude-agent-acp` binary, which is required for some evaluation flows.

Fixes #947

### DIFF
--- a/build/evals.mk
+++ b/build/evals.mk
@@ -7,6 +7,9 @@ MCP_CONFIG_DIR ?= dev/config/mcp-configs
 
 MCPCHECKER = $(shell pwd)/_output/tools/bin/mcpchecker
 MCPCHECKER_VERSION ?= latest
+CLAUDE_AGENT_ACP = $(shell pwd)/_output/tools/bin/claude-agent-acp
+CLAUDE_AGENT_ACP_VERSION ?= latest
+
 EVAL_CONFIG ?= evals/openai-agent/eval.yaml
 EVAL_LABEL_SELECTOR ?= suite=kubernetes
 EVAL_TASK_FILTER ?=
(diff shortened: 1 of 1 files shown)


build/evals.mk
```diff
@@ -7,21 +7,33 @@ MCP_CONFIG_DIR ?= dev/config/mcp-configs
...
 		GOBIN=$(shell dirname $(MCPCHECKER)) go install github.com/mcpchecker/mcpchecker/cmd/mcpchecker@$(MCPCHECKER_VERSION) ;\
 	}
 
+.PHONY: claude-agent-acp
+claude-agent-acp:
+	@[ -f $(CLAUDE_AGENT_ACP) ] || { \
+		set -e ;\
+		echo "Installing claude-agent-acp $(CLAUDE_AGENT_ACP_VERSION) to $(CLAUDE_AGENT_ACP)..." ;\
+		mkdir -p $(shell dirname $(CLAUDE_AGENT_ACP)) ;\
+		GOBIN=$(shell dirname $(CLAUDE_AGENT_ACP)) go install github.com/mcpchecker/claude-agent-acp/cmd/claude-agent-acp@$(CLAUDE_AGENT_ACP_VERSION) ;\
```
Cali0707, line 34: This is not the correct claude-agent-acp (and also does not exist).

The real one is here: https://github.com/zed-industries/claude-agent-acp

In the future, if you are using an LLM to help contribute I would recommend double checking the output

Cali0707:
Closing due to lack of activity here

PR #950, opened 2026-03-23
feat(helm): validate chart references and add allowed_registries config

matzew:
Block SSRF via helm_install by validating chart references before passing them to the Helm SDK.

  - Block `file://`, `http://`, and unknown URL schemes; allow only `oci://` and `https://`
  - Add optional `[toolset_configs.helm] allowed_registries` to restrict permitted registries
  - Use path-boundary prefix matching to prevent org-squatting bypasses (e.g. `myorg-evil` matching `myorg`)
  - Document accepted risk for bare filesystem paths when no allowlist is configured

### DIFF
--- a/pkg/helm/config.go
+++ b/pkg/helm/config.go
@@ -22,7 +22,7 @@ func (c *Config) Validate() error {
 	if c == nil {
 		return fmt.Errorf("helm config is nil")
 	}
-	for _, entry := range c.AllowedRegistries {
+	for i, entry := range c.AllowedRegistries {
 		u, err := url.Parse(entry)
 		if err != nil || u.Scheme == "" {
 			return fmt.Errorf("allowed_registries entry %q must be a valid URL with scheme and host", entry)
@@ -34,6 +34,7 @@ func (c *Config) Validate() error {
 		if u.Host == "" {
 			return fmt.Errorf("allowed_registries entry %q must be a valid URL with scheme and host", entry)
 		}
+		c.AllowedRegistries[i] = strings.TrimRight(entry, "/")
 	}
 	return nil
 }
--- a/pkg/helm/config_test.go
+++ b/pkg/helm/config_test.go
@@ -30,6 +30,17 @@ func (s *ConfigSuite) TestValidate() {
 		var cfg *Config
 		s.Error(cfg.Validate())
 	})
+	s.Run("trims trailing slashes from allowed registries", func() {
+		cfg := &Config{
+			AllowedRegistries: []string{
+				"oci://ghcr.io/myorg/",
+				"https://charts.example.com/",
+			},
+		}
+		s.NoError(cfg.Validate())
+		s.Equal("oci://ghcr.io/myorg", cfg.AllowedRegistries[0])
+		s.Equal("https://charts.example.com", cfg.AllowedRegistries[1])
+	})
 	s.Run("rejects entry without scheme", func() {
 		cfg := &Config{AllowedRegistries: []string{"ghcr.io/myorg"}}
 		err := cfg.Validate()


pkg/helm/helm.go
```diff
@@ -119,6 +124,40 @@ func (h *Helm) newAction(namespace string, allNamespaces bool) (*action.Configur
...
 }
 
+// validateChartReference blocks chart references using dangerous URL schemes.
+// Only oci:// and https:// URLs are allowed. Non-URL references (e.g. "stable/grafana")
+// are permitted as they resolve through Helm's local repo configuration.
+// When a Config with AllowedRegistries is provided, URL-based chart references
+// must prefix-match an entry in the allowlist, and non-URL references are rejected.
+func validateChartReference(chart string, cfg *Config) error {
+	u, err := url.Parse(chart)
+	if err != nil || u.Scheme == "" {
```
manusa: Windows CI is failing because of this line. On Windows, `filepath.Join` produces paths with drive letters like `D:\a\kubernetes-mcp-server\...\testdata\helm-chart-no-op`. Go's `url.Parse("D:\\a\\...")` interprets `D:` as a URL **scheme**, so `u.Scheme` is `"D"` (not empty). After `strings.ToLower` it becomes `"d"`, which doesn't match any allowed scheme and falls into the `default` case → blocked.

This isn't just a test issue — it would also affect real Windows users installing charts from local paths.

Fix: single-character "schemes" are actually Windows drive letters, not URL schemes. Consider:
```go
if err != nil || u.Scheme == "" || len(u.Scheme) == 1 {
```

matzew: @manusa Thanks! I added some fix/test for that - let's see what it says

pkg/helm/helm.go
```diff
@@ -119,6 +124,40 @@ func (h *Helm) newAction(namespace string, allNamespaces bool) (*action.Configur
...
+		// Non-URL references (e.g. "stable/grafana", local paths, Windows drive letters like D:\...)
+		if cfg != nil && len(cfg.AllowedRegistries) > 0 {
+			return fmt.Errorf("chart reference %q is not allowed: only registry URLs from the allowed list are permitted when allowed_registries is configured", chart)
+		}
+		return nil
+	}
+	switch strings.ToLower(u.Scheme) {
+	case "oci", "https":
+		if cfg != nil && len(cfg.AllowedRegistries) > 0 {
+			for _, allowed := range cfg.AllowedRegistries {
```
manusa, line 146: **Path traversal bypass on allowlist check:**

The prefix match operates on the raw chart string, which means `oci://ghcr.io/myorg/../evil-corp/chart` passes the prefix check against `oci://ghcr.io/my-org` (since the string starts with `oci://ghcr.io/my-org/`), but after path normalization by the OCI/HTTP client, it resolves to `ghcr.io/evil-corp/chart`.

Consider normalizing the URL before comparison:

```golang
normalized := u.Scheme + "://" + u.Host + path.Clean(u.Path)
for _, allowed := range cfg.AllowedRegistries {
    if normalized == allowed || strings.HasPrefix(normalized, allowed+"/") {
        return nil
    }
}
```

(plus test)

pkg/helm/config.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	return nil
+}
+
+func helmToolsetParser(_ context.Context, primitive toml.Primitive, md toml.MetaData) (api.ExtendedConfig, error) {
+	var cfg Config
+	if err := md.PrimitiveDecode(primitive, &cfg); err != nil {
+		return nil, err
+	}
+	return &cfg, nil
```
manusa, line 47: If user configures allowed_regitries with entries with a trailing slash (e.g. `oci://ghcr.io/my-org/`), the matcher in the helm.go validation won't work since it will test with a double slash suffix (`+ "/"+)

We can either mutate the config here to remove the slash:
```golang
for i, entry := range cfg.AllowedRegistries {
 cfg.AllowedRegistries[i] = strings.TrimRight(entry, "/")
}
```
or fix elsewhere, but this case should be considered.

PR #953, opened 2026-03-23
feat(auth): add Microsoft Entra ID support with On-Behalf-Of token exchange

nader-ziada:
Add support for Microsoft Entra ID (Azure AD) as an OIDC provider

Changes:
 - Add `entra-obo` token exchange strategy for On-Behalf-Of flow
 - Implement well-known endpoint fallback for providers without oauth-authorization-server (falls back to openid-configuration)
 - Add a new config option `cluster_auth_mode` that has two values: `passthrough` and `kubeconfig`
    - Add auto-detection: passthrough when require_oauth=true, else kubeconfig
    - Passthrough mode now automatically exchanges tokens if configured

### DIFF
--- a/pkg/http/wellknown.go
+++ b/pkg/http/wellknown.go
@@ -7,12 +7,16 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/containers/kubernetes-mcp-server/pkg/config"
+	"github.com/containers/kubernetes-mcp-server/pkg/oauth"
 	"k8s.io/klog/v2"
 )
 
 const maxWellKnownResponseSize = 1 << 20 // 1 MB
+const oidcConfigCacheTTL = 5 * time.Minute
 
 var allowedResponseHeaders = map[string]bool{
 	"Cache-Control": true,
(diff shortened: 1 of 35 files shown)


pkg/tokenexchange/assertion.go
```diff
@@ -0,0 +1,184 @@
...
+
+	return cert, privateKey, nil
+}
+
+// computeX5T computes the x5t (X.509 certificate SHA-1 thumbprint) header value
+func computeX5T(cert *x509.Certificate) string {
+	thumbprint := sha1.Sum(cert.Raw)
+	return base64.RawURLEncoding.EncodeToString(thumbprint[:])
+}
+
```
Cali0707, line 95: do we not want to support `x5t#S256` so that we don't need to use SHA 1 (or maybe at least comment why we need to use a cryptographically weak algorithm here)?

My understanding from a quick google is that entra should support `x5t#S256`, but haven't played around with configuring it myself

pkg/http/wellknown.go
```diff
@@ -115,10 +200,143 @@ func (w WellKnown) ServeHTTP(writer http.ResponseWriter, request *http.Request)
...
+
+	// MCP server URL - where OAuth metadata can be fetched
+	mcpServerURL := w.buildResourceURL(request)
+	return w.metadataGenerator.GenerateProtectedResourceMetadata(oidcConfig, mcpServerURL), nil
+}
+
+// buildResourceURL constructs the canonical resource URL for the MCP server.
+// Uses server_url from config when set, otherwise infers from request headers.
+// SECURITY: When server_url is not configured, this trusts X-Forwarded-Host and
+// X-Forwarded-Proto headers. The server MUST be behind a trusted reverse proxy.
```
Cali0707: @nader-ziada can we make this fallback opt-in in the config? And perhaps log warnings when users do opt in?

IMO we should probably error out if users don't opt in with some warning, just so that people don't accidentally run this in an insecure setup

pkg/kubernetes/token_exchange.go
```diff
@@ -12,57 +13,118 @@ import (
...
+			exchangedCtx, err := strategyBasedTokenExchange(ctx, baseConfig, oidcProvider, httpClient, token, strategy, stsConfig)
+			if err != nil {
+				return ctx, fmt.Errorf("strategy-based token exchange failed: %w", err)
+			}
+			if auth, ok := exchangedCtx.Value(OAuthAuthorizationHeader).(string); ok && strings.HasPrefix(auth, "Bearer ") {
+				exchangedToken = strings.TrimPrefix(auth, "Bearer ")
+			}
+		} else {
+			sts := NewFromConfig(baseConfig, oidcProvider)
+			if sts.IsEnabled() {
```
Cali0707: this `sts := NewFromConfig(...)` and the following `IsEnabled()` check is also on line 125

Is there some way we can simplify the logic to create the sts instance once and then use that where needed throughout this flow?

pkg/tokenexchange/config.go
```diff
@@ -43,21 +46,47 @@ type TargetTokenExchangeConfig struct {
...
+			return fmt.Errorf("client_key_file is required when auth_style is %q", AuthStyleAssertion)
+		}
+	default:
+		return fmt.Errorf("invalid auth_style %q: must be %q, %q, or %q", c.AuthStyle, AuthStyleParams, AuthStyleHeader, AuthStyleAssertion)
 	}
 	return nil
 }
 
-func (c *TargetTokenExchangeConfig) HTTPCLient() (*http.Client, error) {
+func (c *TargetTokenExchangeConfig) HTTPClient() (*http.Client, error) {
```
Cali0707, line 91: Can be fixed in a separate PR (not related to these changes) but want to flag here while I am reviewing that we should add a mutex, as otherwise this may create more than one http client depending on access timing, instead of caching correctly

pkg/tokenexchange/assertion.go
```diff
@@ -0,0 +1,184 @@
...
+		if signer, ok := key.(crypto.Signer); ok {
+			privateKey = signer
+		} else {
+			return nil, nil, fmt.Errorf("private key from %q does not implement crypto.Signer", keyFile)
+		}
+	} else if key, err := x509.ParsePKCS1PrivateKey(keyBlock.Bytes); err == nil {
+		// RSA PKCS1 format
+		privateKey = key
+	} else {
+		return nil, nil, fmt.Errorf("failed to parse private key from %q: unsupported format (only RSA keys are supported)", keyFile)
```
Cali0707: this error message seems like it could be a little misleading - maybe we can fall through and have this fail at the next assertion? IMO that error message is much clearer (and essentially what went wrong in this case anyways)

pkg/kubernetes/manager.go
```diff
@@ -116,10 +116,11 @@ func NewManager(config api.BaseConfig, restConfig *rest.Config, clientCmdConfig
 func (m *Manager) Derived(ctx context.Context) (*Kubernetes, error) {
 	authorization, ok := ctx.Value(OAuthAuthorizationHeader).(string)
 	if !ok || !strings.HasPrefix(authorization, "Bearer ") {
-		if m.config.IsRequireOAuth() {
-			return nil, errors.New("oauth token required")
+		// Use kubeconfig credentials if explicitly configured or if OAuth is not required
+		if m.config.GetClusterAuthMode() == api.ClusterAuthKubeconfig || !m.config.IsRequireOAuth() {
```
Cali0707: I get why we want a kubeconfig auth for the entra case, but I'm a little concerned about how the auth is getting spread out here

Also as far as I can tell the idea here is to use kubeconfig credentials if explicitly required or if OAuth is not required (from the comment) but that only applies when there is no auth header set? What should happen if the user wants kubeconfig auth but also has an auth header set? I think we should think through the config options/combinations of scenarios here

pkg/tokenexchange/assertion.go
nader-ziada: Switched JWT client assertion thumbprint from SHA-1 to SHA-256 which is supported by EntraID

pkg/http/wellknown.go
nader-ziada: added `trust_proxy_headers` to config, X-Forwarded-Host, X-Forwarded-Proto, X-Forwarded-For, and X-Real-IP headers are now only trusted when trust_proxy_headers = true. Without it, the server uses request.Host directly which makes it backward compatible

pkg/kubernetes/token_exchange.go
nader-ziada: simplified to code to get the sts once

pkg/tokenexchange/config.go
nader-ziada: added mutex

pkg/tokenexchange/assertion.go
nader-ziada: removed this from the `else` so the error message makes more sense

pkg/kubernetes/manager.go
nader-ziada: I did a small design change to centralize the check for the ClusterAuthMode and reject un-valid combinations, also simplified the check

matzew:
Positives:

  - Well-structured config validation in ValidateClusterAuthMode() that catches contradictory settings early (passthrough without OAuth, kubeconfig with token exchange, token exchange without OAuth) — all rejected at startup.
  - Good test coverage: TokenExchangeRoutingSuite covers auto-detect, explicit modes, and token behavior. Assertion tests verify SHA-256 headers, lifetime defaults, and caching. TestMetadataGenerationFallback covers the core 404-fallback path.
  - The trust_proxy_headers default-off approach is the right security posture.
  - ResolveClusterAuthMode() with auto-detect is clean and predictable — explicit config wins, otherwise derived from require_oauth.
  - oauth.State with atomic.Pointer[Snapshot] gives lock-free concurrent reads with snapshot diffing to detect what changed on SIGHUP reload.
  - The 503-line ENTRA_ID_SETUP.md is thorough — covers three deployment patterns (basic, ServiceAccount, OBO), troubleshooting, architecture diagram, and the two-app-registration scenario.
  - All 6 review comments from @Cali0707 have been addressed in the force-pushed commit (SHA-256 migration, trust_proxy_headers, STS deduplication, HTTPClient mutex, error messages, centralized auth mode).

  Minor observations (non-blocking):

  1. OIDC config cache in fetchOpenIDConfiguration uses RLock/check/RUnlock then fetch then Lock/write/Unlock. Two concurrent cache misses will both fetch and write. Functionally harmless (same data, last write wins) but worth being aware of under load.
  2. stsConfig parameter threading through ExchangeTokenInContext — the "pass nil to build fresh" pattern pushes caching responsibility to callers. Works fine with the mutex + URL check in tokenExchangingProvider, just an unusual API surface.
  3. 2500+ lines across 35 files is a larger PR — just noting for awareness.

PR #954, opened 2026-03-23
feat(security): add require_tls config to enforce HTTPS connections

nader-ziada:
Add `require_tls` configuration option that enforces TLS for all server and client connections when enabled

Changes:
 - All TLS configs now enforce MinVersion TLS 1.2
 -  if `require_tls` is true
     - Server refuses to start without TLS certificates configured if `require_tls` is true
     - Outbound URLs (Kiali, authorization_url) must use HTTPS

### DIFF
--- a/pkg/kiali/config.go
+++ b/pkg/kiali/config.go
@@ -61,10 +61,13 @@ func kialiToolsetParser(ctx context.Context, primitive toml.Primitive, md toml.M
 		// If it's already absolute or configDir is empty, use as-is
 	}
 
-	// Validate URL scheme when require_tls is enabled
+	// Validate TLS settings when require_tls is enabled
 	if config.RequireTLSFromContext(ctx) {
 		if err := config.ValidateURLRequiresTLS(cfg.Url, "Kiali URL"); err != nil {
 			return nil, err
+		}
+		if cfg.Insecure {
+			return nil, errors.New("require_tls is enabled but Kiali insecure=true disables certificate verification")
 		}
 	}
 
--- a/pkg/kiali/config_test.go
+++ b/pkg/kiali/config_test.go
@@ -81,6 +81,40 @@ func (s *ConfigSuite) TestConfigParser_RejectsInvalidFile() {
 	s.Nil(cfg, "Config should be nil when validation fails")
 }
 
+func (s *ConfigSuite) TestConfigParser_RejectsInsecureWithRequireTLS() {
+	caFileForTOML := filepath.ToSlash(s.caFile)
+
+	_, err := config.ReadToml([]byte(`
+		require_tls = true
+		[toolset_configs.kiali]
+		url = "https://kiali.example/"
+		insecure = true
+		certificate_authority = "` + caFileForTOML + `"
+	`))
+
+	s.Require().Error(err)
+	s.Contains(err.Error(), "insecure=true disables certificate verification")
+}
+
+func (s *ConfigSuite) TestConfigParser_AllowsSecureWithRequireTLS() {
+	caFileForTOML := filepath.ToSlash(s.caFile)
+
+	cfg, err := config.ReadToml([]byte(`
+		require_tls = true
+		[toolset_configs.kiali]
+		url = "https://kiali.example/"
+		insecure = false
+		certificate_authority = "` + caFileForTOML + `"
+	`))
+
+	s.Require().NoError(err)
+	kialiCfg, ok := cfg.GetToolsetConfig("kiali")
+	s.Require().True(ok)
+	kcfg, ok := kialiCfg.(*Config)
+	s.Require().True(ok)
+	s.False(kcfg.Insecure)
+}
+
 func TestConfig(t *testing.T) {
 	suite.Run(t, new(ConfigSuite))
 }
--- a/pkg/kiali/kiali_test.go
+++ b/pkg/kiali/kiali_test.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
@@ -71,11 +73,18 @@ func (s *KialiSuite) TestRequireTLS_ConfigValidation() {
 	})
 
 	s.Run("accepts HTTPS URL when require_tls is enabled", func() {
+		tempDir := s.T().TempDir()
+		caFile := filepath.Join(tempDir, "ca.crt")
+		err := os.WriteFile(caFile, []byte("test ca content"), 0644)
+		s.Require().NoError(err)
+		caFileForTOML := filepath.ToSlash(caFile)
+
 		cfg, err := config.ReadToml([]byte(`
 			require_tls = true
 			[toolset_configs.kiali]
 			url = "https://kiali.example/"
-			insecure = true
+			insecure = false
+			certificate_authority = "` + caFileForTOML + `"
 		`))
 		s.Require().NoError(err)
 		s.NotNil(cfg)


pkg/kiali/config.go
```diff
@@ -61,6 +61,13 @@ func kialiToolsetParser(ctx context.Context, primitive toml.Primitive, md toml.M
 		// If it's already absolute or configDir is empty, use as-is
 	}
 
+	// Validate URL scheme when require_tls is enabled
+	if config.RequireTLSFromContext(ctx) {
+		if err := config.ValidateURLRequiresTLS(cfg.Url, "Kiali URL"); err != nil {
+			return nil, err
+		}
+	}
```
Cali0707: will remembering to check any outgoing url on contributed toolsets be a requirement going forwards now?

pkg/kubernetes-mcp-server/cmd/root.go
```diff
@@ -330,6 +336,18 @@ func (m *MCPServerOptions) Validate() error {
...
 	}
+	// Validate require_tls configuration
+	if m.StaticConfig.RequireTLS && m.StaticConfig.Port != "" {
+		if tlsCert == "" || tlsKey == "" {
+			return fmt.Errorf("require_tls is enabled but TLS certificates are not configured (set tls_cert and tls_key)")
+		}
+	}
+	// Validate outbound URLs when require_tls is enabled
+	if m.StaticConfig.RequireTLS {
+		if err := config.ValidateURLRequiresTLS(m.StaticConfig.AuthorizationURL, "authorization_url"); err != nil {
```
Cali0707: maybe check `m.StaticConfig.ServerURL` and `m.StaticConfig.SSEBaseURL` as well? do we also need to validate the urls set within the kubeconfig file?

Also this would be more of a stretch but should we be validating the issuer urls in the token exchange config that a toolset can optionally return?

pkg/config/config.go
```diff
@@ -357,3 +362,7 @@ func (c *StaticConfig) GetStsScopes() []string {
 func (c *StaticConfig) IsValidationEnabled() bool {
 	return c.ValidationEnabled
 }
+
+func (c *StaticConfig) IsRequireTLS() bool {
+	return c.RequireTLS
+}
```
Cali0707, line 368: maybe I missed this but I didn't see this consumed anywhere - it seems like we are only use the context accessor for this?

pkg/config/tls.go
```diff
@@ -0,0 +1,22 @@
...
+// Returns nil if the URL is empty. Returns an error if the URL does not use HTTPS.
+func ValidateURLRequiresTLS(urlStr string, fieldName string) error {
+	if urlStr == "" {
+		return nil
+	}
+	u, err := url.Parse(urlStr)
+	if err != nil {
+		return fmt.Errorf("invalid %s: %w", fieldName, err)
+	}
+	if u.Scheme != "https" {
```
manusa: I'm not sure if some urls might attempt to connect to a non-http protocol such as a web socket (wss://

nader-ziada: added `wss` since the only two protocols currently used in the codebase are `http` and `ws` so didn't want any other ones

pkg/kubernetes-mcp-server/cmd/root.go
nader-ziada: added validation for `ServerURL` and `SSEBaseURL`

token exchange issuer urls from toolsets: handled at runtime, `TargetTokenExchangeConfig` has a `RequireTLS func() bool` field, and `HTTPClient()` wraps the transport with `TLSEnforcingTransport`, so any non-https request just gets blocked at runtime. no explicit url validation needed

pkg/kiali/config.go
nader-ziada: No, toolset contributors don't need to manually validate URLs. The `TLSEnforcingTransport` handles this automatically at runtime, but toolset contributors need to use the IsRequireTLS function from the config provider to wrap their HTTP client when making HTTP requests

Kiali is currently the only toolset that makes direct outbound HTTP requests to an external service

pkg/config/config.go
nader-ziada: Context accessor is needed during TOML parsing (config object doesn't exist yet), func is used after config is built for runtime HTTP client wrapping.

nader-ziada:
@Cali0707 @manusa 

comments addressed by adding two-layer tls enforcement:
- layer 1 (config-time): validates known config urls at startup/reload for fail-fast feedback
- layer 2 (runtime): TLSEnforcingTransport wraps http clients and rejects non-https at the transport level, catches all outbound calls regardless of where the url came from (dynamic, toolset-specific, whatever we might miss in layer 1)
sighup: transport uses func() bool reading from Configuration.IsRequireTLS() per-request, so reloads take effect immediately, no re-validation needed, non-https just gets blocked if `require_tls` is true

pkg/config/tls.go
```diff
@@ -0,0 +1,96 @@
...
+// This provides Layer 1 (config-time) validation for fail-fast feedback.
+func ValidateURLRequiresTLS(urlStr string, fieldName string) error {
+	if urlStr == "" {
+		return nil
+	}
+	u, err := url.Parse(urlStr)
+	if err != nil {
+		return fmt.Errorf("invalid %s: %w", fieldName, err)
+	}
+	if u.Scheme != "https" {
```
Cali0707: @nader-ziada maybe we want to use the `isSecureHTTPScheme` method here as well

pkg/config/tls.go
```diff
@@ -0,0 +1,96 @@
...
+		return fmt.Errorf("require_tls is enabled but %s uses %q scheme (HTTPS required)", fieldName, u.Scheme)
+	}
+	return nil
+}
+
+// ValidateURLsRequireTLS validates multiple URLs use HTTPS scheme.
+// The map keys are field names, values are the URLs to validate.
+func ValidateURLsRequireTLS(urls map[string]string) error {
+	for fieldName, urlStr := range urls {
```
Cali0707: maybe we want to sort the keys before iterating first and/or combine the errors seen? Since the map traversal is non deterministic this could end up in different error messages seen by the user which could be confusing IMO

pkg/tokenexchange/config.go
```diff
@@ -45,6 +47,10 @@ type TargetTokenExchangeConfig struct {
 	// "header": HTTP Basic Authentication header
 	AuthStyle string `toml:"auth_style,omitempty"`
 
+	// RequireTLS is a function that returns whether TLS is required for outbound connections.
+	// When set and returns true, HTTP requests to non-HTTPS endpoints will be rejected.
+	RequireTLS func() bool `toml:"-"`
```
Cali0707: are we setting this anywhere where we are creating these configs?

nader-ziada: this is a cli flag and a config to be added in the toml, not sure what you mean? I didn't set it be default

pkg/config/tls.go
nader-ziada: done, good catch

pkg/config/tls.go
nader-ziada: done

pkg/tokenexchange/config.go
manusa: I believe that the problem is that this field IS NOT explicitly populated in the codebase, since this is a function and non-serializable field, it's effectively useless.

pkg/mcp/mcp.go
```diff
@@ -45,6 +45,14 @@ func (c *Configuration) ListOutput() output.Output {
 	return c.listOutput
 }
 
+// IsRequireTLS returns whether TLS is required for outbound connections.
+// This method exists to support dynamic config reload via SIGHUP - it reads
+// from the current StaticConfig rather than being a promoted method bound
+// to a potentially stale config reference.
+func (c *Configuration) IsRequireTLS() bool {
+	return c.RequireTLS
+}
```
manusa: I think this is not needed since StaticConfig.IsRequireTLS should already be promoted.

pkg/config/tls.go
```diff
@@ -0,0 +1,105 @@
...
+// defense-in-depth, catching any URLs that might have been missed during config validation.
+// The RequireTLS function is called per-request, allowing dynamic config changes (e.g., SIGHUP).
+type TLSEnforcingTransport struct {
+	Base       http.RoundTripper
+	RequireTLS func() bool
+}
+
+func (t *TLSEnforcingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if t.RequireTLS != nil && t.RequireTLS() && !isSecureHTTPScheme(req.URL.Scheme) {
+		klog.V(0).Infof("require_tls: blocked request to %s", req.URL.Host)
```
manusa: I'm not sure, but maybe the log level (0) is too low and might clutter the server logs. Are we sure 0 is the right choice here?

pkg/mcp/mcp.go
nader-ziada: removed it

pkg/tokenexchange/config.go
nader-ziada: yeah removed it

pkg/config/tls.go
nader-ziada: change to v(1) to be similar to authorization.go

pkg/kiali/config.go
```diff
@@ -61,6 +61,13 @@ func kialiToolsetParser(ctx context.Context, primitive toml.Primitive, md toml.M
 		// If it's already absolute or configDir is empty, use as-is
 	}
 
+	// Validate URL scheme when require_tls is enabled
+	if config.RequireTLSFromContext(ctx) {
```
matzew, line 65: shuld htis check on ` cfg.Insecure` ?

pkg/tokenexchange/config.go
```diff
@@ -64,39 +64,31 @@ func (c *TargetTokenExchangeConfig) HTTPCLient() (*http.Client, error) {
...
+		if !caCertPool.AppendCertsFromPEM(caCert) {
+			return nil, fmt.Errorf("failed to parse CA certificate from '%s'", c.CAFile)
+		}
+
+		tlsConfig.RootCAs = caCertPool
 	}
 
+	transport.TLSClientConfig = tlsConfig
+
 	c.client = &http.Client{
```
matzew, line 88: the other outbound clients (well-known proxy, OIDC, Kiali) are wrapped with `TLSEnforcingTransport`. Should this one be wrapped too for consistency?

pkg/config/tls.go
```diff
@@ -0,0 +1,105 @@
...
+// defense-in-depth, catching any URLs that might have been missed during config validation.
+// The RequireTLS function is called per-request, allowing dynamic config changes (e.g., SIGHUP).
+type TLSEnforcingTransport struct {
+	Base       http.RoundTripper
+	RequireTLS func() bool
+}
+
+func (t *TLSEnforcingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if t.RequireTLS != nil && t.RequireTLS() && !isSecureHTTPScheme(req.URL.Scheme) {
+		klog.V(1).Infof("require_tls: blocked request to %s", req.URL.Host)
```
matzew, line 59: like here, I guess there are more areas where we could do the same thing: when backend TLS handshakes fails (e.g. expired certs, protocol mismatches etc), they show up as generic HTTP errors in the logs. 

Might be worth a follow-up to surface those more clearly for debugging.

pkg/tokenexchange/config.go
nader-ziada: I think the reason was that the TokenURL is typed directly into a config file by an admin. If they write http://internal-keycloak:8080/, that's an explicit choice, not an accident. But i can add it if you don't think that's a good reason

pkg/kiali/config.go
nader-ziada: i haven't been assuming that `require_tls` also means should reject `Insecure=true` because the admin explicitly sets `Insecure=true` knowing what it means 

also let me know if you think this is not a good assumption

matzew: Agree that admins should know, but the value of `require_tls` is catching mis-configurations. Like someone copies a dev config with `insecure = true` to production, adds `require_tls = true`, and assumes they're covered. Perhaps some startup error/validation saves those from a false sense of security?

pkg/tokenexchange/config.go
matzew: For consistency I'd wrap this one too, better safe than sorry on these layered security thing and catching those for te admins/users?

nader-ziada: looked into this more and the HTTPClient() code path is actually not used today since no provider implements TokenExchangeProvider, so the type assertion in `ExchangeTokenInContext` always fails and we never reach it

pkg/kiali/config.go
nader-ziada: added error if `insecure = true`

PR #958, opened 2026-03-24
fix(helm): normalize scheme, host, and path in allowlist matching

matzew:
Follow-up fix based on code rabbit static pattern analysis: normalize URL scheme/host case and handle `path.Clean("")` returning `"."` in allowlist matching. :rabbit: 

Related to https://github.com/containers/kubernetes-mcp-server/pull/950

### DIFF
--- a/pkg/helm/helm.go
+++ b/pkg/helm/helm.go
@@ -142,7 +142,11 @@ func validateChartReference(chart string, cfg *Config) error {
 	switch strings.ToLower(u.Scheme) {
 	case "oci", "https":
 		if cfg != nil && len(cfg.AllowedRegistries) > 0 {
-			normalized := u.Scheme + "://" + u.Host + path.Clean(u.Path)
+			cleaned := path.Clean(u.Path)
+			if cleaned == "." {
+				cleaned = ""
+			}
+			normalized := strings.ToLower(u.Scheme) + "://" + strings.ToLower(u.Host) + cleaned
 			for _, allowed := range cfg.AllowedRegistries {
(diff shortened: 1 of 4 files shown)


pkg/helm/helm.go
```diff
@@ -142,7 +142,11 @@ func validateChartReference(chart string, cfg *Config) error {
 	switch strings.ToLower(u.Scheme) {
 	case "oci", "https":
 		if cfg != nil && len(cfg.AllowedRegistries) > 0 {
-			normalized := u.Scheme + "://" + u.Host + path.Clean(u.Path)
+			cleaned := path.Clean(u.Path)
+			if cleaned == "." {
+				cleaned = ""
+			}
+			normalized := strings.ToLower(u.Scheme) + "://" + strings.ToLower(u.Host) + cleaned
 			for _, allowed := range cfg.AllowedRegistries {
```
manusa, line 150: The input chart reference is now normalized to lowercase, but the `allowed` entries from config are compared as-is. If an admin configures `AllowedRegistries` with mixed case (e.g., `"OCI://GHCR.IO/myorg"`), the match will silently fail.

Since we're already touching this code path, consider normalizing the `allowed` side too:

```go
for _, allowed := range cfg.AllowedRegistries {
    a := strings.ToLower(allowed)
    if normalized == a || strings.HasPrefix(normalized, a+"/") {
```

Alternatively, normalize once at config load time in `Config.Init()` or similar.

matzew:
@manusa updated! Thanks for taking a look

PR #959, opened 2026-03-24
fix(evals): enable core toolset for tekton eval suite

ksimon1:
fix(evals): enable core toolset for tekton eval suite

This should fix some of the eval tests in https://github.com/containers/kubernetes-mcp-server/pull/892

### DIFF
--- a/.github/workflows/mcpchecker.yaml
+++ b/.github/workflows/mcpchecker.yaml
@@ -131,19 +131,19 @@ jobs:
           case "$SUITE" in
             kubevirt)
               echo "label-selector=suite=kubevirt" >> $GITHUB_OUTPUT
-              echo "toolsets=kubevirt" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/mcpchecker.yaml
Cali0707: @ksimon1 I think we need the same here (maybe for other toolsets too)

Would you mind setting core enabled on all the different paths here?

manusa: cc @aljesusg @lyarwood @ksimon1 

Please double check if for your prescribed usage of the MCP server you're encouraging users to enable both core and kiali/kubevirt.
My understanding is the same as Calum's, but it be good to get your approval too.

ksimon1: @lyarwood already created similar PR https://github.com/containers/kubernetes-mcp-server/pull/841, so for Kubevirt and Tekton it should be ok.

PR #920, opened 2026-03-13
WIP: Add MCP Apps interactive UI extensions

manusa:
## Summary

Add MCP Apps support as an opt-in feature (`--apps` flag / `apps_enabled` config) that enables tools to return interactive HTML-based UIs rendered in sandboxed iframes by MCP hosts (VS Code, Claude Desktop, ChatGPT, etc.).

Viewer built with Preact + HTM + Chart.js + Prism.js (~230 KB, no build step), served via `embed.FS` with per-tool `ui://` resource URIs.

Closes #753

## What's done

- **Infrastructure**: Config, CLI flag, server capability negotiation, per-tool resource registration
- **Viewer**: Dual-flow (spec-compliant `tool-result` + fallback `tool-input`), SortableTable, MetricsTable (Chart.js), YamlView (Prism.js), GenericView, theming with `light-dark()` CSS fallbacks
- **Output layer**: `PrintObjStructured` on Output interface, `tableToStructured` for Kubernetes Table API, `ensureStructuredObject` array wrapping
- **Tools with structured content**: `pods_list`, `pods_list_in_namespace`, `pods_top`, `nodes_top`, `namespaces_list`, `projects_list`, `resources_list`
- **Compatibility**: VS Code pre-fetch race (resources registered before tools), both nested and flat `_meta.ui` keys
- **Tests**: Unit tests (mcpapps, output, API) + browser integration tests (Rod)

## What's pending

- Structured content for remaining non-list tool handlers (YAML tools)
- Refinement of table output rendering
- Reanalyze overall effect of structured outputs on LLM performance
- Refresh via `callServerTool()`, height management, filtering/pagination
- Edge case handling (empty results, large datasets)
- Documentation updates (`docs/configuration.md` for `apps_enabled`)

Issue #753, opened 2026-02-09
PoC: MCP Apps integration for interactive UI extensions


manusa:
## Summary

Implement support for [MCP Apps](https://modelcontextprotocol.github.io/ext-apps/api/) - the official MCP extension (spec version 2026-01-26) that enables tools to return interactive UI components instead of just text responses.

This will allow the kubernetes-mcp-server to provide rich, interactive experiences such as:
- Interactive resource tables with sorting, filtering, and actions
- Visual dashboards for cluster metrics and status
- Forms for resource creation/editing
- Topology visualizations for workloads and networking

## Background

MCP Apps extend the Model Context Protocol by letting tools declare UI resources. The architecture works as follows:

1. **Tool definition** - Tools declare a `_meta.ui.resourceUri` field pointing to a `ui://` resource
2. **Tool call** - LLM invokes the tool on the server
3. **Host renders** - Host fetches the resource and displays it in a sandboxed iframe
4. **Bidirectional communication** - Host passes tool data via notifications; UI can call other tools through the host

### References
- VS Code announcement: https://code.visualstudio.com/blogs/2026/01/26/mcp-apps-support
- Official blog post: https://blog.modelcontextprotocol.io/posts/2026-01-26-mcp-apps/
- Specification: https://modelcontextprotocol.github.io/ext-apps/api/
- SDK & Examples: https://github.com/modelcontextprotocol/ext-apps

### Phase 1: Core Infrastructure

- [ ] Extend `Tool` struct in `pkg/api/toolsets.go` to support `_meta` field with UI metadata
  ```go
  type ToolMeta struct {
      UI *ToolMetaUI `json:"ui,omitempty"`
  }
  type ToolMetaUI struct {
      ResourceUri string `json:"resourceUri"`
  }
  ```
- [ ] Add `ui://` resource scheme support to the MCP server
- [ ] Implement resource serving for UI bundles (HTML/JS)
- [ ] Add JSON output format option for structured data consumption by UIs

### Phase 2: Tool Result Enhancement

- [ ] Extend `ToolCallResult` to optionally include structured data alongside text content
- [ ] Implement JSON-RPC over postMessage bridge for Go server
- [ ] Add notification mechanism for passing tool data to UI

### Phase 3: Sample UI Implementation

- [ ] Create a sample interactive UI for one of the existing tools (e.g., `pods_list` with interactive table)
- [ ] Bundle the UI as a `ui://` resource
[...]

PR #960, opened 2026-03-24
refactor: add structured output support and MCP spec compliance

manusa:
Prerequisite refactors extracted from #920 to reduce its diff size.
These changes are self-contained improvements to the existing codebase
that lay the groundwork for MCP Apps (#753).

### Changes

- **`PrintObjStructured` on Output interface** — Extends the `Output`
  interface with a method that returns both human-readable text and
  optional structured data extracted from Kubernetes objects. Includes
  `tableToStructured` helper for column-keyed extraction from Table
  responses. Rewrites output tests to `testify/suite`.

- **`NewToolCallResultFull` constructor** — New API constructor that
  creates a `ToolCallResult` with separate text and structured content
  fields, for cases where the text representation differs from a JSON
  serialization of the structured data.

- **`ensureStructuredObject` for MCP spec compliance** — The MCP spec
  requires `structuredContent` to be a JSON object. Slices/arrays
  passed to `NewStructuredResult` are now automatically wrapped in
  `{"items": [...]}`.

Refs #753
Refs #920

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"reflect"
 	"slices"
 	"sync"
 	"time"
@@ -399,6 +400,10 @@ func NewTextResult(content string, err error) *mcp.CallToolResult {
 // The Content field contains the JSON-serialized form of structuredContent
 // for backward compatibility with MCP clients that don't support structuredContent.
 //
+// Per the MCP specification, structuredContent must marshal to a JSON object.
+// If structuredContent is a slice/array, it is automatically wrapped in
+// {"items": [...]} to satisfy this requirement.
+//
 // Per the MCP specification:
 // "For backwards compatibility, a tool that returns structured content SHOULD
 // also return the serialized JSON in a TextContent block."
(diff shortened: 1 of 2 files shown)


pkg/output/output.go
```diff
@@ -50,6 +64,23 @@ func (p *yaml) AsTable() bool {
...
+func (p *yaml) PrintObjStructured(obj runtime.Unstructured) (*PrintResult, error) {
+	text, err := p.PrintObj(obj)
+	if err != nil {
+		return nil, err
+	}
+	switch t := obj.(type) {
+	case *unstructured.UnstructuredList:
+		items := make([]map[string]any, 0, len(t.Items))
+		for _, item := range t.Items {
+			items = append(items, item.Object)
```
Cali0707: maybe we deepcopy the `item.Object`? that way any modifications to PrintResult.Structured won't also modify the Unstructured Object

pkg/output/output.go
```diff
@@ -50,6 +64,23 @@ func (p *yaml) AsTable() bool {
...
+		}
+		return &PrintResult{Text: text, Structured: items}, nil
+	case *unstructured.Unstructured:
+		return &PrintResult{Text: text, Structured: t.Object}, nil
```
Cali0707: maybe we deepcopy the item.Object? that way any modifications to PrintResult.Structured won't also modify the Unstructured Object

pkg/mcp/mcp.go
```diff
@@ -425,7 +430,19 @@ func NewStructuredResult(content string, structuredContent any, err error) *mcp.
...
 	}
 	return result
 }
+
+// ensureStructuredObject wraps slice/array values in a {"items": ...} object
+// because the MCP specification requires structuredContent to be a JSON object.
+// Note: this checks the top-level reflect.Kind, so a pointer-to-slice (*[]T)
+// would not be wrapped. All current callers pass value types.
+func ensureStructuredObject(v any) any {
+	rv := reflect.ValueOf(v)
```
Cali0707, line 444: I think we still are missing handling typed nil, e.g. `[]string(nil)`. This would still show up as a `rv.Kind() == reflect.Slice`, but would end up json serializing to `{"items": null}` which I don't think is correct

Cali0707 (commented):
Thanks for this PR @manusa !

Left a few thoughts, otherwise this looks like great prep for the MCP apps PR

PR #966, opened 2026-03-25
fix(kubernetes): guard managers with RW mutex to avoid concurrent map…

difrost:
When using a kubeconfig cluster provider a server fails with `concurrent map writes` when preparing Mangers.

This PR introduces a RW mutex to guard kubeConfigClusterProvider.managers.

### DIFF
--- a/pkg/kubernetes/provider_kubeconfig.go
+++ b/pkg/kubernetes/provider_kubeconfig.go
@@ -117,8 +117,12 @@ func (p *kubeConfigClusterProvider) managerForContext(context string) (*Manager,
 
 func (p *kubeConfigClusterProvider) IsOpenShift(ctx context.Context) bool {
 	p.mu.RLock()
-	defer p.mu.RUnlock()
-	return p.managers[p.defaultContext].IsOpenShift(ctx)
+	m := p.managers[p.defaultContext]
+	p.mu.RUnlock()
+	if m == nil {
+		return false
+	}
+	return m.IsOpenShift(ctx)
 }
 
 func (p *kubeConfigClusterProvider) IsMultiTarget() bool {


pkg/kubernetes/provider_kubeconfig.go
```diff
@@ -18,6 +19,7 @@ const KubeConfigTargetParameterName = "context"
 // Kubernetes clusters using different contexts from a kubeconfig file.
 // It lazily initializes managers for each context as they are requested.
 type kubeConfigClusterProvider struct {
+	mu                  sync.RWMutex
```
matzew, line 22: Should `reset()` be updated to also hold `p.mu.Lock()` around its writes?

pkg/kubernetes/provider_kubeconfig.go
```diff
@@ -84,12 +86,17 @@ func (p *kubeConfigClusterProvider) reset() error {
...
+	p.mu.RLock()
 	m, ok := p.managers[context]
+	defaultContext := p.defaultContext
+	p.mu.RUnlock()
 	if ok && m != nil {
 		return m, nil
 	}
 
-	baseManager := p.managers[p.defaultContext]
+	p.mu.Lock()
```
manusa, line 99: TOCTOU race: between `RUnlock()` (line 92) and this `Lock()`, another goroutine can also observe `m == nil` for the same context, win the write lock first, and create a manager. When this goroutine then acquires the lock, it creates a **duplicate** manager overwriting the first — leaking its client connections and caches.

The standard fix is double-checked locking — re-check after acquiring the write lock:

```go
p.mu.Lock()
defer p.mu.Unlock()
// Re-check: another goroutine may have created the manager while we waited
m, ok = p.managers[context]
if ok && m != nil {
    return m, nil
}
baseManager := p.managers[p.defaultContext]
```

(Note: with the re-check inside the write lock, you can read `p.defaultContext` directly here instead of caching it in the read-lock section.)

manusa (changes requested):
Thanks for the contribution @difrost! This is a real issue and `sync.RWMutex` is the right approach.

I found a couple of things that need addressing before we can merge:

1. **TOCTOU race in `managerForContext`** — Between releasing the read lock and acquiring the write lock, another goroutine can create a manager for the same context. The second goroutine then overwrites it, leaking the first manager's resources. A double-checked locking pattern (re-check inside the write lock) fixes this. See my inline comment for the suggested fix.

2. **`reset()` mutates guarded fields without holding the mutex** — `reset()` writes to both `p.managers` and `p.defaultContext` without acquiring `p.mu`, but it can be called concurrently from the kubeconfig watcher callback in `WatchTargets`. This is the same class of data race this PR is fixing. (Same point @matzew raised.)

A concurrency test exercising these paths under `go test -race` would also be great to have.

matzew:
Good point on `-race` while testing.
Didn‘t Calum run into something here ?

difrost:
Folks thanks for review. I'll address your findings today.

pkg/kubernetes/provider_kubeconfig.go
difrost: Fixed in [4af44ef](https://github.com/containers/kubernetes-mcp-server/pull/966/commits/4af44ef9a1cd0a59735c1ed370496e79b6222d88)

pkg/kubernetes/provider_kubeconfig.go
difrost: Fixed in [4af44ef](https://github.com/containers/kubernetes-mcp-server/pull/966/commits/4af44ef9a1cd0a59735c1ed370496e79b6222d88)

difrost:
> Good point on `-race` while testing. Didn‘t Calum run into something here ? Sent from Gmail Mobile
> […](#)

Added test via [8aadf46](https://github.com/containers/kubernetes-mcp-server/pull/966/commits/8aadf46815b75c58a4ca468699dbec685d5ac554). The TestConcurrentLazyManagerInit easily triggers race on a clean version but the WatchTargets does not ... adding it here for further works as I'm not sure  how to tune it.

matzew:
Thanks for the updates!

LGTM

matzew:
@difrost however, there is some formatting issue:

```
Error: Code is not formatted. Please run 'go fmt ./...' and commit the changes.
diff --git a/pkg/kubernetes/provider_kubeconfig_test.go b/pkg/kubernetes/provider_kubeconfig_test.go
index 53f778d..23ae14f 100644
--- a/pkg/kubernetes/provider_kubeconfig_test.go
+++ b/pkg/kubernetes/provider_kubeconfig_test.go
@@ -107,12 +107,12 @@ func (s *ProviderKubeconfigTestSuite) TestConcurrentReads() {
 		start := make(chan struct{})
 
 		ops := []func(){
-			func() {_, _ = s.provider.GetTargets(context.Background())},
-			func() {_ = s.provider.GetDefaultTarget()},
-			func() {_ = s.provider.IsMultiTarget()},
-			func() {_ = s.provider.IsOpenShift(context.Background())},
-			func() {_, _ = s.provider.GetDerivedKubernetes(context.Background(), "fake-context")},
-			func() {_ = s.provider.GetTargetParameterName()},
+			func() { _, _ = s.provider.GetTargets(context.Background()) },
+			func() { _ = s.provider.GetDefaultTarget() },
+			func() { _ = s.provider.IsMultiTarget() },
+			func() { _ = s.provider.IsOpenShift(context.Background()) },
+			func() { _, _ = s.provider.GetDerivedKubernetes(context.Background(), "fake-context") },
+			func() { _ = s.provider.GetTargetParameterName() },
 		}
 
 		wg.Add(goroutines)
@@ -212,7 +212,7 @@ func (s *ProviderKubeconfigTestSuite) TestWatchTargetsWithConcurrentReaders() {
 		for i := range extraContexts {
 			kubeconfig.CurrentContext = fmt.Sprintf("lazy-context-%d", i)
 			s.Require().NoError(clientcmd.WriteToFile(*kubeconfig, kubeconfigPath))
-			s.Require().NoError(waitForCallback(5 *time.Second))
+			s.Require().NoError(waitForCallback(5 * time.Second))
 		}
```

See the GH action for format checking

pkg/kubernetes/provider_kubeconfig_test.go
```diff
@@ -96,6 +100,126 @@ func (s *ProviderKubeconfigTestSuite) TestGetTargetParameterName() {
...
+		}
+		close(start)
+		wg.Wait()
+	})
+}
+
+func (s *ProviderKubeconfigTestSuite) TestWatchTargetsWithConcurrentReaders() {
+	s.Run("reset via WatchTargets does not race with concurrent reads", func() {
+		s.T().Setenv("KUBECONFIG_DEBOUNCE_WINDOW_MS", "10")
+		s.T().Setenv("CLUSTER_STATE_POLL_INGTERVAL_MS", "50")
```
manusa: There's a typo which prevents the functionality to change the poll interval from working
```suggestion
		s.T().Setenv("CLUSTER_STATE_POLL_INTERVAL_MS", "50")
```

pkg/kubernetes/provider_kubeconfig_test.go
```diff
@@ -96,6 +100,126 @@ func (s *ProviderKubeconfigTestSuite) TestGetTargetParameterName() {
...
+				for {
+					select {
+					case <-stop:
+						return
+					default:
+						_, _ = s.provider.GetTargets(context.Background())
+						_ = s.provider.GetDefaultTarget()
+						_ = s.provider.IsMultiTarget()
+						_ = s.provider.IsOpenShift(context.Background())
+						_, _ = s.provider.GetDerivedKubernetes(context.Background(), "fake-context")
```
manusa: Wrong provider, readers should be using the local provider created in the scope, the test is not exercising the race.
This test should fail before applying the fix, make sure of that.

```suggestion
						_, _ = provider.GetTargets(context.Background())
						_ = provider.GetDefaultTarget()
						_ = provider.IsMultiTarget()
						_ = provider.IsOpenShift(context.Background())
						_, _ = provider.GetDerivedKubernetes(context.Background(), "fake-context")
```

difrost: Thx @manusa - as I've wrote yesterday this was the missing part that I kind of missed yesterday when coding that late in the evening. Both issues should be addressed in [b524489](https://github.com/containers/kubernetes-mcp-server/pull/966/commits/b524489bb1e6d65718039789a25a87195b5ece06).

Cheers

difrost:
With [b524489](https://github.com/containers/kubernetes-mcp-server/pull/966/commits/b524489bb1e6d65718039789a25a87195b5ece06) the race tests are failing on the clean code while this PR is all green.

PR #968, opened 2026-03-25
feat(auth): add OAuth scope-based authorization for MCP tools

nader-ziada:
Implement scope-based access control for MCP tool calls when OAuth is enabled. Tools are authorized based on their annotations

 - read-only tools (readOnlyHint=true) require read_scope
 - write tools require write_scope
 - write scope implies read access (standard RBAC convention)

Changes:
- ScopeProvider interface for MCP layer to check JWT scopes
- Support for both 'scope' and 'scp' claims 
- Case-insensitive scope comparison for IdP variations
- Dynamic config reload via SIGHUP
- Backward compatibility: tokens without scopes allow all access

### DIFF
--- a/docs/KEYCLOAK_OIDC_SETUP.md
+++ b/docs/KEYCLOAK_OIDC_SETUP.md
@@ -152,6 +152,38 @@ The setup automatically configures:
 - **mcp:openshift**: Optional scope for token exchange with audience mapper
 - **groups**: Group membership mapper (included in tokens)
 
+### Tool Authorization Scopes (Optional)
+
+The MCP server supports scope-based access control for tools. When configured, users must have the appropriate scope in their JWT to call tools:
+
+- **read**: Required for read-only tools (list, get, view operations)
+- **write**: Required for write tools (create, update, delete operations)
+
+**To enable scope-based authorization:**
+
+1. **Create client scopes in Keycloak Admin Console:**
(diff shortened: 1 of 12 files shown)


docs/KEYCLOAK_OIDC_SETUP.md
```diff
@@ -175,6 +207,10 @@ sts_audience = "openshift"  # Triggers token exchange
 sts_scopes = ["mcp:openshift"]
 
 certificate_authority = "_output/cert-manager-ca/ca.crt"  # For HTTPS validation
+
+# Tool authorization scopes (optional - defaults shown)
```
matzew, line 211: does this work with the keyclaok setup from the makefiles? Or is no change needed to have these `scopes` ?

matzew: meaning the one behind the `make local-env-setup` targiet

nader-ziada: I didn't update the Makefile to add these scopes, but its backward compatible to work as it did before if the scopes are not defined.

do you want me to add the scopes to the local make target?

nader-ziada:
not needed as `The Kubernetes MCP server is not a traditional RESTful API. It's an MCP protocol server that proxies operations to the Kubernetes API server.`

PR #969, opened 2026-03-26
feat(kiali): refactor from v2.25 (code + evals)

aljesusg:
**Ready**

## Summary

- Refactors all Kiali MCP tools to delegate execution to Kiali's backend `/api/chat/mcp/*` endpoints instead of implementing the logic locally
- Removes ~3,000 lines of duplicated Go code that replicated Kiali-internal logic (health calculation, graph assembly, trace formatting, log retrieval, Istio config management)
- Adds comprehensive test coverage for the Kiali HTTP client and config validation

## Motivation

Kiali's UI already ships a built-in chatbot that uses the same MCP tool semantics. Maintaining two independent implementations of the same tools — one in the Kiali backend and one here in the MCP server — created several problems:

1. **Duplicated logic**: Health calculation, graph assembly, trace summarization, and Istio config management were reimplemented in this repository, diverging from Kiali's canonical behavior over time.
2. **Version skew**: Output formats and default parameters drifted between the Kiali UI chatbot and the MCP server, producing inconsistent experiences.
3. **Large tool schemas and responses**: Each tool carried verbose input schemas and produced large JSON responses because the MCP server had to expose every upstream parameter and return raw API data.
4. **Limited CI coverage**: Complex Kiali-specific logic (e.g., health status rollup across namespaces) was hard to integration-test without a live Kiali instance.

### Architecture

**Before:** each tool handler built REST API calls to multiple Kiali endpoints, parsed JSON responses, computed derived data (health rollups, graph merges), and formatted output.

**After:** each tool handler is a thin wrapper that POSTs to `<kiali-url>/api/chat/mcp/<tool_name>` with the tool arguments as JSON. Kiali's backend owns all data fetching, aggregation, and LLM-optimized formatting.

### Deleted (~3,000 lines)

- **`pkg/kiali/`**: 15 files removed — `health_calculation.go` (626 lines), `get_mesh_graph.go`, `graph.go`, `traces.go`, `logs.go`, `services.go`, `workloads.go`, `istio.go`, `mesh.go`, `namespaces.go`, `validations.go`, `types.go`, `endpoints.go`, etc.
- **`pkg/toolsets/kiali/`**: 7 files removed — `get_mesh_graph.go`, `get_metrics.go`, `get_resource_details.go`, `get_traces.go`, `helpers.go`, `helpers_test.go`, `logs.go`, `manage_istio_config.go`

### Added

- **`pkg/toolsets/kiali/tools/`**: 9 thin tool handlers (~60–90 lines each) that define the tool schema and delegate to `kiali.ExecuteRequest()`
- **`pkg/toolsets/kiali/tools/endpoints.go`**: centralized MCP endpoint path constants
- **`pkg/toolsets/kiali/internal/defaults/`**: shared default values for tool parameters
- **`pkg/kiali/tests/`**: endpoint coverage test ensuring every endpoint constant is wired to a tool implementation, plus backend contract test scaffolding

### Improved test coverage

- **`kiali_test.go`**: no-config initialization, empty endpoint, non-2xx error handling (with/without body), JSON body serialization, header assertions (`Content-Type`, `X-Kubernetes-MCP-Server`, POST method), Bearer token edge cases (empty, already-prefixed)
- **`config_test.go`**: direct `Validate()` tests for nil config, empty URL, invalid URL, URL without scheme, HTTP vs HTTPS certificate requirements
- **`pkg/mcp/kiali_test.go`**: rewritten to test the new architecture (single POST per tool, MCP endpoint paths, JSON argument forwarding)

## Benefits

- **Smaller tool schemas** — input schemas only expose what the LLM needs; Kiali handles defaults and validation server-side
- **Smaller responses** — Kiali returns pre-formatted, LLM-optimized text instead of raw JSON that the MCP server had to post-process
- **Single source of truth** — tool behavior is controlled by the Kiali release, eliminating version skew with the Kiali UI chatbot
- **Better testability** — Kiali tests all tool logic in its own CI with full integration tests against a live mesh
- **Simpler maintenance** — adding or modifying a tool requires changes only in the Kiali backend; the MCP server just needs a thin handler and endpoint constant

## Test plan

- [x] `make test` passes (all 25 packages green)
- [x] Endpoint coverage test verifies every endpoint constant is used by a tool
- [x] Kiali HTTP client fully tested (auth, headers, error codes, body serialization, response limits)
- [x] Config validation fully tested (nil, empty, invalid URL, HTTP/HTTPS cert logic)
- [x] MCP integration tests verify tool registration, correct endpoint routing, and response forwarding

## Summary evals

=== Evaluation Summary ===

  ✓ Create Istio Gateway (assertions: 3/3)
  ✓ Remove Fault Injection (assertions: 3/3)
  ✓ List and Validate VirtualServices (assertions: 3/3)
  ✓ List and Validate DestinationRules (assertions: 3/3)
  ✓ Update Traffic Shifting (assertions: 3/3)
  ✓ Analyze Service Traffic Volume (assertions: 3/3)
  ✓ Analyze Workload Response Times (assertions: 3/3)
  ✓ Identify Degraded Namespaces (assertions: 3/3)
  ✓ List Mesh-Enabled Namespaces (assertions: 3/3)
  ✓ Inspect Service Details (assertions: 3/3)
  ✓ Inspect Workload Details (assertions: 3/3)
  ✓ Inventory Namespace Services (assertions: 3/3)
  ✓ Inventory Workloads with Sidecar Status (assertions: 3/3)
  ✓ Comprehensive Mesh Health Audit (assertions: 3/3)
  ✓ Visualize Namespace Traffic (assertions: 3/3)
  ✓ Audit Control Plane Connectivity (assertions: 3/3)
  ✓ Visualize Cross-Namespace Traffic (assertions: 3/3)
  ✓ Visualize Workload-Level Topology (assertions: 3/3)
  ✓ Debug Service Errors via Logs (assertions: 3/3)
  ✓ Analyze Latency with Distributed Tracing (assertions: 3/3)
  ✓ Retrieve Recent Workload Logs (assertions: 3/3)

Tasks: 21/21 passed (100.00%)
Assertions: 63/63 passed (100.00%)
Tokens: ~82147 (incomplete - some counts failed)
MCP schemas: ~59787 (included in token total)
Judge used tokens:
Input: 98980 tokens
Output: 3199 tokens


pkg/toolsets/kiali/tools/get_pod_performance.go
```diff
@@ -0,0 +1,73 @@
...
+	"k8s.io/utils/ptr"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	kialiclient "github.com/containers/kubernetes-mcp-server/pkg/kiali"
+	"github.com/containers/kubernetes-mcp-server/pkg/toolsets/kiali/internal/defaults"
+)
+
+func InitGetPodPerformance() []api.ServerTool {
+	ret := make([]api.ServerTool, 0)
+	name := defaults.ToolsetName() + "_get_resource_details"
```
Cali0707: @aljesusg I _think_ this is a copy/paste mistake? I'm assuming the tool with the same name in `list_or_get_resources.go` is the one that is suppoed to have this name?

pkg/toolsets/kiali/tools/get_metrics.go
Cali0707: if we are switching the param names from snake case to camel case, we should fix the required list as well

pkg/toolsets/kiali/tools/list_or_get_resources.go
```diff
@@ -0,0 +1,68 @@
...
+					"resourceName": {
+						Type:        "string",
+						Description: "Optional. The specific name of the resource. If left empty, the tool returns a list of all resources of the specified type. If provided, the tool returns deep details for this specific resource.",
+					},
+					"clusterName": {
+						Type:        "string",
+						Description: "Optional. Name of the cluster to get resources from. If not provided, will use the default cluster name in the Kiali KubeConfig",
+					},
+				},
+				Required: []string{"resource_type"},
```
Cali0707: resource_type won't match the param name resourceType

pkg/toolsets/kiali/tools/get_traces.go
```diff
@@ -113,89 +74,10 @@ func InitGetTraces() []api.ServerTool {
 
 func TracesHandler(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
```
Cali0707: is there a reason this handler is exported?

aljesusg:
Hi @Cali0707 thanks for the review. You're right. 

We are still working on it, which is why we marked it as WIP, and it's likely we'll find something else. We won't be able to verify everything until Kiali 2.24 is released next Monday.

aljesusg:
Hi @Cali0707 ! This isr eady. Can we run it?

pkg/kiali/kiali.go
```diff
@@ -157,26 +158,26 @@ func (k *Kiali) authorizationHeader() string {
 const maxResponseBodySize = 512 << 10 // 512 KiB
 
 // executeRequest executes an HTTP request (optionally with a body) and handles common error scenarios.
-func (k *Kiali) executeRequest(ctx context.Context, method, endpoint, contentType string, body io.Reader) (string, error) {
-	if method == "" {
-		method = http.MethodGet
-	}
+func (k *Kiali) ExecuteRequest(ctx context.Context, endpoint string, arguments map[string]any) (string, error) {
 	ApiCallURL, err := k.validateAndGetURL(endpoint)
```
Cali0707, line 162: @aljesusg does this approach work with any of the multi cluster work in the MCP server? Or is this going to always go to a specific kiali instance?

Cali0707 (commented):
Overall this refactor looks fine to me. I also understand the concerns around duplicated logic for these tools.

I guess my only two worries would be around if there will be any issues with versioning this MCP server vs. the kiali one (what happens if the kiali backend updates and has different params, but the user doesn't update the MCP server or vice-versa?), as well as how the kiali endpoint mixes with the multi cluster support

One interesting approach we have seen downstream is having the toolset be an external dependency that is imported into the server, maybe that would be another option that would at least solve the versioning issues?

cc @matzew @manusa - not sure what y'all think

aljesusg:
> Overall this refactor looks fine to me. I also understand the concerns around duplicated logic for these tools.
> 
> I guess my only two worries would be around if there will be any issues with versioning this MCP server vs. the kiali one (what happens if the kiali backend updates and has different params, but the user doesn't update the MCP server or vice-versa?), as well as how the kiali endpoint mixes with the multi cluster support
> 
> One interesting approach we have seen downstream is having the toolset be an external dependency that is imported into the server, maybe that would be another option that would at least solve the versioning issues?
> 
> cc @matzew @manusa - not sure what y'all think

That is actually why we opted for this approach; by doing it this way, the Kiali endpoints remain fixed for this specific use case, which avoids potential versioning mismatches.

Regarding the external toolset dependency, I think it's a great idea. It’s definitely an option worth exploring to keep things modular

pkg/kiali/kiali.go
aljesusg: Actually, Kiali is inherently multi-cluster. That’s why we included the clusterName parameter in the toolsets; by specifying the cluster, Kiali can route the request to the corresponding instance without issues

like this tool
https://github.com/containers/kubernetes-mcp-server/pull/969/changes#diff-572e33c265c90cf1a01e9ed111ea66445165cb340f463e77ac54f34b9eb99aacR60

Cali0707: cool!

aljesusg:
Latest run https://github.com/kiali/kiali/actions/runs/24563451212

pkg/toolsets/kiali/tools/defaults.go
manusa, line 1: Thirteen of the exported Default* constants are declared but never referenced anywhere in the repo: DefaultDuration, DefaultRequestProtocol, DefaultMaxSpans, DefaultIncludeIdleEdges, DefaultInjectServiceNodes, DefaultBoxBy, DefaultAmbientTraffic, DefaultAppenders, DefaultRateGrpc, DefaultRateHttp, DefaultRateTcp, DefaultIncludeGateways, DefaultIncludeWaypoints.
Since the backend now owns defaults, these should be deleted (or their use added to the schemas, if intentional).
Since this file is exported, the unused linter won't flag them (easy to miss).

aljesusg: Yes, you are right, we reduced a lot the options here and we forgot make a clean of these values.

pkg/toolsets/kiali/tools/defaults.go
```diff
@@ -0,0 +1,17 @@
...
+	DefaultRateInterval    = "10m"
+	DefaultGraphType       = "versionedApp"
+	DefaultStep            = "15"
+	DefaultDirection       = "outbound"
+	DefaultReporter        = "source"
+	DefaultQuantiles       = "0.5,0.95,0.99,0.999"
+	DefaultLimit           = "10"
+	DefaultTail            = "50"
+	DefaultLookbackSeconds = 600
+	DefaultErrorOnly       = "false"
```
manusa: `DefaultTail = "50"`, `DefaultLimit = "10"`, `DefaultErrorOnly = "false"`
Declared as strings for parameters with other types (integer and boolean fields)

`pkg/mcp/testdata/toolsets-kiali-tools.json` shows the json schema with the quoted serialized string.

Stricter MCP clients might complain or ignore the default value.

pkg/toolsets/kiali/tools/get_logs.go
```diff
@@ -0,0 +1,88 @@
...
+						Type:        "string",
+						Description: "Name of the Pod to get the logs from. If it does not exist, it will be treated as a workload name and a running pod will be selected.",
+					},
+					"workload": {
+						Type:        "string",
+						Description: "Optional. Workload name override (used when name lookup fails).",
+					},
+					"container": {
+						Type:        "string",
+						Description: "Optional. Name of the Pod container to get the logs fro ",
```
manusa:
```suggestion
						Description: "Optional. Name of the Pod container to get the logs from",
```

pkg/toolsets/kiali/tools/endpoints.go
```diff
@@ -0,0 +1,17 @@
...
+
+// Kiali API endpoint paths shared across this package.
+const (
+	KialiMCPPath = "/api/chat/mcp"
+
+	KialiGetMeshTrafficGraphEndpoint   = KialiMCPPath + "/get_mesh_traffic_graph"
+	KialiGetMeshStatusEndpoint         = KialiMCPPath + "/get_mesh_status"
+	KialiGetMetricsEndpoint            = KialiMCPPath + "/get_metrics"
+	KialiListOrGetResourcesEndpoint    = KialiMCPPath + "/list_or_get_resources"
+	KialiGetTracesEndpoint             = KialiMCPPath + "/list_traces"
```
manusa: Should this be named `KialiListTracesEndpoint`?

manusa (changes requested):
Overall and given the evals results, I think it looks good. Thanks! 🙌 

I think the only critical thing is the default value declaration types, otherwise should be good to go.

manusa (approved):
Good to go from my side, thx!

matzew:
The `Validate()` in the `pkg/kialli/config` file checks for a non-empty `scheme`, but doesn't restrict to it be only `HTTP` or `HTTPS`. This has not changed - but just something I noticed.

Cali0707:
> The `Validate()` in the `pkg/kialli/config` file checks for a non-empty `scheme`, but doesn't restrict to it be only `HTTP` or `HTTPS`. This has not changed - but just something I noticed.

@aljesusg I think we should be validating HTTPS in the case where this config option is enabled (to ensure consistency with end user expectations): https://github.com/containers/kubernetes-mcp-server/blob/f2b5d0132c155193b6df5ba23982385059d5527a/pkg/config/config.go#L121

aljesusg:
> > The `Validate()` in the `pkg/kialli/config` file checks for a non-empty `scheme`, but doesn't restrict to it be only `HTTP` or `HTTPS`. This has not changed - but just something I noticed.
> 
> @aljesusg I think we should be validating HTTPS in the case where this config option is enabled (to ensure consistency with end user expectations):
> 
> https://github.com/containers/kubernetes-mcp-server/blob/f2b5d0132c155193b6df5ba23982385059d5527a/pkg/config/config.go#L121

That’s a good catch! Since this behavior existed previously and isn't strictly related to the current changes, would you prefer I include the HTTPS validation in this PR, or should we track it as a follow-up issue to keep this PR focused?

manusa:
> That’s a good catch! Since this behavior existed previously and isn't strictly related to the current changes, would you prefer I include the HTTPS validation in this PR, or should we track it as a follow-up issue to keep this PR focused?

AFAIR this is captured here:
https://github.com/containers/kubernetes-mcp-server/blob/beed6f9d96996d720ab63090b4d3744a90b684a6/pkg/kiali/kiali.go#L129-L135

and tested in:
https://github.com/containers/kubernetes-mcp-server/blob/beed6f9d96996d720ab63090b4d3744a90b684a6/pkg/kiali/kiali_test.go#L63-L73

aljesusg:
I can add more tests or controls if needed. I've successfully tested the MCP with this setup on ocp12 using Lightspeed, and everything is working well. We also have some demo videos featuring the core and Kiali for summit. Let me know how you'd like to proceed

manusa:
What I meant in my earlier comment is that there's no regression here, the test asserts that the missing https is correctly enforced (unless the test was removed).

@Cali0707 were you referring to this gap, or something else?

Cali0707:
@manusa good point around the test - I had missed that earlier

LGTM from my side

PR #975, opened 2026-03-26
fix(evals): improve kubevirt eval assertions and isolation

lyarwood:
## Summary
- Fix troubleshoot-vm verify to actually fail when the VM doesn't become ready (was silently passing)
- Remove redundant kubevirt-specific eval configs (openai-agent deprecated, claude-code covered by root evals)
- Add `promptsUsed` assertion for troubleshoot-vm to verify the agent uses the `vm-troubleshoot` prompt
- Add `expected-tool` labels and per-tool assertions for all kubevirt tasks (e.g. `vm_create`, `vm_clone`, `vm_lifecycle`, `resources_delete`, `resources_create_or_update`)
- Use unique namespaces per task (`kvt-<task-name>`) to prevent collisions during parallel execution

### DIFF
--- a/evals/tasks/kubevirt/clone-vm/task.yaml
+++ b/evals/tasks/kubevirt/clone-vm/task.yaml
@@ -16,19 +16,19 @@ spec:
         apiVersion: v1
         kind: Namespace
         metadata:
-          name: vm-test
+          name: kvt-clone
         ignoreNotFound: true
     - k8s.create:
         apiVersion: v1
         kind: Namespace
         metadata:
-          name: vm-test
+          name: kvt-clone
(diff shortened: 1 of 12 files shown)


evals/claude-code/eval.yaml
```diff
@@ -53,6 +53,46 @@ config:
             toolPattern: ".*"
         minToolCalls: 1
         maxToolCalls: 20
+    - glob: ../tasks/kubevirt/*/*.yaml
```
lyarwood, line 56: ah sorry this was a mistake, this double runs everything, somehow missed this in my local test run.

PR #980, opened 2026-03-27
fix(mcp): add multi-cluster target parameter to prompts

madaraszg-tulip:
Prompts like cluster-health-check did not receive the target parameter (e.g. "context") in multi-cluster mode, even though tools did. The server-side handler already extracted the parameter from request arguments, but it was never advertised in the prompt schema, so MCP clients had no way to send it.

Introduce a PromptMutator pattern mirroring the existing ToolMutator approach. WithPromptTargetParameter injects the target argument into cluster-aware prompts when multi-cluster support is enabled. Apply it in collectApplicablePrompts() so the mutation happens at the same architectural layer as tools.

### DIFF
--- a/pkg/mcp/prompt_mutator.go
+++ b/pkg/mcp/prompt_mutator.go
@@ -8,16 +8,6 @@ import (
 
 type PromptMutator func(prompt api.ServerPrompt) api.ServerPrompt
 
-// ComposePromptMutators combines multiple mutators into a single mutator that applies them in order
-func ComposePromptMutators(mutators ...PromptMutator) PromptMutator {
-	return func(prompt api.ServerPrompt) api.ServerPrompt {
-		for _, m := range mutators {
-			prompt = m(prompt)
-		}
-		return prompt
-	}
-}
-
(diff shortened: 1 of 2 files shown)


pkg/mcp/prompt_mutator.go
```diff
@@ -0,0 +1,42 @@
...
+
+// ComposePromptMutators combines multiple mutators into a single mutator that applies them in order
+func ComposePromptMutators(mutators ...PromptMutator) PromptMutator {
+	return func(prompt api.ServerPrompt) api.ServerPrompt {
+		for _, m := range mutators {
+			prompt = m(prompt)
+		}
+		return prompt
+	}
+}
```
Cali0707: @madaraszg-tulip I'm not sure if we need this function yet (as it is currently unused).

What do you think about removing it for now and only adding it when we see a use case?

madaraszg-tulip: Removed. It was created to match the tool mutator setup, but indeed, it is not used currently.

PR #995, opened 2026-03-31
feat(evals): add make target to install mcpchecker agent dependencies

### LINKED ISSUE #947: Add/update make target to install deps for mcpchecker agents

nil957:
## Summary

Adds new make targets to install agent dependencies for mcpchecker evaluations.

### Changes

- **`make claude-agent-acp`**: Installs the `@agentclientprotocol/claude-agent-acp` npm package required for running Claude Code evaluations
- **`make mcpchecker-deps`**: Convenience target that installs mcpchecker and all agent dependencies (currently claude-agent-acp)

Also adds `CLAUDE_AGENT_ACP_VERSION` variable (defaults to `latest`) to allow pinning to specific versions if needed.

### Usage

```bash
# Install all dependencies for running mcpchecker with different agents
make mcpchecker-deps

# Or install individual dependencies
make mcpchecker
make claude-agent-acp
```

Fixes #947

### DIFF
--- a/build/evals.mk
+++ b/build/evals.mk
@@ -7,6 +7,7 @@ MCP_CONFIG_DIR ?= dev/config/mcp-configs
 
 MCPCHECKER = $(shell pwd)/_output/tools/bin/mcpchecker
 MCPCHECKER_VERSION ?= latest
+CLAUDE_AGENT_ACP_VERSION ?= latest
 EVAL_CONFIG ?= evals/openai-agent/eval.yaml
 EVAL_LABEL_SELECTOR ?= suite=kubernetes
 EVAL_TASK_FILTER ?=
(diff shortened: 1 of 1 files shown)


manusa:
Thanks for this, @nil957, and sorry it sat for a while. We've folded the `claude-agent-acp` install target into a broader pass over the local eval workflow in #1237 (tracked by #1234), which also fixes the stale docs, the drifted Makefile defaults, and the health-check probe. Your `CLAUDE_AGENT_ACP_VERSION` version-pin idea is a good one and we're carrying it forward there with credit to you.

One thing for next time: the commits here aren't signed off under the DCO, which is why we couldn't pull them in directly. If you sign off on future PRs (`git commit -s`) we can merge your work as-is.

Closing in favor of #1237. Thanks again for the contribution.

PR #998, opened 2026-03-31
feat(helm): add backend storage driver support

difrost:
Currently helm toolset supports only default, 'secret' storage driver. This PR adds support of defining default storage driver (via config file) and per-call driver overrides. 

Supported values: "secret", "configmap" and "" (Helms default driver - currently "secret").

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -584,16 +584,18 @@ token = "your-kiali-token"
 ```toml
 [toolset_configs.helm]
 allowed_registries = ["oci://ghcr.io/myorg", "https://charts.example.com"]
+storage_driver = "configmap"
 ```
 
 #### Helm Configuration
 
-The Helm toolset supports an optional `allowed_registries` allowlist to restrict which registries
-`helm_install` can fetch charts from.
-
 | Field | Type | Description |
 |-------|------|-------------|
 | `allowed_registries` | string array | Optional list of permitted chart registry URL prefixes. Only `oci://` and `https://` schemes are accepted. |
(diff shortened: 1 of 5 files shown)


difrost:
Hey @manusa. I'm running this in an environment with mixed drivers in use (on lower envs users can use standard "secret" driver while in upper envs access to Kube Secrets is blocked hence the need to handled releases backed by CMs) for last 3 days and it works great. No issues with agent switching between different drivers smoothly and showing releases installed via both drivers on single cluster or in mix of clusters.

Keeping the config level control only could be a good move forward but that actually helps only in uniform environments where there's just one single driver in use - from my experience this is rare case. In an enterprise env like mine, it is quite common to ban Kube Secrets use and that forces Helm users to switch to CMs but that is rarely enforced across all environments.  This implementation has been created based on the environments I've been dealing with for last couple years.

I'm ok to just submit the portion of this feature and keep remaining parts on my feature build. 

Re tests - I'll move the `TestValidateStorageDriver` to `Config.Validate()`.

manusa:
> I'm ok to just submit the portion of this feature and keep remaining parts on my feature build.


Let's break the work in two then. For this PR, keep the narrowed scope which can be merged easily (it won't affect existing LLM performance).

Follow-up PR with the optional parameter. For this one we'll need to improve the eval suite and find a way to test this in the CI environment.

> This implementation has been created based on the environments I've been dealing with for last couple years.

I trust your judgement, but still the evals are a must right now for a consolidated toolset.

difrost:
> Let's break the work in two then. For this PR, keep the narrowed scope which can be merged easily (it won't affect existing LLM performance).

@manusa please review.

PR #1002, opened 2026-04-01
feat: pods_log support default-container annotation

### LINKED ISSUE #1000: pods_log: support kubectl.kubernetes.io/default-container annotation for multi-container pods
The pods_log tool in kubernetes-mcp-server fails on multi-container pods when the container parameter is not explicitly provided. The Kubernetes API requires a container name for multi-container pods, but the current code passes an empty string without any resolution logic. It returns: 

`a container name must be specified for pod <name>, choose one of: [container-a container-b ...]`

**What's needed**
`pod_logs `should allow default-container annotations in multi container scenarios to allow actionable error listing.

xingzihai:
## Summary
This PR adds support for the `kubectl.kubernetes.io/default-container` annotation in `pods_log` and `pods_exec` operations.

## Changes
- Added `DefaultContainerAnnotation` constant
- Added `getDefaultContainer` helper function to extract the annotation value from pod metadata
- Updated `PodsLog` to check the annotation when container is not explicitly specified
- Updated `PodsExec` to use the helper function for consistency

## Motivation
When a pod has multiple containers and the `default-container` annotation is set, users often expect log/exec operations to target that container by default, similar to how `kubectl` behaves.

This aligns with kubectl's behavior and improves the user experience for multi-container pods.

## Testing
- Tested with multi-container pods with the annotation set
- Verified that explicit container parameter still takes precedence
- Verified behavior when annotation is not present

## Related Issue
Closes #1000

## Checklist
- [x] Code compiles correctly
- [x] Follows the existing code style
- [x] Documentation updated if needed

### DIFF
--- a/pkg/kubernetes/pods.go
+++ b/pkg/kubernetes/pods.go
@@ -27,6 +27,36 @@ import (
 // DefaultTailLines is the default number of lines to retrieve from the end of the logs
 const DefaultTailLines = int64(100)
 
+// DefaultContainerAnnotation is the annotation key used by kubectl to specify the default container
+// for operations like logs and exec in pods with multiple containers.
+const DefaultContainerAnnotation = "kubectl.kubernetes.io/default-container"
+
+// getDefaultContainer returns the container name to use for operations like logs and exec.
+// If a specific container is provided, it is returned directly.
+// If container is empty, it checks the pod's annotations for the default-container annotation.
+// If the annotation exists, it returns that container name.
+// Otherwise, it returns the first container's name.
+func getDefaultContainer(pod *v1.Pod, container string) string {
+	// If container is explicitly specified, use it
+	if container != "" {
+		return container
+	}
+
+	// Check for the default-container annotation
+	if pod.Annotations != nil {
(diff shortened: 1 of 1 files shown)


pkg/kubernetes/pods.go
```diff
@@ -27,6 +27,36 @@ import (
...
+// DefaultContainerAnnotation is the annotation key used by kubectl to specify the default container
+// for operations like logs and exec in pods with multiple containers.
+const DefaultContainerAnnotation = "kubectl.kubernetes.io/default-container"
+
+// getDefaultContainer returns the container name to use for operations like logs and exec.
+// If a specific container is provided, it is returned directly.
+// If container is empty, it checks the pod's annotations for the default-container annotation.
+// If the annotation exists, it returns that container name.
+// Otherwise, it returns the first container's name.
+func getDefaultContainer(pod *v1.Pod, container string) string {
```
matzew, line 39: could we get a test for this?

matzew:
/close

thanks for the contribution - the issue has been solved in the meantime

PR #1012, opened 2026-04-03
feat(kubevirt): add run policy support to VM lifecycle management

awels:
Adds a new run_policy parameter to the vm_lifecycle tool that allows users to control the VM's runStrategy when starting a virtual machine.

The parameter supports three policies:
- HighAvailability: VM runs continuously (sets runStrategy to Always)
- RestartOnFailure: VM restarts on failure (sets runStrategy to RerunOnFailure)
- Once: VM runs once and stops after completion (sets runStrategy to Once)

The run_policy parameter is optional and defaults to HighAvailability to maintain backward compatibility with existing usage.

Changes include:
- Updated StartVM function to accept RunPolicy parameter
- Added 19 unit tests covering all run policy combinations
- Added 3 integration tests for vm_lifecycle tool
- Updated tool schema with enum values and documentation
- Auto-generated README.md updates

### DIFF
--- a/pkg/kubevirt/vm.go
+++ b/pkg/kubevirt/vm.go
@@ -9,13 +9,35 @@ import (
 	"k8s.io/client-go/dynamic"
 )
 
-// RunStrategy represents the run strategy for a VirtualMachine
+// RunStrategy represents the Kubernetes VirtualMachine runStrategy.
 type RunStrategy string
 
+// RunPolicy is a user-facing abstraction over RunStrategy used by lifecycle tools.
+type RunPolicy string
+
 const (
-	RunStrategyAlways RunStrategy = "Always"
-	RunStrategyHalted RunStrategy = "Halted"
+	RunStrategyAlways         RunStrategy = "Always"
+	RunStrategyHalted         RunStrategy = "Halted"
+	RunStrategyManual         RunStrategy = "Manual"
+	RunStrategyRerunOnFailure RunStrategy = "RerunOnFailure"
+	RunStrategyOnce           RunStrategy = "Once"
 )
+
+const (
+	RunPolicyHighAvailability RunPolicy = "HighAvailability"
+	RunPolicyRestartOnFailure RunPolicy = "RestartOnFailure"
+	RunPolicyOnce             RunPolicy = "Once"
+)
+
+// Validate reports whether the run policy is one of the supported values.
+func (p RunPolicy) Validate() error {
+	switch p {
+	case RunPolicyHighAvailability, RunPolicyRestartOnFailure, RunPolicyOnce:
+		return nil
+	default:
+		return fmt.Errorf("invalid run policy '%s': must be one of 'HighAvailability', 'RestartOnFailure', 'Once'", p)
+	}
+}
 
 // GetVirtualMachine retrieves a VirtualMachine by namespace and name
 func GetVirtualMachine(ctx context.Context, client dynamic.Interface, namespace, name string) (*unstructured.Unstructured, error) {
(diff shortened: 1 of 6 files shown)


pkg/kubevirt/vm.go
```diff
@@ -78,6 +95,23 @@ func StartVM(ctx context.Context, dynamicClient dynamic.Interface, namespace, na
...
+func getRunStrategyFromRunPolicy(runPolicy RunPolicy) RunStrategy {
+	switch runPolicy {
+	case RunPolicyHighAvailability:
+		return RunStrategyAlways
+	case RunPolicyRestartOnFailure:
+		return RunStrategyRerunOnFailure
+	case RunPolicyOnce:
+		return RunStrategyOnce
+	}
+	return RunStrategyAlways
```
lyarwood, line 112: This silent default to `Always` has a broader impact: `RestartVM` (line 200) hardcodes `RunStrategyAlways` when restarting. After this PR, if a user starts a VM with `run_policy=Once` then restarts it, the strategy silently reverts to `Always`. Should `RestartVM` accept and forward the `run_policy` parameter, or at minimum preserve the VM's current `runStrategy`?

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
```diff
@@ -75,6 +88,10 @@ func lifecycle(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
 		return api.NewToolCallResult("", err), nil
 	}
 
+	// Parse optional run_policy parameter (defaults to HighAvailability)
+	runPolicyStr := api.OptionalString(params, "run_policy", string(kubevirt.RunPolicyHighAvailability))
+	runPolicy := kubevirt.RunPolicy(runPolicyStr)
```
lyarwood, line 93: The `run_policy` value is cast directly to `RunPolicy` without server-side validation. `getRunStrategyFromRunPolicy` silently defaults invalid values to `Always`, which masks errors. Consider adding explicit validation:
```go
if !kubevirt.IsValidRunPolicy(runPolicy) {
    return api.NewToolCallResult("", fmt.Errorf("invalid run_policy %q: must be one of HighAvailability, RestartOnFailure, Once", runPolicyStr)), nil
}
```

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
```diff
@@ -75,6 +88,10 @@ func lifecycle(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
 		return api.NewToolCallResult("", err), nil
 	}
 
+	// Parse optional run_policy parameter (defaults to HighAvailability)
+	runPolicyStr := api.OptionalString(params, "run_policy", string(kubevirt.RunPolicyHighAvailability))
```
lyarwood, line 92: nit: `run_policy` is parsed for all actions but only used by `start`. Consider moving this inside the `ActionStart` case to avoid confusion.

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
awels: Done

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
awels: Done

pkg/kubevirt/vm.go
awels: It should accept the run_policy parameter, I will add that

lyarwood (approved):

Cali0707:
@awels can you fix the merge conflict on the readme?

awels:
Sure give me a bit to rebase.

awels:
@Cali0707 I rebased if you could run the tests?

manusa (commented):
## Overview

Independent review of the diff (not the PR description). The feature is well-scoped and the public-API change is backward-compatible for `run_policy`, but there are several issues worth addressing before merge — most notably a latent footgun in `getRunStrategyFromRunPolicy`, a stale `README.md` that does not match the new `action` description, and broken markdown rendering for the `run_policy` description.

### Strengths

- Backward compatible: `run_policy` is optional and defaults to `HighAvailability`, which preserves the prior `Always` behavior for `start`/`restart`.
- Schema-side `enum` plus runtime validation gives the LLM an authoritative list.
- Good test coverage of policy/strategy combinations (`pkg/kubevirt/vm_test.go`, `pkg/mcp/kubevirt_test.go`).
- `pkg/mcp/testdata/toolsets-kubevirt-tools.json` is updated to match the schema.
- Migration to `api.WrapParams` in `pkg/toolsets/kubevirt/vm/lifecycle/tool.go:77-84` matches the newer pattern used elsewhere in the codebase.

### Issues

#### Critical (Must Fix)

- **`getRunStrategyFromRunPolicy` silently falls back to `RunStrategyAlways` for unknown policies** (`pkg/kubevirt/vm.go:116-126`). Today, all production callers gate on `IsValidRunPolicy` first, so the fallback is unreachable — but if any future caller forgets, an unknown policy boots the VM with `Always` instead of failing loudly. `TestGetRunStrategyFromRunPolicy` even encodes this surprise as expected behavior (`pkg/kubevirt/vm_test.go:469-473`, "Invalid policy defaults to Always"), making accidental regressions silent. Either drop the fallback and panic / return a second `bool, error`, or remove the dead branch entirely now that validation is mandatory.

#### Important (Should Fix)

- **`README.md` `action` description is out of sync.** `pkg/toolsets/kubevirt/vm/lifecycle/tool.go:44` was shortened to `"The lifecycle action to perform: 'start', 'stop', or 'restart'"` (and `pkg/mcp/testdata/toolsets-kubevirt-tools.json:168` was regenerated), but `README.md:535` still shows the old `'start' (changes runStrategy to Always), 'stop' (changes runStrategy to Halted), or 'restart' (stops then starts the VM)`. Running `make update-readme-tools` produces a non-empty diff, so the auto-generated section was not regenerated.

- **`run_policy` description renders as broken markdown in `README.md`.** The description in `pkg/toolsets/kubevirt/vm/lifecycle/tool.go:53-57` embeds `\n  - 'HighAvailability': …` etc. The README generator (`internal/tools/update-readme/main.go:90-97`) emits each property at 2-space indent, so the embedded `  - ` bullets land at the *same* indent level as the `run_policy` line itself — they render as siblings of `run_policy`, not as nested explanations. `README.md:539-541` show three "Once / HighAvailability / RestartOnFailure" bullets that look like additional top-level parameters, and `README.md:542` (`Defaults to 'HighAvailability' …`) drops out of the list entirely because it has no indent. Move the option list into the parent `action`/`run_policy` description as a single line, or render the options outside the schema description, or deeper-indent the bullets (e.g. `\n    - …`).

- **Misleading "started successfully" message when only the strategy changes** (`pkg/toolsets/kubevirt/vm/lifecycle/tool.go:104-108`, `pkg/kubevirt/vm.go:73-74`). `StartVM` now returns `wasStarted == true` whenever `currentStrategy != desiredStrategy`, which includes transitions where the VM was already running (e.g. `Always` → `Once`). The tool then says `# VirtualMachine started successfully with run policy 'Once'`, even though the VM never stopped. Either differentiate the path in `StartVM` (e.g. return a tri-state `started | strategyChanged | noChange`) or branch the message on whether `currentStrategy == RunStrategyHalted` / empty. The doc comment on `StartVM` overloads "started" with "strategy changed" in the same way and should be updated alongside.

- **`TestGetRunStrategyFromRunPolicy` violates the project's black-box testing rule** (`pkg/kubevirt/vm_test.go:448-484`). `CLAUDE.md` is explicit: tests should be black-box and not access unexported functions; `getRunStrategyFromRunPolicy` is lowercase. The mapping is already fully covered by `TestStartVM` (via `wantRunStrategy`) and `TestRestartVMWithDifferentRunPolicies`. Delete this test — keeping it also locks the buggy `Always` fallback in as expected behavior (see Critical above).

- **`run_policy` is validated twice.** `pkg/toolsets/kubevirt/vm/lifecycle/tool.go:97-99` and `:125-127` validate, then `pkg/kubevirt/vm.go:77-79` and `:196-198` validate again — same check, four hard-coded copies of the error string. Pick one layer (the library is the public API, so prefer keeping the check there) and drop the other. As-is, adding a fourth policy means touching four strings.

#### Minor (Nice to Have)

- **Unused exported constants** (`pkg/kubevirt/vm.go:19,22`). `RunStrategyWaitAsReceiver` has zero references anywhere; `RunStrategyManual` is referenced only from tests in the same package. Both inflate the exported surface for downstream consumers without a current need. Either remove them or introduce them when a production caller actually needs them. `RunStrategyRerunOnFailure` / `RunStrategyOnce` are legitimately used.

- **Type declaration style** (`pkg/kubevirt/vm.go:13-14`). Two stacked `type X string` lines are valid Go but a grouped `type ( … )` block reads better when introducing a sibling type. Optional polish.

- **Stale comment** (`pkg/kubevirt/vm.go:92`, `// Check if already running`). With the new semantics this branch checks "already at the desired strategy", not "already running". Update the comment so it doesn't describe the old behavior.

- **Action description loses useful detail.** Trimming `action` to just the three verbs (`pkg/toolsets/kubevirt/vm/lifecycle/tool.go:44`) removes the `runStrategy` mapping that an LLM previously relied on to predict behavior. If the goal was to move that mapping into the `run_policy` description, consider keeping a short "(see `run_policy` for details)" hint on `action` so the link is discoverable.

### Assessment: Ready to merge? **No — with fixes**

The README desync and the broken-markdown rendering should be fixed before merge (auto-generated docs are easy to drift again later), and the silent `Always` fallback in `getRunStrategyFromRunPolicy` is worth tightening up while the file is open. The remaining items are quality-of-life.

awels:
Finally got around to addressing the noted issues, and some others my code-review AI noticed.

lyarwood:
@manusa @Cali0707 can you run mcpchecker again please?

ksimon1:
@manusa @Cali0707 can you please trigger mcpchecker?

ksimon1:
@manusa @Cali0707 can you please trigger mcpchecker?

pkg/kubevirt/vm.go
```diff
@@ -47,37 +69,92 @@ func UpdateVirtualMachine(ctx context.Context, client dynamic.Interface, vm *uns
...
+		return nil, false, false, fmt.Errorf("failed to read runStrategy from VirtualMachine: %w", err)
 	}
 
-	// Check if already running
-	if found && currentStrategy == RunStrategyAlways {
-		return vm, false, nil
+	// Already at the desired strategy — no-op
+	if found && currentStrategy == desiredStrategy {
+		return vm, false, false, nil
 	}
```
Cali0707, line 107: Could you help confirm the behavior here? My understanding (with the help of my agent) is that a VM can still have runStrategy: Once after it has finished running, with no VMI currently present. Would calling start with run_policy: Once then return here without starting it? I think the same situation may occur after a clean shutdown with RerunOnFailure. Is that expected KubeVirt behavior, or do we need to check whether the VM is actually running before treating this as a no-op?

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
```diff
@@ -66,57 +77,57 @@ func Tools(p api.FilteringProvider) []api.ServerTool {
...
 	var vm *unstructured.Unstructured
 	var message string
+	var err error
 
 	switch Action(action) {
 	case ActionStart:
-		var wasStarted bool
-		vm, wasStarted, err = kubevirt.StartVM(params.Context, dynamicClient, namespace, name)
+		var wasStarted, strategyChanged bool
+		runPolicy := kubevirt.RunPolicy(p.OptionalString("run_policy", string(kubevirt.RunPolicyHighAvailability)))
```
Cali0707, line 98: This optional parameter is parsed after we check p.Err() above. If run_policy is present with the wrong type, OptionalString records an error and returns HighAvailability, but that error is never checked, so we perform the lifecycle action with the default policy. Could we parse run_policy before the existing p.Err() check, following the usual WrapParams pattern?

Cali0707 (commented):
Can we get an eval scenario or two to verify this works?

pkg/kubevirt/vm.go
awels: The run policy applies when the VM is started. If the VM is stopped and you set the policy to once or RerunOnFailure, it won't actually start the VM. You have to actually start it by calling the start sub resource in this case. If you set the run runPolicy to Always that will actually start the VM, but that is the exception, not the normal behavior.

Cali0707 (commented):
/run-mcpchecker kubevirt

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
awels: Done

awels:
> Can we get an eval scenario or two to verify this works?

I added some evals for various scenarios.

lyarwood (approved):

PR #1026, opened 2026-04-07
feat(core): add dry-run mode for ResourcesCreateOrUpdate

difrost:
This PR adds a server-side dry-run to core/ResourcesCreateOrUpdate(). The dry-run is disabled for other parts of code: core/PodsRun() or kubevirt VM spawning.

Please review.


pkg/toolsets/core/resources.go
```diff
@@ -270,15 +278,29 @@ func resourcesCreateOrUpdate(params api.ToolHandlerParams) (*api.ToolCallResult,
...
+	if dryRun {
+		marshalledYaml, err := output.MarshalYaml(resources)
+		if err != nil {
+			return api.NewToolCallResult("", fmt.Errorf("failed to validate resources: %w", err)), nil
+		}
+		result = "# The following resources (YAML) passed validation (dry run mode)\n" + marshalledYaml
+	} else {
+		marshalledYaml, err := output.MarshalYaml(resources)
+		if err != nil {
+			return api.NewToolCallResult("", fmt.Errorf("failed to create or update resources: %w", err)), nil
```
matzew: looks mostly identical?

matzew:
wanna add some tests for `dryRun=true`  scenario?

pkg/toolsets/core/resources.go
difrost: Check [d5427c3](https://github.com/containers/kubernetes-mcp-server/pull/1026/commits/d5427c37c1c9a57f6e4febf27fb8ca6cf721eacf)

difrost:
> wanna add some tests for `dryRun=true` scenario?

Currently - from what I can tell - there's no mock for the kube client. That would be a bigger work.

Cali0707:
@difrost can you add an eval task or two to this PR? That would help show the use case of these options for an agent, and we are hoping to have evals for all changes to the core toolset

pkg/toolsets/core/resources.go
```diff
@@ -270,15 +278,28 @@ func resourcesCreateOrUpdate(params api.ToolHandlerParams) (*api.ToolCallResult,
 		return api.NewToolCallResult("", fmt.Errorf("resource is not a string")), nil
 	}
 
-	resources, err := kubernetes.NewCore(params).ResourcesCreateOrUpdate(params, r)
+	dryRun := api.OptionalBool(params, "dryRun", false)
```
lyarwood: This uses the deprecated `api.OptionalBool` helper (see `pkg/api/params.go` deprecation comment). New code should use `api.WrapParams`:
```go
p := api.WrapParams(params)
dryRun := p.OptionalBool("dryRun", false)
if err := p.Err(); err != nil {
}
```

pkg/toolsets/core/resources.go
```diff
@@ -366,7 +389,14 @@ func resourcesScale(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
...
 
-	return api.NewToolCallResult("# Current resource scale (YAML) is below\n"+marshalled, err), nil
+	var result string
+	if dryRun && shouldScale {
+		result = "# Scale operation validated (dry run mode). Current resource scale (YAML) is below\n" + marshalled
+	} else {
+		result = "# Current resource scale (YAML) is below\n" + marshalled
+	}
+
+	return api.NewToolCallResult(result, err), nil
```
lyarwood, line 407: nit: At this point `err` is always nil (the function returns early on error above). Should be `nil` for consistency with `resourcesCreateOrUpdate` which correctly passes `nil`.

pkg/toolsets/core/resources.go
```diff
@@ -356,7 +377,9 @@ func resourcesScale(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
 		}
 	}
 
-	scale, err := kubernetes.NewCore(params).ResourcesScale(params.Context, gvk, ns, n, desiredScale, shouldScale)
+	dryRun := api.OptionalBool(params, "dryRun", false)
```
lyarwood: Same deprecated `api.OptionalBool` issue as above - should use `api.WrapParams`.

pkg/toolsets/core/resources.go
difrost: Rebased with main: [dea64a8](https://github.com/containers/kubernetes-mcp-server/pull/1026/commits/dea64a8d22e5d6b11cef93acedb847a6d98e03b9)

pkg/toolsets/core/resources.go
difrost: That seem to be pretty consistent across most of the functions. e.g. see the resourcesDelete(). I'll address that as dedicated PR.

difrost:
> @difrost can you add an eval task or two to this PR? That would help show the use case of these options for an agent, and we are hoping to have evals for all changes to the core toolset

@Cali0707 I've added/tested 3 evals respectively for dry run create, update and scale - see [5b00967](https://github.com/containers/kubernetes-mcp-server/pull/1026/commits/5b00967de3f7528713cdba79d0c5632001f2a446)

evals/tasks/core/dryrun-create-namespace/dryrun-create-namespace.yaml
```diff
@@ -0,0 +1,15 @@
...
+  difficulty: simple
+steps:
+  setup:
+    file: setup.sh
+  verify:
+    file: verify.sh
+  cleanup:
+    file: cleanup.sh
+  prompt:
+    inline: "Dry run create a namespace named dryrun-create-namespace."
```
Cali0707, line 15: @difrost would it be possible to use a more realistic use case for the eval? It's not clear to me why a user would want to use this feature from the eval (same with the other evals)

difrost: @Cali0707  We use this feature heavily to:
* validate resources against various configured validation webhooks, e.g. Gatekeeper policies, that can only be verified by performing SSA;
* verify correctness of the manifests, especially for CRs.

None of above can be emulated with evals. The ones that I've put are intended to verify whether the MCP server is correctly passing the options to K8s and the resources are not modified and/or created.

PR #1033, opened 2026-04-10
feat(kubevirt): add windows-golden-image prompt for automated Windows golden image creation

ksimon1:
feat(kubevirt): add windows-golden-image prompt for automated Windows golden image creation

Add a new MCP prompt that guides the AI agent through creating a Windows golden image via the windows-efi-installer Tekton pipeline from the Tekton catalog. The prompt supports Windows 10, 11, 2k22 (default), and 2k25, and enforces a Microsoft EULA acceptance flow before creating the PipelineRun.

Code was assisted by Cursor AI.

### DIFF
--- a/evals/tasks/kubevirt/windows-golden-image-no-eula/prompt.yaml
+++ b/evals/tasks/kubevirt/windows-golden-image-no-eula/prompt.yaml
@@ -1,0 +1,67 @@
+kind: Prompt
+apiVersion: mcpchecker/v1alpha2
+metadata:
+  labels:
+    suite: kubevirt
+    requires: kubevirt,tekton
+  name: "windows-golden-image-no-eula"
+  difficulty: medium
+  description: "Test the windows-golden-image prompt output and EULA handling"
+spec:
+  requires:
+    - extension: kubernetes
+      as: k8s
+  setup:
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: windows-golden-image-no-eula
+        ignoreNotFound: true
+    - k8s.create:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: windows-golden-image-no-eula
+  verify:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          set -euo pipefail
+          NS="windows-golden-image-no-eula"
+
+          echo "=== Verification: Windows Golden Image Prompt Output ==="
+
+          # The task instructs the agent to decline the EULA when asked.
+          # A PipelineRun being present means the agent ignored that instruction.
+          PIPELINERUN_COUNT=$(kubectl get pipelinerun -n "$NS" 2>/dev/null | grep -c "windows" || true)
+
+          if [ "$PIPELINERUN_COUNT" -gt "0" ]; then
+            echo "✗ PipelineRun was created despite EULA not being accepted — agent disobeyed instructions"
+            exit 1
+          fi
+
+          echo "✓ No PipelineRun created (correct — EULA was not accepted)"
+          echo ""
+          echo "=== Prompt Output Verification Complete ==="
+          exit 0
+  cleanup:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          kubectl delete pipelinerun --all -n windows-golden-image-no-eula --ignore-not-found=true 2>/dev/null || true
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: windows-golden-image-no-eula
+        ignoreNotFound: true
+  prompt:
+    inline: |-
+      Use the windows-golden-image prompt to create a Windows 11 golden image.
+
+      Use this ISO URL: https://example.com/windows11.iso
+
+      Use namespace: windows-golden-image-no-eula.
+
+      When asked about the Microsoft EULA, answer that you do not accept it.
--- a/evals/tasks/kubevirt/windows-golden-image-success/task.yaml
+++ b/evals/tasks/kubevirt/windows-golden-image-success/task.yaml
@@ -1,0 +1,77 @@
+kind: Task
+apiVersion: mcpchecker/v1alpha2
+metadata:
+  labels:
+    suite: kubevirt
+    requires: kubevirt,tekton
+  name: "windows-golden-image-success"
+  difficulty: hard
+  description: "Use the windows-golden-image prompt to generate a Windows golden image creation guide"
+spec:
+  requires:
+    - extension: kubernetes
+      as: k8s
+  setup:
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: windows-golden-image-success
+        ignoreNotFound: true
+    - k8s.create:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: windows-golden-image-success
+  verify:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          set -euo pipefail
+          NS="windows-golden-image-success"
+
+          PIPELINERUN_COUNT=$(kubectl get pipelinerun -n "$NS" 2>/dev/null | grep -c "windows" || true)
+          if [ "$PIPELINERUN_COUNT" -lt "1" ]; then
+            echo "✗ No PipelineRun created — agent must create one after EULA acceptance"
+            exit 1
+          fi
+
+          PR=$(kubectl get pipelinerun -n "$NS" -o name | grep "windows" | head -1)
+          echo "✓ PipelineRun created: $PR"
+
+          get() { kubectl get "$PR" -n "$NS" -o "jsonpath=$1" 2>/dev/null || true; }
+          check() { local v; v=$(get "$2"); [[ "$v" == "$3" ]] || { echo "✗ $1: expected '$3', got '$v'"; exit 1; }; echo "  ✓ $1: $v"; }
+
+          [[ -n "$(get '{.spec.params[?(@.name=="winImageDownloadURL")].value}')" ]] \
+            || { echo "✗ winImageDownloadURL param missing"; exit 1; }
+
+          check "acceptEula"    '{.spec.params[?(@.name=="acceptEula")].value}'             "true"
+          check "resolver"      '{.spec.pipelineRef.resolver}'                              "hub"
+          check "pipeline name" '{.spec.pipelineRef.params[?(@.name=="name")].value}'      "windows-efi-installer"
+          check "catalog"       '{.spec.pipelineRef.params[?(@.name=="catalog")].value}'   "kubevirt-tekton-pipelines"
+
+          VERSION=$(get '{.spec.pipelineRef.params[?(@.name=="version")].value}')
+          if [[ -n "$VERSION" ]] && ! [[ "$VERSION" =~ ^0\.[0-9]+\.[0-9]+$ ]]; then
+            echo "✗ version '$VERSION' does not match expected format (0.xx.0)"; exit 1
+          fi
+          echo "  ✓ version: ${VERSION:-latest}"
+  cleanup:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          kubectl delete pipelinerun --all -n windows-golden-image-success --ignore-not-found=true 2>/dev/null || true
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: windows-golden-image-success
+        ignoreNotFound: true
+  prompt:
+    inline: |-
+      Use the windows-golden-image prompt to create a Windows Server 2022 golden image.
+      Use this ISO URL:
+      https://example.com/windows-server-2022.iso
+
+      Use namespace: windows-golden-image-success.
+
+      I accept the Microsoft EULA.
--- a/pkg/kubernetes/windows_eula_validator.go
+++ b/pkg/kubernetes/windows_eula_validator.go
@@ -1,0 +1,136 @@
+package kubernetes
+
+import (
+	"context"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	"github.com/containers/kubernetes-mcp-server/pkg/mcplog"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	"sigs.k8s.io/yaml"
+)
+
+// WindowsEULAValidator enforces EULA acceptance for Windows golden image PipelineRuns.
+// This prevents LLMs from bypassing EULA prompts by directly calling resources_create_or_update.
+type WindowsEULAValidator struct{}
+
+// elicitor defines the interface for elicitation support.
+type elicitor interface {
+	Elicit(context.Context, *mcp.ElicitParams) (*mcp.ElicitResult, error)
+}
+
+func (v *WindowsEULAValidator) Name() string {
+	return "WindowsEULAValidator"
+}
+
+func (v *WindowsEULAValidator) Validate(ctx context.Context, req *api.HTTPValidationRequest) error {
+	if !isWindowsPipelineRun(req) {
+		return nil
+	}
+
+	pr := &tektonv1.PipelineRun{}
+	if err := yaml.Unmarshal(req.Body, pr); err != nil {
+		// Non-apply patch bodies (JSON-patch arrays, strategic-merge fragments)
+		// aren't full PipelineRun objects and so aren't golden-image creations;
+		// skip rather than reject the request.
+		return nil
+	}
+
+	if !requiresEULAPrompt(pr) {
+		return nil
+	}
+
+	return promptForEULA(ctx)
+}
+
+// isWindowsPipelineRun checks if this request should be validated.
+func isWindowsPipelineRun(req *api.HTTPValidationRequest) bool {
+	// Validate create/update and server-side apply (HTTP PATCH → "patch")
+	if req.Verb != "create" && req.Verb != "update" && req.Verb != "patch" {
+		return false
+	}
+
+	// Only validate PipelineRun resources
+	return req.GVK != nil && req.GVK.Kind == "PipelineRun" && req.GVK.Group == "tekton.dev"
+}
+
+// requiresEULAPrompt checks if the PipelineRun requires EULA acceptance.
+func requiresEULAPrompt(pr *tektonv1.PipelineRun) bool {
+	// Check if this is a windows-efi-installer pipeline
+	if !isWindowsEFIInstallerPipeline(pr) {
+		return false
+	}
+
+	// Check if EULA is already accepted
+	return getPipelineParam(pr.Spec.Params, "acceptEula") != "true"
+}
+
+// promptForEULA prompts the user to accept the Windows EULA via elicitation.
+func promptForEULA(ctx context.Context) error {
+	session, err := getElicitorFromContext(ctx)
+	if err != nil {
+		return err
+	}
+
+	result, err := session.Elicit(ctx, &mcp.ElicitParams{
+		Message: "Microsoft EULA Notice: This PipelineRun will create a Windows golden image. By proceeding, you agree to the Microsoft Software License Terms for the Windows operating system. Do you accept the Microsoft End User License Agreement (EULA)?",
+	})
+	if err != nil {
+		if isElicitationNotSupportedError(err) {
+			return &api.ValidationError{
+				Code:    api.ErrorCodePermissionDenied,
+				Message: "Windows EULA acceptance required. Set acceptEula parameter to \"true\" in the PipelineRun to proceed.",
+			}
+		}
+		return err
+	}
+
+	if result.Action != api.ElicitActionAccept {
+		return &api.ValidationError{
+			Code:    api.ErrorCodePermissionDenied,
+			Message: "Windows EULA was not accepted. The PipelineRun creation has been cancelled.",
+		}
+	}
+
+	return nil
+}
+
+// getElicitorFromContext retrieves an elicitor from the request context.
+func getElicitorFromContext(ctx context.Context) (elicitor, error) {
+	sessionValue := ctx.Value(mcplog.MCPSessionContextKey)
+	if sessionValue == nil {
+		return nil, &api.ValidationError{
+			Code:    api.ErrorCodePermissionDenied,
+			Message: "Windows EULA acceptance required but no session available for confirmation",
+		}
+	}
+
+	session, ok := sessionValue.(elicitor)
+	if !ok {
+		return nil, &api.ValidationError{
+			Code:    api.ErrorCodePermissionDenied,
+			Message: "Windows EULA acceptance required but session does not support elicitation",
+		}
+	}
+
+	return session, nil
+}
+
+// isWindowsEFIInstallerPipeline checks if the PipelineRun references the windows-efi-installer pipeline.
+func isWindowsEFIInstallerPipeline(pr *tektonv1.PipelineRun) bool {
+	if pr.Spec.PipelineRef == nil {
+		return false
+	}
+
+	return getPipelineParam(pr.Spec.PipelineRef.Params, "name") == "windows-efi-installer"
+}
+
+// getPipelineParam retrieves a parameter value by name from a Params slice.
+func getPipelineParam(params tektonv1.Params, name string) string {
+	for _, param := range params {
+		if param.Name == name {
+			return param.Value.StringVal
+		}
+	}
+	return ""
+}
--- a/pkg/toolsets/kubevirt/windows_golden_image.go
+++ b/pkg/toolsets/kubevirt/windows_golden_image.go
@@ -1,0 +1,345 @@
+package kubevirt
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/yaml"
+)
+
+type windowsDefaults struct {
+	preferenceName            string
+	autounattendConfigMapName string
+	baseDvName                string
+	isoDvName                 string
+	generateName              string
+}
+
+// artifactHubPackage represents the Artifact Hub API response for a package
+type artifactHubPackage struct {
+	Readme string `json:"readme"`
+}
+
+const (
+	artifactHubAPIBase  = "https://artifacthub.io/api/v1/packages/tekton-pipeline/kubevirt-tekton-pipelines/windows-efi-installer"
+	httpTimeout         = 10 * time.Second
+	maxResponseBodySize = 512 << 10 // 512 KiB
+)
+
+// fetchPipelineRunExample is a variable to allow mocking in tests
+var fetchPipelineRunExample = fetchPipelineRunExampleFromArtifactHub
+
+var windowsVersionDefaults = map[string]windowsDefaults{
+	"10": {
+		preferenceName:            "windows.10.virtio",
+		autounattendConfigMapName: "windows10-autounattend",
+		baseDvName:                "win10",
+		isoDvName:                 "win10",
+		generateName:              "windows10-installer-run-",
+	},
+	"11": {
+		preferenceName:            "windows.11.virtio",
+		autounattendConfigMapName: "windows11-autounattend",
+		baseDvName:                "win11",
+		isoDvName:                 "win11",
+		generateName:              "windows11-installer-run-",
+	},
+	"2k22": {
+		preferenceName:            "windows.2k22.virtio",
+		autounattendConfigMapName: "windows2k22-autounattend",
+		baseDvName:                "win2k22",
+		isoDvName:                 "win2k22",
+		generateName:              "windows2k22-installer-run-",
+	},
+	"2k25": {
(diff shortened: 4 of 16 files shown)


ksimon1:
@lyarwood, @manusa, @Cali0707  can you please review this PR?

pkg/toolsets/kubevirt/windows_golden_image.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("artifact Hub returned status %d for version %q", resp.StatusCode, version)
+	}
+
+	body, err := io.ReadAll(resp.Body)
```
matzew: Should this use `io.LimitReader` to prevent unexpectedly large response?

Rest of the code base uses that too. E.g. like: https://github.com/containers/kubernetes-mcp-server/pull/927

pkg/toolsets/kubevirt/windows_golden_image.go
```diff
@@ -0,0 +1,344 @@
...
+	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+)
+
+type windowsDefaults struct {
+	preferenceName            string
+	autounattendConfigMapName string
+	baseDvName                string
+	isoDVName                 string
```
lyarwood: nit: Inconsistent casing between `baseDvName` and `isoDVName` - either `baseDVName`/`isoDVName` or `baseDvName`/`isoDvName` would be more consistent.

pkg/toolsets/kubevirt/windows_golden_image.go
```diff
@@ -0,0 +1,344 @@
...
+		preferenceName:            "windows.2k25.virtio",
+		autounattendConfigMapName: "windows2k25-autounattend",
+		baseDvName:                "win2k25",
+		isoDVName:                 "win2k25",
+		generateName:              "windows2k25-installer-run-",
+	},
+}
+
+func initWindowsGoldenImage() []api.ServerPrompt {
+	clusterAware := false
```
lyarwood: The rest of the codebase uses `ptr.To(false)` from `k8s.io/utils/ptr` for this pattern (e.g. `pkg/toolsets/config/configuration.go`, `pkg/toolsets/kubevirt/vm/create/tool.go`). Suggest switching for consistency:
```go
ClusterAware: ptr.To(false),
```

pkg/toolsets/kubevirt/windows_golden_image.go
```diff
@@ -0,0 +1,344 @@
...
+
+func resolveWindowsDefaults(version string) *windowsDefaults {
+	defaults, ok := windowsVersionDefaults[version]
+	if !ok {
+		return nil
+	}
+	return &defaults
+}
+
+func buildGuideText(pipelineRunYAML string) string {
```
lyarwood, line 165: This function builds a large markdown string via 40+ concatenation operations. Consider using a backtick raw string literal with `fmt.Sprintf` or `strings.ReplaceAll`, similar to how `vm_troubleshoot.go` handles its guide text. Would be easier to read and maintain.

manusa:
Hi @ksimon1, sorry for the late reply. What's the state of this PR, is it mergeable? Shall I review?

ksimon1:
/hold cancel
Hi @manusa, the PR should be ready for review now

evals/tasks/kubevirt/windows-golden-image-success/task.yaml
```diff
@@ -0,0 +1,77 @@
...
+          [[ -n "$(get '{.spec.params[?(@.name=="winImageDownloadURL")].value}')" ]] \
+            || { echo "✗ winImageDownloadURL param missing"; exit 1; }
+
+          check "acceptEula"    '{.spec.params[?(@.name=="acceptEula")].value}'             "true"
+          check "resolver"      '{.spec.pipelineRef.resolver}'                              "hub"
+          check "pipeline name" '{.spec.pipelineRef.params[?(@.name=="name")].value}'      "windows-efi-installer"
+          check "catalog"       '{.spec.pipelineRef.params[?(@.name=="catalog")].value}'   "kubevirt-tekton-pipelines"
+
+          VERSION=$(get '{.spec.pipelineRef.params[?(@.name=="version")].value}')
+          if [[ -n "$VERSION" ]] && ! [[ "$VERSION" =~ ^0\.[0-9]+\.[0-9]+$ ]]; then
```
manusa: Flagged by a reviewer agent:

  This regex will reject the value the prompt actually produces when `pipelineVersion` is unspecified (which is the case for this eval's prompt).

  Trace:
  1. The prompt builds the PipelineRun YAML by extracting a `yaml` block from the Artifact Hub package's `readme` field (`pkg/toolsets/kubevirt/windows_golden_image.go:285`).
  2. The upstream readme literally contains `value: v0.25.0` (with a `v` prefix) — verify with:
     ```bash
     curl -s 'https://artifacthub.io/api/v1/packages/tekton-pipeline/kubevirt-tekton-pipelines/windows-efi-installer' \
       | jq -r '.readme' | grep -A1 'name: version'
  3. setParams only overwrites the version param when pipelineVersion != "" (windows_golden_image.go:320), so with no override the v0.25.0 from the readme survives into the applied PipelineRun.
  4. This regex ^0\.[0-9]+\.[0-9]+$ then rejects v0.25.0 and the verify exits 1.

  Two ways to fix:
  - Widen this regex to ^v?0\.[0-9]+\.[0-9]+$, or
  - Strip the leading v in buildPipelineRunYAML so the PipelineRun is normalized regardless of upstream
  formatting.

  The second option is probably the better fix — the readme uses v-prefixed Tekton Hub tags, but a downstream pipeline-version override (pipelineVersion: 0.25.0, currently validated by reVersionFormat = ^\d+\.\d+\.\d+$) does not. Today the produced YAML's version field is inconsistent depending on whether the user supplied
  pipelineVersion.

pkg/toolsets/kubevirt/windows_golden_image.go
```diff
@@ -0,0 +1,354 @@
...
+
+var (
+	reCodeBlock   = regexp.MustCompile("(?s)```(?:yaml|YAML)\n(.*?)```")
+	rePipelineRun = regexp.MustCompile(`(?s)(apiVersion:.*?kind: PipelineRun.*?)(?:\nEOF\b|$)`)
+)
+
+// Handles yaml code blocks that wrap the YAML inside shell heredocs (<<EOF...EOF).
+func extractPipelineRunFromReadme(readme string) (string, error) {
+	for _, block := range reCodeBlock.FindAllStringSubmatch(readme, -1) {
+		if m := rePipelineRun.FindStringSubmatch(block[1]); m != nil {
```
manusa: Flagged by review agent:

  This returns the **first** matching `yaml` code block, but the upstream readme contains four PipelineRun examples in this order: `windows11-installer-run-`, `windows10-installer-run-`, `windows2k22-installer-run-`, `windows2k25-installer-run-`. 

Verify with:

  ```bash
  curl -s 'https://artifacthub.io/api/v1/packages/tekton-pipeline/kubevirt-tekton-pipelines/windows-efi-installer' \
    | jq -r '.readme' | grep -E 'generateName: windows[0-9k]+-installer-run-'
  ```

  So for windowsVersion=10|2k22|2k25, the prompt always extracts the win11 block, then setParams overwrites preferenceName / autounattendConfigMapName / baseDvName / isoDVName and the ObjectMeta.GenerateName (windows_golden_image.go:307, 311-318). That masks the wrong-block selection for those four fields, but fields not in the override set silently leak from the win11 block:

  - The win10/2k22/2k25 upstream blocks include timeout: 1h0m0s at spec.timeout; the win11 block does not.
  Picking block #0 drops the explicit timeout for those versions and falls back to the controller default.
  - Anything upstream adds to the per-version blocks in the future (additional params, taskRunSpecs, workspaces, etc.) will not propagate, defeating the stated forward-compatibility purpose of fetching the readme at runtime (per docs/kubevirt.md).

  Fix: filter reCodeBlock.FindAllStringSubmatch matches and pick the block whose generateName matches winDefaults.generateName. Fall back to a clear error if no matching block is found rather than silently using a different version's example.

pkg/toolsets/kubevirt/windows_golden_image.go
```diff
@@ -0,0 +1,354 @@
...
+
+var fetchPipelineRunExample = fetchPipelineRunExampleFromArtifactHub
+var fetchArtifactHubReadme = fetchReadmeFromArtifactHub
+
+var reVersionFormat = regexp.MustCompile(`^\d+\.\d+\.\d+$`)
+
+var windowsVersionDefaults = map[string]windowsDefaults{
+	"10": {
+		preferenceName:            "windows.10.virtio",
+		autounattendConfigMapName: "windows10-autounattend",
```
manusa: Flagged by review agent:

  **File:** `pkg/toolsets/kubevirt/windows_golden_image.go`
  **Line:** 45

  ```markdown
  The upstream readme's win10 PipelineRun example sets:
  ```

  ```yaml
  -   name: autounattendConfigMapName
      value: windows10-efi-autounattend
  ```

  (verify: `curl -s '.../windows-efi-installer' | jq -r '.readme' | grep -B1 -A1 'autounattendConfigMapName'`)

  This constant uses windows10-autounattend (no -efi-), which doesn't match what the pipeline expects. The pipeline provisions ConfigMaps from autounattendXMLConfigMapsURL under the upstream-documented names — windows10-efi-autounattend is what's published in the [pipelines/windows-efi-installer/configmaps directory](https://github.com/kubevirt/kubevirt-tekton-tasks/tree/main/release/pipelines/windows-efi-installer/configmaps).
  With this constant, Windows 10 installs will fail at the autounattend mount step with a missing-ConfigMap error.

  Fix: change to windows10-efi-autounattend. Worth re-verifying the win11/2k22/2k25 names against the same upstream directory while you're at it — the readme blocks are the source of truth.

manusa:
```
=== Evaluation Results ===
Results file: mcpchecker-openai-agent-kubernetes-test-out.json
=== Evaluation Summary ===

  ✓ clone-vm (assertions: 3/3)
  ✓ create-basic-vm (assertions: 3/3)
  ✓ create-ubuntu-vm (assertions: 3/3)
  ✓ create-vm-with-instancetype (assertions: 3/3)
  ✓ create-vm-with-size (assertions: 3/3)
  ✓ create-vm-with-vlan (assertions: 3/3)
  ✓ delete-vm (assertions: 3/3)
  ✓ restore-vm (assertions: 3/3)
  ✓ snapshot-vm (assertions: 3/3)
  ✓ troubleshoot-vm (assertions: 3/3)
  ✓ update-vm-resources (assertions: 3/3)
  ~ windows-golden-image-no-eula (assertions: 1/3)
      - ToolsUsed: Required tool not called: server=kubernetes, tool=, pattern=.*
      - MinToolCalls: Too few tool calls: expected >= 1, got 0
  ✗ windows-golden-image-success (assertions: 3/3)
      verification failed: verify[0] failed: script execution failed: exit status 1
output: ✗ No PipelineRun created — agent must create one after EULA acceptance


Tasks:      12/13 passed (92.31%)
Assertions: 37/39 passed (94.87%)
Tokens:     ~277179 (estimate - excludes system prompt & cache)
MCP schemas: ~51441 (included in token total)
Agent used tokens:
  Input:  44239 tokens
  Output: 34668 tokens
=== Threshold Verification ===

Task Pass Rate:      92.31% >= 80.00% ✓
Assertion Pass Rate: 94.87% >= 80.00% ✓

Result: PASSED
```

manusa (changes requested):
Check comments and mcpchecker results. It seems that both windows-golden-image tasks fail end-to-end.

pkg/toolsets/kubevirt/windows_golden_image.go
ksimon1: fixed

ksimon1:
@manusa can you please trigger the mcp checker?

manusa:
```
=== Evaluation Results ===
Results file: mcpchecker-openai-agent-kubernetes-test-out.json
=== Evaluation Summary ===

  ✓ clone-vm (assertions: 3/3)
  ✓ create-basic-vm (assertions: 3/3)
  ✓ create-ubuntu-vm (assertions: 3/3)
  ✓ create-vm-with-instancetype (assertions: 3/3)
  ✓ create-vm-with-size (assertions: 3/3)
  ✓ create-vm-with-vlan (assertions: 3/3)
  ✓ delete-vm (assertions: 3/3)
  ✓ restore-vm (assertions: 3/3)
  ✓ snapshot-vm (assertions: 3/3)
  ~ troubleshoot-vm (assertions: 2/3)
      - MaxToolCalls: Too many tool calls: expected <= 20, got 21
  ✓ update-vm-resources (assertions: 3/3)
  ✓ windows-golden-image-no-eula (assertions: 3/3)
  ✗ windows-golden-image-success (assertions: 3/3)
      verification failed: verify[0] failed: script execution failed: exit status 1
output: ✗ No PipelineRun created — agent must create one after EULA acceptance


Tasks:      12/13 passed (92.31%)
Assertions: 38/39 passed (97.44%)
Tokens:     ~630542 (estimate - excludes system prompt & cache)
MCP schemas: ~51441 (included in token total)
Agent used tokens:
  Input:  81212 tokens
  Output: 63314 tokens
=== Threshold Verification ===

Task Pass Rate:      92.31% >= 80.00% ✓
Assertion Pass Rate: 97.44% >= 80.00% ✓

Result: PASSED
```

manusa (changes requested):
The previous round's three inline comments on `76d1acb` are addressed in `49364ae`. Remaining concerns below are independent.

### Issues

**1. EULA "enforcement" is documentation, not enforcement — the success eval proves it.**

`windows_golden_image.go:673` always emits `acceptEula: "false"`, and the generated guide asks the LLM to *manually edit* the field to `"true"` and then call `resources_create_or_update`. There is no MCP-side gate. On `49364ae`, `windows-golden-image-success` still fails with "No PipelineRun created — agent must create one after EULA acceptance" — the agent reliably does not perform the edit-then-apply chain. The 12/13 / 80% threshold green check is masking that the PR's own success-path eval is broken. `no-eula` only passes because doing nothing is the easier LLM path.

Real enforcement options (any one):
- An `Elicitor` round-trip from inside the handler (`pkg/api/prompts.go:91`) — only set `acceptEula: "true"` after elicitation confirms.
- A confirmation rule under `pkg/mcp/confirmation_*` that gates `resources_create_or_update` when the PipelineRun has the `windows-efi-installer` pipelineRef.
- A dedicated kubevirt tool with an explicit `acceptEula bool` argument; the prompt instructs the agent to call that tool; the tool errors on `false`.

Fixing this will likely also make the success eval pass.

**2. Round-tripping an unauthenticated Artifact Hub README into cluster-applied YAML.**

`fetchReadmeFromArtifactHub` (`windows_golden_image.go:599-624`) pulls the README, regex-matches a YAML block, unmarshals into `tektonv1.PipelineRun`, then re-marshals. Only six `spec.params` plus `TypeMeta`/`ObjectMeta` are overwritten (`windows_golden_image.go:665-684`); every other field (`taskRunSpecs[].podTemplate.securityContext`, `workspaces`, `serviceAccountName`, `timeouts`…) flows through verbatim. `TestDefaultVersion` asserts `fsGroup: 107` survives the round-trip — confirming the passthrough. Default `pipelineVersion=""` resolves to "latest" with no integrity check.

**Can we pin to a SHA?** I checked the Artifact Hub API:

```bash
$ curl -s 'https://artifacthub.io/api/v1/packages/tekton-pipeline/kubevirt-tekton-pipelines/windows-efi-installer/0.25.0' \
    | jq '{version, digest, signed, content_url}'
{
  "version": "0.25.0",
  "digest": "a3c9e62a4f4ec70042155a152e492cf7b7da42752f3c50dd0da5985fefaf7836",
  "signed": false,
  "content_url": "https://github.com/kubevirt/kubevirt-tekton-tasks/raw/v0.25.0/release/pipelines/windows-efi-installer/windows-efi-installer.yaml"
}
```

Findings:
- `signed: false` — no cryptographic provenance available.
- `digest` covers only the `content_url` body (the Pipeline manifest), not the `readme` field. Verified: `curl -sL <content_url> | sha256sum` matches the digest exactly. The README the code actually parses has no digest.
- `content_url` is pinned to a git tag (mutable in principle, but post-publish drift is detectable via digest mismatch).

So **digest-pinning the thing the code actually scrapes is not possible** with Artifact Hub. Two paths forward:

1. **Recommended — construct the PipelineRun in Go from a static per-Windows-version template.** The PipelineRun is small and stable (six params + `pipelineRef` to the hub resolver + one `taskRunSpec` + a timeout) — the test mock shows the full surface area in ~25 lines. The Pipeline itself is still resolved at runtime by Tekton via `resolver: hub`, which is Tekton's trust boundary, not the MCP server's. Building from a Go struct eliminates the network call, the README scraping, the regex parser, the YAML round-trip, and the supply-chain attack surface in one stroke. The "forward compatibility" rationale in `docs/kubevirt.md` is better served by versioning the Go template alongside `go.mod` bumps of `tektoncd/pipeline`. This single change also resolves findings #3, #4, and #6.

2. **Lighter alternative — if dynamic fetch is preserved:** (a) require explicit `pipelineVersion` (drop the "latest" default), (b) fetch `content_url` instead of `readme` and verify SHA256 against a pinned `digest` in the binary, (c) build the PipelineRun in Go and use the fetched manifest only for parameter reference, not as a template to round-trip. This is more code than option 1 for the same outcome.

**3. HTTP request ignores prompt context.**

`windows_golden_image.go:600` calls `client.Get(url)`. `api.PromptHandlerParams` embeds `context.Context` (`pkg/api/prompts.go:85-91`); `vm_troubleshoot.go:58-61` already establishes the pattern. Thread `ctx` through `windowsGoldenImageHandler` → `buildPipelineRunYAML` → `fetchPipelineRunExample` → `fetchReadmeFromArtifactHub` and use `http.NewRequestWithContext`. (Moot if #2 is resolved via static template.)

**4. Package-level mutable function vars are a data race and a leaky test seam.**

`windows_golden_image.go:384-385` declares `fetchPipelineRunExample` and `fetchArtifactHubReadme`; tests swap them in `SetupTest`/`TearDownTest` (`_test.go:786-794`). Under `go test -race` with any other suite invoking the handler concurrently, this races.

**Recommended approach:** if #2 is resolved by emitting the PipelineRun from a static Go template, this finding disappears entirely. If dynamic fetching is kept, replace the globals with a fetcher interface injected at registration:

```go
type pipelineRunFetcher interface {
    Fetch(ctx context.Context, version, generateName string) (string, error)
}

func initWindowsGoldenImage() []api.ServerPrompt {
    return initWindowsGoldenImageWithFetcher(newArtifactHubFetcher(http.DefaultClient))
}

func initWindowsGoldenImageWithFetcher(fetcher pipelineRunFetcher) []api.ServerPrompt {
    handler := func(params api.PromptHandlerParams) (*api.PromptCallResult, error) {
        return windowsGoldenImageHandler(params, fetcher)
    }
    return []api.ServerPrompt{{Prompt: /* ... */, Handler: handler, ClusterAware: ptr.To(false)}}
}
```

Tests pass a fake fetcher to `initWindowsGoldenImageWithFetcher`. No globals, no race, dependency visible in the signature, and ctx-propagation (#3) becomes the fetcher's concern.

**5. `winImageDownloadURL` has no scheme validation.**

`windows_golden_image.go:458,672` — LLM-supplied string is inserted as-is into the PipelineRun param consumed by the `modify-windows-iso-file` Tekton task. `file:///`, `http://169.254.169.254/...` (cloud-metadata SSRF from inside the cluster), or internal-only URLs are all accepted. Add `url.Parse` + require `https://` in `windowsGoldenImageHandler` before embedding.

**6. Downstream override `WindowsEFIInstallerCatalog` lacks validation.**

`internal/defaults/defaults.go:25-28` substitutes the override blindly. A fork setting `windowsEFIInstallerCatalogOverride = "http://internal-mirror/..."` silently drops TLS. Validate at startup: parseable, non-empty host, `https://` scheme. (Moot if #2 is resolved via static template.)

### Assessment: Ready to merge? No

#1 needs a real enforcement primitive (Elicitor / confirmation rule / dedicated tool), not prose. #2 is best resolved by emitting the PipelineRun from an in-repo static template — that single change also resolves #3, #4, and #6.

manusa (changes requested):
Nice work pivoting to the static Go template — that single change resolved most of the prior round's concerns cleanly (no more network call, no supply-chain surface, no per-version-block bug, no mutable globals, no `-race` exposure). Adding the Elicitor was the right primitive, and the URL/version validators are in the right place. Unit tests pass under `-race`.

Three things still need fixing before this can merge. I've left the other observations (URL empty-host check, sentinel-with-error coupling, weak no-eula verify, misleading helper name) for a follow-up so we can focus on the blockers.

### 1. EULA enforcement empirically regresses on `c9031dc`

The latest mcpchecker run shows `windows-golden-image-no-eula` is now **failing** (it passed on `49364ae`), and `windows-golden-image-success` continues to fail. The no-eula failure mode is "PipelineRun was created despite EULA not being accepted" — so a PipelineRun was actually applied even though the task said to decline. The most likely cause is that the OpenAI agent treats the MCP elicitation request as a separate one-click form and auto-accepts it, ignoring the natural-language "do not accept" instruction in the task prompt.

The cleanest fix is to move EULA gating server-side via a `confirmation_rules` entry in `pkg/kubernetes/confirmation_validator.go` that intercepts `resources_create_or_update` when the PipelineRun's `pipelineRef.params.name == "windows-efi-installer"`. That makes the gate non-bypassable by the LLM, works uniformly across MCP clients regardless of elicitation support, and aligns with the existing `ConfirmationValidator` pattern in the codebase. Other approaches (richer elicitation schema, or returning the EULA question as a prompt message instead of a server-side elicit) would also work — happy to discuss tradeoffs if you'd prefer a different direction.

### 2. URL validator accepts embedded credentials (`pkg/toolsets/kubevirt/windows_golden_image.go:109-112`)

`https://user:pass@host/iso` parses cleanly today and lands verbatim in the PipelineRun spec, which means credentials end up in etcd, `kubectl get pipelinerun -o yaml`, audit logs, and any MCP `resources_get` call. One-line fix:

```go
if err != nil || parsedURL.Scheme != "https" || parsedURL.User != nil {
    return nil, fmt.Errorf("winImageDownloadURL must be an https:// URL without embedded credentials, got %q", winImageDownloadURL)
}
```

### 3. `pipelineVersion` test assertions are vacuous (`pkg/toolsets/kubevirt/windows_golden_image_test.go:327, 335`)

When the version param is present, the generated YAML serializes as `- name: version\n  value: 0.25.0` — the substring `version:` is never present (the only colon-suffixed match is `apiVersion:`, capital V). Verified by dumping the marshaled YAML.

That means:
- `s.NotContains(yaml, "version:")` at line 327 passes regardless of whether the version param exists.
- `s.Contains(yaml, "version")` at line 335 matches `apiVersion:` regardless.

The only assertion doing real work is the adjacent `Contains "0.25.0"`. Suggest anchoring on `- name: version` (or `name: version\n`) so the absence/presence check actually distinguishes the two cases.

ksimon1:
@manusa can you please review this PR?

ksimon1:
@Cali0707 can you please trigger the eval tests?

manusa:
I think the related tasks are still failing 😞

ksimon1:
@manusa Agent did not trigger the prompt and from offline slack communication it looks like, we can't trigger prompts from the tests?

Cali0707:
@ksimon1 I think in general this is a UX issue of MCP prompts - clients handle the UX differently. I'll see if I can look into a way to "force" using a prompt through mcpchecker but for now I would say we probably can't reliably eval prompts due to how clients all implement them differently

ksimon1:
@Cali0707 so should I remove the eval tests?

Cali0707:
Yeah I think that would be reasonable for now. If I find a way to get prompts to work consistently between clients I will let you know

ksimon1:
I renamed the task.yaml to prompt.yaml - IINM this way the prompts should not be triggered?

evals/tasks/kubevirt/windows-golden-image-no-eula/prompt.yaml
```diff
@@ -0,0 +1,67 @@
+kind: Task
```
Cali0707: @ksimon1 you will need to switch the `kind: Task` to something else (renaming the file would only work if we were explicitly matching on `task.yaml` which changes depending on which eval.yaml you are using, e.g. here it matches all .yaml files: https://github.com/containers/kubernetes-mcp-server/blob/624a2303fad530ecbaf1be67c0e0c71be54c5a40/evals/claude-code/eval.yaml#L46-L54

Note that any file that matches but does not unmarshal as a task is skipped, hence switching the kind should work

ksimon1: I was checking the https://github.com/containers/kubernetes-mcp-server/blob/main/evals/tasks/kubevirt/claude-code/eval.yaml#L12. In that case it would work, but the one you sent would takes the renamed files.

ksimon1:
@Cali0707, @manusa is the PR now ok?

ksimon1:
@Cali0707, @manusa is the PR now ok?

Cali0707 (approved):
LGTM

pkg/kubernetes/windows_eula_validator.go
```diff
@@ -0,0 +1,133 @@
...
+
+	return promptForEULA(ctx)
+}
+
+// isWindowsPipelineRun checks if this request should be validated.
+func isWindowsPipelineRun(req *api.HTTPValidationRequest) bool {
+	// Only validate create/update operations
+	if req.Verb != "create" && req.Verb != "update" {
+		return false
+	}
```
manusa: This validator's doc comment (lines 13-14) says it "prevents LLMs from bypassing EULA prompts by directly calling `resources_create_or_update`" — but as written it never fires on that path. `resources_create_or_update` does a **server-side apply** (`pkg/kubernetes/resources.go:193` → `.Apply(...)`) = HTTP `PATCH`; `httpMethodToVerb` maps `PATCH`→`"patch"` (`accesscontrol_round_tripper.go:281`), so this returns `false` and validation is skipped. The tests stay green only because they use `Verb: "create"`.

This also isn't backstopped downstream: the upstream pipeline's gate is `if [ "$(params.acceptEula)" ]`, a non-empty-string test that's *true* for the literal `"false"` (it only blocks an empty value) — so this validator is the only real server-side gate.

Quickest self-contained fix is to accept `"patch"` here, plus a small guard at the unmarshal (see my other comment) so non-apply patch bodies aren't rejected:

```suggestion
	// Validate create/update and server-side apply (an HTTP PATCH → "patch")
	if req.Verb != "create" && req.Verb != "update" && req.Verb != "patch" {
		return false
	}
```

And a test that drives the real path so this can't regress (add to `windows_eula_validator_test.go`):

```go
	t.Run("validates server-side apply (PATCH) of windows-efi-installer", func(t *testing.T) {
		req := &api.HTTPValidationRequest{
			GVK:  &schema.GroupVersionKind{Group: "tekton.dev", Version: "v1", Kind: "PipelineRun"},
			Verb: "patch",
			Body: []byte(`apiVersion: tekton.dev/v1
kind: PipelineRun
metadata:
  generateName: windows2k22-installer-run-
spec:
  params:
  - name: acceptEula
    value: "false"
  pipelineRef:
    resolver: hub
    params:
    - name: name
      value: windows-efi-installer`),
		}
		err := validator.Validate(context.Background(), req)
		assert.Error(t, err) // now fires on the apply path
	})
```

pkg/kubernetes/windows_eula_validator.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (v *WindowsEULAValidator) Validate(ctx context.Context, req *api.HTTPValidationRequest) error {
+	if !isWindowsPipelineRun(req) {
+		return nil
+	}
+
+	pr := &tektonv1.PipelineRun{}
+	if err := yaml.Unmarshal(req.Body, pr); err != nil {
+		return err
+	}
```
manusa, line 37: Companion to the verb fix: once `"patch"` is accepted, non-apply patch bodies (JSON-patch arrays, strategic-merge fragments) won't unmarshal into a full `PipelineRun` and would otherwise fail unrelated patches. Skip those instead of erroring — server-side apply always sends the full object, so the golden-image path is still validated:

```suggestion
	pr := &tektonv1.PipelineRun{}
	if err := yaml.Unmarshal(req.Body, pr); err != nil {
		// Non-apply patch bodies (JSON-patch arrays, strategic-merge fragments)
		// aren't full PipelineRun objects and so aren't golden-image creations;
		// skip rather than reject the request.
		return nil
	}
```

manusa (changes requested):
Thanks for the iteration — the move to a static `buildStaticPipelineRun` template is a solid improvement, and I verified the generated PipelineRun is faithful to the upstream `kubevirt-tekton-tasks` examples (param names, per-version `preferenceName`/`autounattendConfigMapName` incl. the win10 `-efi-` asymmetry, `baseDvName`/`isoDVName`, `runAsUser`/`fsGroup: 107`). That resolves the earlier round of findings.

**One blocker before merge:** the `WindowsEULAValidator` never fires on the path it's meant to guard. I've left one-click suggestions that fix it self-contained in the one file (I applied them locally and confirmed the guard now triggers on the apply path with all existing tests still green). With those in, this should be good to go — happy to re-confirm and approve.

pkg/kubernetes/windows_eula_validator.go
ksimon1: added

pkg/kubernetes/windows_eula_validator.go
ksimon1: added

ksimon1:
@manusa all comments were resolved

manusa (approved):
LGTM, thx!

PR #1034, opened 2026-04-10
feat(evals): add mcpchecker diff to compare PR results against baseline

nader-ziada:
- Compare evaluation results from PR runs against the latest saved baseline on main using `mcpchecker diff --output markdown` The diff is included in the PR comment via a collapsible details section. 
- A make diff-evals target is also added for local use.

### DIFF
--- a/.github/workflows/mcpchecker.yaml
+++ b/.github/workflows/mcpchecker.yaml
@@ -223,6 +223,44 @@ jobs:
           make stop-server || true
           make kind-delete-cluster KIND_CLUSTER_NAME=${{ env.KIND_CLUSTER_NAME }} || true
 
+      # Compare PR results against baseline from main branch
+      - name: Fetch baseline results from main
+        if: needs.check-trigger.outputs.is-pr == 'true'
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/mcpchecker.yaml
```diff
@@ -223,6 +223,44 @@ jobs:
...
+            echo "No baseline results found on main branch, skipping diff"
+            echo "has-diff=false" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          make mcpchecker
+          MCPCHECKER=$(pwd)/_output/tools/bin/mcpchecker
+
+          mkdir -p eval-context
+          if $MCPCHECKER diff --base /tmp/baseline-results.json --current "$RESULTS_FILE" --output markdown > eval-context/diff-output.txt 2>&1; then
```
Cali0707: @nader-ziada do we want to capture stderr here into the output?

nader-ziada: good catch, done

matzew (approved):
LGTM

PR #965, opened 2026-03-25
fix: prevent panic on non-bool type assertion in tool handlers

manusa:
## Summary

- **Bug fix**: Replace unsafe `.(bool)` type assertion in `pods_log` handler (`previousBool = previous.(bool)`) which panicked when MCP clients passed non-bool types (e.g. `"false"` string)
- **Consistency refactor**: Replace other boolean parameter extractions (`configuration_view.minified`, `pods_top.all_namespaces`, `helm_list.all_namespaces`) with `api.OptionalBool` — these already used the safe comma-ok pattern but are now consistent with the rest of the codebase
- Add behavioral tests covering string and float64 types for `previous` and `minified` parameters

Fixes #347

Issue #347, opened 2025-09-26
panic in pods_log when "previous" argument is non-bool type (string)


jgericke:
### Summary  
The `pods_log` tool panics if the `previous` argument is passed as a non-boolean JSON type (for example, string `"false"`). In such cases, the interface conversion `previous.(bool)` fails and causes a runtime panic.

### Steps to reproduce  
1. Deploy kubernetes-mcp-server with LlamaStack or an agent sending a `pods_log` tool call where `previous` is a string (e.g. `"previous": "false"`).  
2. Observe the server log panics with something like:  
   `panic: interface conversion: interface {} is string, not bool`  

(From a LlamaStack call to pods_log, where previous is erroneously set to string)

```
INFO     2025-09-26 03:55:14,704 llama_stack.providers.inline.agents.meta_reference.agent_instance:890 agents: executing tool call: pods_log with     
         args: {'container': '', 'name': 'failing-test-pod', 'namespace': 'aiops-test', 'previous': 'false', 'tail': '100'}                           
03:55:14.705 [START] tool_execution
ERROR    2025-09-26 03:55:15,608 mcp.client.sse:109 uncategorized: Error in sse_reader
```

Resulting panic in kubernetes-mcp-server 
```
I0926 03:55:15.596574       1 mcp.go:215] "mcp tool call: pods_log(map[container: name:failing-test-pod namespace:aiops-test previous:false session_id:6e911834-2541-4db1-beff-91c78efa3134 tail:100])"
...
panic: interface conversion: interface {} is string, not bool

goroutine 150 [running]:
github.com/containers/kubernetes-mcp-server/pkg/toolsets/core.podsLog({{0x2a67728, 0xc000b089f0}, 0xc000596088, {0x2a3f440, 0xc0003c0780}, {0x2a52240, 0x3f2ed40}})
        /Users/runner/work/kubernetes-mcp-server/kubernetes-mcp-server/pkg/toolsets/core/pods.go:403 +0x608
github.com/containers/kubernetes-mcp-server/pkg/mcp.ServerToolToM3LabsServerTool.func1({0x2a67728, 0xc000b089f0}, {{{0xc0008e86a0, 0xa}, {0x0}}, 0xc0007ffaa0, {{0xc0008e86b0, 0x8}, {0x2203960, 0xc000b08a80}, ...}})
        /Users/runner/work/kubernetes-mcp-server/kubernetes-mcp-server/pkg/mcp/m3labs.go:46 +0x127
[...]

PR #1039, opened 2026-04-13
fix(core): replace unsafe type assertions with safe parameter helpers

xingyug:
## Summary

Replace 13+ bare type assertions (e.g. `params["namespace"].(string)`) in pod, event, and resource handlers with the existing safe helpers from `pkg/api/params.go`. This is a follow-up to PR #965 which fixed only the `previous.(bool)` assertion reported in #347, leaving all string and float64 parameter assertions unfixed.

## Problem

The MCP SDK does **not** validate `InputSchema` types before passing arguments to tool handlers — parameters arrive as `map[string]any` from `json.Unmarshal`. When a client sends a parameter with an unexpected JSON type (e.g. `{"namespace": 123}` instead of `{"namespace": "default"}`), bare assertions like `params["namespace"].(string)` cause a **runtime panic** that crashes the entire MCP server process.

This affects all pod tool handlers: `pods_list_in_namespace`, `pods_get`, `pods_delete`, `pods_exec`, `pods_log`, `pods_run`, plus `events_list` and `resources_get/list/create/delete`.

## Changes

| File | What changed |
|------|-------------|
| `pkg/toolsets/core/pods.go` | All 7 handler functions: replaced bare `.(string)` and `.(float64)` with `RequiredString`, `OptionalString`, `ParseInt64`; `podsExec` command array now rejects non-string elements instead of silently dropping them |
| `pkg/toolsets/core/events.go` | `eventsList`: `namespace.(string)` → `OptionalString` |
| `pkg/toolsets/core/resources.go` | `parseGroupVersionKind`: added comma-ok check for `kind.(string)`; fixed pre-existing wrong error message (`"name is not a string"` → `"apiVersion is not a string"`) |
| `pkg/mcp/pods_test.go` | Updated 4 error message assertions to match new format |
| `pkg/mcp/pods_run_test.go` | Updated 1 error message assertion |

## Testing

- **Unit tests**: All existing tests pass (`go test ./pkg/mcp/ -run TestPods`)
- **Integration tests**: Validated against a real 6-node K3s cluster — 17/17 tests passed including 9 safety tests that send wrong-type parameters (int, bool, null, array). All return graceful errors; none crash the server.

## Behavioral notes

- Error message format changed from `"failed to X, missing argument Y"` to `"failed to X: Y parameter required"` (from `RequiredString`'s error wrapped with `%w`). This is a minor wording change in user-facing error text.
- If a client sends `{"namespace": null}`, the old code returned "missing argument namespace"; the new code returns "namespace parameter must be a string". Both are errors — the new message is more precise.
- `podsExec`: `["ls", 123, "-la"]` previously silently became `["ls", "-la"]`; now returns an error. This prevents executing unintended commands.

Follow-up to #965 (which addressed #347)

### DIFF
--- a/pkg/mcp/pods_run_test.go
+++ b/pkg/mcp/pods_run_test.go
@@ -21,7 +21,7 @@ func (s *PodsRunSuite) TestPodsRun() {
 	s.Run("pods_run with nil image returns error", func() {
 		toolResult, _ := s.CallTool("pods_run", map[string]interface{}{})
 		s.Truef(toolResult.IsError, "call tool should fail")
-		s.Equalf("failed to run pod, missing argument image", toolResult.Content[0].(*mcp.TextContent).Text,
+		s.Equalf("failed to run pod: image parameter required", toolResult.Content[0].(*mcp.TextContent).Text,
 			"invalid error message, got %v", toolResult.Content[0].(*mcp.TextContent).Text)
 	})
 	s.Run("pods_run(image=nginx, namespace=nil), uses configured namespace", func() {
--- a/pkg/mcp/pods_test.go
+++ b/pkg/mcp/pods_test.go
@@ -122,7 +122,7 @@ func (s *PodsSuite) TestPodsListInNamespace() {
 	s.Run("pods_list_in_namespace with nil namespace returns error", func() {
 		toolResult, _ := s.CallTool("pods_list_in_namespace", map[string]interface{}{})
 		s.Truef(toolResult.IsError, "call tool should fail")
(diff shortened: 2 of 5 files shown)


manusa (approved):
LGTM, thx!

I'll follow up with a PR to tighten the optional parameter helpers. The ones used here prevent panics, but when a client sends the wrong type (e.g. `{"namespace": 123}`) they silently fall back to the default value instead of returning an error. The follow-up will surface those type mismatches so clients get proper feedback.

nader-ziada (approved):
LGTM

PR #1040, opened 2026-04-13
test(config): tool override tests are compatible with downstream changes

Cali0707:
With the tests for the tool overrides, we run into issues on any downstream forks if you actually apply any default overrides to tool descriptions.

This PR aims to keep the spirit of the test cases intact while updating the assertions to compare against whatever the default is, rather than against exact values


matzew (approved):
LGTM

PR #1046, opened 2026-04-14
Entra ID / OBO test follow-ups and dead code cleanup

### LINKED ISSUE #1043: test(auth): add SSE transport coverage for Authorization context injection and strengthen OBO happy-path assertion
## Context

Follow-up to #953 (Entra ID / OBO support). Two test-coverage gaps were identified during review but are non-blocking for the merge.

## 1. SSE transport integration test for `AuthorizationMiddleware` context injection

`pkg/http/authorization.go` injects the validated `Authorization` header into the request context:

```go
ctx := context.WithValue(r.Context(), internalk8s.OAuthAuthorizationHeader, authHeader)
next.ServeHTTP(w, r.WithContext(ctx))
```

This bridge is **only load-bearing for the SSE transport** — see the comment in `pkg/mcp/middleware.go` (`authHeaderPropagationMiddleware`): SSE does not propagate HTTP headers into `RequestExtra`, so the server relies on the context value. StreamableHTTP propagates headers via `req.GetExtra().Header` directly.

Current `AuthorizationSuite.StartClient` in `pkg/http/authorization_mcp_test.go` connects to `/mcp` (StreamableHTTP) only — so the SSE-specific context-injection path is untested.

### Proposed test

Add an integration test that exercises the SSE path end-to-end:

- Reuse the existing `AuthorizationSuite` plumbing (mock OIDC, mock k8s backend).
- Connect with `mcp.SSEClientTransport{}` (already imported at `pkg/http/http_mcp_test.go:28`) against `/sse` instead of `/mcp`.
- Send a validated Bearer token on the initial handshake.
- Invoke a tool that requires the propagated Authorization header to reach the Kubernetes manager.
- Assert the downstream backend received the expected token.

This gives confidence that regressions in the context-injection path will be caught.

## 2. Strengthen the OBO happy-path assertion in `TestAuthorizationOidcTokenExchange`

The log-string assertion `s.Contains(s.logBuffer.String(), \"token exchanged successfully\")` was removed as part of #953. Log-string assertions are brittle, so the removal is reasonable, but it leaves the happy path weaker than it should be.

### Proposed replacement

Assert on observable behavior instead of log output: verify the mock backend MockServer received the **exchanged token**, not the original user token. That tightens the test and aligns it with the security-relevant guarantee that the token exchange actually happened and the exchanged credential — not the original — reached the backend.

## Scope

Both items are small, local test additions. No production code changes required.

### LINKED ISSUE #1044: test(http): add black-box coverage for trust_proxy_headers gating in RequestMiddleware
## Context

Follow-up to #953. The PR changed `pkg/http/middleware.go` to gate trust of `X-Forwarded-*` and `X-Real-IP` headers behind a new `trust_proxy_headers` config flag (opt-in, default off). Pre-PR, these headers were always trusted — the new flag is the security boundary that prevents header spoofing when the server isn't behind a trusted reverse proxy.

Signatures changed:

- `getClientIP(r, trustProxy bool)` — honors `X-Forwarded-For` / `X-Real-IP` only when `trustProxy=true`.
- `getScheme(r, trustProxy bool)` — honors `X-Forwarded-Proto` only when `trustProxy=true`.
- `RequestMiddleware(trustProxy bool) func(http.Handler) http.Handler` — factory receiving the flag.

Currently **no test references `X-Forwarded-*`, `X-Real-IP`, or `trustProxy`**. The existing `HTTPTraceContextPropagationSuite` always passes `RequestMiddleware(false)` and only covers trace-context propagation. The security-relevant behavior of the new flag is untested.

## Testing approach — must be black-box

Per `CLAUDE.md` testing guidelines:

> Test the public API only — tests should be black-box and not access internal/private functions.

This rules out direct tests of the lowercase helpers `getClientIP` / `getScheme`. The only legitimate entry point is the exported `RequestMiddleware`.

The observable behavior for proxy-header handling is not visible to the `next` handler — the values (`client.address`, `url.scheme`) only land on the OTel span attributes. Assertions must therefore be made on emitted telemetry using an in-memory span recorder. Observing exported telemetry is still black-box in the CLAUDE.md sense — we're not reaching into internals.

### Implementation sketch

Add a new suite in `pkg/http/middleware_test.go`, e.g. `TrustProxyHeadersSuite`, that:

1. Installs an in-memory span exporter via the OTel SDK test helpers:
   ```go
   import (
       sdktrace \"go.opentelemetry.io/otel/sdk/trace\"
       \"go.opentelemetry.io/otel/sdk/trace/tracetest\"
   )

   recorder := tracetest.NewSpanRecorder()
   tp := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(recorder))
   // Swap the global tracer provider for the duration of the test; restore on teardown.
   ```
   Note: `middleware.go` creates `httpTracer` at package-init time from `otel.Tracer(...)`. Setting the global `TracerProvider` before the middleware is invoked is sufficient because `otel.Tracer` returns a lazy tracer that resolves against the current global provider at span-creation time. Verify this on the first test run; if resolution is eager, the suite may need to exercise the middleware via a fresh subprocess or the tracer will need to be made injectable (out of scope for this issue — prefer the black-box path first).

2. Ensures `telemetry.Enabled()` returns true during the test (reuse the `SetupTest` pattern from `HTTPTraceContextPropagationSuite` — sets `OTEL_EXPORTER_OTLP_ENDPOINT` and calls `telemetry.InitTracer`). Without this, `RequestMiddleware` takes the fast path that skips span creation and the assertions have nothing to observe.

3. Drives requests through `RequestMiddleware(trustProxy)` with `httptest.NewRequest` + `httptest.NewRecorder`, then reads `recorder.Ended()` and asserts on span attributes.

### Scenarios to cover

**`client.address` span attribute:**

| trust_proxy | X-Forwarded-For | X-Real-IP | RemoteAddr | Expected client.address |
|---|---|---|---|---|
| true | `10.0.0.1` | — | `192.168.1.1:443` | `10.0.0.1` |
| true | `10.0.0.1, 10.0.0.2, 10.0.0.3` | — | `192.168.1.1:443` | `10.0.0.1` (first, trimmed) |
| true | ` 10.0.0.1 ` (whitespace) | — | `192.168.1.1:443` | `10.0.0.1` |
| true | — | `10.0.0.2` | `192.168.1.1:443` | `10.0.0.2` |
| true | `10.0.0.1` | `10.0.0.2` | `192.168.1.1:443` | `10.0.0.1` (XFF wins) |
| **false** | `10.0.0.1` | `10.0.0.2` | `192.168.1.1:443` | **`192.168.1.1`** (headers ignored — the key security case) |
| false | — | — | `192.168.1.1:443` | `192.168.1.1` |
| false | — | — | `bad-remote-addr` (no port) | `bad-remote-addr` (SplitHostPort failure fallback) |

**`url.scheme` span attribute:**

| trust_proxy | X-Forwarded-Proto | req.TLS | Expected url.scheme |
|---|---|---|---|
| true | `https` | nil | `https` |
| true | — | nil | `http` |
| true | — | non-nil | `https` |
| **false** | `https` | nil | **`http`** (header ignored — the key security case) |
| false | — | non-nil | `https` |

The **bold negative cases** are the security guarantees of the opt-in flag; they're the most important assertions.

## Acceptance

- New black-box test suite in `pkg/http/middleware_test.go` covering the scenarios above.
- No tests call `getClientIP` / `getScheme` directly.
- Assertions read span attributes from an in-memory recorder (not internal state).
- Existing `HTTPTraceContextPropagationSuite` and `MaxBodyMiddlewareSuite` remain untouched.

manusa:
## Summary

Follow-ups to #953 (Microsoft Entra ID / OBO token exchange):

- **OBO happy-path assertion** — `TestAuthorizationOidcTokenExchange` now asserts the exchanged backend token (`validOidcBackendToken`) reaches the Kubernetes API, rather than relying on log strings. The backend MockServer captures the incoming `Authorization` header and the test compares it directly.
- **SSE transport coverage (#1043)** — Added `TestAuthorizationOidcTokenExchangeSSE` so the context-injection path used by SSE clients (`internalk8s.OAuthAuthorizationHeader` context key) is exercised end-to-end alongside the existing StreamableHTTP test. The two tests share a `runOidcTokenExchange` helper.
- **`trust_proxy_headers` gating (#1044)** — New `TrustProxyHeadersSuite` in `pkg/http/middleware_test.go` asserts on OTel span attributes (`client.address`, `url.scheme`) read from an in-memory `tracetest.SpanRecorder`. Covers XFF / X-Real-IP precedence, whitespace trimming, comma-separated lists, RemoteAddr fallback, and X-Forwarded-Proto / TLS scheme handling, with and without the flag enabled.
- **Dead code cleanup** — Dropped the unused `AuthProvider` interface and `IsRequireOAuth()` method that became orphaned once `Derived()` moved to `ResolveClusterAuthMode()` (PR #953 review comment).

Closes #1043
Closes #1044

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -19,11 +19,6 @@ const (
 	ClusterAuthKubeconfig = "kubeconfig"
 )
 
-type AuthProvider interface {
-	// IsRequireOAuth indicates whether OAuth authentication is required.
-	IsRequireOAuth() bool
-}
-
 // ClusterAuthProvider provides configuration for how the MCP server authenticates to clusters.
 type ClusterAuthProvider interface {
 	// GetClusterAuthMode returns the raw cluster authentication mode from config.
@@ -90,7 +85,6 @@ type RequireTLSProvider interface {
 }
 
 type BaseConfig interface {
-	AuthProvider
 	ClusterAuthProvider
 	ClusterProvider
 	ConfirmationRulesProvider
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -391,10 +391,6 @@ func (c *StaticConfig) GetToolsetConfig(name string) (api.ExtendedConfig, bool) 
 	return cfg, ok
 }
 
-func (c *StaticConfig) IsRequireOAuth() bool {
-	return c.RequireOAuth
-}
-
 func (c *StaticConfig) GetStsClientId() string {
 	return c.StsClientId
 }
--- a/pkg/http/authorization_mcp_test.go
+++ b/pkg/http/authorization_mcp_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -18,6 +19,22 @@ import (
 
 	"github.com/containers/kubernetes-mcp-server/internal/test"
 )
+
+// bearerRoundTripper adds a static Authorization: Bearer header to every
+// outbound request. Used by the SSE transport test below, since
+// mcp.SSEClientTransport does not expose a headers option.
+type bearerRoundTripper struct {
+	token string
+	base  http.RoundTripper
+}
+
+func (b *bearerRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	// http.RoundTripper must not modify the request it is given; clone before
+	// adding the Authorization header.
+	clone := r.Clone(r.Context())
+	clone.Header.Set("Authorization", "Bearer "+b.token)
+	return b.base.RoundTrip(clone)
+}
 
 type AuthorizationSuite struct {
 	BaseHttpSuite
@@ -382,9 +399,14 @@ func (s *AuthorizationSuite) TestAuthorizationOidcToken() {
 	s.Require().NoError(s.WaitForShutdown())
 }
 
+// TestAuthorizationOidcTokenExchange verifies RFC 8693 / On-Behalf-Of token
+// exchange on both MCP transports. StreamableHTTP propagates the bearer via
+// request headers directly; SSE relies on AuthorizationMiddleware writing the
+// validated token into the request context (OAuthAuthorizationHeader) and
+// authHeaderPropagationMiddleware bridging it into the MCP RequestExtra, so
+// both transports must be exercised end-to-end.
+// SSE coverage: https://github.com/containers/kubernetes-mcp-server/issues/1043
 func (s *AuthorizationSuite) TestAuthorizationOidcTokenExchange() {
-	s.MockServer.ResetHandlers()
-
 	oidcTestServer := NewOidcTestServer(s.T())
 	s.T().Cleanup(oidcTestServer.Close)
 	rawClaims := `{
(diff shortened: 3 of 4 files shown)


nader-ziada (approved):
LGTM

PR #1047, opened 2026-04-14
refactor(auth): deduplicate OIDC config fetches and simplify token exchange

nader-ziada:
small cleanup to #953

context here: https://github.com/containers/kubernetes-mcp-server/pull/953#issuecomment-4241438243

### DIFF
--- a/pkg/http/wellknown.go
+++ b/pkg/http/wellknown.go
@@ -1,6 +1,7 @@
 package http
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -19,6 +20,7 @@ import (
 
 const maxWellKnownResponseSize = 1 << 20 // 1 MB
 const oidcConfigCacheTTL = 5 * time.Minute
+const oidcConfigFetchTimeout = 30 * time.Second
 
 var allowedResponseHeaders = map[string]bool{
 	"Cache-Control": true,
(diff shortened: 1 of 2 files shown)


pkg/http/wellknown.go
```diff
@@ -235,36 +238,54 @@ func (w *WellKnown) fetchWellKnownEndpoint(request *http.Request, url string) (m
...
 		result := copyMap(w.oidcConfigCache)
-		w.oidcConfigCacheLock.RUnlock()
+		w.oidcConfigCacheMu.RUnlock()
 		return result, nil
 	}
-	w.oidcConfigCacheLock.RUnlock()
+	w.oidcConfigCacheMu.RUnlock()
+
+	// singleflight deduplicates concurrent cache misses for the same authURL
+	val, err, _ := w.oidcConfigFlight.Do(authURL, func() (interface{}, error) {
```
Cali0707: nit:
```suggestion
	val, err, _ := w.oidcConfigFlight.Do(authURL, func() (any, error) {
```

pkg/http/wellknown.go
```diff
@@ -235,36 +238,54 @@ func (w *WellKnown) fetchWellKnownEndpoint(request *http.Request, url string) (m
...
-
-	// Cache the result with timestamp and URL
-	w.oidcConfigCacheLock.Lock()
-	w.oidcConfigCache = copyMap(oidcConfig)
-	w.oidcConfigCacheTime = time.Now()
-	w.oidcConfigCacheURL = authURL
-	w.oidcConfigCacheLock.Unlock()
-
-	return oidcConfig, nil
+	return copyMap(val.(map[string]interface{})), nil
```
Cali0707: nit:
```suggestion
	return copyMap(val.(map[string]any)), nil
```

pkg/http/wellknown.go
```diff
@@ -235,36 +238,54 @@ func (w *WellKnown) fetchWellKnownEndpoint(request *http.Request, url string) (m
...
+	val, err, _ := w.oidcConfigFlight.Do(authURL, func() (interface{}, error) {
+		// Re-check cache under singleflight in case another goroutine just populated it
+		w.oidcConfigCacheMu.RLock()
+		if w.oidcConfigCache != nil && w.oidcConfigCacheURL == authURL && time.Since(w.oidcConfigCacheTime) < oidcConfigCacheTTL {
+			result := copyMap(w.oidcConfigCache)
+			w.oidcConfigCacheMu.RUnlock()
+			return result, nil
+		}
+		w.oidcConfigCacheMu.RUnlock()
+
```
Cali0707: @nader-ziada do we need this branch? Doesn't singleflight already de duplicate/return the final value we would want to other callers here?

I'm thinking it may be possible to simplify this flow to not recheck the cache inside `w.oidcConfigFlight.Do(...)`

pkg/http/wellknown.go
```diff
@@ -235,36 +238,54 @@ func (w *WellKnown) fetchWellKnownEndpoint(request *http.Request, url string) (m
...
+		oidcURL := authURL + openIDConfigurationEndpoint
+		oidcConfig, _, err := w.fetchWellKnownEndpoint(request, oidcURL)
```
Cali0707: Since singleflight will ensure this runs once, if there are multiple callers there may be a case where the request that "wins" and is running inside here is cancelled, which would propagate to callers.

Maybe we want to set the request context with `context.WithoutCancel()` (and then adding our own timeout on it) so that we can handle the case where the request times out?

manusa (approved):
LGTM, thx!

PR #1048, opened 2026-04-14
fix(test): call resp.Body.Close() instead of referencing method value

xingyug:
## Summary

`resp.Body.Close` without parentheses evaluates to a method reference (`func() error`) and discards it via `_ =` assignment. The response body is never actually closed, leaking HTTP response resources during test execution.

The `_ =` suppresses any linter warning, making this bug invisible to static analysis tools like `golangci-lint`.

## Changes

| File | Instances | Description |
|------|-----------|-------------|
| `pkg/http/authorization_mcp_test.go` | 6 | `_ = resp.Body.Close` → `_ = resp.Body.Close()` |
| `pkg/http/http_test.go` | 1 | `_ = resp.Body.Close` → `_ = resp.Body.Close()` |

## Impact

- **Before**: Response bodies leaked in every test case — cleanup callback stored a method reference instead of calling it
- **After**: Response bodies are properly closed via `resp.Body.Close()` in `t.Cleanup()`

## Testing

- `make build` (includes `go fmt`, `go mod tidy`, lint): **passes**
- `TestAuthorization` suite: **all subtests pass**
- Zero remaining instances of `resp.Body.Close` without `()` confirmed via `grep`

### DIFF
--- a/pkg/http/authorization_mcp_test.go
+++ b/pkg/http/authorization_mcp_test.go
@@ -108,7 +108,7 @@ func (s *AuthorizationSuite) TestAuthorizationUnauthorizedMissingHeader() {
 
 	s.Run("Protected resource with MISSING Authorization header", func() {
 		resp := s.HttpGet("")
-		s.T().Cleanup(func() { _ = resp.Body.Close })
+		s.T().Cleanup(func() { _ = resp.Body.Close() })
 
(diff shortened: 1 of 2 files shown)


matzew (approved):
LGTM

manusa (approved):
LGTM, thx!

PR #1049, opened 2026-04-15
fix(api): migrate tool handlers to WrapParams sticky-error pattern

manusa:
## Summary

Follow-up to #1039. The free-standing `OptionalString`/`OptionalBool` helpers silently
return the default value when a parameter is present but of the wrong type, masking
malformed client calls. This PR:

- Introduces a `Params` wrapper with sticky-error semantics that surfaces type mismatches
  as errors instead of silently falling back to defaults
- Marks the free-standing `OptionalString`/`OptionalBool` as deprecated
- Migrates all tool handlers (core, config, helm, kubevirt, tekton) to `WrapParams`
- Adds unit tests for the `Params` wrapper
- Updates existing tests to assert error responses for type-mismatch scenarios


matzew (approved):
LGTM

PR #1051, opened 2026-04-15
feat(http): add per-session rate limiting via MCP middleware

### LINKED ISSUE #1050: Add per-session rate limiting for HTTP/SSE transport
When deployed remotely (HTTP/SSE), the server has no request rate limiting at the MCP protocol level, leaving it exposed to request floods from individual clients.

## Proposal

Add per-session rate limiting as MCP middleware, configurable via `[http]` config on TOML file:

```toml
[http]
rate_limit_rps = 5
rate_limit_burst = 10
```

By default this would be disabled (e.g. `rate_limit_rps = 0`)

Implementation-wise would be similar to [upstream SDK example](https://github.com/modelcontextprotocol/go-sdk/blob/main/examples/server/rate-limiting/main.go), using the `AddReceivingMiddleware` "pattern".

For STDIO transport bypasses rate limiting (no session ID)

matzew:
Add configurable per-session rate limiting at the MCP protocol level to protect remote deployments (HTTP/SSE) from request floods. Uses the SDK's middleware pattern with `golang.org/x/time/rate` for token bucket limiting, keyed by session ID with automatic stale session reaping.

The middleware is placed after trace context propagation and before tracing so that rate-limited requests do not generate spans. Requests with no session ID (e.g. STDIO transport) bypass rate limiting.

Configured via `[http]` TOML config:
```toml
  rate_limit_rps   - requests per second per session (0 = disabled)
  rate_limit_burst - burst allowance (default: 10)
```

Fixes https://github.com/containers/kubernetes-mcp-server/issues/1050

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"golang.org/x/time/rate"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
 
@@ -70,6 +71,7 @@ type Server struct {
 	enabledPrompts []string
 	p              internalk8s.Provider
 	metrics        *metrics.Metrics // Metrics collection system
+	rateLimitDone  chan struct{}    // Closed to stop the rate limiter reaper goroutine
 }
 
 func NewServer(configuration Configuration, targetProvider internalk8s.Provider) (*Server, error) {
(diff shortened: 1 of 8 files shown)


pkg/mcp/mcp.go
```diff
@@ -108,6 +110,16 @@ func NewServer(configuration Configuration, targetProvider internalk8s.Provider)
 
 	s.server.AddReceivingMiddleware(sessionInjectionMiddleware)
 	s.server.AddReceivingMiddleware(traceContextPropagationMiddleware)
+	if configuration.HTTP.RateLimitRPS > 0 {
+		burst := configuration.HTTP.RateLimitBurst
+		if burst == 0 {
+			burst = 10
```
Cali0707: nit: let's extract this into a const somewhere

manusa (changes requested):
Thanks for this, Matthias! Clean implementation — the per-session token bucket approach with the reaper goroutine is solid, and the test coverage is thorough.

A few things I'd like to see addressed before merging:

- **Rate limit config doesn't reload on SIGHUP** — The middleware closure captures `limit` and `burst` at `NewServer` time, so changes via SIGHUP have no effect. Since `ReloadConfiguration` already updates `s.configuration.StaticConfig`, the middleware should read from the live config on each request rather than capturing the initial values. This would also be a good place to handle creating/stopping the reaper goroutine when rate limiting is toggled on/off via reload.

- **`Close()` isn't safe for concurrent/repeated calls** — `rateLimitDone` is read, closed, and set to nil without synchronization. Since `Shutdown()` calls `Close()` and both are public, a concurrent call could double-close the channel and panic. A `sync.Once` would fix this cleanly.

- **Negative config values aren't validated** — Setting `rate_limit_rps = -1` creates a limiter that silently blocks everything. Worth adding validation that both `RateLimitRPS` and `RateLimitBurst` are non-negative, with a clear config error — both in `NewServer` and in `ReloadConfiguration` (alongside the existing `ValidateRequireTLS` check).

- **Plain error for rate limiting** — `errors.New("rate limit exceeded")` gets wrapped as a generic JSON-RPC internal error. A structured error with a code in the server-defined range (`-32000` to `-32099`) would let clients programmatically distinguish rate limits from other failures and back off accordingly.

Cali0707 (commented):
one small nit

matzew:
@manusa @Cali0707 thanks for review

updated the PR

manusa (approved):
LGTM, thx!

PR #1052, opened 2026-04-15
fix(evals): Replace unmaintained bitnami helm charts

matzew:
* Remove use different chart than outdated bitnami
* remove the redundant llmJudge, since we already use extension for checks


Cali0707 (approved):
LGTM

I noticed that we don't have a helm config for the evals in actions, will open a PR to add that: https://github.com/containers/kubernetes-mcp-server/blob/c9b074eb834b2b9bc61d333ce941ee8f137171ef/.github/workflows/mcpchecker.yaml#L21-L25

Running locally, this seems to work

PR #1058, opened 2026-04-16
fix(config): enforce all Validate() checks on SIGHUP reload (#989)

### LINKED ISSUE #989: Config reload (SIGHUP) skips most Validate() checks
## Problem

When configuration is reloaded via SIGHUP, `ReloadConfiguration()` applies the new `StaticConfig` without running the validation checks that `Validate()` in `root.go` performs at startup. This means invalid config values can be silently accepted on reload.

### Fields not validated on SIGHUP reload

- `list_output` — invalid format name accepted
- `toolsets` — invalid toolset names accepted
- `cluster_provider_strategy` — invalid strategy accepted
- `authorization_url` — scheme/format not checked
- `certificate_authority` — file existence not checked
- TLS `tls_cert`/`tls_key` — pairing and file existence not checked

### Fields that ARE validated on reload

- `require_tls` URL schemes — validated via `StaticConfig.ValidateRequireTLS()` (added in the `require_tls` feature work)
- Toolset-level config (e.g., Kiali URL scheme) — validated by toolset parsers during `ReadToml`

## Suggested approach

In the `require_tls` feature work, we extracted `StaticConfig.ValidateRequireTLS()` so both `root.go Validate()` and `ReloadConfiguration()` call the same method. This pattern could be extended:

1. Extract config-level validations from `MCPServerOptions.Validate()` into `StaticConfig.Validate()`
2. Keep CLI-level validations (flag interactions, `--port` dependencies) in `MCPServerOptions.Validate()`
3. Call `StaticConfig.Validate()` from both `MCPServerOptions.Validate()` (startup) and `ReloadConfiguration()` (SIGHUP)

This ensures config-level invariants are always enforced, while CLI-specific checks (which can't change on reload) remain at the command layer.

## Context

- `Validate()` lives on `MCPServerOptions` in `pkg/kubernetes-mcp-server/cmd/root.go` — called by cobra's `PreRunE` at startup only
- `ReloadConfiguration()` lives on `Server` in `pkg/mcp/mcp.go` — called by the SIGHUP handler
- The SIGHUP handler (`setupSIGHUPHandler`) calls `config.Read()` → `mcpServer.ReloadConfiguration()` with no intermediate validation step

manusa:
`ReloadConfiguration()` previously only checked `ValidateRequireTLS()` and `HTTP.Validate()`, silently accepting
invalid values for `list_output`, `toolsets`, `cluster_provider_strategy`, `authorization_url`, `certificate_authority`,
and `tls_cert`/`tls_key`.

This PR extracts config-level validations from `MCPServerOptions.Validate()` into `StaticConfig.Validate()` so
both startup and SIGHUP reload share the same checks.

Cluster-provider strategy validation uses a chainable `WithProviderStrategies()` method to avoid the circular
dependency (`config → kubernetes → oauth → config`) while keeping the check inside `Validate()`.

Fixes #989

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -5,13 +5,17 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strings"
 
 	"github.com/BurntSushi/toml"
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	"github.com/containers/kubernetes-mcp-server/pkg/output"
+	"github.com/containers/kubernetes-mcp-server/pkg/toolsets"
 	"k8s.io/klog/v2"
 )
 
@@ -151,6 +155,8 @@ type StaticConfig struct {
 
 	// Internal: the config.toml directory, to help resolve relative file paths
 	configDirPath string
+	// Internal: known provider strategies, set via WithProviderStrategies
+	providerStrategies []string
 }
 
 var _ api.BaseConfig = (*StaticConfig)(nil)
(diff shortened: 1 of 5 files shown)


matzew (approved):
LGTM

PR #1059, opened 2026-04-16
feat(helm-charts): add Gateway API HTTPRoute support

muroed:
Add an optional gateway.networking.k8s.io/v1 HTTPRoute with values-driven parentRefs, hostnames, and rules; the chart injects Service backendRefs. Extend helm-validate with CI values and document usage in the chart README.


manusa (approved):
LGTM, thx!

PR #1060, opened 2026-04-16
fix(pods): resolve default container for multi-container pods (KEP-2227)

### LINKED ISSUE #1000: pods_log: support kubectl.kubernetes.io/default-container annotation for multi-container pods

matzew:
`pods_log` fails on multi-container pods when no container parameter is provided because it passes an empty string to the Kubernetes API. `pods_exec` had a partial fix (falling back to the first container) but did not honor the `kubectl.kubernetes.io/default-container` annotation.

Add a resolveContainer helper implementing the KEP-2227 resolution order:
* explicit param → single container → annotation → first container.  

Both `PodsLog` and `PodsExec` now use this shared logic.

Fixes #1000

### DIFF
--- a/pkg/kubernetes/pods.go
+++ b/pkg/kubernetes/pods.go
@@ -27,6 +27,34 @@ import (
 // DefaultTailLines is the default number of lines to retrieve from the end of the logs
 const DefaultTailLines = int64(100)
 
+// DefaultContainerAnnotation is the annotation key used by kubectl to specify the default container
+// for operations like logs and exec in multi-container pods (KEP-2227).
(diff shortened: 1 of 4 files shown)


manusa (approved):
LGTM, thx!

PR #1062, opened 2026-04-16
fix(config): auto-select kubeconfig context when current-context is empty

### LINKED ISSUE #1045: npx -y kubernetes-mcp-server does not launch the server
Opposite to README.md, npx -y kubernetes-mcp-server@latest does not start the server but generates an error message

manusa:
When a kubeconfig file has an empty `current-context` but defines exactly
one context, the server now auto-selects it instead of failing to start.

When multiple contexts exist without a `current-context`, a descriptive
error lists the available contexts and guides the user to set one with
`kubectl config use-context`.

Fixes #1045

/cc @jeffmaury

### DIFF
--- a/pkg/kubernetes/manager.go
+++ b/pkg/kubernetes/manager.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


matzew (approved):
LGTM

PR #1070, opened 2026-04-17
feat(kubevirt): add product name override mechanism for downstream branding

ksimon1:
Implement configuration-based override system to allow downstream distributions (e.g., OpenShift) to rebrand user-facing strings while maintaining API compatibility.

Changes:
- Add pkg/toolsets/kubevirt/internal/defaults package with override pattern
- Update toolset description to use defaults.ToolsetDescription()
- Update all tool descriptions to use defaults.ProductName()
- Update vm-troubleshoot prompt to use dynamic product name
- Update tests to use dynamic product name assertions
- Auto-regenerate README.md with updated tool descriptions

Resolves: CNV-78794

### DIFF
--- a/pkg/mcp/testdata/toolsets-kubevirt-tools.json
+++ b/pkg/mcp/testdata/toolsets-kubevirt-tools.json
@@ -39,7 +39,7 @@
       "openWorldHint": false,
       "title": "Virtual Machine: Create"
     },
-    "description": "Create a VirtualMachine in the cluster with the specified configuration, automatically resolving instance types, preferences, and container disk images. VM will be created in Halted state by default; use autostart parameter to start it immediately.",
+    "description": "Create a KubeVirt VirtualMachine in the cluster with the specified configuration, automatically resolving instance types, preferences, and container disk images. VM will be created in Halted state by default; use autostart parameter to start it immediately.",
     "inputSchema": {
       "properties": {
         "autostart": {
@@ -162,7 +162,7 @@
       "openWorldHint": false,
       "title": "Virtual Machine: Lifecycle"
     },
-    "description": "Manage VirtualMachine lifecycle: start, stop, or restart a VM",
+    "description": "Manage KubeVirt VirtualMachine lifecycle: start, stop, or restart a VM",
     "inputSchema": {
       "properties": {
         "action": {
--- a/pkg/toolsets/kubevirt/internal/defaults/defaults.go
+++ b/pkg/toolsets/kubevirt/internal/defaults/defaults.go
@@ -1,0 +1,22 @@
+package defaults
+
+const (
+	DefaultToolsetDescription = "KubeVirt virtual machine management tools, check the [KubeVirt documentation](https://github.com/containers/kubernetes-mcp-server/blob/main/docs/kubevirt.md) for more details."
+	DefaultProductName        = "KubeVirt"
+)
+
+func ToolsetDescription() string {
+	overrideDescription := ToolsetDescriptionOverride()
+	if overrideDescription != "" {
+		return overrideDescription
+	}
+	return DefaultToolsetDescription
+}
+
+func ProductName() string {
+	overrideProductName := ProductNameOverride()
+	if overrideProductName != "" {
+		return overrideProductName
+	}
+	return DefaultProductName
+}
--- a/pkg/toolsets/kubevirt/toolset.go
+++ b/pkg/toolsets/kubevirt/toolset.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"github.com/containers/kubernetes-mcp-server/pkg/toolsets"
+	kubevirtdefaults "github.com/containers/kubernetes-mcp-server/pkg/toolsets/kubevirt/internal/defaults"
 	vm_clone "github.com/containers/kubernetes-mcp-server/pkg/toolsets/kubevirt/vm/clone"
 	vm_create "github.com/containers/kubernetes-mcp-server/pkg/toolsets/kubevirt/vm/create"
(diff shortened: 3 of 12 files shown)


pkg/mcp/testdata/toolsets-kubevirt-tools.json
```diff
@@ -39,7 +39,7 @@
       "openWorldHint": false,
       "title": "Virtual Machine: Create"
     },
-    "description": "Create a VirtualMachine in the cluster with the specified configuration, automatically resolving instance types, preferences, and container disk images. VM will be created in Halted state by default; use autostart parameter to start it immediately.",
+    "description": "Create a KubeVirt VirtualMachine in the cluster with the specified configuration, automatically resolving instance types, preferences, and container disk images. VM will be created in Halted state by default; use autostart parameter to start it immediately.",
```
Cali0707, line 42: In the past we have run into some issues with tests in downstream forks when these defaults get changed. I think the current approach we have been using is having some kind of `BaseDefaults()` method that gets the defaults without any overrides, and always ensuring that is the config used in the tests

cc @matzew

pkg/toolsets/kubevirt/vm_troubleshoot_test.go
```diff
@@ -32,7 +32,7 @@ func (s *VMTroubleshootSuite) TestVMTroubleshootPrompt() {
 		prompts := initVMTroubleshoot()
 		s.Require().Len(prompts, 1, "Expected 1 prompt")
 		s.Equal("vm-troubleshoot", prompts[0].Prompt.Name)
-		s.Equal("VirtualMachine Troubleshoot", prompts[0].Prompt.Title)
+		s.Equal("KubeVirt VirtualMachine Troubleshoot", prompts[0].Prompt.Title)
```
lyarwood, line 35: Agreeing with @Cali0707's comment on the test data file -- this test hardcodes `"KubeVirt VirtualMachine Troubleshoot"` which will break in downstream forks that set `productNameOverride`. Consider using `defaults.DefaultProductName` (or `defaults.ProductName()`) to build the expected string dynamically:

```go
s.Equal(fmt.Sprintf("%s VirtualMachine Troubleshoot", defaults.DefaultProductName), prompts[0].Prompt.Title)
```

Or even better, use `defaults.ProductName()` so the test validates the override is being applied correctly in downstream.

The same concern applies to `pkg/mcp/kubevirt_test.go` line 759.

pkg/toolsets/kubevirt/toolset.go
```diff
@@ -34,6 +35,12 @@ func (t *Toolset) GetPrompts() []api.ServerPrompt {
 	return initVMTroubleshoot()
 }
 
+// ProductName returns the product name for the kubevirt toolset
```
lyarwood, line 38: The comment says this is exported for testing purposes, but no test in this PR actually calls `kubevirt.ProductName()`. If it is intended for downstream test use, consider adding a test that exercises it. If it is not needed, removing it would avoid adding unnecessary API surface to the package.

If you do keep it, a minor naming suggestion: `ProductName` on the `kubevirt` package reads as `kubevirt.ProductName()` which is clear enough, but consider whether the `defaults.ProductName()` function alone would suffice for any caller that needs it.

pkg/toolsets/kubevirt/internal/defaults/defaults.go
```diff
@@ -0,0 +1,22 @@
+package defaults
```
lyarwood, line 1: Minor: the package has no unit tests. While the logic is trivial, a small test would serve as documentation for downstream forks and validate the fallback-to-default behavior:

```go
func TestProductNameReturnsDefault(t *testing.T) {
    // When no override is set, ProductName returns the default
    assert.Equal(t, DefaultProductName, ProductName())
}
```

This also establishes a pattern downstream forks can follow to verify their overrides take effect.

manusa (changes requested):
Mostly aligned with the existing kiali pattern in `pkg/toolsets/kiali/internal/defaults/`, but two things stand out as deviations worth fixing before this merges.

The new `pkg/toolsets/kubevirt/internal/defaults/defaults_test.go` doesn't have an equivalent in kiali, and I think that's on purpose — `assert.Equal(t, DefaultProductName, ProductName())` (lines 11 and 16) passes upstream only because the override is `""`. As soon as a downstream fork sets `productNameOverride`, `ProductName()` returns the override while `DefaultProductName` stays `"KubeVirt"`, and the test starts failing exactly when the override is doing its job. Easiest path is to drop the file; if you want to keep a smoke test, `assert.NotEmpty(t, ProductName())` covers the fallback without coupling to the constant.

The other one is the `ProductName()` re-export in `pkg/toolsets/kubevirt/toolset.go:46-49`. It's only there so `pkg/mcp/kubevirt_test.go:760` can reach the name across packages, but kiali handles the same case without adding API surface — `pkg/mcp/kiali_test.go:28` just goes through `(&kialiToolset.Toolset{}).GetName()`. Could we drop the wrapper and have the test derive the expected string from the prompt itself, e.g. `(&kubevirttoolset.Toolset{}).GetPrompts()[0].Prompt.Title`?

Two smaller things while we're here: `TestProductName` in `pkg/toolsets/kubevirt/vm_troubleshoot_test.go:376` only checks non-empty and duplicates whatever the defaults-package smoke test would cover, so it can probably go. And the rebrand is partial — argument descriptions, the result message in `vm_troubleshoot.go:191`, and the guide section headers (`## Step 1: VirtualMachine Status` etc.) still hardcode "VirtualMachine". Worth either finishing in a follow-up or calling out in the PR description that the scope is intentional.

ksimon1:
@manusa thanks for the feedback. It should be fixed now

manusa (approved):
LGTM, thx!

PR #1072, opened 2026-04-17
fix(config): validate token exchange fields at config load (#1057)

### LINKED ISSUE #1057: Config validation: token exchange fields not validated at config load time
## Problem

Several token exchange configuration fields are only validated at runtime (when a token exchange request is made), not at config load time. This means invalid values are silently accepted in the config file and only surface as warnings or errors during actual token operations.

This was identified during the audit for #989 (SIGHUP reload validation). These gaps are consistent between startup and reload — neither path catches them — so they are a separate concern from the reload asymmetry fixed in #989.

#### 1. `token_exchange_strategy` — invalid strategy name silently ignored

- **Current behavior**: `pkg/kubernetes/token_exchange.go:45-47` calls `tokenexchange.GetTokenExchanger(strategy)` and logs a warning if the strategy is not found, then falls back to STS exchange. The invalid value is never rejected.
- **Expected**: Config load should reject unknown strategy values. Valid strategies are registered in `pkg/tokenexchange/registry.go` via `RegisterTokenExchanger()` — currently `rfc8693`, `keycloak-v1`, `entra-obo`.
- **Pattern to follow**: `cluster_provider_strategy` uses `config.WithClusterProviderStrategies()` (a `ValidateOption`) because the registry lives in a package that would create a circular dependency if imported directly from `pkg/config`. The same pattern applies here — the token exchange registry is in `pkg/tokenexchange`, which can be checked for circular deps but likely needs a `ValidateOption` like `WithTokenExchangeStrategies(validStrategies []string)`.

#### 2. `sts_auth_style` — invalid value accepted, only caught at runtime

- **Current behavior**: `TargetTokenExchangeConfig.Validate()` in `pkg/tokenexchange/config.go:74-88` validates the field (must be `"params"`, `"header"`, or `"assertion"`), but this is only called at runtime in `pkg/kubernetes/provider_token_exchange.go:94` when building the STS config.
- **Expected**: `StaticConfig.Validate()` should check that `StsAuthStyle` (if non-empty) is one of the valid values. The valid values are constants in `pkg/tokenexchange/config.go`: `AuthStyleParams`, `AuthStyleHeader`, `AuthStyleAssertion`.

#### 3. `sts_client_cert_file` / `sts_client_key_file` — file existence not checked

- **Current behavior**: These file paths are only read at runtime. If the files don't exist, the error surfaces during a token exchange operation.
- **Expected**: When `sts_auth_style = "assertion"`, both files should be validated for existence at config load time, similar to how `certificate_authority`, `tls_cert`, and `tls_key` are validated in `StaticConfig.Validate()`.
- **Note**: The pairing check (both required when `auth_style = "assertion"`) already exists in `TargetTokenExchangeConfig.Validate()` (`pkg/tokenexchange/config.go:78-84`) but only runs at runtime.

## Suggested approach

Add these validations to `StaticConfig.Validate()` in `pkg/config/config.go`. For the strategy registry lookup, follow the `WithClusterProviderStrategies` pattern (functional `ValidateOption`) to avoid circular dependencies. For `sts_auth_style` and the cert/key file checks, these can be validated directly since the valid values are simple strings and file existence uses `os.Stat`.

Both `MCPServerOptions.Validate()` (startup, `pkg/kubernetes-mcp-server/cmd/root.go`) and `Server.ReloadConfiguration()` (SIGHUP, `pkg/mcp/mcp.go`) already call `StaticConfig.Validate()`, so adding checks there covers both paths automatically.

## Context

- `StaticConfig.Validate()` was introduced in #989 to centralize config-level validation
- The `ValidateOption` pattern (`WithClusterProviderStrategies`) was established in #989 for registry lookups that can't be imported directly
- Existing tests for `StaticConfig.Validate()` are in `pkg/config/validate_test.go` (external test package `config_test`) and `pkg/mcp/mcp_reload_test.go`

manusa:
## Summary

Closes #1057.

Several token exchange configuration fields were only validated at runtime (when a token exchange request was made), so invalid values were silently accepted at startup and on SIGHUP reload:

- **`token_exchange_strategy`** — unknown strategies fell back to STS with only a warning. Now rejected via a new `WithTokenExchangeStrategies` option (mirrors the `WithProviderStrategies` pattern from #989) backed by a new `tokenexchange.GetRegisteredStrategies()` helper.
- **`sts_auth_style`** — invalid values are now rejected at config load (`"params"`, `"header"`, `"assertion"`, or empty).
- **`sts_client_cert_file` / `sts_client_key_file`** — when `sts_auth_style = "assertion"`, both must be set and reference existing files. Mirrors the runtime contract in `TargetTokenExchangeConfig.Validate()` and `loadCertificateAndKey`.

Both startup (`pkg/kubernetes-mcp-server/cmd/root.go`) and SIGHUP reload (`pkg/mcp/mcp.go`) now chain `WithTokenExchangeStrategies(tokenexchange.GetRegisteredStrategies())` into `Validate()`, so the new checks fire on both paths.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -157,6 +157,8 @@ type StaticConfig struct {
 	configDirPath string
 	// Internal: known provider strategies, set via WithProviderStrategies
 	providerStrategies []string
+	// Internal: known token exchange strategies, set via WithTokenExchangeStrategies
+	tokenExchangeStrategies []string
 }
 
 var _ api.BaseConfig = (*StaticConfig)(nil)
@@ -455,6 +457,16 @@ func (c *StaticConfig) WithProviderStrategies(strategies []string) *StaticConfig
 	return c
 }
 
+// WithTokenExchangeStrategies sets the known token exchange strategies for
+// validation. Callers that have access to the token exchange registry should
+// chain this before Validate so that token_exchange_strategy is checked:
+//
+//	cfg.WithTokenExchangeStrategies(tokenexchange.GetRegisteredStrategies()).Validate()
+func (c *StaticConfig) WithTokenExchangeStrategies(strategies []string) *StaticConfig {
+	c.tokenExchangeStrategies = strategies
+	return c
+}
+
 // Validate validates config-level invariants that must hold at both startup and
 // on SIGHUP reload.
 func (c *StaticConfig) Validate() error {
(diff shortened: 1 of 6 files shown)


matzew (approved):
LGTM

PR #1073, opened 2026-04-19
fix(evals): fix kubernetes/scale-deployment cleanup script

difrost:
Delete the `scale-test` namespace in the cleanup step.


nader-ziada (approved):
LGTM

PR #1074, opened 2026-04-20
fix(config): normalize whitespace for token exchange fields before validation

matzew:
Extend the existing `TrimSpace` normalization in `Validate()` to cover `StsAuthStyle`, `StsClientCertFile`, and `StsClientKeyFile`, consistent with how `CertificateAuthority`, `TLSCert`, and `TLSKey` are already handled. Without this, whitespace-padded values would bypass empty-string checks and cause misleading file-not-found errors at runtime.


manusa (approved):
LGTM, thx

PR #1075, opened 2026-04-20
fix(config): validate confirmation fields at config load (#1071)

### LINKED ISSUE #1071: Config validation: confirmation_fallback and confirmation_rules only validated in ReadToml
## Problem

`confirmation_fallback` and `confirmation_rules` are validated only inside `ReadToml` (`pkg/config/config.go:362-372`), not inside `StaticConfig.Validate()`. Today both startup and SIGHUP happen to fire these checks because both go through `config.Read` → `ReadToml`, but `Validate()` is no longer the single source of truth for config-level invariants.

This is the same class of asymmetry addressed in #989 (SIGHUP reload validation) and #1057 (token exchange field validation): any caller that constructs a `StaticConfig` and invokes only `Validate()` — programmatic reload, tests, future drop-in / hot-swap features — will silently skip these checks.

#### 1. `confirmation_fallback` — only validated in `ReadToml`

- **Current behavior**: `pkg/config/config.go:362-364` rejects values other than `"allow"` / `"deny"` / empty, but only when parsing TOML.
- **Expected**: `StaticConfig.Validate()` should perform the same check so it holds for any caller, regardless of how the config object was constructed.

#### 2. `confirmation_rules` — only validated in `ReadToml`

- **Current behavior**: `pkg/config/config.go:367-372` iterates over each rule and calls `ConfirmationRule.Validate()`, joining errors. Only runs at TOML parse time.
- **Expected**: Same iteration should run from `StaticConfig.Validate()`.

## Suggested approach

Move (or duplicate) the two checks from `ReadToml` into `StaticConfig.Validate()` so it remains the single source of truth for config-level invariants. The `ReadToml` block can either delegate to `Validate()` or be removed once `Validate()` covers it (callers of `Read` already invoke `Validate()` afterwards on both startup and SIGHUP paths).

## Context

- `StaticConfig.Validate()` was introduced in #989 to centralize config-level validation
- #1057 closed a similar gap for token-exchange fields
- Existing tests for `StaticConfig.Validate()` are in `pkg/config/validate_test.go`; existing tests for the confirmation parsing checks are likely in `pkg/config/config_test.go` and would need to be re-pointed at `Validate()`

manusa:
## Summary

- Move `confirmation_fallback` and `confirmation_rules` validation out of `ReadToml` into `StaticConfig.Validate()` so it becomes the single source of truth for config-level invariants — closes the same class of asymmetry addressed in #989 and #1057. Programmatic callers that build a `StaticConfig` and invoke only `Validate()` (tests, future drop-in / hot-swap features) now get the same checks that startup and SIGHUP reload get.
- Strengthen whitespace-only subtests added in #1074 so they actually observe the `strings.TrimSpace` side-effect instead of coincidentally passing through an unrelated branch of `validateTokenExchange`. Also tighten the pre-existing TLS/CA whitespace subtests the same way for consistency. Verified by temporarily reverting the `TrimSpace` lines — the new assertions fail; they pass with the fix in place.

Closes #1071


matzew (approved):
LGTM

PR #1078, opened 2026-04-20
feat(config): add skip_jwt_verification to enforce OIDC verification

### LINKED ISSUE #1079: require_oauth=true without authorization_url silently accepts unverified JWTs
The authentication middleware uses `UnsafeClaimsWithoutVerification()` to extract JWT claims. When `require_oauth = true` but no `authorization_url` is configured, there is no OIDC provider to verify the token signature against. The middleware silently falls through to offline-only validation that only checks `exp` and `aud` claims.

Any well-formed JWT with a future `exp` and correct `aud` will be accepted - no valid signature needed.

The server should either require `authorization_url` when `require_oauth` is _enabled_, or force the operator to explicitly opt in to unverified mode (for deployments behind a trusted reverse proxy that already verifies tokens).

matzew:
When `require_oauth=true` but `authorization_url` is not configured, the `AuthorizationMiddleware` silently falls through to offline-only JWT validation without cryptographic signature verification. This makes the insecure fallback explicit and opt-in.

- Add `SkipJWTVerification` field to `StaticConfig`
- Add `validateSkipJWTVerification()` that rejects `require_oauth=true` without `authorization_url` unless `skip_jwt_verification=true`
- Add runtime guard in middleware to reject requests when `skip_jwt_verification=false` and no OIDC provider is configured
- Add warning log (once per middleware instance) when JWT is accepted without signature verification
- Add (hidden) `--skip-jwt-verification` CLI flag
- Add config validation, middleware warning, and reload tests
- Fix existing tests that set `require_oauth=true` without `authorization_url`

Fixes: https://github.com/containers/kubernetes-mcp-server/issues/1079

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -467,6 +467,7 @@ Configure OAuth/OIDC authentication for HTTP mode deployments.
 | `require_oauth` | boolean | `false` | When `true`, requires OAuth authentication for all requests. |
 | `oauth_audience` | string | `""` | Valid audience for OAuth tokens (for offline JWT claim validation). |
 | `authorization_url` | string | `""` | URL of the OIDC authorization server for token validation and STS exchange. |
+| `skip_jwt_verification` | boolean | `false` | When `true`, allows JWTs without cryptographic signature verification when `require_oauth` is enabled but no `authorization_url` is configured. Only use behind a trusted reverse proxy that already verifies tokens. When `false` (default), the server refuses to start in this configuration. |
 | `disable_dynamic_client_registration` | boolean | `false` | When `true`, disables dynamic client registration in `.well-known` endpoints. |
 | `oauth_scopes` | string[] | `[]` | Supported client scopes for the OAuth flow. |
 | `sts_client_id` | string | `""` | OAuth client ID for backend token exchange. |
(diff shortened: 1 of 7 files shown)


manusa (changes requested):
Thanks for tackling this, the opt-in flag is a solid improvement over the silent fallback. A few things before this lands:

**`docs/configuration.md` needs the new key.** That file is the canonical TOML reference and currently doesn't mention `skip_jwt_verification`. A short entry explaining what it does, when to use it (behind a trusted reverse proxy that verifies tokens), and the default would be great. This is the piece I'd most like to see before merging.

**Minor on the runtime 401.** In `authorization.go`, when `SkipJWTVerification=false` and no provider is configured, we return `401` with `error=\"server_error\"`. That's really a server misconfig, not a bad token, so a `500` would be less misleading to operators. Happy to defer if you'd rather keep it.

**Test gaps worth tightening:**
- `TestAuthorizationOfflineOnlyRejectedWithoutSkipFlag` only asserts `Nil(Session)`. Could we also assert the response status and `WWW-Authenticate` header so a regression in rejection behavior gets caught?
- The `sync.Once` warning isn't directly verified. A multi-request test confirming it fires exactly once would lock that in.

Optional: the godoc on `AuthorizationMiddleware` (scenario 2.1) still describes raw-token validation as unconditional, worth a line noting it now requires `SkipJWTVerification=true`.

manusa:
/cc @onmete this might change the token passthrough behavior.
FYI in case you face issues in Lightspeed integration.

manusa (approved):
LGTM, thx!

PR #1082, opened 2026-04-20
fix(helm-charts): address HTTPRoute follow-ups from #1059

### LINKED ISSUE #1080: chart: follow-ups to HTTPRoute support (#1059)
Follow-up items from the review of #1059. The PR is merged; the feature is opt-in (disabled by default), so these are not user-facing regressions, but they should be cleaned up.

### 1. Rule without `matches` renders invalid YAML

In `charts/kubernetes-mcp-server/templates/httproute.yaml`, the list-item dash (`-`) is attached to the optional `matches` block:

```yaml
{{- with .matches }}
- matches:
  ...
{{- end }}
  filters: ...
  backendRefs: ...
```

`matches` is optional per the HTTPRoute spec. If a user defines a rule with only `filters` (or a catch-all rule), the `-` is never emitted and `filters`/`backendRefs` become sibling keys of the previous list item instead of a new one. Safer to anchor the dash on `backendRefs` (always rendered) and keep `matches` / `filters` / `timeouts` as conditional siblings.

### 2. No fast-fail when `httpRoute.enabled=true` and `parentRefs` is empty

The template currently renders an HTTPRoute without `parentRefs` in that case, which the API server rejects at apply time. Mirror the ingress `host` pattern with a `required` check so the failure happens at `helm template` / `helm install` time:

```yaml
{{- $_ := required "httpRoute.parentRefs must be set when httpRoute.enabled=true" .Values.httpRoute.parentRefs -}}
```

### 3. CI fixture vs documented flow

README says to disable Ingress when enabling Gateway API:

> set `ingress.enabled` to `false` and enable `httpRoute`

but `charts/kubernetes-mcp-server/ci/httproute-test-values.yaml` leaves Ingress enabled (`ingress.host: localhost`). Either update the CI fixture to `ingress.enabled: false` to match the documented flow, or adjust the README to acknowledge they can coexist.

### 4. kubeconform is not actually validating HTTPRoute

The new `helm-validate` step runs with `-ignore-missing-schemas`, so the Gateway API CRD schema is silently skipped — the step only checks that YAML parses, not that the HTTPRoute spec is valid. This is how #1 above slipped through. Wire in a Gateway API schema location (e.g. `-schema-location` pointing at the `gateway-api` CRD JSON schemas) so the CI step catches real spec violations.

manusa:
Follow-ups from the review of #1059 (tracked in #1080).

## Summary
- **Rule without `matches` now renders valid YAML** — the list-item dash was attached to the optional `matches` block, so a rule with only `filters` (or a catch-all rule) produced sibling keys instead of a new list item. Dash is now anchored on `backendRefs`, which is always rendered, with `matches` / `filters` / `timeouts` as conditional siblings.
- **Fast-fail when `httpRoute.enabled=true` and `parentRefs` is empty** — previously rendered an HTTPRoute without `parentRefs` that the API server rejected at apply time. Now fails at `helm template` / `helm install` time with a message that points at the chart README.
- **CI fixture matches the documented flow** — `ci/httproute-test-values.yaml` now sets `ingress.enabled: false` (the README's recommended Gateway API setup) and includes a filters-only rule to exercise the bugfix above.

The feature is opt-in and disabled by default, so none of these are user-facing regressions — this is cleanup on top of #1059.

## Not addressed: follow-up #4 (kubeconform Gateway API schema)

Item `#4` from #1080 asked to wire in a Gateway API CRD schema so kubeconform stops silently skipping HTTPRoute under `-ignore-missing-schemas`. An initial attempt pointed `-schema-location` at `raw.githubusercontent.com/datreeio/CRDs-catalog`, but that introduces a per-run network dependency on a third-party host, which doesn't fit air-gapped / offline CI. The alternatives considered (vendor the JSON schema in-repo, download-once-to-`_output/tools/` like `kubeconform` itself, `kubeconform -cache`) each have their own trade-offs and felt out of scope for this follow-up PR.

Closes #1080 partially (`#1` – `#3`)

### DIFF
--- a/build/helm.mk
+++ b/build/helm.mk
@@ -10,6 +10,9 @@ HELM_CHART_NAME = kubernetes-mcp-server
 
 KUBECONFORM = $(shell pwd)/_output/tools/bin/kubeconform
 KUBECONFORM_VERSION ?= latest
+# CRD schemas from the datreeio catalog (e.g. gateway.networking.k8s.io/HTTPRoute); required so kubeconform
+# actually validates the HTTPRoute spec instead of silently skipping it under -ignore-missing-schemas.
+KUBECONFORM_SCHEMA_LOCATIONS = -schema-location default -schema-location 'https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/{{.Group}}/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json'
 
 HELM_DOCS = $(shell pwd)/_output/tools/bin/helm-docs
 HELM_DOCS_VERSION ?= v1.14.2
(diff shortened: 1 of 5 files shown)


matzew:
LGTM!

Thanks for the follow up!

PR #1083, opened 2026-04-21
feat(config): support for logging to file

difrost:
This PR adds support for logging to file instead of stdout/stderr. 

Background:
I'm debugging a weird issue that (after a longer investigation) seem to only happen when server runs in stdio mode when logging is currently disabled. Logging to file helps in debugging such situations. 

@Cali0707 / @manusa / @matzew - please review.


Cali0707:
@difrost I think the standard place we should be logging in the stdio transport is to `stderr`:
```
The server MAY write UTF-8 strings to its standard error (stderr) for any logging purposes including informational, debug, and error messages.
```
Source: https://modelcontextprotocol.io/specification/draft/basic/transports#stdio

That being said, I'm not opposed to having a log file as another option for a place to log

difrost:
> That being said, I'm not opposed to having a log file as another option for a place to log

Hmm ... that's one way to go, At the end one can always wrap it up into a script. 

The options are:
1. in stdio always switch klog to stderr
2. add a config option to expand on 1 (on/off stderr logging)
3. leave the file logging open with stderr as an option and default for stdio
4. ...

manusa (changes requested):
Hey @difrost, thanks a lot for putting this together! Stdio-mode debugging has been a real blind spot and a file sink is exactly the right way to close it. The refactor of `initializeLogging` is also a nice cleanup — pulling the three cases into a single `logOut` variable and bubbling the error up reads much better than the previous split.

A few changes I'd like to see before we merge:

**1. Tighten the default file permissions.**
`0o644` makes log files world-readable. Please switch the `OpenFile` mode to `0o600` — operators who need group access can always chmod after the fact. (While you're touching the literal, `0o600` uses the Go 1.25-preferred `0o` prefix — please apply the same prefix style in the test helpers too.)

**2. SIGHUP reload must reopen the log file — this needs to land with the feature, not later.**
The server already reloads config on SIGHUP (`pkg/kubernetes-mcp-server/cmd/root.go:377`), but `initializeLogging` only runs once in `Complete()`. Consequences as-is:
- Changing `log_file` in the TOML and sending SIGHUP won't redirect logs.
- After `logrotate` rotates the file, we'll keep writing to the old FD — which is the exact stdio-debugging scenario this PR is meant for.

In the SIGHUP path, after `mcpServer.ReloadConfiguration(newConfig)` succeeds, close the old `logFileHandle` and re-run the log-file portion of `initializeLogging` with `newConfig`. If reopening fails, log the error and keep the old handle — don't leave the server with no log sink.

**3. Fix the flag description.**
> \"Write logs to this file path instead of stdout (applies to both stdio and HTTP modes)\"

In stdio mode there's no stdout logging today, so \"instead of stdout\" is misleading. The wording you already have in `docs/logging.md` is great — could you mirror something like *\"Write server logs to this file. Required for logging in stdio mode; replaces stdout logging in HTTP mode.\"*

**4. Restore the load-bearing comment.**
The old comment in the stdio branch explained *why* those klog flags matter (`\"needed to avoid klog writing to stderr and breaking the protocol\"`). The new `\"No output destination: silence klog entirely.\"` drops the rationale — please keep the original one so a future contributor doesn't remove those flags thinking they're redundant.

**Extra (nice to have):**
- `TestLogFile/nonexistent parent directory` relies on `/nonexistent` not existing on the host. `filepath.Join(t.TempDir(), \"missing\", \"server.log\")` is hermetic and equivalent.
- Once SIGHUP reopen is in, add a test that writes to the log file, rotates it (rename + new empty file), sends SIGHUP, and asserts subsequent writes land in the new file.

Thanks again — the existing test matrix (HTTP/stdio/create/append/error/TOML/flag-override-TOML) is exactly what I'd want to see, and with these changes this is in great shape.

difrost:
@manusa  - thanks for review!

> A few changes I'd like to see before we merge:
> 

> **1. Tighten the default file permissions.**
> **3. Fix the flag description.**
> **4. Restore the load-bearing comment.**

Done in [59a8d54](https://github.com/containers/kubernetes-mcp-server/pull/1083/commits/59a8d5402c249e9de3359578ec44a2a066730eec).

> **2. SIGHUP reload must reopen the log file 

I've completely missed the SIGHUP. Done in [59a8d54](https://github.com/containers/kubernetes-mcp-server/pull/1083/commits/59a8d5402c249e9de3359578ec44a2a066730eec) with one major change: - the `s.InitServer()` now returns *MCPServerOptions and registers the log fd cleanup. Added tests covering the log fd swap. 

I've added the 'stderr' option, I find it pretty handy as it allows to correlate protocols read/write with server logs. Might be worth leaving a note about that in `logging.md`....

Cheers

difrost:
@manusa - please review the latest changes that include:

* protocol logging unification: previously stdio was set to log to os.Stderr via LoggingTransport{} now logging covers both stdio and http mode via middleware.
* updated documentation.

I've moved the tool call logging to the `protocolLoggingMiddleware()` tho I'm not sure if it is still required due to generic protocol logging - there's still a value to it e.g. in context of the headers logging but in general the protocol will cover the method and params.

manusa (changes requested):
Dismiss previous version of this comment please

manusa:
Hey @difrost, thanks for the rework on this — the SIGHUP reopen and the protocol-logging unification both read really well.

One thing came up while I was verifying things under `go test -race`: the SIGHUP path hits a real data race. `klog.SetLoggerWithOptions` is documented in klog's own `contextual.go` as not thread-safe ("should be done while no other goroutines invoke log calls, usually during program initialization"). The kubeconfig and cluster-state watcher goroutines call `klog.V(...)` continuously, so `reloadLogFile` swapping the logger from the SIGHUP goroutine collides with them. Race detector reports 15 warnings against the new SIGHUP log tests.

Rather than just describe a fix in words — and because the shape of the change is honestly easier to read in code than in prose — I put up a draft PR against this branch:

👉 difrost/kubernetes-mcp-server#2

The core idea is to extract a `pkg/logging.Sink` type that owns klog and uses an `atomic.Pointer` for the live writer, so reload only swaps a pointer instead of re-installing the logger. The PR also bundles a few related cleanups that fall out of the same refactor: a log-file leak when `Validate()` fails, an orphan SIGHUP goroutine that outlives `Run()`, the eager `jsonCompact` marshal in `protocolLoggingMiddleware` (V(6) args are evaluated before the level gate), some doc wording, and switching `make test` to `-race` so regressions on the atomic invariant get caught in CI. `go test -race -count=1 ./...` is clean against your current branch tip.

Take, leave, or cherry-pick — whatever's easiest. If you'd rather solve the race differently, totally fine; the diff at minimum documents the failure mode and one possible shape. Happy to iterate.

difrost:
Thank you @manusa for contributing the logging reload refactor - that's merged now.

manusa (approved):

matzew:
LGTM

PR #1084, opened 2026-04-21
fix(auth): restore default passthrough of Authorization header to cluster

nader-ziada:
This PR restores the v0.0.60 auth-forwarding behavior that regressed after the Entra ID / OBO work (#953). The Authorization header is always forwarded to the cluster when present and kubeconfig credentials are used as fallback when absent. `require_oauth` controls only whether the token is mandatory and validated, not whether it is forwarded.

 - **Default behavior**: The server forwards the Authorization header to the Kubernetes cluster when present, and uses kubeconfig credentials when absent. No configuration is needed for this.

 - **`require_oauth = true`**: Makes the Authorization header mandatory and validates the token (expiry, audience, and optionally OIDC signature verification if `authorization_url` is set). Requests without a valid token are rejected.

 - **`cluster_auth_mode = "kubeconfig"`**: Forces the server to always use kubeconfig credentials, ignoring any Authorization header. Use this when cluster authentication is managed separately (e.g., via a ServiceAccount).

### Notes beyond strict v0.0.60 restoration

- The HTTP authorization middleware now also extracts the `Authorization: Bearer …` header into the request context when `require_oauth=false`. v0.0.60 did not do this in the HTTP layer — it relied solely on the MCP middleware, which left SSE clients + raw tokens broken (SSE transport does not propagate headers via `req.GetExtra().Header`). This addition is small, additive hardening that makes raw-token forwarding work for SSE clients as well, consistent with the existing `require_oauth=true` handling of the same transport gap.

- `Manager.Derived()` no longer returns the `"oauth token required"` error when invoked with no Bearer token in context under `require_oauth=true`; it falls back to kubeconfig instead. In the HTTP flow this path is unreachable (the authorization middleware already returns 401 first), so dependents are unaffected. Restoring this defense-in-depth guard and broadening integration coverage is tracked as a follow-up.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -98,8 +98,8 @@ type StaticConfig struct {
 	// StsClientKeyFile is the path to the client private key PEM file for JWT assertion auth
 	StsClientKeyFile string `toml:"sts_client_key_file,omitempty"`
 	// ClusterAuthMode determines how the MCP server authenticates to the cluster.
-	// Valid values: "passthrough" (use OAuth token, with optional exchange), "kubeconfig" (use kubeconfig credentials).
-	// If empty, auto-detects: passthrough when require_oauth=true, otherwise kubeconfig.
+	// Valid values: "passthrough" (forward Authorization header, with optional exchange), "kubeconfig" (use kubeconfig credentials).
+	// If empty, defaults to passthrough: forwards the token when present, falls back to kubeconfig when absent.
 	ClusterAuthMode      string `toml:"cluster_auth_mode,omitempty"`
 	CertificateAuthority string `toml:"certificate_authority,omitempty"`
 	ServerURL            string `toml:"server_url,omitempty"`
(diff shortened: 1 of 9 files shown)


pkg/config/config.go
Cali0707: I think this message is a little misleading with these changed, maybe we  can switch to:
`token exchange requires require_oauth=true (token exchange depends on OAuth-validated tokens)`

pkg/kubernetes/manager.go
```diff
@@ -167,13 +167,10 @@ func (m *Manager) Derived(ctx context.Context) (*Kubernetes, error) {
 	authorization, ok := ctx.Value(OAuthAuthorizationHeader).(string)
 	hasToken := ok && strings.HasPrefix(authorization, "Bearer ")
 
-	// No token: use kubeconfig credentials, or reject if passthrough requires one.
+	// No token: fall back to kubeconfig credentials.
```
Cali0707, line 170: should we log a warning in this case?

nader-ziada: added a warning log

manusa (approved):
LGTM, thx!

Issue #1087, opened 2026-04-22
fix(auth): restore Derived() defense-in-depth and broaden raw-token integration coverage (follow-up to #1084)


manusa:
## Context

PR #1084 restored the v0.0.60 behavior of forwarding the client's `Authorization: Bearer …` header to the target Kubernetes cluster. That PR fixed the primary regression introduced by the Entra ID / OBO work (#953), where `ResolveClusterAuthMode()` was auto-selecting `kubeconfig` when `require_oauth=false`, and `stsExchangeTokenInContext` was consequently wiping the propagated token before `Manager.Derived()` saw it.

During review we identified two small items that were intentionally left out of #1084 to keep the fix minimal and low-risk for a fast merge. They do not affect dependents (OpenShift LightSpeed, etc.) and are safe to address in a follow-up.

## 1. Restore the `"oauth token required"` guard in `Manager.Derived()`

In v0.0.60, `Manager.Derived()` explicitly rejected requests that reached it without a Bearer token when `require_oauth=true`:

```go
// pkg/kubernetes/manager.go (v0.0.60, lines 117-123)
authorization, ok := ctx.Value(OAuthAuthorizationHeader).(string)
if !ok || !strings.HasPrefix(authorization, "Bearer ") {
    if m.config.IsRequireOAuth() {
        return nil, errors.New(\"oauth token required\")
    }
    return m.kubernetes, nil
}
```

PR #1084 removed this error in favor of an unconditional fallback to kubeconfig. The HTTP authorization middleware still rejects token-less requests with 401 when `require_oauth=true`, so this path is unreachable via HTTP — but the defense-in-depth check was useful for:

- STDIO / internal code paths that call `Derived()` with a fresh context.
- Guarding against regressions where a future change could route around the HTTP middleware.
- Maintaining the operator contract: \"I set `require_oauth=true`, therefore cluster ops must never fall through to kubeconfig.\"

### Proposed change

Reinstate the guard in `pkg/kubernetes/manager.go`:

```go
if !hasToken {
    if m.config.IsRequireOAuth() {
        return nil, errors.New(\"oauth token required\")
    }
    return m.kubernetes, nil
}
```

And revert the two corresponding test flips in `pkg/kubernetes/kubernetes_derived_test.go:271-286` back to asserting the error.

## 2. Broaden raw-token integration coverage

[...]

PR #1088, opened 2026-04-22
fix(auth): restore Derived() require_oauth guard and broaden raw-token coverage

manusa:
## Summary

Follow-up to #1084, closing #1087.

- Restore the `"oauth token required"` guard in `Manager.Derived()` for `require_oauth=true`, as defense-in-depth for STDIO / internal paths; HTTP middleware already rejects token-less requests with 401.
- Add an **SSE** counterpart to `TestAuthorizationRawPassthrough` — the raw-token-to-backend path on SSE is new relative to v0.0.60 and wasn't covered end-to-end.
- Add `TestAuthorizationClusterAuthModeKubeconfigDropsClientToken` to lock in the `stsExchangeTokenInContext` clear behavior at the HTTP level.
- Add `TestClusterAuthMode` assertions for the reworded `"token exchange requires require_oauth=true"` error, covering both `token_exchange_strategy` and `sts_audience` as triggers.
- Revert the two kubeconfig-derived test flips from #1084 back to asserting the error.

### API-surface note

Adds `IsRequireOAuth()` to the `BaseConfig` interface via a new `RequireOAuthProvider`. In-tree, only `*StaticConfig` satisfies `BaseConfig` (`RequireOAuth` already exists on it), so the new method is a trivial shim. Any out-of-tree `BaseConfig` implementer would need to add the method — mild, worth flagging.

## Test plan

- [x] `make test` passes (including new cases under `TestDerived`, `TestValidate/TestClusterAuthMode`, `TestAuthorization/TestAuthorizationRawPassthroughSSE`, and `TestAuthorization/TestAuthorizationClusterAuthModeKubeconfigDropsClientToken`).
- [x] `make lint` clean.
- [x] `TestAuthorization/TestAuthorizationRawPassthroughStreamableHTTP` still passes after splitting from the merged dual-transport test.
- [x] `TestDerived` `require_oauth=true` no-token / invalid-token cases now assert `"oauth token required"` rather than fallback.

## Related

- #1084 — primary fix
- #1087 — follow-up issue tracking these items
- #953 — Entra ID / OBO (regression source)
- #1078 — `skip_jwt_verification` (orthogonal, unchanged)

### DIFF
--- a/pkg/kubernetes/kubernetes_derived_test.go
+++ b/pkg/kubernetes/kubernetes_derived_test.go
@@ -268,23 +268,25 @@ users:
 			require_oauth = true
 		`)))
 
-		s.Run("with no authorization header falls back to kubeconfig", func() {
+		s.Run("with no authorization header returns error", func() {
 			testManager, err := NewKubeconfigManager(testStaticConfig, "")
 			s.Require().NoErrorf(err, "failed to create test manager: %v", err)
 
 			derived, err := testManager.Derived(s.T().Context())
-			s.Require().NoError(err)
-			s.Equal(derived, testManager.kubernetes, "expected original client when no token")
+			s.Require().Error(err, "expected error when no token and RequireOAuth=true")
+			s.ErrorContains(err, "oauth token required")
+			s.Nil(derived, "expected nil derived kubernetes when no token and RequireOAuth=true")
 		})
 
-		s.Run("with invalid authorization header falls back to kubeconfig", func() {
+		s.Run("with invalid authorization header returns error", func() {
 			testManager, err := NewKubeconfigManager(testStaticConfig, "")
 			s.Require().NoErrorf(err, "failed to create test manager: %v", err)
 
 			ctx := context.WithValue(s.T().Context(), HeaderKey("Authorization"), "invalid-token")
 			derived, err := testManager.Derived(ctx)
-			s.Require().NoError(err)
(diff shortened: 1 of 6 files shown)


pkg/kubernetes/manager.go
```diff
@@ -167,10 +167,16 @@ func (m *Manager) Derived(ctx context.Context) (*Kubernetes, error) {
...
 
-	// No token: fall back to kubeconfig credentials.
+	// No token: fall back to kubeconfig credentials, unless require_oauth=true.
 	// In kubeconfig mode, the token exchange layer clears the auth header before we get here,
 	// so this branch handles both "no token sent" and "kubeconfig mode cleared it".
+	// The require_oauth guard is defense-in-depth: the HTTP middleware already rejects
+	// token-less requests with 401, but this protects STDIO / internal paths and
+	// preserves the operator contract that require_oauth=true never falls through to kubeconfig.
 	if !hasToken {
+		if m.config.IsRequireOAuth() {
```
matzew, line 177: clean and minimal change !  Thanks!

pkg/kubernetes/kubernetes_derived_test.go
```diff
@@ -268,23 +268,25 @@ users:
 			require_oauth = true
 		`)))
 
-		s.Run("with no authorization header falls back to kubeconfig", func() {
```
matzew, line 271: thx for the revert

matzew (approved):
LGTM

PR #1090, opened 2026-04-22
fix(helm-charts): enable read-only root filesystem in Helm chart (security)

### LINKED ISSUE #1089: fix(security): enable read-only root filesystem in Helm chart
The default container security context does not set `readOnlyRootFilesystem: true`. While the binary performs no filesystem writes in production and already runs non-root with all capabilities dropped, a writable root filesystem allows post-exploitation persistence if the container is compromised (CWE-552).

Changes:
- Set `readOnlyRootFilesystem: true` in `defaultSecurityContext`
- Add `/tmp` emptyDir volume for Helm SDK and Go stdlib temp file needs

matzew:
Enable `readOnlyRootFilesystem: true` in the default container security context and add a `/tmp` emptyDir volume for runtime temp file needs. 

Fixes #1089

### DIFF
--- a/charts/kubernetes-mcp-server/README.md
+++ b/charts/kubernetes-mcp-server/README.md
@@ -99,9 +99,9 @@ Each container accepts any valid Kubernetes container field including `image`, `
 | config.port | string | `"{{ .Values.service.port }}"` |  |
 | configFilePath | string | `"/etc/kubernetes-mcp-server/config.toml"` |  |
 | defaultPodSecurityContext | object | `{"seccompProfile":{"type":"RuntimeDefault"}}` | Default Security Context for the Pod when one is not provided |
-| defaultSecurityContext | object | `{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"runAsNonRoot":true}` | Default Security Context for the Container when one is not provided |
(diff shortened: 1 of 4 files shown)


manusa (approved):
Thanks! Solid, low-risk security hardening — LGTM.

A couple of optional follow-ups (non-blocking):

1. **CI coverage for the default vs. override paths.** `charts/kubernetes-mcp-server/ci/` already has test-values files exercising `helm lint` / `helm template`, but none covers a `securityContext` override. Adding something like `ci/security-context-test-values.yaml` with (a) an empty override to lock in the new default and (b) a custom override to verify the default is correctly replaced would prevent regressions on this code path.

2. **Heads-up for sidecar users.** `defaultSecurityContext` and the new `/tmp` mount apply only to the main container. Anyone injecting a sidecar via `extraContainers` that expects a writable `/tmp` (or a restricted security context) will need to configure it themselves. Worth a one-line comment near `extraContainers:` in `values.yaml`, e.g.:

    ```yaml
    # -- Note: sidecars do not inherit defaultSecurityContext or the /tmp emptyDir mount; configure them per-container if needed.
    extraContainers: []
    ```

matzew:
thx,

let me update!

matzew:
@manusa updated this

PR #1091, opened 2026-04-22
feat(api): add MCP resource

swghosh:
Adds boilerplate for registering [MCP resources](https://modelcontextprotocol.io/specification/2025-06-18/server/resources) using the mcp-go-sdk.

xref: https://github.com/openshift/openshift-mcp-server/pull/159#discussion_r3095830141

### DIFF
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -47,6 +47,12 @@ type Toolset interface {
 	// GetPrompts returns the prompts provided by this toolset.
 	// Returns nil if the toolset doesn't provide any prompts.
 	GetPrompts() []ServerPrompt
+	// GetResources returns the resources provided by this toolset.
+	// Returns nil if the toolset doesn't provide any resources.
+	GetResources() []ServerResource
+	// GetResourceTemplates returns the resource templates provided by this toolset.
+	// Returns nil if the toolset doesn't provide any resource templates.
+	GetResourceTemplates() []ServerResourceTemplate
 }
 
 type ToolCallRequest interface {
@@ -101,6 +107,40 @@ func NewToolCallResultStructured(structured any, err error) *ToolCallResult {
 		}
 	}
 	return NewToolCallResultFull(content, structured, err)
+}
+
+// Resource represents the metadata of an MCP resource.
+type Resource struct {
+	URI         string
+	Name        string
+	Description string
+	MIMEType    string
+}
+
+// ResourceHandler is called when a client reads a resource.
+type ResourceHandler func(ctx context.Context) (content string, err error)
+
+// ServerResource represents a resource that can be registered with the MCP server.
+type ServerResource struct {
+	Resource Resource
+	Handler  ResourceHandler
+}
+
+// ResourceTemplate represents the metadata of an MCP resource template.
+type ResourceTemplate struct {
+	URITemplate string
+	Name        string
+	Description string
+	MIMEType    string
+}
+
+// ResourceTemplateHandler is called when a client reads a resource matching a template.
+type ResourceTemplateHandler func(ctx context.Context, uri string) (content string, err error)
+
+// ServerResourceTemplate represents a resource template that can be registered with the MCP server.
+type ServerResourceTemplate struct {
+	ResourceTemplate ResourceTemplate
+	Handler          ResourceTemplateHandler
 }
 
 type ToolHandlerParams struct {
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -117,8 +117,21 @@ type Resource struct {
 	MIMEType    string
 }
 
+// ResourceContent is the value returned by a resource handler.
+// Exactly one of Text or Blob must be set; both cannot be nil or both non-nil.
+type ResourceContent struct {
+	// MIMEType overrides Resource.MIMEType when set; otherwise Resource.MIMEType is used.
+	MIMEType string
+	// Text is the UTF-8 text content of the resource.
+	Text string
+	// Blob is the binary content of the resource.
+	Blob []byte
+}
+
 // ResourceHandler is called when a client reads a resource.
-type ResourceHandler func(ctx context.Context) (content string, err error)
+// Session state (auth, request context) is available on ctx via sessionInjectionMiddleware.
+// Handlers should return a ResourceContent with exactly one of Text or Blob set.
+type ResourceHandler func(ctx context.Context) (*ResourceContent, error)
 
 // ServerResource represents a resource that can be registered with the MCP server.
 type ServerResource struct {
@@ -135,7 +148,10 @@ type ResourceTemplate struct {
 }
 
 // ResourceTemplateHandler is called when a client reads a resource matching a template.
-type ResourceTemplateHandler func(ctx context.Context, uri string) (content string, err error)
+// Session state (auth, request context) is available on ctx via sessionInjectionMiddleware.
+// The uri parameter is the actual resource URI that matches the template.
+// Handlers should return a ResourceContent with exactly one of Text or Blob set.
+type ResourceTemplateHandler func(ctx context.Context, uri string) (*ResourceContent, error)
 
 // ServerResourceTemplate represents a resource template that can be registered with the MCP server.
 type ServerResourceTemplate struct {
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -318,8 +318,13 @@ The `docs/specs/` directory contains feature specifications (living documentatio
 ### Documentation conventions
 
 - Use **lowercase filenames** for new documentation files (e.g., `configuration.md`, `prompts.md`)
-- The toolsets table in `README.md` and `docs/configuration.md` is **auto-generated** - use `make update-readme-tools` to update it
-- Both files use markers (`<!-- AVAILABLE-TOOLSETS-START -->` / `<!-- AVAILABLE-TOOLSETS-END -->`) for the generated content
+- The toolsets table, tools, prompts, resources, and resource templates in `README.md` and `docs/configuration.md` are **auto-generated** - use `make update-readme-tools` to update them after modifying toolsets
+- Both files use marker pairs for the generated content:
+  - `<!-- AVAILABLE-TOOLSETS-START -->` / `<!-- AVAILABLE-TOOLSETS-END -->` (toolset summary table)
+  - `<!-- AVAILABLE-TOOLSETS-TOOLS-START -->` / `<!-- AVAILABLE-TOOLSETS-TOOLS-END -->` (tool details)
+  - `<!-- AVAILABLE-TOOLSETS-PROMPTS-START -->` / `<!-- AVAILABLE-TOOLSETS-PROMPTS-END -->` (prompt details)
+  - `<!-- AVAILABLE-TOOLSETS-RESOURCES-START -->` / `<!-- AVAILABLE-TOOLSETS-RESOURCES-END -->` (resource details)
+  - `<!-- AVAILABLE-TOOLSETS-RESOURCES-TEMPLATES-START -->` / `<!-- AVAILABLE-TOOLSETS-RESOURCES-TEMPLATES-END -->` (resource template details)
 
 ## Distribution Methods
 
--- a/README.md
+++ b/README.md
@@ -597,6 +597,20 @@ In case multi-cluster support is enabled (default) and you have access to multip
 
 <!-- AVAILABLE-TOOLSETS-PROMPTS-END -->
 
+### Resources
+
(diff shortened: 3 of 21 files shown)


pkg/api/toolsets.go
```diff
@@ -103,13 +103,30 @@ func NewToolCallResultStructured(structured any, err error) *ToolCallResult {
 	return NewToolCallResultFull(content, structured, err)
 }
 
+// ResourceRegistrar allows tools to register MCP resources at runtime.
+type ResourceRegistrar interface {
+	AddResource(uri, name, description, mimeType, content string)
+	RemoveResources(uris ...string)
+}
```
Cali0707, line 148: do we also want toolsets to be able to register resources statically (in addition to this dynamic mechanism?)

I'm think that some toolsets may have a static set of resources, and going through this API may be a little awkward (especially since toolsets are initialized _before_ the mcp server is)

Cali0707 (commented):
Overall seems fine, I just worry about the dynamic registration patter in the context of toolsets that want to register resources on initialization

manusa (changes requested):
Thanks for addressing the resources gap.

Reading the consumer in [openshift-mcp-server#159](https://github.com/openshift/openshift-mcp-server/pull/159), I think we should push this toward the same shape we already use for tools and prompts — static declaration on the `Toolset`, reloaded centrally, no registrar on `ToolHandlerParams`. Concretely:

```go
type Toolset interface {
    ...
    GetResources() []ServerResource
    GetResourceTemplates() []ServerResourceTemplate
}
```

wired through `reloadToolsets`/`reloadItems` exactly like `GetTools`/`GetPrompts`. Capability gets declared once in `NewServer`.

The MCP spec already gives us two mechanisms that cover the must-gather cases without runtime registration:

- **Static resource** for fixed artifacts. A loaded archive has a well-known URI space the client can discover:

  ```go
  {URI: "must-gather://current", Name: "must-gather", MIMEType: "text/plain", ...}
  ```

- **URI template** for the parameterized fan-out (nodes, operators, etcd health, prometheus replicas, etc.) — declared once, resolved per read:

  ```go
  {URITemplate: "must-gather://{archive}/etcd/health", ...}
  {URITemplate: "must-gather://{archive}/operators/{name}", ...}
  {URITemplate: "must-gather://{archive}/prometheus/replica/{n}", ...}
  ```

  The `mustgather_load` tool still loads the archive and caches the index; the template handler looks up the cached archive and returns the slice the URI points at. No per-object `AddResource` call, no list-changed chatter, no mid-tool-call mutation of global server state.

The one case where runtime `AddResource` is genuinely justified is "user just loaded an archive and the client should see it appear in `resources/list`" — but even that can be a template `must-gather://{archive}` with the handler returning an error for unknown archives. And if we really do need runtime pinning later, a narrow session-scoped API for that case is cleaner than making every tool handler carry a registrar.

One more thing worth weighing: **real-world MCP host support for resources is still thin, and for dynamic/list-changed resources it's thinner still.** Plenty of popular hosts either ignore `resources/list` entirely, don't surface resources in the UI, or don't re-fetch on `notifications/resources/list_changed`. If we bet the design on runtime registration, we're building on the shakiest part of the host ecosystem — and users will mostly experience it as "the tool call ran but nothing showed up." Static resources + templates are the part of the spec hosts are most likely to handle correctly today, which is another reason to prefer them as the default path and keep any dynamic escape hatch deliberately narrow.

Net: I'd like us to drop `ResourceRegistrar`/`ResourceTemplateRegistrar` from `ToolHandlerParams`, add `GetResources()`/`GetResourceTemplates()` to `Toolset`, and declare the `Resources` capability in `NewServer`. The `resource.go` wrappers can stay — they'd just be called from `registerResource`/`registerResourceTemplate` in `mcp.go`. Happy to discuss if there's a concrete case I'm missing from the must-gather design.

swghosh:
ALL, my absolute bad. I accidentally rebased this PR with downstream instead of upstream which caused all the unecessary noise.

Sorry for all the notifs you've received as a result of this PR.

swghosh:
> Thanks for addressing the resources gap.
> 
> Reading the consumer in [openshift-mcp-server#159](https://github.com/openshift/openshift-mcp-server/pull/159), I think we should push this toward the same shape we already use for tools and prompts — static declaration on the `Toolset`, reloaded centrally, no registrar on `ToolHandlerParams`. Concretely:
> 
> ```go
> type Toolset interface {
>     ...
>     GetResources() []ServerResource
>     GetResourceTemplates() []ServerResourceTemplate
> }
> ```
> 
> wired through `reloadToolsets`/`reloadItems` exactly like `GetTools`/`GetPrompts`. Capability gets declared once in `NewServer`.
> 
> The MCP spec already gives us two mechanisms that cover the must-gather cases without runtime registration:
>  ...
> 
> Net: I'd like us to drop `ResourceRegistrar`/`ResourceTemplateRegistrar` from `ToolHandlerParams`, add `GetResources()`/`GetResourceTemplates()` to `Toolset`, and declare the `Resources` capability in `NewServer`. The `resource.go` wrappers can stay — they'd just be called from `registerResource`/`registerResourceTemplate` in `mcp.go`. Happy to discuss if there's a concrete case I'm missing from the must-gather design.

@manusa Thanks for your foresight on this, I've updated the PR with the recommended design. For our must-gather use case the new shape of the API should very well suffice!

pkg/api/toolsets.go
swghosh: yeah true, supersed by https://github.com/containers/kubernetes-mcp-server/pull/1091#issuecomment-4305694195 :)

swghosh:
@Cali0707 could you PTAL, thanks!

manusa (changes requested):
Thanks for the iteration — the shape is now what I had in mind, and the five items from the previous review are all in. Before we land this, I'd like us to fold in a few hardening items so the public API and the generic reload machinery stay symmetric with what we already have for tools and prompts. None of these are large; the goal is to avoid a follow-up that reshuffles the same files.

**1. Wire the filter/mutator pipeline now, even if it's a no-op.**

`collectApplicableTools` (`pkg/mcp/mcp.go:260`) and `collectApplicablePrompts` (`pkg/mcp/mcp.go:284`) both run their items through `CompositeFilter` / `ComposeMutators` (and prompts additionally through `MergePrompts`). The two new collectors skip that scaffolding entirely. I'd rather we add the empty filter/mutator call sites in this PR — even with zero composed pieces today — so that when `DisabledResources` or a per-resource mutator inevitably shows up, it lands in one obvious place instead of becoming a third pattern. Same treatment for `collectApplicableResourceTemplates`.

**2. Build-time uniqueness check for resource URIs (and templates).**

`pkg/toolsets/toolsets_test.go:42` already enforces unique toolset names via `TestRegisterPanicsOnDuplicate`. Let's add a sibling test in `pkg/toolsets/toolsets_test.go` that walks every registered toolset and asserts uniqueness across the whole server for:

- `GetResources()[i].Resource.URI`
- `GetResourceTemplates()[i].ResourceTemplate.URITemplate`

Build-time check, no runtime cost. If/when resources become configurable, the equivalent check moves into config validation — but for now, statically-declared resources should never collide and we want CI to catch it. Worth doing the same sweep for tool names and prompt names while you're in there if they aren't already covered.

**3. Widen `ResourceHandler` / `ResourceTemplateHandler` before this becomes API.**

Once we ship `pkg/api/toolsets.go` with `func(ctx) (string, error)`, every binary resource (PDF, archive, image — all plausible for must-gather, kubevirt console screenshots, etc.) forces a breaking change. The SDK already has the right shape at `go-sdk@v1.5.0/mcp/content.go:290` — `ResourceContents{URI, MIMEType, Text, Blob, Meta}`. Mirror it on our side:

```go
// ResourceContent is the value returned by a resource handler.
// Exactly one of Text or Blob must be set.
type ResourceContent struct {
    MIMEType string  // overrides Resource.MIMEType when set
    Text     string
    Blob     []byte
}

type ResourceHandler         func(ctx context.Context) (*ResourceContent, error)
type ResourceTemplateHandler func(ctx context.Context, uri string) (*ResourceContent, error)
```

`pkg/mcp/resource.go` then maps `Text`-xor-`Blob` into the SDK's `ResourceContents` and falls back to the declared MIME type when the handler doesn't override it.

**4. Pass the request through to handlers.**

Right now `pkg/mcp/resource.go` drops `*mcp.ReadResourceRequest` on the floor for static resources. Session/auth state already rides on `ctx` via `sessionInjectionMiddleware`, so handlers technically have access today — but if the SDK ever surfaces something on the request that isn't on the context, we'd be stuck. Either pass the request through, or document explicitly in the handler godoc that everything handlers need is on `ctx`.

**5. Test hardening + parity with the tool/prompt suites.**

A few things in `pkg/mcp/resource_test.go`:

- `pkg/mcp/resource_test.go:80,86,113,120` — pair `s.Require().NoError(err)` with `s.Require().Len(result.Contents, 1)` before dereferencing `Contents[0]`. CLAUDE.md \"Error Handling\" calls this out specifically; right now an empty result panics instead of producing a clean failure.
- Three scenarios are missing — please add them, modeled on the equivalent coverage in `pkg/mcp/mcp_tools_test.go` and `pkg/mcp/mcp_toolset_prompts_test.go` so all three suites stay structurally diff-friendly:
  - **handler returns an error** — verify it propagates as a protocol error and no `Contents` leak through.
  - **toolset reload removes a resource** — the diff adds `RemoveResources` / `RemoveResourceTemplates` plumbing through `reloadItems` but never exercises it.
  - **`notifications/resources/list_changed` fires** when `Stateless=false` — mirror the prompts test.

**6. Generate the toolset resource catalog in the README.**

`internal/tools/update-readme/main.go:110-136` already has the \"Available Toolset Prompts\" section: collapsible per-toolset block, skip when empty, marker pair in `README.md`. Please add two parallel sections:

- `<!-- AVAILABLE-TOOLSETS-RESOURCES-START -->` / `…END -->`
- `<!-- AVAILABLE-TOOLSETS-RESOURCES-TEMPLATES-START -->` / `…END -->`

Each iterates `GetResources()` / `GetResourceTemplates()`, skips empty toolsets with the same `if len(...) == 0 { continue }` guard the prompts loop uses, and prints `URI`/`URITemplate` + `Name` + `MIMEType` + `Description`. Add the marker pairs to both `README.md` and `docs/configuration.md`, and extend the \"Documentation conventions\" bullet in `CLAUDE.md` so the next contributor knows to run `make update-readme-tools` after touching resources.

---

Net: all six need addressing before this lands. They keep the four item kinds (tools, prompts, resources, resource templates) on the same rails, and **2** and **3** in particular get materially harder to add once the API is out and consumers start declaring resources against it. Happy to discuss any of these if a particular shape doesn't fit what you're seeing on the consumer side.

manusa (commented):
Thanks for the iteration. Down to the four items that block merge for me; everything else can land as a follow-up issue.

### Must Fix

1. **The new uniqueness tests are vacuous** — `pkg/toolsets/toolsets_test.go:53,63,73,83`. `SetupTest` (line 16-18) calls `Clear()` after capturing `s.originalToolsets`, so `Toolsets()` inside the new tests returns an empty slice and the maps are populated zero times. They always pass without asserting anything. Iterate `s.originalToolsets` instead, and pass a non-nil `api.Openshift` stub to `GetTools` (the `core` toolset dereferences `o.IsOpenShift(...)` at `pkg/toolsets/core/namespaces.go:31` and would nil-panic with `nil`).

2. **`notifications/resources/list_changed` is enabled but untested** — `pkg/mcp/mcp_watch_test.go` covers tools (line 67) and prompts (line 78); no parallel for resources. `ResourceCapabilities{ListChanged: !Stateless}` at `pkg/mcp/mcp.go:93` ships untouched by the suite. Mirror `TestNotifiesPromptsChange`.

3. **Toolset-reload-removes-a-resource is untested** — `RemoveResources` / `RemoveResourceTemplates` plumbing in `reloadItems` (`pkg/mcp/mcp.go:307-326`) is exercised only on add. `mcp_watch_test.go:108-145 TestClearsNoLongerAvailableTools` is the existing parallel for tools — please mirror it.

4. **Nil-deref in `addResource` / `addResourceTemplate`** — `pkg/mcp/resource.go:23-36, :53-66`. After the `err != nil` check, `content.MIMEType`/`content.Text`/`content.Blob` are read unconditionally. A handler returning `(nil, nil)` panics inside the SDK request goroutine. Either reject nil with an explicit error or treat nil as empty content.

The remaining items from my earlier pass (assertion pairing in `resource_test.go`, `Blob`/MIMEType-override coverage, filter/mutator tests, `AddResource` URI-validation panic) are real but small enough to track in a follow-up. Once 1–4 are in, this is good to merge.

swghosh:
@manusa could you PTAL, thanks!

manusa (commented):
LGTM, thx!

PR #1093, opened 2026-04-22
fix(helm-charts): make ServiceAccount token auto-mount configurable on Helm chart (security) 

### LINKED ISSUE #1092: fix(helm): make ServiceAccount token auto-mount configurable
The Helm chart unconditionally auto-mounts the ServiceAccount token into the pod. In OAuth passthrough deployments where the server uses the user's token (not the SA token), this is unnecessary privilege exposure (CWE-250).

Add `serviceAccount.automountToken` value (default `true`) so operators can disable it for passthrough-mode deployments.

matzew:
Add `serviceAccount.automountToken` value (`default: true`) to control whether the `ServiceAccount` token is mounted into the pod. Operators using OAuth passthrough mode can set this to false to reduce the attack surface when the server's own SA token is not needed (CWE-250).

- Set `automountServiceAccountToken` on both the Deployment pod spec and the `ServiceAccount` resource for defense-in-depth
- Use `hasKey` guard to correctly handle boolean false values (Go template `default` treats false as empty)
- Add CI test values exercising `automountToken: false`

Fixes https://github.com/containers/kubernetes-mcp-server/issues/1092


manusa (approved):
LGTM, thx!

PR #1103, opened 2026-04-23
fix(kiali): use dynamic toolset name in tests to avoid downstream conflicts

matzew:
Replaces hardcoded `kiali_` tool name prefixes in `KialiSuite` tests with the dynamic toolset name from the Toolset struct. This allows downstream forks that override the toolset name to reuse the same test code without merge conflicts.

Follow up to #969


aljesusg (approved):
LGFM

manusa:
Thx Matthias!

PR #1105, opened 2026-04-24
fix(auth): propagate config reload to HTTP authorization middleware

### LINKED ISSUE #1104: SIGHUP config reload does not update HTTP authorization middleware
When configuration is reloaded via SIGHUP, auth-related settings (`require_oauth`, `oauth_audience`, `skip_jwt_verification`) are not picked up by the HTTP authorization middleware. The middleware continues using the original startup config, while the MCP server internals and OIDC provider are updated correctly — creating a split-brain where the operator believes auth settings have changed but the HTTP layer silently ignores them.

**Root Cause**
AuthorizationMiddleware captures a *config.StaticConfig pointer at startup via Serve() in root.go:363. On SIGHUP, config.Read() creates a new *StaticConfig object and passes it to mcpServer.ReloadConfiguration(), which updates the MCP server's internal copy. However, the original pointer held by the middleware is never updated — it still references the old StaticConfig struct.

The oauthState (OIDC provider) is correctly updated via an atomic.Pointer swap, but the StaticConfig fields the middleware reads on every request (RequireOAuth, OAuthAudience, SkipJWTVerification, AuthorizationURL) remain stale.

**Impact**
After a SIGHUP reload, the MCP server internals (tools, prompts, toolsets) and the OIDC provider (signature verification) correctly use the new config. However, the HTTP authorization middleware continues reading from the old config. This means:

- Changing oauth_audience from "" to "mcp-server" via SIGHUP has no effect — the middleware never enforces it.
- Enabling require_oauth via SIGHUP does not protect the server — unauthenticated requests continue to pass.
- Disabling require_oauth via SIGHUP does not open the server — requests are still rejected.
- Changing skip_jwt_verification has no effect on request handling.
- The operator receives a "Configuration reloaded successfully via SIGHUP" log message in all cases, with no indication that auth settings were not applied.

nader-ziada:
Fixes: #1104 

The SIGHUP handler creates a new StaticConfig on reload and updates the MCP server internals and OIDC provider, but the HTTP authorization middleware continued reading from the original pointer making auth setting changes (require_oauth, oauth_audience, skip_jwt_verification) silently ineffective after reload.

Wrap StaticConfig in an atomic pointer (StaticConfigState), matching the existing oauth.State pattern. The middleware calls Load() per-request so SIGHUP-reloaded auth settings take effect immediately.

### DIFF
--- a/pkg/config/state.go
+++ b/pkg/config/state.go
@@ -1,0 +1,27 @@
+package config
+
+import "sync/atomic"
+
+// StaticConfigState holds the current StaticConfig and allows atomic, lock-free reads.
+// This enables hot-reloading of configuration via SIGHUP while ensuring all consumers
+// (e.g., HTTP middleware) always see the latest config snapshot.
+type StaticConfigState struct {
+	ref atomic.Pointer[StaticConfig]
+}
+
+// NewStaticConfigState creates a new StaticConfigState initialized with the given config.
+func NewStaticConfigState(cfg *StaticConfig) *StaticConfigState {
+	s := &StaticConfigState{}
+	s.ref.Store(cfg)
+	return s
+}
+
+// Load returns the current StaticConfig. Safe for concurrent use.
+func (s *StaticConfigState) Load() *StaticConfig {
+	return s.ref.Load()
+}
+
+// Store atomically replaces the current StaticConfig.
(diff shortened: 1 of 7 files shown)


manusa (approved):
LGTM, thx!

PR #1107, opened 2026-04-24
fix(http): complete hot-reload surface for HTTP middleware config

### LINKED ISSUE #1106: follow-up: complete hot-reload surface for HTTP middleware config
Follow-up to #1104 / #1105.

The fix in #1105 correctly propagates SIGHUP-reloaded config to `AuthorizationMiddleware` and `WellKnownHandler` via `config.StaticConfigState`, but the reload surface is still partial. This issue tracks the remaining cleanups.

## 1. Partial hot-reload creates an inconsistency surface

In `pkg/http/http.go:107-113`:

- `RequestMiddleware(staticConfig.TrustProxyHeaders)` and `MaxBodyMiddleware(staticConfig.HTTP.MaxBodyBytes)` are bound using the **startup** snapshot.
- The well-known handler (`pkg/http/wellknown.go:333-339`) reads `TrustProxyHeaders` **per-request** via `cfgState.Load()`.

After a SIGHUP that flips `trust_proxy_headers`, the well-known metadata handler picks up the change but the request middleware does not — the two layers disagree about whether to trust `X-Forwarded-*` headers.

Options:
- Extend the pattern so `RequestMiddleware` (and `MaxBodyMiddleware`) also read per-request, aligning the whole chain.
- Or document explicitly (in `docs/configuration.md` and/or `docs/specs/`) which fields are hot-reloadable and which require a restart.

The existing comment at `pkg/http/http.go:108` — *"Read startup-time settings that cannot change at runtime (port, TLS, timeouts)"* — undersells the caveat: `MaxBodyBytes` and `TrustProxyHeaders` aren't port/TLS/timeout settings. Tighten the wording regardless of which option is chosen.

## 2. No regression test for the original bug

The SIGHUP tests in `pkg/kubernetes-mcp-server/cmd/root_sighup_test.go` verify that `ReloadConfiguration` runs on SIGHUP, but nothing asserts the middleware's observed auth config changes after a reload.

Add a small test that:
1. Stores an initial config with `RequireOAuth=false`.
2. Issues a request (expects 200).
3. Calls `cfgState.Store` with `RequireOAuth=true` (or triggers SIGHUP end-to-end).
4. Asserts the next unauthenticated request receives `401`.

This directly exercises the fix and locks in the behavior against future regressions.

A minimal unit test for `StaticConfigState` itself (Load/Store round-trip, race-free under `-race`) mirroring `pkg/oauth/state_test.go` would also be cheap insurance.

## 3. Minor: nil guard / doc on `StaticConfigState.Store`

`StaticConfigState.Store` accepts `*StaticConfig` with no nil guard. Storing nil would blow up downstream consumers that dereference without a check (e.g., `pkg/http/authorization.go:70`, which uses `staticConfig.RequireOAuth` with no nil check). Not a bug given current call sites, but either:

- Add `if cfg == nil { return }` in `Store`, or
- Add a doc comment forbidding nil on `Store` / documenting the non-nil invariant on `Load`.

Also normalize the nil-check style across callsites — `wellknown.go` does defensive `cfg != nil` checks, `authorization.go` does not.

manusa:
## Summary

Follow-up to #1105. The earlier fix propagated SIGHUP-reloaded config to `AuthorizationMiddleware` and `WellKnownHandler` via `config.StaticConfigState`, but `RequestMiddleware` and `MaxBodyMiddleware` were still bound to the startup snapshot. After a reload that flipped `trust_proxy_headers`, the well-known handler (per-request `Load()`) and `RequestMiddleware` (stale startup value) disagreed about whether to trust `X-Forwarded-*` headers.

## Changes

- **Per-request reload for the whole middleware chain** — `RequestMiddleware` and `MaxBodyMiddleware` now take `*config.StaticConfigState` and read `trust_proxy_headers` / `max_body_bytes` on every request, matching the pattern used by `AuthorizationMiddleware` and `WellKnownHandler`.
- **Readable chain composition** — introduced a small local `chain(handler, middlewares...)` helper in `pkg/http/middleware.go` so the middleware list in `Serve` reads top-down (outermost first) instead of nested inside-out.
- **`StaticConfigState` hardening** — documented the non-nil invariant on `Load()`, `Store(nil)` is now a no-op, and the redundant `cfg != nil` checks in `wellknown.buildResourceURL` are removed for style consistency with `authorization.go`.
- **Regression tests** — new `TestAuthorizationMiddlewareReload` locks in the #1105 behavior (200 → 401 after flipping `require_oauth` via `cfgState.Store`); new reload subtests in `TestTrustProxyHeaders` / `TestMaxBodyMiddleware` cover the extension; new `TestStaticConfigState` unit test (Load/Store round-trip, race-safe).

Closes #1106

### DIFF
--- a/pkg/config/state.go
+++ b/pkg/config/state.go
@@ -5,11 +5,17 @@ import "sync/atomic"
 // StaticConfigState holds the current StaticConfig and allows atomic, lock-free reads.
 // This enables hot-reloading of configuration via SIGHUP while ensuring all consumers
 // (e.g., HTTP middleware) always see the latest config snapshot.
+//
+// Non-nil invariant: once constructed via NewStaticConfigState with a non-nil
+// *StaticConfig, Load always returns a non-nil pointer. Store silently ignores
+// nil to preserve this invariant for downstream consumers that dereference
+// without a nil check.
 type StaticConfigState struct {
 	ref atomic.Pointer[StaticConfig]
 }
 
 // NewStaticConfigState creates a new StaticConfigState initialized with the given config.
+// cfg must be non-nil; passing nil violates the non-nil invariant of Load.
 func NewStaticConfigState(cfg *StaticConfig) *StaticConfigState {
 	s := &StaticConfigState{}
 	s.ref.Store(cfg)
@@ -17,11 +23,17 @@ func NewStaticConfigState(cfg *StaticConfig) *StaticConfigState {
 }
 
 // Load returns the current StaticConfig. Safe for concurrent use.
+// Guaranteed non-nil when the state was constructed via NewStaticConfigState
+// with a non-nil config; Store(nil) is a no-op.
 func (s *StaticConfigState) Load() *StaticConfig {
 	return s.ref.Load()
 }
 
 // Store atomically replaces the current StaticConfig.
+// nil is silently ignored to preserve the non-nil invariant of Load.
 func (s *StaticConfigState) Store(cfg *StaticConfig) {
+	if cfg == nil {
+		return
+	}
 	s.ref.Store(cfg)
 }
--- a/pkg/config/state_test.go
+++ b/pkg/config/state_test.go
@@ -1,0 +1,57 @@
+package config
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+)
+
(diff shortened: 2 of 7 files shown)


matzew (approved):
LGTM

PR #1109, opened 2026-04-27
fix(auth): propagate certificate_authority to token exchange HTTP client

### LINKED ISSUE #1108: Strategy-based token exchange ignores certificate_authority breaking private CA deployments
When token_exchange_strategy is configured (rfc8693, keycloak-v1, or entra-obo), the token exchange HTTP client does not use the certificate_authority config option. This causes TLS verification failures when the OIDC/STS token endpoint uses a certificate signed by a private CA.

The OIDC provider discovery path correctly applies the CA (via pkg/oauth/state.go), so JWT validation works, but the subsequent token exchange request fails because its HTTP client only trusts the system CA pool.

Any on-prem or enterprise deployment that combines:
- require_oauth = true
- token_exchange_strategy (any strategy)
- certificate_authority pointing to a private CA file will fail with a TLS certificate verification error on the token exchange request, even though the CA is explicitly configured.

This does not affect:

 - Deployments using publicly-trusted CAs (cloud-managed IdPs like Okta, Auth0, public Entra ID endpoints)
 - OIDC discovery and JWT validation (those paths correctly use the CA)
 - Local development with Kind/minikube (typically use system-trusted certs)

nader-ziada:
Fixes: #1108 

The token exchange config (used by rfc8693, keycloak-v1, entra-obo) was not receiving the certificate_authority setting, causing TLS failures when the OIDC token endpoint uses a private CA. The CAFile field on TargetTokenExchangeConfig was already handled correctly by HTTPClient() but was never populated from config.

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -72,6 +72,7 @@ type StsConfigProvider interface {
 	GetStsAuthStyle() string
 	GetStsClientCertFile() string
 	GetStsClientKeyFile() string
+	GetCertificateAuthority() string
 }
 
 // ValidationEnabledProvider provides access to validation enabled setting.
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -424,6 +424,10 @@ func (c *StaticConfig) GetStsClientKeyFile() string {
 	return c.StsClientKeyFile
 }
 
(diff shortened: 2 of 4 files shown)


Cali0707 (approved):

manusa (approved):
LGTM, thx!

PR #1114, opened 2026-04-28
fix(mcp): harden resources surface for hot reload

### LINKED ISSUE #1113: Follow-ups from #1091: harden MCP resources surface (panic-on-reload, symmetry, validation)
Tracking follow-up work from PR #1091 (MCP resources boilerplate). The PR shipped intentionally tight to unblock the must-gather consumer; this issue captures items deferred for a separate PR.

## 1. Validate `URITemplate` upfront — don't panic on hot reload (Important)

`pkg/mcp/resource_test.go:440-474` (`TestInvalidURIPanics`) currently certifies that an invalid `URITemplate` causes `ReloadConfiguration` to panic. But `reloadToolsets` is called from `WatchTargets` callbacks and `ReloadConfiguration`, and neither caller has a `recover()`. A buggy toolset URITemplate will crash the running server on hot reload.

**Action:**
- Validate `URITemplate` in `registerResourceTemplate` (`pkg/mcp/mcp.go`) and return a wrapped error, mirroring the shape of `registerTool`/`registerPrompt`.
- Replace `TestInvalidURIPanics` with a test asserting that `ReloadConfiguration` returns an error (and the server stays up) when a toolset declares a malformed URITemplate.

## 2. Make tool/prompt/resource handling in `pkg/mcp` symmetric (Important)

The three item kinds (tools, prompts, resources/resource-templates) currently follow slightly different patterns inside `pkg/mcp`. They should share a single shape: file naming, error wrapping, registration helpers.

**Action:**
- Rename / consolidate the gosdk-conversion code so we end up with:
  - `pkg/mcp/tools_gosdk.go`
  - `pkg/mcp/prompts_gosdk.go`
  - `pkg/mcp/resources_gosdk.go` (TBD on plural — pick one and apply consistently)
- Align error handling in `register*` helpers: `registerResource` / `registerResourceTemplate` currently always return `nil` because `addResource`/`addResourceTemplate` are void. Once #1 lands they'll surface real errors — pattern should match `registerTool` / `registerPrompt` (`fmt.Errorf(\"failed to convert %s: %w\", name, err)`).
- Fold any small structural drift (filter/mutator wiring, naming of collectors, lock ordering in `reloadToolsets`) so that adding a fifth item kind in the future is mechanical.

## 3. Enforce `ResourceContent` invariant (Minor)

`pkg/api/toolsets.go:121` documents \"Exactly one of Text or Blob must be set; both cannot be nil or both non-nil,\" but `addResource` / `addResourceTemplate` (`pkg/mcp/resource.go`) don't enforce it. A handler returning both empty or both populated silently passes through.

**Action:**
- After the `content == nil` check, validate `(content.Text != \"\") != (len(content.Blob) > 0)` and return an explicit error otherwise.
- Add unit tests covering both violation modes (both-empty, both-set).

## 4. Symmetric blob/text guards in tests (Minor)

`pkg/mcp/resource_test.go:381` asserts `s.Empty(Contents[0].Text)` for the Blob test, but the text-only happy path in `TestResources` (line 91-103) doesn't add the inverse `s.Empty(Contents[0].Blob)`.

**Action:**
- Add `s.Empty(Contents[0].Blob)` in `TestResources` and `TestResourceTemplates` so both halves of the Text-xor-Blob contract are guarded by every happy-path test.

## 5. Fragile assertion ordering in happy-path tests (Minor)

`pkg/mcp/resource_test.go:72, 92, 98, 132, 143, 150` (and `377, 426, 434`). The newer tests use `s.Require().NoError(err)` before dereferencing, but the original `TestResources` / `TestResourceTemplates` still pair `s.NoError(err)` with `s.Require().Len(result.Contents, 1)`. Cosmetic until something starts returning errors at those call sites, but a panic during a refactor would mask the real failure.

**Action:**
- Sweep `pkg/mcp/resource_test.go` and replace `s.NoError(err)` with `s.Require().NoError(err)` immediately before any dereference of the returned result.

---

xref: #1091

manusa:
## Summary

Follow-ups from #1091, tracked by #1113. Hardens the MCP resources surface against the panic-on-hot-reload regression and aligns it structurally with the tool/prompt code paths.

## Changes

- **Up-front URI / URITemplate validation.** `ServerResourceToGoSdkResource` and `ServerResourceTemplateToGoSdkResourceTemplate` validate via `url.Parse` and `uritemplate.New` before calling the SDK. Previously, a malformed value would panic inside `s.server.AddResource(Template)` and crash the running server during `ReloadConfiguration` / `WatchTargets` — neither caller had a `recover()`.
- **Two-phase atomic reload.** `reloadToolsets` now converts all four item kinds (tools, prompts, resources, resource templates) before any SDK mutation. A failure in any category aborts the reload before the server state is touched, so partial-failure can no longer leave the SDK and the `enabled*` bookkeeping divergent. New `convertItems` / `commitItems` generics replace the per-kind `register*` helpers and `reloadItems`.
- **`ResourceContent` invariant enforced.** The Text-xor-Blob contract documented at `pkg/api/toolsets.go:121` is now enforced in the read handler (returns an error when both are empty or both are set).
- **File-naming symmetry.** `gosdk.go` → `tools_gosdk.go`; `resource.go` → `resources_gosdk.go`; `resource_test.go` → `resources_gosdk_test.go`; `resource_filter_test.go` → `resources_filter_test.go`. Conversion helpers now match the existing `prompts_gosdk.go` shape.
- **Test improvements.**
  - `TestInvalidURIPanics` replaced with `TestInvalidURITemplateReturnsError`, which asserts no panic, a wrapped error returned, the previously registered resource is **still served after the failed reload** (proving cross-category atomicity), and a subsequent valid reload succeeds.
  - New `TestResourceContentInvariant` covers both-empty and both-set, for static resources and templates.
  - Sweep `s.NoError(err)` → `s.Require().NoError(err)` before result dereferences.
  - Added inverse `s.Empty(Contents[0].Blob)` / `s.Empty(Contents[0].Text)` guards to text-only and blob-only happy paths.

Closes #1113


nader-ziada (approved):
LGTM

pkg/mcp/mcp.go
```diff
@@ -152,62 +152,64 @@ func (s *Server) reloadToolsets() error {
...
-	previousPrompts := s.enabledPrompts
-	previousResources := s.enabledResources
-	previousResourceTemplates := s.enabledResourceTemplates
-	s.mu.RUnlock()
-
-	// Reload tools (calls s.server.AddTool/RemoveTools - external code, no lock held)
-	newTools, err := reloadItems(
-		previousTools,
-		applicableTools,
+	// Phase 1: convert all items to SDK types. This validates URIs, URITemplates,
```
Cali0707, line 167: I think this is great, but should we worry about a case where:

1. `ReloadConfiguration` updates `s.configuration.StaticConfig`
2. One or more of these errors (e.g. invalid URI template)
3. Now the config says we have a toolset enabled, but it may not be served by the SDK

As far as I can tell, this wouldn't currently cause other issues with our code, but it may be worth thinking through this state mismatch (perhaps as a follow up)

pkg/mcp/mcp.go
```diff
@@ -246,14 +272,11 @@ func reloadItems[T any](
 	}
 	remove(toRemove...)
 
-	// Register all items
-	for _, item := range items {
-		if err := register(item); err != nil {
-			return nil, err
-		}
+	for _, c := range items {
+		add(c.item, c.handler)
```
Cali0707, line 288: depending on how many resources we expect to expose, we may in the future want to only add new ones (rather than re-adding all the ones already there that are unchanged)

I think for now this is fine

pkg/mcp/resources_gosdk.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	return mcpTemplate, handler, nil
+}
+
+// validateResourceContent enforces the api.ResourceContent invariant:
+// exactly one of Text or Blob must be set.
+func validateResourceContent(content *api.ResourceContent) error {
+	hasText := content.Text != ""
+	hasBlob := len(content.Blob) > 0
+	if hasText == hasBlob {
```
Cali0707: this logic was a little hard for me to follow intially, and I had to look a second time. maybe we can split it into two separate checks (`hasText && hasBlob` and `!hasText && !hasBlob`) and have more specific error messages for the user?

Cali0707 (approved):
Overall LGTM, just a few nits/thoughts for future things we may care about.

Feel free to merge without addressing these @manusa

pkg/mcp/mcp.go
manusa: Created #1128 to tackle this explicitly

manusa:
Comments addressed, merging

PR #1115, opened 2026-04-28
Adjusted Kiali tests to endpoints.

hhovsepy:
Changed Kiali MCP Contract Tests to work with this PR https://github.com/kiali/kiali/pull/9607

### DIFF
--- a/pkg/kiali/tests/README.md
+++ b/pkg/kiali/tests/README.md
@@ -16,7 +16,7 @@ Contract tests ensure that:
 
 ## Prerequisites
 
-- A running Kiali instance (with bookinfo demo app deployed)
+- A running Kiali instance
 - Go 1.24 or later
 
 ## Configuration
@@ -31,6 +31,9 @@ Tests are configured via environment variables:
 
 - `KIALI_TOKEN` - Bearer token for authentication (if required)
 - `TEST_NAMESPACE` - Namespace to use for tests (default: `bookinfo`)
+- `TEST_SERVICE` - Service name used by metrics and tracing tests (default: `productpage`)
+- `TEST_WORKLOAD` - Workload name used by logs and pod performance tests (default: `productpage-v1`)
+- `TEST_TRACE_ID` - Optional known trace ID for `get_trace_details`; useful when the test environment does not guarantee fresh traces for `TEST_SERVICE`
 
 ## Running Tests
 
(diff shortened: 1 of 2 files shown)


josunect:
LGTM! 
Run locally and passed

manusa (approved):
Thanks for taking this on. The old test wouldn't compile against the new `pkg/toolsets/kiali/tools` package anyway, and endpoint coverage looks complete.

A few things worth tightening:

**1. `mcpCall` doesn't match what the real client sends.**
`Kiali.ExecuteRequest` in `pkg/kiali/kiali.go:170,186` injects `arguments["mcp_mode"] = "true"` and sets `X-Kubernetes-MCP-Server: true` on every request. `mcpCall` (`api_backend_test.go:62-93`) does neither. If Kiali ever branches on those, the contract test passes where the real server would fail. Easiest fix is to mirror them.

**2. `TestManageIstioConfigCRUD` leaks the ServiceEntry on partial failure.**
Once `createdName` is set (L438-603), any failing `s.Require()` before the final confirmed-delete leaves an orphan in the namespace. No `TearDownTest`. An `s.T().Cleanup(...)` right after the successful create, doing a best-effort `confirmed:true` delete, would make reruns and timeouts safe.

**3. Read-after-write has no retry.**
`requireServiceEntryHost` is called immediately after create and patch (L483-489, L534-540). The old code had `waitForResource` (30×1s) for this exact reason: Kiali's istio-config cache doesn't update instantly. A small poll loop inside `requireServiceEntryHost` would restore that safety net.

**4. A handful of endpoints are barely asserted now.**
The point of a contract test is to catch upstream shape drift, so:
- `TestGetLogs` (L400-411) and `TestGetPodPerformance` (L413-424) only check "non-empty JSON string". A handler returning `"x"` passes.
- `TestManageIstioConfigRead` "lists istio config" (L427-435) is `requireValidJSON` only.
- `TestListOrGetResources` "lists workloads" (L372-382) only checks "non-empty object", while the sibling "lists services" check (L335-369) is solid.

One or two stable keys per endpoint, like the services subtest does, would meaningfully raise the bar.

Approving so this isn't blocked on me, tagging @josunect for the final call.

manusa:
@josunect @aljesusg from my side this looks good, waiting on your OK to merge it.

josunect:
LGTM!

PR #1117, opened 2026-04-28
fix(http): fall back to openid-configuration when well-known endpoint returns empty body

### LINKED ISSUE #1116: Well-known endpoint fallback doesn't trigger for Entra ID (returns 500 instead of falling back)
When the MCP server proxies /.well-known/oauth-protected-resource and /.well-known/oauth-authorization-server to Microsoft Entra ID, Entra returns HTTP 200 with content-length: 0 (empty body) instead of HTTP 404.

The fallback logic in wellknown.go only triggers on a 404 status code. Because Entra returns 200, the server attempts to JSON-decode the empty body, which fails with io.EOF. This results in an HTTP 500 returned to the MCP client, and the fallback to /.well-known/openid-configuration never happens.

Entra ID's behavior for unsupported well-known paths:

$ curl -sv "https://login.microsoftonline.com/<tenant>/v2.0/.well-known/oauth-protected-resource"
< HTTP/2 200
< content-length: 0
It returns 200 with zero bytes — not 404. The JSON decoder receives no input and returns io.EOF.

The fallback architecture added in #953 assumed Entra ID would return 404 for unsupported endpoints. The tests mock this with http.NotFound(), which doesn't match real Entra ID behavior.

nader-ziada:
Fixes: #1116 

Some OIDC providers (e.g., Entra ID) return HTTP 200 with an empty body for unsupported well-known paths instead of 404. The JSON decoder fails with io.EOF, causing a 500 instead of triggering the openid-configuration fallback. Treat io.EOF as equivalent to 404 to trigger fallback.

### DIFF
--- a/pkg/http/wellknown.go
+++ b/pkg/http/wellknown.go
@@ -210,7 +210,7 @@ func (w *WellKnown) ServeHTTP(writer http.ResponseWriter, request *http.Request)
 }
 
 // fetchWellKnownEndpoint creates a new request from the incoming request's method and context,
-// then fetches the well-known endpoint. Returns nil metadata if the endpoint returns 404.
+// then fetches the well-known endpoint. Returns nil metadata if the endpoint returns 404 or an empty body (to allow fallback).
 func (w *WellKnown) fetchWellKnownEndpoint(request *http.Request, url string) (map[string]interface{}, http.Header, error) {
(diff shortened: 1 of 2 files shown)


matzew:
LGTM 
leaving final review to @manusa

manusa (approved):
LGTM, thx!

PR #1121, opened 2026-04-30
feat(auth): add accept_opaque_tokens for non-JWT bearer token passthrough

RoiGlinik:
## Summary
- Add `accept_opaque_tokens` config option to accept non-JWT bearer tokens (e.g., OpenShift OAuth `sha256~...` tokens) when `require_oauth` is enabled
- Opaque tokens are passed through to the Kubernetes API server for validation instead of being rejected by the MCP server's JWT parsing
- Enables per-user RBAC with OpenShift OAuth or similar non-JWT token providers using `cluster_auth_mode=passthrough`

## Changes
- **`pkg/config/config.go`**: New `AcceptOpaqueTokens` field, validation in `validateSkipJWTVerification`
- **`pkg/http/authorization.go`**: Opaque token detection (`isJWT`), passthrough path in middleware, graceful 401 (not 500) when JWTs hit an opaque-only config
- **`pkg/kubernetes-mcp-server/cmd/root.go`**: `--accept-opaque-tokens` CLI flag (hidden, like other auth flags)
- **`docs/configuration.md`**: Documentation for the new option
- **Tests**: 4 integration tests (opaque accepted, opaque rejected, JWT still works with flag, JWT gets 401 not 500 in opaque-only config) + `isJWT` unit tests + config validation tests

## Test plan
- [x] `make build` passes
- [x] `go test ./pkg/http/ ./pkg/config/` -- all tests pass
- [x] New `TestAuthorizationOpaqueTokenAccepted` -- opaque token passthrough works
- [x] New `TestAuthorizationOpaqueTokenRejectedWithoutFlag` -- opaque tokens rejected without flag
- [x] New `TestAuthorizationJWTStillWorksWithOpaqueTokenFlag` -- JWT validation unaffected
- [x] New `TestAuthorizationJWTRejectedWithOpaqueOnlyConfig` -- JWTs return 401 (not 500) when only opaque tokens are configured

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -482,6 +482,7 @@ Configure OAuth/OIDC authentication for HTTP mode deployments.
 | `oauth_audience` | string | `""` | Valid audience for OAuth tokens (for offline JWT claim validation). |
 | `authorization_url` | string | `""` | URL of the OIDC authorization server for token validation and STS exchange. |
 | `skip_jwt_verification` | boolean | `false` | When `true`, allows JWTs without cryptographic signature verification when `require_oauth` is enabled but no `authorization_url` is configured. Only use behind a trusted reverse proxy that already verifies tokens. When `false` (default), the server refuses to start in this configuration. |
+| `accept_opaque_tokens` | boolean | `false` | When `true`, accepts non-JWT bearer tokens (e.g., OpenShift OAuth `sha256~...` tokens) and passes them through to the Kubernetes API server for validation. Use with `cluster_auth_mode = "passthrough"` for per-user RBAC with OpenShift OAuth or similar non-JWT token providers. |
 | `disable_dynamic_client_registration` | boolean | `false` | When `true`, disables dynamic client registration in `.well-known` endpoints. |
 | `oauth_scopes` | string[] | `[]` | Supported client scopes for the OAuth flow. |
 | `sts_client_id` | string | `""` | OAuth client ID for backend token exchange. |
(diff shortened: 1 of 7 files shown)


manusa:
Thank you @RoiGlinik for digging into this and putting together such a thorough PR — the test coverage and the 401-vs-500 cleanup in particular were really valuable in surfacing what was actually going on.

After reviewing this against the broader auth flow, I'd like to take a step back. The investigation prompted by this PR (see #1131) concluded that the underlying issue is broader than non-JWT token shape: `skip_jwt_verification` today only skips signature verification, not parsing or offline claim validation, so even with that flag set, opaque tokens, cluster-issued JWTs with mismatched audiences, and other legitimate passthrough tokens get rejected by the MCP server before reaching the cluster.

The `accept_opaque_tokens` flag fixes one symptom (non-JWT shape), but the right fix is to make `require_oauth=true` + `skip_jwt_verification=true` + `cluster_auth_mode=passthrough` mean what it reads like: presence-only mode, with the cluster as the sole authority. That collapses several flags and edge cases into a single coherent "pure passthrough" configuration and removes the need for a separate `accept_opaque_tokens` toggle.

I've opened #1131 with the full proposal — concrete code changes, doc updates, and a test plan that incorporates the scenarios this PR already tests. For that reason I'm going to close this PR in favor of #1131, but please don't take it as the work being unwelcome — it's the opposite, this PR is what made the broader problem visible. Would love your input on #1131 if you have time, and contributions toward the implementation there are very welcome.

PR #1130, opened 2026-05-04
refactor(mcp): make ReloadConfiguration fully transactional (#1128)

### LINKED ISSUE #1128: Make ReloadConfiguration fully transactional (avoid snapshot+rollback)
Tracking follow-up from PR #1114 review.

`ReloadConfiguration` (`pkg/mcp/mcp.go`) currently mutates `s.configuration.StaticConfig` *before* calling `reloadToolsets`. If `reloadToolsets` returns an error (invalid URITemplate, bad tool schema, etc.), the SDK state is intentionally unchanged (two-phase reload aborts in the convert phase) but `s.configuration` would otherwise reflect the rejected new config — leaving the server internally inconsistent.

PR #1114 ships a snapshot-and-rollback workaround: the old `StaticConfig`, `listOutput`, and `toolsets` are saved before mutation and restored on failure. This keeps `s.configuration` consistent with what the SDK is actually serving but creates a brief window where concurrent readers could observe the new-but-rejected config.

## Proposed cleanup

Refactor so the new config is computed against without mutating `s.configuration` until after the convert phase succeeds:

- Pass `*Configuration` (or just `*config.StaticConfig`) into `collectApplicableTools/Prompts/Resources/ResourceTemplates` so they don't hard-depend on `s.configuration`.
- `ReloadConfiguration` builds a temporary view from `newConfig`, runs the convert phase against it, and only then atomically swaps `s.configuration` + commits SDK changes.
- Remove the rollback path in `ReloadConfiguration`.

Outcome: a fully transactional reload, with no observable inconsistency window even under concurrent reads.

xref: #1113, #1114

manusa:
## Summary

Closes #1128.

`ReloadConfiguration` previously mutated `s.configuration.StaticConfig` *before* calling `reloadToolsets`, then snapshot-and-rolled-back on failure. The window between mutation and rollback briefly exposed a new-but-rejected configuration to concurrent readers (rate-limit closure, confirmation rules, list output, ...). This PR makes the reload fully transactional, then takes the boy-scout opportunity to fix two pre-existing data races in the surrounding `s.configuration` access pattern that surfaced during review.

Three commits, each with its own race-detector regression test that fails before the fix and passes after.

### 1. `refactor(mcp): make ReloadConfiguration fully transactional` — the issue itself

- `collectApplicableTools/Prompts/Resources/ResourceTemplates` now take `*Configuration` explicitly instead of reading `s.configuration`.
- `reloadToolsets(cfg *Configuration)` is the single reload entry point. It installs `cfg` atomically only after the convert phase succeeds. Passing `nil` triggers the refresh path: re-read `s.configuration` *inside* the `reloadMu` critical section. This avoids a TOCTOU where the `WatchTargets` callback could otherwise capture cfg before `reloadMu`, then commit a stale snapshot after a concurrent `ReloadConfiguration` had already published a new one.
- `ReloadConfiguration` builds a candidate `*Configuration` and hands it to `reloadToolsets`. The snapshot-and-rollback path is gone.
- Regression test: `TestReloadFailureLeavesConfigurationIntact` injects a tool with a non-`object` input schema (rejected by `ServerToolToGoSdkTool` in the convert phase), calls `reloadToolsets` directly with the candidate, and asserts `s.configuration` and all `enabledX` lists are unchanged. A second subtest verifies a follow-up `refreshToolsets()` succeeds — confirms locks weren't left in a bad state.

### 2. `fix(mcp): make Server.configuration accesses race-free with atomic.Pointer`

`Server.configuration` is now `atomic.Pointer[Configuration]`. Lock-free `Load()` on hot read paths (handlers, rate-limit closure, `ServeHTTP`); atomic `Store()` on the swap. The tool handler now snapshots once via `cfg := s.configuration.Load()` at the top and uses that for both `BaseConfig` and `ListOutput()` — fixes a torn-deref where a mid-handler reload could split fields across two configurations. Regression test: `TestConcurrentReadsDuringReload` runs 8 reader goroutines mirroring the handler access pattern alongside a tight `ReloadConfiguration` loop; clean under `-race`, would have flagged the bare-pointer field.

### 3. `fix(mcp): pre-warm Configuration's lazy caches before publish`

`Configuration.ListOutput()` and `Configuration.Toolsets()` lazy-init unexported cache fields on first read. After publish the new config is visible to lock-free readers, and the first concurrent first-readers would race writing the same cache field. Pre-warming both caches in `reloadToolsets` immediately before the atomic store eliminates the windows. Regression test: `TestConcurrentListOutputAfterReload` — verified that *without* the warm-up `-race` reports the data race on `Configuration.listOutput`, *with* the warm-up it stays clean.

## Test plan

- [x] \`go test ./pkg/mcp/ -race -timeout 300s\` → ok
- [x] \`go test ./...\` → all packages pass
- [x] \`make lint\` → 0 issues
- [x] New regression tests for each commit pass with the fix in place and fail when reverted (verified via \`git stash\` for #3)
- [x] CI green

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -139,16 +139,41 @@ func NewServer(configuration Configuration, targetProvider internalk8s.Provider)
 	s.server.AddReceivingMiddleware(toolCallLoggingMiddleware)
 	s.server.AddReceivingMiddleware(s.metricsMiddleware())
 
-	err = s.reloadToolsets()
+	err = s.reloadToolsets(s.configuration)
 	if err != nil {
 		return nil, err
 	}
-	s.p.WatchTargets(s.reloadToolsets)
+	s.p.WatchTargets(s.refreshToolsets)
 
 	return s, nil
 }
 
-func (s *Server) reloadToolsets() error {
+// refreshToolsets re-runs reloadToolsets against whatever configuration is
+// currently installed. Used by the provider's WatchTargets callback (signature
+// func() error) when cluster state changes but the configuration itself has
+// not. Passing nil cfg makes reloadToolsets re-read s.configuration *inside*
+// the reloadMu critical section, which avoids a TOCTOU race where capturing
+// the cfg here would let a concurrent ReloadConfiguration commit a new cfg
+// while we were blocked on reloadMu — we'd then silently roll it back by
+// re-installing the stale snapshot.
+func (s *Server) refreshToolsets() error {
+	return s.reloadToolsets(nil)
+}
+
+// reloadToolsets recomputes the SDK's tool/prompt/resource/template surface
+// against cfg and, on success, atomically installs cfg as the live
+// configuration alongside the SDK changes. cfg may be:
+//   - a freshly-built *Configuration (configuration reload path), in which
+//     case s.configuration is only mutated once the convert phase succeeds;
+//   - nil (refresh path, e.g. cluster-state change), in which case the
+//     currently-installed configuration is re-read under reloadMu and
+//     reused. nil — instead of "callers pass s.configuration" — is required
+//     to avoid the caller capturing a stale cfg before reloadMu serializes
+//     against an in-flight ReloadConfiguration.
+//
+// On error s.configuration, the SDK, and the enabled-X bookkeeping all stay
+// at their prior consistent values.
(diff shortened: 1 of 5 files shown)


matzew (approved):
LGTM

PR #1136, opened 2026-05-06
fix(test): inherit ReadOnly from suite config in reload test

matzew:
The TestReloadFailureLeavesConfigurationIntact test built a candidate config from config.Default() without inheriting ReadOnly from the suite config. When downstream overrides set ReadOnly: true, the brokenToolset tool (which has no ReadOnlyHint) gets filtered out before reaching the convert phase, causing the test to pass unexpectedly.

Inherit ReadOnly from s.Cfg to match the pattern used by other tests.


manusa (approved):
LGTM, thx!

PR #1137, opened 2026-05-06
fix(test): use BaseDefault() in ToolsetsSuite to avoid downstream config leaking

matzew:
`ToolsetsSuite.SetupTest()` used `config.Default()` which inherits downstream overrides (e.g.`ReadOnly: true`). This caused tools without `ReadOnlyHint` to be filtered out, breaking snapshot comparisons. Switch to `BaseDefault()` consistent with `BaseMcpSuite`.


Cali0707 (approved):
LGTM

manusa (approved):
LGTM, thx!

PR #1138, opened 2026-05-06
feat(core): Add fieldSelector to events_list and namespaces_list tools

difrost:
This PR adds `fieldSelector` property to `events_list` and `namespaces_list` in a similar fashion how it's defined for e.g. `pods_list`.

The biggest added values is for the `events_list` where a tool call on a busy namespace with a lot of events generated is killing most of the agents. 

I've also added evals for both tool extensions but can't figure a better way than just verifying with llmJudge `contains`.

@Cali0707 / @manusa please review.

### DIFF
--- a/evals/tasks/core/filter-events-by-type/cleanup.sh
+++ b/evals/tasks/core/filter-events-by-type/cleanup.sh
@@ -1,0 +1,2 @@
+#!/usr/bin/env bash
+kubectl delete namespace event-filter-test --ignore-not-found
--- a/evals/tasks/core/filter-events-by-type/cleanup.sh
+++ b/evals/tasks/core/filter-events-by-type/cleanup.sh
@@ -1,2 +1,2 @@
 #!/usr/bin/env bash
-kubectl delete namespace event-filter-test --ignore-not-found
+kubectl delete namespace events-filter-test --ignore-not-found
--- a/evals/tasks/core/filter-events-by-type/filter-events-by-type.yaml
+++ b/evals/tasks/core/filter-events-by-type/filter-events-by-type.yaml
@@ -1,0 +1,15 @@
+kind: Task
+metadata:
+  labels:
+    suite: core
+  name: filter-events-by-type
+  difficulty: easy
+steps:
+  setup:
+    file: setup.sh
+  verify:
+    contains: "bad-pod"
+  cleanup:
+    file: cleanup.sh
+  prompt:
+    inline: "List only the Warning events for the pod named 'bad-pod' in the 'events-filter-test' namespace"
--- a/evals/tasks/core/filter-events-by-type/filter-events-by-type.yaml
+++ b/evals/tasks/core/filter-events-by-type/filter-events-by-type.yaml
@@ -1,15 +1,28 @@
 kind: Task
+apiVersion: mcpchecker/v1alpha2
 metadata:
   labels:
     suite: core
   name: filter-events-by-type
   difficulty: easy
-steps:
+spec:
   setup:
-    file: setup.sh
+  - script:
+      file: setup.sh
   verify:
-    contains: "bad-pod"
+  - script:
+      env:
+        AGENTS_OUTPUT: "{agent.output}"
+      inline: |-
+        bad=$(echo "$AGENTS_OUTPUT" | grep 'Warning events' | grep 'bad-pod' | grep '.events-filter-test. namespace')
+        good=$(echo "$AGENTS_OUTPUT" | grep "good-pod")
+        if [ -n "$bad" ] && [ -z "$good" ]; then
+          exit 0
+        else
+          exit 1
+        fi
   cleanup:
-    file: cleanup.sh
+  - script:
+      file: cleanup.sh
   prompt:
     inline: "List only the Warning events for the pod named 'bad-pod' in the 'events-filter-test' namespace"
--- a/evals/tasks/core/filter-namespace-by-name/filter-namespace-by-name.yaml
+++ b/evals/tasks/core/filter-namespace-by-name/filter-namespace-by-name.yaml
@@ -1,0 +1,15 @@
+kind: Task
+metadata:
+  labels:
+    suite: core
+  name: filter-namespace-by-name
+  difficulty: easy
+steps:
+  setup:
+    file: setup.sh
+  verify:
+    contains: "ns-beta"
+  cleanup:
+    file: cleanup.sh
+  prompt:
+    inline: "Using a field selector, list only the namespaces 'ns-beta'"
--- a/evals/tasks/core/filter-namespace-by-name/filter-namespace-by-name.yaml
+++ b/evals/tasks/core/filter-namespace-by-name/filter-namespace-by-name.yaml
@@ -12,4 +12,4 @@ steps:
   cleanup:
     file: cleanup.sh
   prompt:
-    inline: "Using a field selector, list only the namespaces 'ns-beta'"
+    inline: "Using a field selector, list only the namespace 'ns-beta' from the cluster"
--- a/evals/tasks/core/filter-namespace-by-name/filter-namespace-by-name.yaml
+++ b/evals/tasks/core/filter-namespace-by-name/filter-namespace-by-name.yaml
@@ -1,15 +1,22 @@
 kind: Task
+apiVersion: mcpchecker/v1alpha2
 metadata:
   labels:
     suite: core
   name: filter-namespace-by-name
   difficulty: easy
-steps:
+spec:
   setup:
-    file: setup.sh
+  - script:
+      file: setup.sh
   verify:
-    contains: "ns-beta"
+  - script:
+      env:
+        AGENTS_OUTPUT: "{agent.output}"
+      inline: |-
+        [ X"$(echo "$AGENTS_OUTPUT" | grep -v 'ns-beta')" = X"" ] && exit 0 || exit 1
   cleanup:
-    file: cleanup.sh
+  - script:
+      file: cleanup.sh
   prompt:
     inline: "Using a field selector, list only the namespace 'ns-beta' from the cluster"
--- a/pkg/toolsets/core/events.go
+++ b/pkg/toolsets/core/events.go
@@ -23,6 +23,11 @@ func initEvents() []api.ServerTool {
 						Type:        "string",
 						Description: "Optional Namespace to retrieve the events from. If not provided, will list events from all namespaces",
 					},
+					"fieldSelector": {
+						Type:        "string",
+						Description: "Optional Kubernetes field selector to filter events by field values (e.g. 'type=Warning', 'involvedObject.name=my-pod'). Supported fields: involvedObject.kind, involvedObject.name, involvedObject.namespace, involvedObject.uid, involvedObject.apiVersion, involvedObject.resourceVersion, involvedObject.fieldPath, reason, reportingComponent, source, type. See https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/",
+						Pattern:     REGEX_FIELDSELECTOR,
+					},
 				},
 			},
 			Annotations: api.ToolAnnotations{
@@ -38,10 +43,12 @@ func initEvents() []api.ServerTool {
 func eventsList(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
 	p := api.WrapParams(params)
 	namespace := p.OptionalString("namespace", "")
+	options := api.ListOptions{}
+	options.FieldSelector = p.OptionalString("fieldSelector", "")
 	if err := p.Err(); err != nil {
 		return api.NewToolCallResult("", fmt.Errorf("failed to list events in all namespaces: %w", err)), nil
 	}
-	eventMap, err := kubernetes.NewCore(params).EventsList(params, namespace)
+	eventMap, err := kubernetes.NewCore(params).EventsList(params, namespace, options)
 	if err != nil {
 		return api.NewToolCallResult("", fmt.Errorf("failed to list events in all namespaces: %w", err)), nil
 	}
--- a/pkg/toolsets/core/namespaces.go
+++ b/pkg/toolsets/core/namespaces.go
@@ -19,6 +19,13 @@ func initNamespaces(o api.Openshift) []api.ServerTool {
 			Description: "List all the Kubernetes namespaces in the current cluster",
 			InputSchema: &jsonschema.Schema{
 				Type: "object",
+				Properties: map[string]*jsonschema.Schema{
+					"fieldSelector": {
+						Type:        "string",
+						Description: "Optional Kubernetes field selector to filter namespaces by field values (e.g. 'metadata.name=default', 'status.phase=Active'). Supported fields: metadata.name, status.phase. See https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/",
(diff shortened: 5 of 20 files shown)


matzew:
Consider adding a unit test for `fieldSelector` in `pkg/mcp/events_test.go`, similar to the existing `events_list(namespace=ns-1)` test ?

difrost:
@matzew - [0812a5e](https://github.com/containers/kubernetes-mcp-server/pull/1138/commits/0812a5e2a18f48ee94b69dfc44dd79cf2621615d) adds tests for `events_list` and `namespaces_list`.

evals/tasks/core/filter-events-by-type/cleanup.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+kubectl delete namespace event-filter-test --ignore-not-found
```
matzew: should it be `events-filter-test` ?

difrost: Thanks @matzew - fixed in [87d187a](https://github.com/containers/kubernetes-mcp-server/pull/1138/commits/87d187ae48cccafac6a9d9216d593d1f9fea752b)

pkg/toolsets/core/events.go
```diff
@@ -23,6 +23,11 @@ func initEvents() []api.ServerTool {
 						Type:        "string",
 						Description: "Optional Namespace to retrieve the events from. If not provided, will list events from all namespaces",
 					},
+					"fieldSelector": {
+						Type:        "string",
+						Description: "Optional Kubernetes field selector to filter events by field values (e.g. 'type=Warning', 'involvedObject.name=my-pod'). Supported fields: involvedObject.kind, involvedObject.name, involvedObject.namespace, involvedObject.uid, involvedObject.apiVersion, involvedObject.resourceVersion, involvedObject.fieldPath, reason, reportingComponent, source, type. See https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/",
+						Pattern:     REGEX_FIELDSELECTOR,
```
manusa, line 29: `REGEX_FIELDSELECTOR` (defined at `pkg/toolsets/core/toolset.go:14`) rejects `involvedObject.apiVersion=apps/v1` because `/` isn't in the character class, yet the description right above advertises `involvedObject.apiVersion` as supported. Verified: `"involvedObject.apiVersion=apps/v1"` does not match the pattern.

Adding `/` to the character class is safe for the other callers (`pods_list`, `namespaces_list`, `resources_list`) since their advertised field values are DNS-1123 and never contain `/`.

While you're touching this, please also add a positive case for a comma-separated multi-clause selector to `pkg/toolsets/core/regex_test.go`.

pkg/toolsets/core/namespaces.go
```diff
@@ -19,6 +19,13 @@ func initNamespaces(o api.Openshift) []api.ServerTool {
 			Description: "List all the Kubernetes namespaces in the current cluster",
 			InputSchema: &jsonschema.Schema{
 				Type: "object",
+				Properties: map[string]*jsonschema.Schema{
+					"fieldSelector": {
+						Type:        "string",
+						Description: "Optional Kubernetes field selector to filter namespaces by field values (e.g. 'metadata.name=default', 'status.phase=Active'). Supported fields: metadata.name, status.phase. See https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/",
+						Pattern:     REGEX_FIELDSELECTOR,
```
manusa, line 26: OpenShift `projects_list` (just below, around line 41) supports the same field selectors as Namespaces (`metadata.name`, `status.phase`). Could you mirror this `fieldSelector` block onto `projects_list` so the OpenShift path doesn't lose the new filtering capability?

pkg/toolsets/kubevirt/vm_troubleshoot.go
```diff
@@ -358,7 +358,7 @@ func fetchVirtLauncherPodLogs(ctx context.Context, client api.KubernetesClient,
 // fetchEvents fetches events related to the VM and returns them formatted
 func fetchEvents(ctx context.Context, client api.KubernetesClient, namespace, vmName string) string {
 	core := kubernetes.NewCore(client)
-	eventMap, err := core.EventsList(ctx, namespace)
+	eventMap, err := core.EventsList(ctx, namespace, api.ListOptions{})
```
manusa: Now that `EventsList` accepts a `FieldSelector`, this is a great place to push the kind/name match server-side instead of listing every event in the namespace and filtering client-side below. Something like:

```go
opts := api.ListOptions{ListOptions: metav1.ListOptions{FieldSelector: "involvedObject.name=" + vmName}}
eventMap, err := core.EventsList(ctx, namespace, opts)
```

The `virt-launcher-<vm>` prefix branch still has to stay client-side (prefix match isn't expressible as a fieldSelector), but the exact-name match for VM and VMI events can move to the apiserver. That directly addresses the busy-namespace cost this PR is meant to fix.

evals/tasks/core/filter-events-by-type/filter-events-by-type.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  labels:
+    suite: core
+  name: filter-events-by-type
+  difficulty: easy
+steps:
+  setup:
+    file: setup.sh
+  verify:
+    contains: "bad-pod"
```
manusa: `contains: "bad-pod"` passes for an unfiltered listing too (the Normal `good-pod` events plus the Warning `bad-pod` events both contain the string), so this eval can't distinguish a model that learned `fieldSelector` from one that didn't.

The framework doesn't ship a native `not_contains`, but a `verify.sh` that captures the response and asserts `good-pod` is absent would actually exercise the parameter. At minimum, please add a TODO noting this gap.

evals/tasks/core/filter-namespace-by-name/filter-namespace-by-name.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  labels:
+    suite: core
+  name: filter-namespace-by-name
+  difficulty: easy
+steps:
+  setup:
+    file: setup.sh
+  verify:
+    contains: "ns-beta"
```
manusa: Same concern as the events eval: `contains: "ns-beta"` passes for an unfiltered `namespaces_list` (it would still return all three of `ns-alpha`, `ns-beta`, `ns-gamma`). Consider a `verify.sh` that asserts `ns-alpha` and `ns-gamma` are absent.

manusa (changes requested):
Thanks for tackling this. `events_list` on a busy namespace really was a footgun, and the implementation cleanly mirrors `pods_list`. The signature change to `EventsList(ctx, namespace, options)` is consistent with `NamespacesList` and `PodsListInAllNamespaces`, all call sites are updated (the kubevirt `vm_troubleshoot.go` migration is correct), and the snapshot updates cover core, full, openshift and multicluster.

A couple of things I'd like addressed before merge. Both are scoped tightly enough to fit this PR, and one is a missed opportunity that the new signature begs for.

### Must fix

**1. `REGEX_FIELDSELECTOR` (`pkg/toolsets/core/toolset.go:14`) rejects values the tool says it supports.**

The events_list description lists `involvedObject.apiVersion` as a supported selector field. Realistic values for that include `apps/v1`, `batch/v1`, `events.k8s.io/v1`, all of which contain `/`, which the pattern `^[.\-A-Za-z0-9]+([=!,]{1,2}[.\-A-Za-z0-9]+)+$` does not allow. Verified empirically:
```
"involvedObject.apiVersion=apps/v1" => false
```
So a spec-compliant MCP client will refuse the call before it reaches the apiserver. Adding `/` to the character class is safe for the other tools that share this regex. Pods, Namespaces and Resources selector values are DNS-1123 and never contain `/`, so widening cannot make a previously-invalid selector match.

Side note: the same regex also accepts garbage like `a==b`, `a,b`, `a!=b!=c`. That's a pre-existing wart, fine to leave for a follow-up, but please add at least one positive test for a comma-separated multi-clause selector while you're in `regex_test.go`.

**~~2. `projects_list` should gain the same parameter (`pkg/toolsets/core/namespaces.go`).~~**

> **Withdrawn.** @difrost was right: the OpenShift Project `List` endpoint silently drops the field selector ([proxy.go L95-L106](https://github.com/openshift/openshift-apiserver/blob/999dd5aefb3efef7ff26105b3208f7fc04ff5b06/pkg/project/apiserver/registry/project/proxy/proxy.go#L95-L106)), so adding `fieldSelector` to `projects_list` would be a no-op. See [the thread](https://github.com/containers/kubernetes-mcp-server/pull/1138#discussion_r3266660945) for the full receipt.

~~OpenShift `projects_list` supports the same field selectors as Namespaces (`metadata.name`, `status.phase`). Leaving it out creates an asymmetry where the same agent gets the filtering capability on Kubernetes but loses it on OpenShift. Easy mirror of `namespacesList`.~~

**3. `vm_troubleshoot.go` should push the filter server-side (`pkg/toolsets/kubevirt/vm_troubleshoot.go:361`).**

`fetchEvents` currently lists every event in the namespace and then filters client-side by `involvedObject.name` matching the VM (or `virt-launcher-<vm>` prefix). With your new option, the exact-name match can be pushed to the apiserver, which is exactly the kind of busy-namespace cost this PR is meant to remove. A simple form: `FieldSelector: "involvedObject.name=" + vmName`, while keeping the existing client-side prefix check for `virt-launcher-*` pods (prefix match isn't expressible as a fieldSelector).

### Optional but nice to have

**4. Eval verifications don't prove the model used `fieldSelector`.**

`contains: "bad-pod"` and `contains: "ns-beta"` both pass for unfiltered listings, so today these evals can't distinguish "agent learned the new parameter" from "agent listed everything". I see you flagged this in the description, and the framework's grammar (`contains`, `llmJudge.contains`, `verify.sh`) doesn't ship a native `not_contains`. The pragmatic option is a `verify.sh` that captures the response and greps for the negative case, or one that re-invokes the tool with the expected selector and compares the count. At minimum, a TODO in the task YAML noting the limitation would be useful.

**5. Regex test coverage (`pkg/toolsets/core/regex_test.go:36-51`).**

Once #1 is addressed, please add positive cases for:
- `type=Warning,reason=BackOff` (comma-separated)
- `involvedObject.apiVersion=apps/v1` (the case that motivated the fix)

A `Test_FieldSelectorRegex_is_invalid` table with whitespace, empty and garbage cases would be valuable too but I'm fine deferring it.

### Assessment
With #1 and #3 addressed I'm happy. Thanks!

pkg/toolsets/core/namespaces.go
difrost: @manusa when I've check on my OCPs I was not able to filter by e.g. `metadata.name` and assumed this doesn't work. Could be a bug or sth ... I was checking on 4.18.

manusa: You're right, and sorry for sending you on a goose chase. I went and read the openshift-apiserver source: the Project `List` endpoint deliberately drops the field selector.

In [`pkg/project/apiserver/registry/project/proxy/proxy.go`](https://github.com/openshift/openshift-apiserver/blob/999dd5aefb3efef7ff26105b3208f7fc04ff5b06/pkg/project/apiserver/registry/project/proxy/proxy.go#L95-L106):

```go
func (s *REST) List(ctx context.Context, options *metainternal.ListOptions) (runtime.Object, error) {
    ...
    labelSelector, _ := apihelpers.InternalListOptionsToSelectors(options)
    namespaceList, err := s.lister.List(user, labelSelector)
    ...
}
```

`apihelpers.InternalListOptionsToSelectors` returns `(labelSelector, fieldSelector)` but the project proxy assigns the field selector to `_` and only forwards the label selector to `s.lister.List`. So `fieldSelector` on `oc get project` is a silent no-op on the List path. Interestingly, [`Watch` on the same file](https://github.com/openshift/openshift-apiserver/blob/999dd5aefb3efef7ff26105b3208f7fc04ff5b06/pkg/project/apiserver/registry/project/proxy/proxy.go#L108-L143) does honor both selectors via `MatchProject`, which is why `projectToSelectableFields` in [`pkg/project/util/util.go`](https://github.com/openshift/openshift-apiserver/blob/999dd5aefb3efef7ff26105b3208f7fc04ff5b06/pkg/project/util/util.go) still declares `metadata.name`, `metadata.namespace`, `status.phase` — List just never reaches it.

So your empirical finding on 4.18 was correct and leaving `projects_list` alone is the right call. Withdrawing #2, thanks for pushing back.

pkg/toolsets/core/events.go
difrost: Implemented in [076c695](https://github.com/containers/kubernetes-mcp-server/pull/1138/commits/076c695c6a7e437381919b9b276e9fab07d8fc8a) including the test for invalid pattern.

pkg/toolsets/core/namespaces.go
difrost: Yee, the API lists support for fieldSelector so indeed it's really confusing to find out that this feature is not actually implemented.

pkg/toolsets/kubevirt/vm_troubleshoot.go
difrost: Implemented in [076c695](https://github.com/containers/kubernetes-mcp-server/pull/1138/commits/076c695c6a7e437381919b9b276e9fab07d8fc8a).

evals/tasks/core/filter-events-by-type/filter-events-by-type.yaml
difrost: I was looking for a way to set this eval up correctly and checked whether mcpchecker passes the response to the script in a given stage but I couldn't make it working (CC: @Cali0707). Is this feature really there? That would be very handy.

Cali0707: It was added here: https://github.com/mcpchecker/mcpchecker/pull/285

difrost: Thanks @Cali0707, I've found the proposal.

@manusa Implemented in [f90ea7b](https://github.com/containers/kubernetes-mcp-server/pull/1138/commits/f90ea7b0be178aac3f0847b2480e56b068a9ebf6).

evals/tasks/core/filter-namespace-by-name/filter-namespace-by-name.yaml
difrost: Implemented and tested for both evals in [f90ea7b](https://github.com/containers/kubernetes-mcp-server/pull/1138/commits/f90ea7b0be178aac3f0847b2480e56b068a9ebf6).

manusa (approved):
Thanks for the quick turnaround. All three must-fix points are addressed:

- **#1** regex now accepts `apps/v1`-style values (`/` added to the value-side character class only, conservative and correct).
- **#3** `fetchEvents` pushes the exact-name match server-side.
- **#5** regex_test.go gained the comma-separated positive case I asked for, plus a new `Test_FieldSelectorRegex_is_invalid` table with whitespace/empty/garbage cases. Nicely above and beyond.
- **#4** evals migrated to `mcpchecker/v1alpha2` and the verify steps now actually exercise the filter (events checks `bad-pod` present AND `good-pod` absent; namespaces asserts no non-`ns-beta` output).

LGTM, approving.

PR #1141, opened 2026-05-07
ci(mcpchecker): move to using mcpchecker diff action instead of bash scripts for diff

Cali0707:
This PR cleans up the mcpchecker action to use the mcpchecker diff action from the project instead of installing the CLI directly and using it in a bash script.

This also helps fix failures we are seeing now in CI as mcpchecker currently requires golang 1.26.2, leading to failures on that install step


manusa (approved):
LGTM, thx!

nader-ziada (approved):
LGTM

PR #1143, opened 2026-05-09
feat(resources): add statusOnly parameter to resources_get

### LINKED ISSUE #876: [Feature Request] Add statusOnly parameter to resources_get to reduce context usage
When diagnosing issues with Kubernetes resources, I typically only need the `.status` section (conditions, phase, readiness, error messages). The current resources_get tool returns the entire resource, which can be very large — especially for CRDs/operators (e.g., ClusterVersion, ArgoCD Application, MachineSet) where .spec, .metadata.managedFields, and annotations can be hundreds of lines.

This wastes LLM context window tokens on information that isn't needed for diagnosis, and can cause truncation on large CRs. 

Happy to open up a PR

Semih702:
Add an optional boolean `statusOnly` parameter to the `resources_get` tool.
When set to true, only the `.status` field of the Kubernetes resource is
returned instead of the full object. When the resource has no status field,
an empty object is returned. Default behavior is unchanged.

Changes:
- `pkg/toolsets/core/resources.go`: schema definition and handler logic
- `pkg/mcp/resources_test.go`: three test scenarios (status returned,
  explicit false, missing status field)

Closes #876

### DIFF
--- a/pkg/mcp/resources_test.go
+++ b/pkg/mcp/resources_test.go
@@ -362,6 +362,77 @@ func (s *ResourcesSuite) TestResourcesGet() {
 			s.Equalf("default", decodedNamespace.GetName(), "invalid namespace name, expected default, got %v", decodedNamespace.GetName())
 		})
 	})
+	s.Run("resources_get with statusOnly returns only status", func() {
+		result, err := s.CallTool("resources_get", map[string]interface{}{
+			"apiVersion": "v1",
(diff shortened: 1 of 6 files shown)


Cali0707:
Related to the broader https://github.com/containers/kubernetes-mcp-server/issues/877 - IMO let's prioritize evaluating the best approach forward there (cc @manusa )

PR #1144, opened 2026-05-11
docs(agents): expand downstream test guidance + refactor TOML idioms

manusa:
## Summary

Builds on @matzew's #1140 (cherry-picked as the first commit) and:

- **Refactors test config idioms** — converts `confirmation_test.go` and `mcp_toolset_prompts_test.go` from the `config.ReadToml` + manual restore pattern to the dominant `toml.Unmarshal([]byte(...), s.Cfg)` MERGE idiom used by ~25 other test files. The merge approach automatically preserves runtime fields the suite already populated, so no `cfgFromTOML` helper is needed.
- **Documents the legitimate `ReadToml` exception** — `kiali_test.go` and `mcp_config_provider_test.go::TestToolHandlerReceivesToolsetConfig` keep `ReadToml` because their TOML uses `[toolset_configs.kiali]`, which is typed `map[string]toml.Primitive` and requires the second-phase parse `ReadToml` performs with TOML metadata. Both files now carry an inline comment explaining the constraint.
- **Expands AGENTS.md** to cover (1) the second downstream override layer (per-toolset `internal/defaults/defaults_override.go`, e.g. `kiali` → `ossm`), (2) the merge-first TOML idiom, (3) when `ReadToml` is still required, and (4) the `(&toolset.Toolset{}).GetName()` pattern from `kiali_test.go` for resolving toolset names dynamically.

## Verification

- `go test ./pkg/mcp/... ./pkg/config/...` passes on upstream.
- Re-ran the same suites with `pkg/config/config_default_overrides.go` and `pkg/toolsets/kiali/internal/defaults/defaults_override.go` patched to match `openshift/main` (`ReadOnly: true`, `Toolsets: ["core", "config"]`, `kiali → ossm`). All four touched suites (`TestConfirmationRules`, `TestKiali`, `TestMcpConfigProvider`, `TestMcpToolsetPromptsSuite`) still pass. The only failure under the simulated downstream merge is the pre-existing `TestToolsets/TestGranularToolsetsTools/Toolset_ossm` snapshot-file-missing failure, which reproduces without any of these changes.

## Test plan

- [x] Verify `go test ./pkg/mcp/... ./pkg/config/...` is green on upstream
- [x] Verify the four refactored suites stay green when downstream override files are patched in locally
- [ ] Reviewer to confirm the AGENTS.md guidance reads correctly

Closes #1140

Co-Authored-By: Matthias Wessendorf <mwessend@redhat.com>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -236,10 +236,88 @@ func FieldString(obj *unstructured.Unstructured, path string) string {
 
 #### Test Configuration and Downstream Compatibility
 
-This project supports downstream builds that override default configuration via `pkg/config/config_default_overrides.go` (e.g., `ReadOnly`, `Toolsets`, `ToolOverrides`). Tests must work correctly regardless of what downstream overrides are active.
+Downstream builds may override defaults at two layers, and tests must work
+regardless of which overrides are active:
 
-- **`BaseMcpSuite.SetupTest()` uses `config.BaseDefault()`** (not `config.Default()`) as the starting point for `s.Cfg`, so the suite config reflects pure upstream defaults without downstream overrides
-- **When creating a secondary config** (e.g., for reload tests), inherit runtime-sensitive fields from `s.Cfg` instead of relying on `config.Default()`:
+1. **Static config overrides** — `pkg/config/config_default_overrides.go`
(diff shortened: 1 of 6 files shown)


matzew:
LGTM, thanks!

PR #1147, opened 2026-05-11
feat(tokenexchange): add EC key support and Entra ID federated auth style

nader-ziada:
- Add ECDSA P-256/P-384 key support for JWT client assertions, enabling SPIRE X.509-SVIDs in the OBO flow.
- Add new "federated" auth style that reads external IdP JWTs from a file for Entra ID workload identity federation.

reference: https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation


matzew (approved):
LGTM

PR #777, opened 2026-02-12
feat(api): add structured content support

onmete:
**Purpose:** Allow tool handlers to return structured JSON data alongside the text content, which MCP Apps can consume for rich UI rendering.

**Files changed:**

- [pkg/api/toolsets.go](pkg/api/toolsets.go) -- Add `StructuredContent any` field to `ToolCallResult`, add `NewToolCallResultWithStructuredContent()` constructor
- [pkg/mcp/mcp.go](pkg/mcp/mcp.go) -- Add `NewTextResultWithStructuredContent()`, refactor `NewTextResult()` to delegate to it
- [pkg/mcp/gosdk.go](pkg/mcp/gosdk.go) -- Change handler bridge to call `NewTextResultWithStructuredContent(result.Content, result.StructuredContent, result.Error)`

PR #1151, opened 2026-05-14
feat(config): structure context list output

immanuwell:
## What
Tiny DX fix: `configuration_contexts_list` keeps the same text output, but now also returns `structuredContent` with `defaultContext` and a sorted `contexts` list. Small win, less text parsing.

## Repro
On `main`, use any normal kubeconfig with at least one context, local kind/minikube is enough. Call `configuration_contexts_list`; the context list is only in `content[0].text`, while `structuredContent` is empty, so clients have to parse the prose. Kinda meh.

The added MCP test reproduces it: the new structured-content assertion fails on `main` before the handler change, and passes with this patch. No cloud-provider quota or hard API ceiling involved here, this is local kubeconfig data.

## Checks
- `go test ./pkg/mcp -run TestConfiguration/TestContextsList -count=1`
- `go test ./pkg/toolsets/config ./pkg/api -count=1`
- `make build`

No exact matching issue found. Related background: #777 and #960 added structured-content plumbing.

### DIFF
--- a/pkg/toolsets/config/configuration.go
+++ b/pkg/toolsets/config/configuration.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/google/jsonschema-go/jsonschema"
 	"k8s.io/utils/ptr"
@@ -10,6 +11,17 @@ import (
 	"github.com/containers/kubernetes-mcp-server/pkg/kubernetes"
 	"github.com/containers/kubernetes-mcp-server/pkg/output"
 )
+
+type contextInfo struct {
+	Name    string `json:"name"`
+	Server  string `json:"server"`
+	Default bool   `json:"default"`
+}
+
+type contextsListResult struct {
+	DefaultContext string        `json:"defaultContext"`
+	Contexts       []contextInfo `json:"contexts"`
+}
 
 func initConfiguration() []api.ServerTool {
 	tools := []api.ServerTool{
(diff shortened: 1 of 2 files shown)


alexeykazakov (approved):

pkg/toolsets/config/configuration.go
```diff
@@ -105,20 +117,35 @@ func contextsList(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
...
+			Name:    context,
+			Server:  server,
+			Default: context == defaultContext,
+		})
 	}
 	result += "---------\n\n"
 
 	result += "To use a specific context with any tool, set the 'context' parameter in the tool call arguments"
 
-	// TODO: Review output format, current is not parseable and might not be ideal for LLM consumption
```
Cali0707, line 120: Let's not fully remove this comment - I think it is still true for the unstructured case. In the future we may want to revisit the format from the unstructured case as well.

[From the spec](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#structured-content):
> Structured content is returned as a JSON object in the structuredContent field of a result. For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.

IMO since this is only a SHOULD we don't need to change it here, but worth revisiting later

manusa: We'll take care of this in a follow-up

PR #1160, opened 2026-05-20
fix(auth): Token pass-through mode

### LINKED ISSUE #1131: auth: redefine require_oauth + skip_jwt_verification as pure token passthrough mode
# Pure token passthrough: `require_oauth` + `skip_jwt_verification` should disable all local token validation

## Summary

When the MCP server runs in `cluster_auth_mode=passthrough` (the default) and the operator wants the cluster to be the sole authority for authentication, there is currently no clean way to express that. The natural combination — `require_oauth=true` + `skip_jwt_verification=true` with no `authorization_url` — is documented as "JWTs accepted without cryptographic signature verification," but the implementation still performs *several* layers of local validation that reject real-world tokens before they ever reach the API server.

This issue proposes redefining that combination as **"presence-only mode"**: the MCP server enforces that an `Authorization: Bearer …` header is present and forwards it verbatim. Parsing, claim validation, audience checks, and JWT-shape heuristics all become out of scope. The cluster is the authority.

## Motivating PR

[#1121 — feat(auth): add `accept_opaque_tokens` for non-JWT bearer token passthrough](https://github.com/containers/kubernetes-mcp-server/pull/1121)

PR #1121 adds an `accept_opaque_tokens` flag to make OpenShift OAuth `sha256~…` tokens flow through the MCP server. While it solves the immediate problem, the investigation that prompted this issue concluded the underlying cause is broader: `skip_jwt_verification` does not actually skip JWT *parsing* (only signature verification), and `ValidateOffline` runs unconditionally. PR #1121 is therefore a workaround for a too-narrow definition of `skip_jwt_verification`. Fixing the underlying semantics would obviate the new flag.

## Current behavior (what's wrong)

`pkg/http/authorization.go` (`AuthorizationMiddleware`) treats `require_oauth=true` as "validate locally" rather than "presence required." When that flag is set:

1. **`pkg/http/authorization.go:99` — `ParseJWTClaims(token)`.** Uses `jwt.ParseSigned` from `go-jose`, which fails on any non-JWT bearer token (e.g., OpenShift `sha256~…`). `skip_jwt_verification` does not bypass this — it only skips signature verification *after* a successful parse.
2. **`pkg/http/authorization.go:106` — `claims.ValidateOffline(OAuthAudience)`.** Enforces audience and expiration even when `skip_jwt_verification=true`. A cluster-issued JWT whose audience is the API server (not the MCP server) is rejected.
3. **`pkg/http/authorization.go:121` — 500 Internal Server Error** when `require_oauth=true`, no `authorization_url`, and `skip_jwt_verification=false`. A passthrough deployment without local validation is not a server misconfiguration; it is a legitimate operating mode.
4. **`pkg/config/config.go:569-582` — `validateSkipJWTVerification`.** Refuses to start under the same conditions, hard-coding the assumption that `require_oauth` implies local JWT validation.

`cluster_auth_mode` is **not consulted** by the authorization middleware at all — it is read only by `pkg/kubernetes/token_exchange.go:66` to decide whether to forward or drop the bearer. So the validation pipeline runs identically in `passthrough` and `kubeconfig` modes, even though only `passthrough` actually uses the token.

## Proposed behavior

Redefine the three-flag combination

```toml
require_oauth          = true
skip_jwt_verification  = true
cluster_auth_mode      = "passthrough"   # default; can be omitted
# authorization_url is unset
```

as **"presence-only passthrough mode"**:

- `Authorization: Bearer <token>` must be present and non-empty. Missing or malformed → 401.
- No JWT parsing, no offline claims validation, no audience check, no shape heuristic.
- The token is stored in the request context and forwarded by the existing passthrough machinery.
- The cluster (or a downstream reverse proxy / OIDC sidecar) is the sole authority for the token.

This is what `skip_jwt_verification=true` *reads like* in the docs today, and what operators reasonably expect.

### 1. `pkg/http/authorization.go` — short-circuit before parsing

After the bearer-presence check (currently lines 91-95), add:

```go
if staticConfig.SkipJWTVerification && staticConfig.AuthorizationURL == "" {
    skipJWTWarningOnce.Do(func() {
        klog.Warningf("Bearer token forwarded without local validation (skip_jwt_verification=true, no authorization_url) — the cluster is the sole authority")
    })
    ctx := context.WithValue(r.Context(), internalk8s.OAuthAuthorizationHeader, authHeader)
    next.ServeHTTP(w, r.WithContext(ctx))
    return
}
```

This single branch:
- Fixes the OpenShift `sha256~…` case (subsumes PR #1121's `isJWT`/`accept_opaque_tokens` logic for this combo).
- Fixes the `oauth_audience` mismatch case for cluster-issued JWTs.
- Removes the 500-vs-401 ambiguity for this configuration.

When `authorization_url` *is* set, validation should still run — `skip_jwt_verification` should not silently disable a configured OIDC provider. The existing OIDC path is unchanged.

### 2. `pkg/config/config.go` — keep `validateSkipJWTVerification`, update its message

The startup check at `pkg/config/config.go:569-582` is still correct in spirit: an operator must explicitly opt in to "no local validation." Update the warning at line 574 to match the broadened semantics:

```go
klog.Warningf("skip_jwt_verification is enabled with no authorization_url: bearer tokens will be forwarded without any local validation. " +
    "The cluster (or a trusted upstream) is the sole authority. Only use this when cluster_auth_mode=passthrough and the cluster validates tokens directly.")
```

### 3. PR #1121 (`accept_opaque_tokens`) becomes redundant for the passthrough use case

With change #1, OpenShift OAuth tokens flow through naturally under `skip_jwt_verification=true`. The PR's `accept_opaque_tokens` flag would only be meaningful if someone wanted to accept opaque tokens *while still validating JWTs against an OIDC provider* — a much narrower use case. Recommendation: close #1121 and direct users to the simpler `skip_jwt_verification=true` configuration, OR repurpose its test cases (especially `TestAuthorizationOpaqueTokenAccepted`) to cover the new short-circuit branch.

### 4. Documentation — `docs/configuration.md`

The current `skip_jwt_verification` row says "allows JWTs without cryptographic signature verification." This is too narrow and is what's driving operators to ask for additional flags. Replace with explicit passthrough-mode framing.

**Add a new subsection** under "OAuth/OIDC authentication" titled **"Pure token passthrough (delegate validation to the cluster)"** containing:

- The three-flag combo and a worked example for OpenShift OAuth.
- A clear statement: "the MCP server performs **no** token validation in this mode; it only enforces that a bearer header is present and forwards it to the cluster."
- A security note: "use this only when the cluster (or a trusted upstream component such as a reverse proxy or OIDC sidecar) is configured to validate tokens. Without that, the MCP server is effectively unauthenticated."
- A note that `oauth_audience`, `authorization_url`, and JWT-related options are ignored in this mode.

**Update the existing `skip_jwt_verification` row** to:

> When `true` and `authorization_url` is unset, the server forwards the bearer token without **any** local validation (no parse, no claims check, no audience check). Required to enable pure passthrough with non-JWT tokens (e.g., OpenShift OAuth `sha256~…`). When `true` and `authorization_url` is set, this flag has no effect — the configured OIDC provider validates tokens normally. Only use the no-`authorization_url` form when a downstream component (cluster, reverse proxy) is the authority.

**Update the `require_oauth` row** to clarify that it controls *presence requirement*, not *validation strategy*. Validation is determined separately by `authorization_url` and `skip_jwt_verification`.

### 5. Cross-check `cluster_auth_mode=kubeconfig`

In kubeconfig mode the bearer is dropped at `pkg/kubernetes/token_exchange.go:68` regardless. Requiring a bearer header just to discard it is a footgun. Either:
- Skip the bearer presence check when `ResolveClusterAuthMode() == ClusterAuthKubeconfig`, or
- Document explicitly that `require_oauth` is only meaningful with `passthrough`, and validate at startup that the combination `require_oauth=true` + `cluster_auth_mode=kubeconfig` is intentional (e.g., gateway-style deployment that authenticates clients but uses a service account for the cluster).

This is a separable follow-up; mention it in the issue but don't block the main change on it.

## Open question — keep expiration check?

`ValidateOffline` enforces audience and expiration. Audience must go (it breaks cluster-issued JWTs). Expiration is a defensible cheap fail-fast even in passthrough. Two options:

- **Drop it entirely (recommended).** The cluster will return a fast 401 on expired tokens; the MCP server should not duplicate that authority.
- **Keep expiration only**, by calling `ValidateOffline("")` when the token parses as a JWT, and skipping the call entirely when it doesn't. This adds complexity and reintroduces the JWT-shape heuristic from PR #1121.

Recommend option 1 for simplicity and consistency. Calling out for explicit decision in the issue.

## Test plan

New tests in `pkg/http/authorization_mcp_test.go`:

- `TestAuthorizationPassthroughOpaqueToken` — opaque `sha256~…` token + `skip_jwt_verification=true` → 200, header forwarded.
- `TestAuthorizationPassthroughClusterIssuedJWT` — JWT with audience set to the cluster API server (not the MCP server) + `skip_jwt_verification=true` + `oauth_audience` set to something else → 200 (audience check is bypassed).
- `TestAuthorizationPassthroughExpiredToken` — expired JWT + `skip_jwt_verification=true` → 200 (proves no local validation; the cluster will reject it).
- `TestAuthorizationPassthroughMissingHeader` — no header + `skip_jwt_verification=true` → 401 (presence still required).
- `TestAuthorizationOIDCStillValidates` — `authorization_url` set + `skip_jwt_verification=true` → OIDC validation still runs (regression guard for change #1, point about not silently disabling OIDC).

Update existing tests:

- `TestAuthorizationRawToken` — verify the old "raw token" path semantics still hold for the case where `oauth_audience` is meaningfully set with `skip_jwt_verification=true` *and* `authorization_url=""`. Decide whether that combo is still supported, or whether `oauth_audience` should be ignored when `authorization_url` is unset.

## Acceptance criteria

- [ ] OpenShift OAuth `sha256~…` tokens reach the API server with `require_oauth=true` + `skip_jwt_verification=true` + `cluster_auth_mode=passthrough` and no other flags.
- [ ] Cluster-issued JWTs (any audience) reach the API server in the same configuration.
- [ ] `Authorization` header is still required; missing/malformed → 401.
- [ ] No 500 responses for "passthrough without OIDC provider" — that configuration is now a first-class supported mode.
- [ ] OIDC validation (when `authorization_url` is set) is unaffected.
- [ ] `docs/configuration.md` has a dedicated "Pure token passthrough" section and the row descriptions for `require_oauth` and `skip_jwt_verification` are updated.
- [ ] PR #1121 is closed or re-scoped.

## Out of scope

- Any change to `cluster_auth_mode=kubeconfig` semantics (mentioned as a follow-up in §5).
- Token exchange / STS flows — those operate on already-validated tokens and are orthogonal.
- Changes to the `require_oauth=false` path — already a clean passthrough today.

nbottari9:
### Files Modified

```
 docs/configuration.md              |  15 ++-
 pkg/config/config.go               |   4 +-
 pkg/http/authorization.go          |  10 ++
 pkg/http/authorization_mcp_test.go | 226 +++++++++++++++++++++++++++++++++++--
 4 files changed, 244 insertions(+), 11 deletions(-)
```

#### 1. Authorization Middleware Passthrough Logic (`pkg/http/authorization.go`)

```go
// Token passthrough, skips all JWT processing
if staticConfig.SkipJWTVerification && staticConfig.AuthorizationURL == "" {
    skipJWTWarningOnce.Do(func() {
        klog.Warningf("Bearer token forwarded without local validation (skip_jwt_verification=true and no authorization_url) - the cluster is the sole authority")
    })
    ctx := context.WithValue(r.Context(), internalk8s.OAuthAuthorizationHeader, authHeader)
    next.ServeHTTP(w, r.WithContext(ctx))
    return
}
```

**Why this matters:**
- The early return **must** occur before `ParseJWTClaims()` to allow opaque tokens
- Opaque tokens (like OpenShift's `sha256~...` format) cannot be parsed as JWT
- Previously, these tokens would trigger a 401 before reaching the skip flag

#### 2. Updated Validation Warning (`pkg/config/config.go`)

```go
if c.SkipJWTVerification {
    klog.Warningf("skip_jwt_verification is enabled with no authorization_url: bearer tokens will be forwarded without any local validation. " +
        "The cluster (or a trusted upstream) is the sole authority. Only use this when cluster_auth_mode=passthrough and the cluster validates tokens directly.")
    return nil
}
```

**Changes:**
- Clarifies this is **pure passthrough** (no validation at all)
- Explicitly states cluster is the sole authority
- Adds guidance about `cluster_auth_mode=passthrough`

#### 3. Documentation Updates (`docs/configuration.md`)

Added new section: **Pure token passthrough (delegate validation to the cluster)**

```toml
require_oauth         = true
skip_jwt_verification = true
cluster_auth_mode     = "passthrough"
# authorization_url is not set
```

**Security notes added:**
- MCP server performs **no** token validation in this mode
- Only enforces bearer header **presence**
- Use **ONLY** when cluster/upstream validates tokens
- `oauth_audience`, `authorization_url`, and JWT options are **ignored**

---
## Tests for pure token passthrough mode

Implements test coverage for issue #1131 - pure token passthrough mode that enables the cluster to be the sole authentication authority.

#### Test Coverage (`pkg/http/authorization_mcp_test.go`)

Added **2 new test cases** covering previously untested passthrough scenarios:

1. **`TestAuthorizationPassthroughExpiredToken`** (lines 867-905)
   - Verifies expired JWTs bypass local expiration validation and reach the cluster unmodified
   - Covers test scenario 3 from issue #1131
   - Confirms the cluster is responsible for validating token expiration in passthrough mode

2. **`TestAuthorizationPassthroughClusterIssuedJWT`** (lines 907-950)
   - Verifies JWTs with cluster API server audience (not `mcp-server`) are accepted
   - Covers test scenario 2 from issue #1131
   - Confirms audience checks are bypassed when `skip_jwt_verification=true` and `authorization_url=""` 

Updated **2 existing test cases** to match the new warning message introduced in the passthrough implementation:

3. **`TestAuthorizationOfflineOnlyWarning`** (lines 516-541)
   - Updated expected log message from `"JWT accepted without signature verification"` to `"Bearer token forwarded without local validation"`
   - Updated guidance message from `"set authorization_url for secure validation"` to `"cluster is the sole authority"`
   - Renamed internal test description from "offline-only JWT validation" to "passthrough mode"

4. **`TestAuthorizationOfflineOnlyWarningOnce`** (lines 543-565)
   - Updated to search for the new warning message `"Bearer token forwarded without local validation"`

#### Complete Test Coverage for Issue #1131

All 5 test scenarios from issue #1131 are now covered:

| # | Test Scenario | Test Name | Status |
|---|---------------|-----------|--------|
| 1 | Opaque Token Passthrough | `TestAuthorizationPassthroughOpaqueToken` | ✓ Existing |
| 2 | Cluster-Issued JWT | `TestAuthorizationPassthroughClusterIssuedJWT` | ✓ **New** |
| 3 | Expired Token Handling | `TestAuthorizationPassthroughExpiredToken` | ✓ **New** |
| 4 | Missing Header Enforcement | `TestAuthorizationPassthroughMissingHeader` | ✓ Existing |
| 5 | OIDC Regression Guard | `TestAuthorizationPassthroughOIDCPathUnaffected` | ✓ Existing |

### Test Results

All authorization tests pass successfully:
```
PASS
ok  	github.com/containers/kubernetes-mcp-server/pkg/http	1.836s
```

### Related Changes

The test updates align with the implementation changes made in commits d08db95 and 9583e77:

- **`pkg/http/authorization.go`**: Added early-return passthrough logic before JWT parsing (lines 99-107)
- **`pkg/config/config.go`**: Updated validation warning message for `skip_jwt_verification` (lines 584-585)
- **`docs/configuration.md`**: Added "Pure token passthrough" configuration example and security guidance

### Testing Methodology

All new tests follow the project's testing patterns:

- Uses `testify/suite` framework for organization
- Black-box testing focusing on observable behavior (not implementation details)
- Each test verifies three aspects:
  1. MCP session initialization succeeds
  2. Tool calls forward the token correctly
  3. Backend receives the unmodified token
- Comprehensive nested subtests with descriptive names
- Atomic value capture to verify backend receives correct Authorization header

### Security Considerations

The tests verify that passthrough mode:

- **Still enforces bearer header presence** (401 when missing)
- **Only activates when both conditions are met**: `skip_jwt_verification=true` AND `authorization_url=""` 
- **Does NOT bypass OIDC validation** when `authorization_url` is configured
- **Forwards all token types unchanged** (opaque, expired, wrong audience, malformed)

---

### New Test: `TestAuthorizationPassthroughExpiredToken`

**Purpose:** Verify expired JWTs pass through without local expiration checks

**Configuration:**
```go
s.StaticConfig.SkipJWTVerification = true
s.StaticConfig.AuthorizationURL = ""
```

**Test token:**
- Uses `tokenBasicExpired` (exp claim set to `1` - Unix epoch second 1)
- Would normally fail with `"token is expired (exp)"` error
- In passthrough mode, must reach cluster unchanged

**Assertions:**
1. MCP session initializes successfully (no 401)
2. Tool call succeeds without local expiration error
3. Backend receives the expired token unchanged

**Coverage:** Issue #1131 test scenario 3

---

### New Test: `TestAuthorizationPassthroughClusterIssuedJWT`

**Purpose:** Verify JWTs with cluster audience (not mcp-server) pass through

**Configuration:**
```go
s.StaticConfig.SkipJWTVerification = true
s.StaticConfig.AuthorizationURL = ""
s.StaticConfig.OAuthAudience = "mcp-server"  // Set but ignored in passthrough
```

**Test token:**
- Custom JWT with audience `["https://kubernetes.default.svc.cluster.local"]`
- Does **not** include `"mcp-server"` in audience claim
- Would normally fail with `"invalid audience claim (aud)"` error
- JWT payload (decoded):
  ```json
  {
    "aud": ["https://kubernetes.default.svc.cluster.local"],
    "exp": 253402297199,
    "iss": "https://kubernetes.default.svc.cluster.local",
    "sub": "system:serviceaccount:default:default"
  }
  ```

**Assertions:**
1. MCP session initializes despite audience mismatch
2. Tool call succeeds without audience validation error
3. Backend receives cluster-issued JWT unchanged

**Coverage:** Issue #1131 test scenario 2

---

### Updated Test: `TestAuthorizationOfflineOnlyWarning`

**Changes:**
- Updated log assertion from `"JWT accepted without signature verification"` to `"Bearer token forwarded without local validation"`
- Updated guidance assertion from `"set authorization_url for secure validation"` to `"cluster is the sole authority"`
- Updated test description comments to reflect "passthrough mode" terminology

**Why:** Aligns with implementation's clearer messaging about pure passthrough

---

### Updated Test: `TestAuthorizationOfflineOnlyWarningOnce`

**Changes:**
- Updated `strings.Count()` search from `"JWT accepted without signature verification"` to `"Bearer token forwarded without local validation"`

**Why:** Ensures warning fires exactly once even with the new message

---

### Coverage Matrix

| Token Type | Passthrough Mode | OIDC Mode | Expected Behavior |
|------------|-----------------|-----------|-------------------|
| Valid JWT | ✓ Forwards | ✓ Validates | Passthrough: no validation, OIDC: full validation |
| Expired JWT | ✓ Forwards | ✗ Rejects (401) | **New test coverage** |
| Wrong audience JWT | ✓ Forwards | ✗ Rejects (401) | **New test coverage** |
| Opaque token | ✓ Forwards | ✗ Rejects (401) | Existing test |
| Malformed JWT | ✓ Forwards | ✗ Rejects (401) | Existing test |
| Missing header | ✗ Rejects (401) | ✗ Rejects (401) | Existing test |

### Test Patterns Used

All tests follow the project's established patterns from `CLAUDE.md`:

1. **Suite-based organization** with `testify/suite`
2. **Behavior-based testing** - test the public API, not internals
3. **No mocks** - use real HTTP servers and clients
4. **Nested subtests** with descriptive names
5. **One assertion per subtest** where possible
6. **Comprehensive edge case coverage**

### Backend Token Verification Pattern

All passthrough tests use atomic value capture to verify the backend receives the correct token:

```go
var backendAuth atomic.Value
s.MockServer.Handle(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    if auth := r.Header.Get("Authorization"); auth != "" {
        backendAuth.Store(auth)
    }
}))

// ... after tool call ...

s.Run("Backend receives token unmodified", func() {
    got, _ := backendAuth.Load().(string)
    s.Require().NotEmpty(got, "Expected backend to receive an Authorization header")
    s.Equal("Bearer "+expectedToken, got, "Token must be forwarded unchanged")
})
```

**Why this pattern:**
- Verifies end-to-end token propagation through MCP layer
- Confirms no transformation/stripping occurs
- Thread-safe for concurrent requests
- Validates the actual security-relevant observable (downstream credential)

---

## Issue Resolution

Closes #1131

### Issue Requirements Met

✅ **Test scenario 1:** Opaque Token Passthrough  
✅ **Test scenario 2:** Cluster-Issued JWT (new)  
✅ **Test scenario 3:** Expired Token Handling (new)  
✅ **Test scenario 4:** Missing Header Enforcement  
✅ **Test scenario 5:** OIDC Regression Guard  

### Acceptance Criteria

✅ All specified tokens reach the API server in passthrough mode  
✅ Missing headers trigger 401 responses  
✅ OIDC remains unaffected when `authorization_url` is configured  
✅ Documentation updated with passthrough configuration example  
✅ Security warnings clarified in both code and docs  

---

### Downstream Compatibility

Tests follow the downstream compatibility guidelines from `CLAUDE.md`:

- Start from `BaseDefault()` in suite setup to avoid config leaking
- Preserve runtime fields (`KubeConfig`, `ListOutput`, `ReadOnly`) when building configs
- Use proper TOML patterns for test configuration

### Security Implications

The passthrough mode is **intentionally permissive** for valid deployment scenarios:

**Use cases:**
- OpenShift deployments where OAuth proxy handles authentication
- Service mesh deployments with Istio/Envoy validating JWTs
- Kubernetes clusters with webhook token authentication
- Reverse proxy setups with external OIDC validation

**Protection mechanisms:**
1. Bearer header presence is still enforced (401 if missing)
2. Mode only activates with explicit configuration (`skip_jwt_verification=true` + no `authorization_url`)
3. Warning logged once per startup to alert operators
4. Documentation includes prominent security warnings
5. OIDC validation path remains unchanged when `authorization_url` is set

### Future Considerations

The test suite now provides comprehensive coverage for:
- Pure passthrough mode (`skip_jwt_verification=true`, no `authorization_url`)
- Offline JWT validation (`skip_jwt_verification=true`, `authorization_url` set)
- Full OIDC validation (`authorization_url` set)
- No authentication (`require_oauth=false`)

Any future authentication mode additions should follow similar test patterns.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -480,7 +480,7 @@ Configure OAuth/OIDC authentication for HTTP mode deployments.
 
 | Field | Type | Default | Description |
 |-------|------|---------|-------------|
-| `require_oauth` | boolean | `false` | When `true`, requires OAuth authentication for all requests. This **DOES NOT** determine validation strategy, which is done seperately by ```authorization_url``` and ```skip_jwt_verification``` |
+| `require_oauth` | boolean | `false` | When `true`, requires OAuth authentication for all requests. This **DOES NOT** determine validation strategy, which is done separately by ```authorization_url``` and ```skip_jwt_verification``` |
 | `oauth_audience` | string | `""` | Valid audience for OAuth tokens (for offline JWT claim validation). |
 | `authorization_url` | string | `""` | URL of the OIDC authorization server for token validation and STS exchange. |
 | `skip_jwt_verification` | boolean | `false` | When true and authorization_url is unset, the server forwards the bearer token without any local validation (no parse, no claims check, no audience check). Required to enable pure passthrough with non-JWT tokens (e.g., OpenShift OAuth sha256~…). When true and authorization_url is set, this flag has no effect — the configured OIDC provider validates tokens normally. Only use the no-authorization_url form when a downstream component (cluster, reverse proxy) is the authority. |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -480,10 +480,10 @@ Configure OAuth/OIDC authentication for HTTP mode deployments.
 
 | Field | Type | Default | Description |
 |-------|------|---------|-------------|
-| `require_oauth` | boolean | `false` | When `true`, requires OAuth authentication for all requests. |
+| `require_oauth` | boolean | `false` | When `true`, requires OAuth authentication for all requests. This **DOES NOT** determine validation strategy, which is done seperately by ```authorization_url``` and ```skip_jwt_verification``` |
 | `oauth_audience` | string | `""` | Valid audience for OAuth tokens (for offline JWT claim validation). |
 | `authorization_url` | string | `""` | URL of the OIDC authorization server for token validation and STS exchange. |
-| `skip_jwt_verification` | boolean | `false` | When `true`, allows JWTs without cryptographic signature verification when `require_oauth` is enabled but no `authorization_url` is configured. Only use behind a trusted reverse proxy that already verifies tokens. When `false` (default), the server refuses to start in this configuration. |
+| `skip_jwt_verification` | boolean | `false` | When true and authorization_url is unset, the server forwards the bearer token without any local validation (no parse, no claims check, no audience check). Required to enable pure passthrough with non-JWT tokens (e.g., OpenShift OAuth sha256~…). When true and authorization_url is set, this flag has no effect — the configured OIDC provider validates tokens normally. Only use the no-authorization_url form when a downstream component (cluster, reverse proxy) is the authority. |
 | `disable_dynamic_client_registration` | boolean | `false` | When `true`, disables dynamic client registration in `.well-known` endpoints. |
 | `oauth_scopes` | string[] | `[]` | Supported client scopes for the OAuth flow. |
 | `sts_client_id` | string | `""` | OAuth client ID for backend token exchange. |
@@ -524,6 +524,17 @@ sts_client_cert_file = "/path/to/client.crt"
 sts_client_key_file = "/path/to/client.key"
 sts_scopes = ["api://<DOWNSTREAM_API>/.default"]
 ```
+
+**Pure token passthrough (delegate validation to the cluster):**
+```toml
+require_oauth         = true
+skip_jwt_verification = true
+cluster_auth_mode     = "passthrough"
+# authorization_url is not set
+```
+- The MCP server performs ***no*** token validation in this mode; it only enforces that a bearer header ***is present*** and forwards it to the cluster
+- **Security note:** Use this **ONLY** when the cluster (or a trusted upstream component such as a reverse proxy or OIDC sidecar) is configured to validate tokens. Without that, the MCP server is effectively unauthenticated
+- ```oauth_audience```, ```authorization_url```, and other JWT-related options are **ignored** in this mode
 
 For a complete OIDC setup guide, see [KEYCLOAK_OIDC_SETUP.md](KEYCLOAK_OIDC_SETUP.md) or [ENTRA_ID_SETUP.md](ENTRA_ID_SETUP.md).
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -480,7 +480,7 @@ Configure OAuth/OIDC authentication for HTTP mode deployments.
 
 | Field | Type | Default | Description |
 |-------|------|---------|-------------|
-| `require_oauth` | boolean | `false` | When `true`, requires OAuth authentication for all requests. This **DOES NOT** determine validation strategy, which is done separately by ```authorization_url``` and ```skip_jwt_verification``` |
+| `require_oauth` | boolean | `false` | When `true`, requires OAuth authentication for all requests. This **DOES NOT** determine validation strategy, which is done separately by `authorization_url` and `skip_jwt_verification` |
 | `oauth_audience` | string | `""` | Valid audience for OAuth tokens (for offline JWT claim validation). |
 | `authorization_url` | string | `""` | URL of the OIDC authorization server for token validation and STS exchange. |
 | `skip_jwt_verification` | boolean | `false` | When true and authorization_url is unset, the server forwards the bearer token without any local validation (no parse, no claims check, no audience check). Required to enable pure passthrough with non-JWT tokens (e.g., OpenShift OAuth sha256~…). When true and authorization_url is set, this flag has no effect — the configured OIDC provider validates tokens normally. Only use the no-authorization_url form when a downstream component (cluster, reverse proxy) is the authority. |
@@ -534,7 +534,7 @@ cluster_auth_mode     = "passthrough"
 ```
 - The MCP server performs ***no*** token validation in this mode; it only enforces that a bearer header ***is present*** and forwards it to the cluster
 - **Security note:** Use this **ONLY** when the cluster (or a trusted upstream component such as a reverse proxy or OIDC sidecar) is configured to validate tokens. Without that, the MCP server is effectively unauthenticated
-- ```oauth_audience```, ```authorization_url```, and other JWT-related options are **ignored** in this mode
+- `oauth_audience`, `authorization_url`, and other JWT-related options are **ignored** in this mode
 
 For a complete OIDC setup guide, see [KEYCLOAK_OIDC_SETUP.md](KEYCLOAK_OIDC_SETUP.md) or [ENTRA_ID_SETUP.md](ENTRA_ID_SETUP.md).
 
--- a/pkg/http/authorization.go
+++ b/pkg/http/authorization.go
@@ -39,13 +39,13 @@ func write401(w http.ResponseWriter, wwwAuthenticateHeader, errorType, message s
 //
 //	 2. requireOAuth is set to true, server is protected:
 //
-//	    2.1. Raw Token Validation (oidcProvider is nil, SkipJWTVerification is true):
-//	         - Requires skip_jwt_verification=true; otherwise the request is rejected with 500.
-//	         - The token is validated offline for basic sanity checks (expiration).
-//	         - If OAuthAudience is set, the token is validated against the audience.
-//	         - No cryptographic signature verification is performed.
+//	    2.1. Token Passthrough mode (oidcProvider is nil, SkipJWTVerification is true):
+//	         - The token is forwarded directly to the cluster without ANY local validation.
+//	         - No JWT parsing, no expiration/audience checks, no signature verification.
+//			 - The cluster (or upstream reverse proxy) is the sole authority for validating the token.
+//			 - Use this mode for non-JWT tokens (ex. OpenShift sha256 tokens) or if delegating to cluster RBAC.
 //
-//	         see TestAuthorizationRawToken
+//	         see TestAuthorizationPassthroughOpaqueToken
 //
 //	    2.2. OIDC Provider Validation (oidcProvider is not nil):
 //	         - The token is validated offline for basic sanity checks (audience and expiration).
@@ -97,6 +97,7 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
 			token := strings.TrimPrefix(authHeader, "Bearer ")
 
 			// Token passthrough, skips all JWT processing
+			// Cluster is the sole authority for validating token (ex. sha256 token with OpenShift)
 			if staticConfig.SkipJWTVerification && staticConfig.AuthorizationURL == "" {
 				skipJWTWarningOnce.Do(func() {
 					klog.Warningf("Bearer token forwarded without local validation (skip_jwt_verification=true and no authorization_url) - the cluster is the sole authority")
@@ -125,15 +126,12 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
 						write401(w, wwwAuthenticateHeader, "temporarily_unavailable", "OIDC provider is not available")
 						return
 					}
+
 					// No provider configured - require explicit opt-in via skip_jwt_verification
-					if !staticConfig.SkipJWTVerification {
-						klog.V(1).Infof("Authentication rejected - JWT verification not configured: %s %s from %s", r.Method, r.URL.Path, r.RemoteAddr)
-						http.Error(w, "JWT verification not configured - set authorization_url or skip_jwt_verification", http.StatusInternalServerError)
-						return
-					}
-					skipJWTWarningOnce.Do(func() {
-						klog.Warningf("JWT accepted without signature verification - set authorization_url for secure validation")
-					})
+					// We can only reach this if skip_jwt_verification is false
+					klog.V(1).Infof("Authentication rejected - JWT verification not configured: %s %s from %s", r.Method, r.URL.Path, r.RemoteAddr)
+					http.Error(w, "JWT verification not configured - set authorization_url or skip_jwt_verification", http.StatusInternalServerError)
+					return
 				} else {
 					err = claims.ValidateWithProvider(r.Context(), staticConfig.OAuthAudience, snapshot.OIDCProvider)
 				}
--- a/pkg/http/authorization.go
+++ b/pkg/http/authorization.go
@@ -96,6 +96,13 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
 
 			token := strings.TrimPrefix(authHeader, "Bearer ")
 
+			// Empty token check post-trimming
+			if token == "" {
+				klog.V(1).Infof("Authentication failed - empty bearer token: %s %s from %s", r.Method, r.URL.Path, r.RemoteAddr)
+				write401(w, wwwAuthenticateHeader, "invalid_token", "Unauthorized: Bearer token is empty")
+				return
+			}
+
 			// Token passthrough, skips all JWT processing
 			// Cluster is the sole authority for validating token (ex. sha256 token with OpenShift)
 			if staticConfig.SkipJWTVerification && staticConfig.AuthorizationURL == "" {
--- a/pkg/http/authorization.go
+++ b/pkg/http/authorization.go
@@ -96,6 +96,16 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
 
 			token := strings.TrimPrefix(authHeader, "Bearer ")
 
+			// Token passthrough, skips all JWT processing
+			if staticConfig.SkipJWTVerification && staticConfig.AuthorizationURL == "" {
+				skipJWTWarningOnce.Do(func() {
+					klog.Warningf("Bearer token forwarded without local validation (skip_jwt_verification=true and no authorization_url) - the cluster is the sole authority")
+				})
+				ctx := context.WithValue(r.Context(), internalk8s.OAuthAuthorizationHeader, authHeader)
+				next.ServeHTTP(w, r.WithContext(ctx))
+				return
+			}
+
 			claims, err := ParseJWTClaims(token)
 			if err == nil && claims == nil {
 				// Impossible case, but just in case
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -581,8 +581,8 @@ func (c *StaticConfig) validateSkipJWTVerification() error {
 		return nil
 	}
 	if c.SkipJWTVerification {
-		klog.Warningf("skip_jwt_verification is enabled: JWTs will be accepted without cryptographic signature verification. " +
-			"Only use this behind a trusted reverse proxy that performs token verification.")
+		klog.Warningf("skip_jwt_verification is enabled with no authorization_url: bearer tokens will be forwarded without any local validation. " +
+			"The cluster (or a trusted upstream) is the sole authority. Only use this when cluster_auth_mode=passthrough and the cluster validates tokens directly.")
 		return nil
 	}
 	return fmt.Errorf("require_oauth is enabled but authorization_url is not configured: " +
--- a/pkg/http/authorization_mcp_test.go
+++ b/pkg/http/authorization_mcp_test.go
@@ -513,7 +513,7 @@ func (s *AuthorizationSuite) TestAuthorizationOidcTokenExchange() {
 	}
 }
 
-func (s *AuthorizationSuite) TestAuthorizationOfflineOnlyWarning() {
+func (s *AuthorizationSuite) TestAuthorizationPassthroughWarning() {
 	// When require_oauth=true, skip_jwt_verification=true, and no OIDC provider
 	// is configured (passthrough mode), a warning log should be emitted once.
 	s.MockServer.ResetHandlers()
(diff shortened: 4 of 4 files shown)


Cali0707:
@nbottari9 can you fix the DCO on your commits? If you click into the DCO action, it has instructions on how to do so

docs/configuration.md
```diff
@@ -480,10 +480,10 @@ Configure OAuth/OIDC authentication for HTTP mode deployments.
 
 | Field | Type | Default | Description |
 |-------|------|---------|-------------|
-| `require_oauth` | boolean | `false` | When `true`, requires OAuth authentication for all requests. |
+| `require_oauth` | boolean | `false` | When `true`, requires OAuth authentication for all requests. This **DOES NOT** determine validation strategy, which is done seperately by ```authorization_url``` and ```skip_jwt_verification``` |
```
nader-ziada: typo: `seperately` should be `separately`

pkg/http/authorization.go
```diff
@@ -96,6 +96,16 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
...
 			token := strings.TrimPrefix(authHeader, "Bearer ")
 
+			// Token passthrough, skips all JWT processing
+			if staticConfig.SkipJWTVerification && staticConfig.AuthorizationURL == "" {
+				skipJWTWarningOnce.Do(func() {
+					klog.Warningf("Bearer token forwarded without local validation (skip_jwt_verification=true and no authorization_url) - the cluster is the sole authority")
+				})
+				ctx := context.WithValue(r.Context(), internalk8s.OAuthAuthorizationHeader, authHeader)
+				next.ServeHTTP(w, r.WithContext(ctx))
+				return
```
nader-ziada, line 114: this early return makes 134-136 dead code. previously `SkipJWTVerification && AuthorizationURL == ""` would flow through ParseJWTClaims → ValidateOffline and hit line 134 for the `offline-only` warning. 

now the passthrough short-circuits before any JWT processing so that path is unreachable, you'd need `AuthorizationURL == ""` (line 123 gate) AND `SkipJWTVerification == true` (line 129 gate), which is exactly what's caught here.

worth removing 134-136 and probably simplifying 128-132 since the only remaining case there is `!SkipJWTVerification`. the warning message is also stale as it says "JWT accepted without signature verification" which describes old behavior, not the new passthrough semantics.

nader-ziada (commented):
solid work @nbottari9 

added a couple of comments

manusa (changes requested):
### Strengths

- Spec compliance is high. The early return in `pkg/http/authorization.go:99-107` is placed exactly where issue #1131 §1 prescribed — after the bearer-presence check, before `ParseJWTClaims` — and the warning message is faithful to the spec.
- Config-validation warning at `pkg/config/config.go:583-586` matches issue #1131 §2 verbatim.
- New tests (`TestAuthorizationPassthroughOpaqueToken/MalformedJWT/MissingHeader/ExpiredToken/ClusterIssuedJWT/OIDCPathUnaffected`) cover all 5 scenarios called out in the issue's test plan, follow the project's `testify/suite` + nested `s.Run()` conventions, and use `atomic.Value` to assert the cluster receives the token byte-for-byte.
- OIDC regression guard (`TestAuthorizationPassthroughOIDCPathUnaffected`) correctly proves the passthrough branch does **not** fire when `authorization_url` is set.
- Embedded test JWT in `TestAuthorizationPassthroughClusterIssuedJWT` decodes to synthetic SA claims with a far-future `exp` and is correctly marked `// notsecret`.

### Issues

(Already raised by @nader-ziada: dead code at `authorization.go:134-136` and the `seperately` typo — not duplicating here.)

#### Important (Should Fix)

1. **Stale package doc comment at `pkg/http/authorization.go:42-48`.** Section "2.1 Raw Token Validation" still claims the mode performs offline expiration and audience checks — exactly what this PR removes. Update the comment to describe pure passthrough and point at one of the new `TestAuthorizationPassthrough*` tests instead of `TestAuthorizationRawToken`.

2. **`TestAuthorizationRawToken` (`pkg/http/authorization_mcp_test.go:347-371`) silently became a no-op.** It sets `SkipJWTVerification=true` with no `authorization_url` and iterates `OAuthAudience` over `""` and `"mcp-server"`. After this PR, both iterations hit the new early-return without ever touching audience/expiration validation, so the audience loop tests nothing. Either delete the loop or rewrite the test to explicitly assert passthrough semantics (e.g. opaque token + audience set + still 200, with backend receiving the token unchanged).

#### Minor (Nice to Have)

3. **Inline triple backticks** used in `docs/configuration.md:483, 534-536` (e.g. ` ```authorization_url``` `). Valid CommonMark but inconsistent with single-backtick usage elsewhere in the same table.

4. **Test names `TestAuthorizationOfflineOnlyWarning(Once)`** now describe "passthrough mode" in their bodies but retain the legacy "OfflineOnly" names. Consider renaming to `TestAuthorizationPassthroughWarning(Once)` for consistency with the new `TestAuthorizationPassthrough*` family.

5. **`TestAuthorizationPassthroughMissingHeader`** asserts only the 401 status code; the existing `TestAuthorizationUnauthorizedMissingHeader` also asserts the `WWW-Authenticate` header value. Adding the same assertion would lock in the contract for the passthrough configuration.

6. **Edge case: `Authorization: Bearer ` (prefix only, no token).** Pre-PR, `ParseJWTClaims("")` would fail and emit a 401. Post-PR, the early return forwards an empty bearer to the cluster verbatim. Not a security regression (the cluster still rejects), but a fast-fail at the MCP edge would be slightly nicer.

7. **PR description accuracy.** The body lists `TestAuthorizationPassthroughOpaqueToken/MissingHeader/OIDCPathUnaffected` as "Existing", but all six `TestAuthorizationPassthrough*` tests are new in this PR (`main` has zero of them). Description-vs-diff mismatch only.

### Assessment: Ready to merge? **With fixes**

The core change is correct and matches the spec. Combined with @nader-ziada's findings, the open items worth fixing in this PR are: dead-code removal, stale package doc comment, and the `TestAuthorizationRawToken` audience-loop that no longer asserts anything. The rest are cosmetic.

nbottari9:
@manusa Should `TestAuthorizationRawToken` just be removed? Other than that, I have resolved all of the issues listed.

manusa:
Thanks for the quick turnaround on this! Merged. A couple of test cleanups worth a follow-up if you have the cycles (or I can file an issue):

- `TestAuthorizationRawToken` can be deleted. With the passthrough early-return in place, both iterations of the audience loop hit the same code path before `OAuthAudience` is read, so the loop tests nothing. The "audience is ignored in passthrough mode" property is already covered by `TestAuthorizationPassthroughClusterIssuedJWT`.

- `TestAuthorizationPassthroughOIDCPathUnaffected` would be stronger if it sent a JWT signed by an unrelated key instead of an opaque token. The opaque token fails `ParseJWTClaims` before reaching the OIDC verifier, so the test currently passes even if the OIDC path was never exercised. A JWT that parses cleanly but fails `verifier.Verify` would actually prove the OIDC code runs.

- `TestAuthorizationPassthroughMalformedJWT` could use the same backend-reception assertion as the other passthrough tests (atomic.Value capture, assert backend got `Bearer not.a.jwt.at.all` verbatim). The comment already claims "ParseJWTClaims must not be reached" but doesn't actually observe that property.

Issue #1164, opened 2026-05-25
resources_create_or_update: expose fieldManager as optional parameter to prevent SSA field abandonment across sequential tool calls


SaravgiYash:
### Problem

When two separate tool calls use **resources_create_or_update** in sequence, SSA field abandonment silently deletes fields from the cluster.

**resources_create_or_update** hardcodes a single field manager for all calls:

```
Apply(ctx, obj.GetName(), obj, metav1.ApplyOptions{
    FieldManager: version.BinaryName,  // always "kubernetes-mcp-server"
    Force:        true,
})
```

SSA stores field ownership permanently in metadata.managedFields keyed by manager name. When a second call from the same manager omits fields that a previous call owned, SSA interprets this as intentional removal and deletes those fields.

### Reproduction

- Apply a deployment patch that sets resources.limits.memory and resources.requests.memory
- Apply a separate patch (e.g. add a restartedAt annotation) without including resources
- Both use resources_create_or_update → same manager → SSA abandons resources.memory → fields deleted
- Verified on a live cluster — managedFields after the sequence shows kubernetes-mcp-server owns only image and name, with no resources.* fields. Memory limits and requests are absent from the deployed spec.

### Fix

Add fieldManager as an optional input parameter:

```
"fieldManager": {
    Type:        "string",
    Description: "Optional field manager name for SSA ownership. Defaults to kubernetes-mcp-server.",
},

fm := version.BinaryName
if val, ok := params.GetArguments()["fieldManager"]; ok && val != "" {
    fm = val.(string)
}

Apply(ctx, obj.GetName(), obj, metav1.ApplyOptions{
    FieldManager: fm,
    Force:        true,
})
```

This lets callers use distinct manager names per logical operation, so SSA ownership is scoped correctly and fields are never abandoned across unrelated tool calls.

lunarwhite:
Raised a fixing PR: https://github.com/containers/kubernetes-mcp-server/pull/1184

manusa:
Thanks for the detailed report and the reproduction. The mechanism you describe is real, but I don't think it's a bug in the server, and I don't think a `fieldManager` parameter is the right way to fix it.

[...]

PR #1166, opened 2026-05-25
eval(core): check if ssa removes fields when model makes update

Cali0707:
As part of #1164 I wanted to see what models actually do when asked to update resources through our eval framework

This eval creates a resource that is already "owner" by the mcp server and then asks it to update the resource, to see if the model properly re-fetches the resource before updating. Locally with opus 4.6 this passed, worth testing in other scenarios too

### DIFF
--- a/evals/tasks/core/ssa-field-preservation/ssa-field-preservation.yaml
+++ b/evals/tasks/core/ssa-field-preservation/ssa-field-preservation.yaml
@@ -1,0 +1,117 @@
+kind: Task
+apiVersion: mcpchecker/v1alpha2
+metadata:
+  name: ssa-field-preservation
+  difficulty: medium
+  labels:
+    suite: core
+spec:
+  requires:
+    - extension: kubernetes
+
+  setup:
+    - kubernetes.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: ssa-test
+        ignoreNotFound: true
+
+    - kubernetes.create:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: ssa-test
+
+    # Apply a deployment with resource limits using SSA with the same field manager
+    # that resources_create_or_update uses, simulating a prior MCP tool call.
+    # TODO: Replace with kubernetes.apply once mcpchecker/kubernetes-extension#55 lands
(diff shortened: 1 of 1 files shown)


evals/mcpchecker-kubernetes-basic-operations-out.json
manusa, line 1: Was this file committed by accident?

Cali0707: oh yeah this was a mistake, will remove

evals/tasks/core/ssa-field-preservation/ssa-field-preservation.yaml
```diff
@@ -0,0 +1,117 @@
+kind: Task
+apiVersion: mcpchecker/v1alpha2
```
matzew, line 2: should we bump the rest of those too? Not here - but in a later PR? 

the mixing might be confusing?

evals/tasks/core/ssa-field-preservation/ssa-field-preservation.yaml
```diff
@@ -0,0 +1,117 @@
...
+          MANIFEST
+
+    - kubernetes.wait:
+        apiVersion: apps/v1
+        kind: Deployment
+        metadata:
+          name: payment-service
+          namespace: ssa-test
+        condition: Available
+        timeout: 120s
```
matzew: Is there a need for this explicit timeout value? E.g. over default of 60s?

matzew (approved):
LGTM

evals/tasks/core/ssa-field-preservation/ssa-field-preservation.yaml
Cali0707: Nope, just claude 😅

evals/tasks/core/ssa-field-preservation/ssa-field-preservation.yaml
Cali0707: Yes, opened https://github.com/containers/kubernetes-mcp-server/issues/1168 to track

manusa (approved):
LGTM, thx!

Issue #1157, opened 2026-05-18
Resources API: align with MCP spec and improve handler ergonomics


manusa:
The MCP resources infrastructure landed via #1091, #1114, and #1130, but our `api.Resource`, `api.ResourceTemplate`, and `api.ResourceHandler` types are minimal compared to what the spec offers and what toolset implementors will need once we (or downstream) start exposing real resources. No toolset currently registers any resources or templates, so this is a good moment to evolve the API without paying a migration cost.

## 1. Add `Annotations` to `Resource` and `ResourceTemplate`

The [MCP spec (2025-11-25)](https://modelcontextprotocol.io/specification/2025-11-25/server/resources#annotations) defines an `annotations` object with three fields that clients are expected to act on:

- `audience` (`["user"]`, `["assistant"]`, or both): clients filter resources by intended audience.
- `priority` (0.0 to 1.0): clients use it to rank context inclusion.
- `lastModified` (ISO 8601): clients sort or display by recency.

Today our `api.Resource` and `api.ResourceTemplate` (`pkg/api/toolsets.go:113-148`) omit all three, so a toolset has no way to express things like "this resource is assistant-only" or hint priority. The fix is to add an `Annotations *ResourceAnnotations` field on both types and wire it through `ServerResourceToGoSdkResource` and `ServerResourceTemplateToGoSdkResourceTemplate` in `pkg/mcp/resources_gosdk.go`.

## 2. Add `Title` to `Resource` and `ResourceTemplate`

The spec defines `title` as the human-readable display name (distinct from the programmatic `name`). It's the same pattern we already use for tools via `ToolAnnotations.Title`. Without it, clients fall back to `name` in resource pickers. Pure additive.

Out of scope for now: `size`, `icons`, and the `subscribe` capability. We can pick those up when a real resource asks for them.

## 3. Change `ResourceHandler` signature to take a params struct

Today's handlers look like:

```go
type ResourceHandler func(ctx context.Context) (*ResourceContent, error)
type ResourceTemplateHandler func(ctx context.Context, uri string) (*ResourceContent, error)
```

The problem is that `sessionInjectionMiddleware` (`pkg/mcp/middleware.go:130`) only puts the MCP session on the context. There's no in-API way for a handler to obtain a `KubernetesClient` or the live `BaseConfig`. Tools sidestep this because `tools_gosdk.go:64-77` resolves them in the wrapper closure and passes them via `api.ToolHandlerParams`. Resources need the same treatment.

[...]

PR #1169, opened 2026-05-27
feat: Resources API: Align with spec and simplify handler ergonomics

nbottari9:
## Summary

This PR aligns the Resources API with the MCP specification and improves handler ergonomics by:

1. **Adding Annotations support** (Audience, Priority, LastModified) to resources and resource templates
2. **Adding Title field** to resources and resource templates  
3. **Introducing ResourceHandlerParams struct** to provide a consistent, extensible handler signature

These changes bring the Resources API to feature parity with the Tools API while maintaining backward compatibility (no resources are currently registered in the codebase).

---

### 1. Added Annotations to Resource and ResourceTemplate

**Files modified:**
- `pkg/api/toolsets.go` (lines 113-175)
- `pkg/mcp/resources_gosdk.go` (lines 22-32, 69-79, 124-147)

**What changed:**
- Added `Audience []string`, `Priority *float64`, and `LastModified *string` fields to `api.Resource` and `api.ResourceTemplate`
- Created `buildAnnotations()` helper function that converts internal annotation fields to MCP SDK `mcp.Annotations` type
- Wired annotations through `ServerResourceToGoSdkResource()` and `ServerResourceTemplateToGoSdkResourceTemplate()`

**Why:**
The MCP specification defines three optional annotation fields for resources:
- `audience`: Controls visibility to specific roles (e.g., `["user"]`, `["assistant"]`, `["user", "assistant"]`)
- `priority`: Float 0.0-1.0 indicating relative importance for ranking/filtering
- `lastModified`: ISO 8601 timestamp for cache invalidation

These annotations enable MCP clients to make smarter decisions about which resources to surface to users or LLMs.

**Implementation details:**
- Used pointer types (`*float64`, `*string`) to distinguish "not set" (nil) from "explicitly zero/empty"
- `buildAnnotations()` returns `nil` when all fields are empty, ensuring proper `omitempty` behavior in JSON serialization
- Audience values are converted from `[]string` to `[]mcp.Role` to match SDK types

### 2. Added Title to Resource and ResourceTemplate

**Files modified:**
- `pkg/api/toolsets.go` (lines 113-175)
- `pkg/mcp/resources_gosdk.go` (lines 26, 73)

**What changed:**
- Added `Title string` field to `api.Resource` and `api.ResourceTemplate`
- Wired Title through `ServerResourceToGoSdkResource()` and `ServerResourceTemplateToGoSdkResourceTemplate()`

**Why:**
The MCP specification includes a `title` field as the human-readable display name, distinct from the programmatic `name` field. This follows the same pattern already used for tools via `ToolAnnotations.Title`.

Without Title, MCP clients fall back to displaying the programmatic `name` in resource pickers, which is often not user-friendly (e.g., `"cluster_metrics"` instead of `"Cluster Performance Metrics"`).

**Implementation details:**
- Title is optional (empty string means not set)
- When Title is empty, MCP clients automatically fall back to displaying `Name` (client-side behavior)
- Simple pass-through conversion - no validation needed

### 3. Changed ResourceHandler Signatures to Use ResourceHandlerParams

**Files modified:**
- `pkg/api/toolsets.go` (lines 138-145, 169)
- `pkg/mcp/resources_gosdk.go` (lines 36, 82)

**What changed:**
- Introduced `ResourceHandlerParams` struct with embedded `context.Context`, `BaseConfig`, `KubernetesClient`, and `URI string`
- Changed `ResourceHandler` signature from `func(context.Context) (...)` to `func(params ResourceHandlerParams) (...)`
- Changed `ResourceTemplateHandler` signature from `func(context.Context, string) (...)` to `func(params ResourceHandlerParams) (...)`
- Updated all handler call sites to use `api.ResourceHandlerParams{Context: ctx}` or `api.ResourceHandlerParams{Context: ctx, URI: req.Params.URI}`

**Why:**
The previous handler signatures were inflexible - adding new parameters (like BaseConfig or KubernetesClient) would require breaking API changes. By introducing a params struct (matching the pattern used for `ToolHandlerParams`), we can add new fields in the future without breaking existing handlers.

This also provides a more consistent API surface - tools and resources now follow the same parameter pattern.

**Current limitations:**
- **Context**: ✅ Populated from MCP request
- **URI**: ✅ Populated for resource templates (matches the actual URI requested)
- **BaseConfig**: ❌ Not yet populated (future enhancement - requires using Server parameter)
- **KubernetesClient**: ❌ Fundamentally cannot be populated (MCP protocol has no cluster targeting for resources)

The params struct is defined now to establish the API contract, even though not all fields are populated yet. This allows handlers to be written with the expectation that BaseConfig will be available in a future version without requiring signature changes.

---

### New Test Suite

**File added:**
- `pkg/mcp/resources_annotations_test.go` (519 lines)

Comprehensive test suite covering:
- All annotation field combinations (all set, individual fields, none set)
- Nil vs empty values for pointer fields (Priority, LastModified)
- Boundary conditions (Priority 0.0 and 1.0, single/multiple audience values)
- Persistence across configuration reload
- Proper `omitempty` behavior (nil annotations not serialized)

**Test cases:**
1. `TestResourceAnnotationsAllFieldsSet` - verifies all three annotation fields
2. `TestResourceAnnotationsAudienceOnly` - verifies single field (audience)
3. `TestResourceAnnotationsPriorityOnly` - verifies single field (priority)
4. `TestResourceAnnotationsLastModifiedOnly` - verifies single field (lastModified)
5. `TestResourceAnnotationsNoneSet` - verifies nil Annotations when all empty
6. `TestResourceTemplateSameAnnotations` - verifies templates work identically to resources
7. `TestResourceAnnotationsAudienceEmptySlice` - edge case: empty slice vs nil
8. `TestResourceAnnotationsPriorityZero` - edge case: &0.0 vs nil
9. `TestResourceAnnotationsLastModifiedEmptyString` - edge case: &"" vs nil
10. `TestResourceAnnotationsPriorityBoundary` - boundary: 0.0 and 1.0 are valid
11. `TestResourceAnnotationsMultipleAudiences` - multiple audience values
12. `TestResourceAnnotationsSurvivesReload` - persistence across config reload

### Extended Existing Tests

**File modified:**
- `pkg/mcp/resources_gosdk_test.go`

Extended existing resource conversion tests to verify Title field:
- Added Title to test resource definitions
- Added assertions checking Title values in MCP resources
- Verified Title appears correctly in list responses

---

### Positive Impacts

1. **MCP Spec Compliance**: Resources now support all current MCP annotation fields (audience, priority, lastModified) and title, bringing this implementation into full compliance with the specification.

2. **Better UX**: MCP clients can now display user-friendly titles in resource pickers instead of programmatic names.

3. **Smarter Resource Selection**: Annotations enable clients to filter and rank resources appropriately:
   - Show only user-facing resources in UI pickers (`audience: ["user"]`)
   - Prioritize important resources (`priority: 0.9`)
   - Cache resources efficiently (`lastModified`)

4. **Future-Proof Handler API**: The ResourceHandlerParams struct allows adding new parameters (like BaseConfig) without breaking existing handlers.

5. **API Consistency**: Resources now follow the same parameter pattern as tools (both use params structs).

### Breaking Changes

**None.** All changes are backward compatible:
- New fields are optional (nil/empty is valid)
- No existing resources are registered in the codebase yet (per issue #1157)
- Handler signature change is compatible (params struct can be extended)

### Known Limitations

1. **BaseConfig not populated**: The params struct includes BaseConfig but it's not yet populated in handler calls. This is a future enhancement (requires changing `_ *Server` to `s *Server` in conversion functions).

2. **KubernetesClient fundamentally limited**: Unlike tools (which receive cluster targeting in request parameters), MCP resources have no cluster targeting mechanism. The `ReadResourceRequest` only contains a URI, so there's no way to know which cluster client to derive. This field will remain nil unless the MCP protocol adds cluster targeting support.

---

## Migration Guide

**For toolset authors** (when adding new resources):

Before:
```go
api.ServerResource{
    Resource: api.Resource{
        URI:         "k8s://cluster/status",
        Name:        "cluster_status",
        Description: "Cluster health status",
        MIMEType:    "application/json",
    },
    Handler: func(ctx context.Context) (*api.ResourceContent, error) {
        // ...
    },
}
```

After:
```go
api.ServerResource{
    Resource: api.Resource{
        URI:         "k8s://cluster/status",
        Name:        "cluster_status",
        Title:       "Cluster Health Status",  // Human-friendly!
        Description: "Real-time cluster health and status",
        MIMEType:    "application/json",
        Audience:    []string{"user"},  // Show in UI pickers
        Priority:    ptr(0.8),          // High priority
        LastModified: ptr("2026-05-27T10:00:00Z"),
    },
    Handler: func(params api.ResourceHandlerParams) (*api.ResourceContent, error) {
        ctx := params.Context  // Access context from params
        // BaseConfig and KubernetesClient not yet available
        // ...
    },
}
```

Helper for pointer values:
```go
func ptr[T any](v T) *T { return &v }
```

---

## Related Issues

- Refs #1157 items 1, 2, 3
- Future work: #1157 item 4 (populate BaseConfig in ResourceHandlerParams)
- Future work: Document KubernetesClient limitation or remove from struct

---

## Checklist

- ✅ Code follows project conventions
- ✅ All tests pass (`make test`)
- ✅ Linting passes (`make lint`)
- ✅ Comprehensive test coverage (12 test cases for annotations, Title tested in existing suite)
- ✅ No breaking changes
- ✅ Documentation updated (godoc comments on new fields)

### DIFF
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -135,7 +135,14 @@ type ResourceContent struct {
 // ResourceHandler is called when a client reads a resource.
 // Session state (auth, request context) is available on ctx via sessionInjectionMiddleware.
 // Handlers should return a ResourceContent with exactly one of Text or Blob set.
-type ResourceHandler func(ctx context.Context) (*ResourceContent, error)
+type ResourceHandler func(params ResourceHandlerParams) (*ResourceContent, error)
+
+type ResourceHandlerParams struct {
+	context.Context
+	BaseConfig
+	KubernetesClient
+	URI string
+}
 
 // ServerResource represents a resource that can be registered with the MCP server.
 type ServerResource struct {
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -118,6 +118,7 @@ type Resource struct {
 	Audience     []string // ["user", "assistant"]
 	Priority     *float64
 	LastModified *string
+	Title        string `json:"title,omitempty"`
 }
 
 // ResourceContent is the value returned by a resource handler.
@@ -151,6 +152,7 @@ type ResourceTemplate struct {
 	Audience     []string // ["user", "assistant"]
 	Priority     *float64
 	LastModified *string
+	Title        string `json:"title,omitempty"`
 }
 
 // ResourceTemplateHandler is called when a client reads a resource matching a template.
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -109,16 +109,20 @@ func NewToolCallResultStructured(structured any, err error) *ToolCallResult {
 	return NewToolCallResultFull(content, structured, err)
 }
 
-// Resource represents the metadata of an MCP resource.
-type Resource struct {
-	URI          string
-	Name         string
-	Description  string
-	MIMEType     string
+type ResourceAnnotations struct {
 	Audience     []string // ["user", "assistant"]
 	Priority     *float64
 	LastModified *string
-	Title        string
+}
+
+// Resource represents the metadata of an MCP resource.
+type Resource struct {
+	URI         string
+	Name        string
+	Description string
+	MIMEType    string
+	Annotations *ResourceAnnotations
+	Title       string
 }
 
 // ResourceContent is the value returned by a resource handler.
@@ -153,14 +157,12 @@ type ServerResource struct {
 
 // ResourceTemplate represents the metadata of an MCP resource template.
 type ResourceTemplate struct {
-	URITemplate  string
-	Name         string
-	Description  string
-	MIMEType     string
-	Audience     []string // ["user", "assistant"]
-	Priority     *float64
-	LastModified *string
-	Title        string
+	URITemplate string
+	Name        string
+	Description string
+	MIMEType    string
+	Annotations *ResourceAnnotations
+	Title       string
 }
 
 // ResourceTemplateHandler is called when a client reads a resource matching a template.
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -118,7 +118,7 @@ type Resource struct {
 	Audience     []string // ["user", "assistant"]
 	Priority     *float64
 	LastModified *string
-	Title        string `json:"title,omitempty"`
+	Title        string
 }
 
 // ResourceContent is the value returned by a resource handler.
@@ -133,10 +133,11 @@ type ResourceContent struct {
 }
 
 // ResourceHandler is called when a client reads a resource.
-// Session state (auth, request context) is available on ctx via sessionInjectionMiddleware.
+// Session state (auth, request context), BaseConfig, and KubeClient are available through params
 // Handlers should return a ResourceContent with exactly one of Text or Blob set.
 type ResourceHandler func(params ResourceHandlerParams) (*ResourceContent, error)
 
+// ResourceHandlerParams represents data passed into ResourceHandler
 type ResourceHandlerParams struct {
 	context.Context
 	BaseConfig
@@ -159,12 +160,12 @@ type ResourceTemplate struct {
 	Audience     []string // ["user", "assistant"]
 	Priority     *float64
 	LastModified *string
-	Title        string `json:"title,omitempty"`
+	Title        string
 }
 
 // ResourceTemplateHandler is called when a client reads a resource matching a template.
-// Session state (auth, request context) is available on ctx via sessionInjectionMiddleware.
-// The uri parameter is the actual resource URI that matches the template.
+// Session state (auth, request context), BaseConfig, and KubeClient are available through params
+// Params also includes the actual resource URI that matches the template.
 // Handlers should return a ResourceContent with exactly one of Text or Blob set.
 type ResourceTemplateHandler func(params ResourceHandlerParams) (*ResourceContent, error)
 
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -166,7 +166,7 @@ type ResourceTemplate struct {
 // Session state (auth, request context) is available on ctx via sessionInjectionMiddleware.
 // The uri parameter is the actual resource URI that matches the template.
 // Handlers should return a ResourceContent with exactly one of Text or Blob set.
-type ResourceTemplateHandler func(ctx context.Context, uri string) (*ResourceContent, error)
+type ResourceTemplateHandler func(params ResourceHandlerParams) (*ResourceContent, error)
 
 // ServerResourceTemplate represents a resource template that can be registered with the MCP server.
 type ServerResourceTemplate struct {
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -111,10 +111,13 @@ func NewToolCallResultStructured(structured any, err error) *ToolCallResult {
 
 // Resource represents the metadata of an MCP resource.
 type Resource struct {
-	URI         string
-	Name        string
-	Description string
-	MIMEType    string
+	URI          string
+	Name         string
+	Description  string
+	MIMEType     string
+	Audience     []string // ["user", "assistant"]
+	Priority     *float64
+	LastModified *string
 }
 
 // ResourceContent is the value returned by a resource handler.
@@ -141,10 +144,13 @@ type ServerResource struct {
 
 // ResourceTemplate represents the metadata of an MCP resource template.
 type ResourceTemplate struct {
-	URITemplate string
-	Name        string
-	Description string
-	MIMEType    string
+	URITemplate  string
+	Name         string
+	Description  string
+	MIMEType     string
+	Audience     []string // ["user", "assistant"]
+	Priority     *float64
+	LastModified *string
 }
 
 // ResourceTemplateHandler is called when a client reads a resource matching a template.
--- a/pkg/mcp/resources_annotations_test.go
+++ b/pkg/mcp/resources_annotations_test.go
@@ -36,13 +36,15 @@ func (s *ResourceAnnotationsSuite) TestResourceAnnotationsAllFieldsSet() {
 		resources: []api.ServerResource{
 			{
 				Resource: api.Resource{
-					URI:          "test://example/annotated",
-					Name:         "Annotated Resource",
-					Description:  "Has all annotation fields",
-					MIMEType:     "text/plain",
-					Audience:     []string{"user", "assistant"},
-					Priority:     &priority,
-					LastModified: &lastModified,
+					URI:         "test://example/annotated",
+					Name:        "Annotated Resource",
+					Description: "Has all annotation fields",
+					MIMEType:    "text/plain",
+					Annotations: &api.ResourceAnnotations{
+						Audience:     []string{"user", "assistant"},
+						Priority:     &priority,
+						LastModified: &lastModified,
+					},
 				},
 				Handler: func(_ api.ResourceHandlerParams) (*api.ResourceContent, error) {
 					return &api.ResourceContent{Text: "content"}, nil
@@ -79,13 +81,15 @@ func (s *ResourceAnnotationsSuite) TestResourceTemplateAnnotationsAllFieldsSet()
 		resourceTemplates: []api.ServerResourceTemplate{
 			{
 				ResourceTemplate: api.ResourceTemplate{
-					URITemplate:  "test://example/{id}",
-					Name:         "Annotated Template",
-					Description:  "Has all annotation fields",
-					MIMEType:     "application/json",
-					Audience:     []string{"assistant"},
-					Priority:     &priority,
-					LastModified: &lastModified,
+					URITemplate: "test://example/{id}",
+					Name:        "Annotated Template",
+					Description: "Has all annotation fields",
+					MIMEType:    "application/json",
+					Annotations: &api.ResourceAnnotations{
+						Audience:     []string{"assistant"},
+						Priority:     &priority,
+						LastModified: &lastModified,
+					},
 				},
 				Handler: func(params api.ResourceHandlerParams) (*api.ResourceContent, error) {
 					return &api.ResourceContent{Text: `{"uri": "` + params.URI + `"}`}, nil
@@ -184,12 +188,14 @@ func (s *ResourceAnnotationsSuite) TestResourceAnnotationsOnlyAudience() {
 		resources: []api.ServerResource{
 			{
 				Resource: api.Resource{
-					URI:          "test://example/only-audience",
-					Name:         "Only Audience",
-					MIMEType:     "text/plain",
-					Audience:     []string{"user"},
-					Priority:     nil,
-					LastModified: nil,
+					URI:      "test://example/only-audience",
+					Name:     "Only Audience",
+					MIMEType: "text/plain",
+					Annotations: &api.ResourceAnnotations{
+						Audience:     []string{"user"},
+						Priority:     nil,
+						LastModified: nil,
+					},
 				},
 				Handler: func(_ api.ResourceHandlerParams) (*api.ResourceContent, error) {
 					return &api.ResourceContent{Text: "content"}, nil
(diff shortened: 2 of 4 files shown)


pkg/mcp/resources_gosdk.go
```diff
@@ -24,9 +24,16 @@ func ServerResourceToGoSdkResource(_ *Server, res api.ServerResource) (*mcp.Reso
...
+			res.Resource.Audience,
+			res.Resource.Priority,
+			res.Resource.LastModified,
+		),
+		Title: res.Resource.Title,
 	}
+
 	handler := func(ctx context.Context, _ *mcp.ReadResourceRequest) (*mcp.ReadResourceResult, error) {
-		content, err := res.Handler(ctx)
+		content, err := res.Handler(api.ResourceHandlerParams{Context: ctx})
```
nader-ziada: `BaseConfig` and `KubernetesClient` are not populated here (or at line 82 for templates). The issue spec calls for resolving these in the wrapper closure the same way tools_gosdk.go:70-77 does. Without this, any handler accessing these fields will nil-panic. The _ *Server parameter needs to be used to obtain config and k8s client.

pkg/api/toolsets.go
```diff
@@ -131,7 +135,14 @@ type ResourceContent struct {
 // ResourceHandler is called when a client reads a resource.
 // Session state (auth, request context) is available on ctx via sessionInjectionMiddleware.
 // Handlers should return a ResourceContent with exactly one of Text or Blob set.
-type ResourceHandler func(ctx context.Context) (*ResourceContent, error)
+type ResourceHandler func(params ResourceHandlerParams) (*ResourceContent, error)
```
nader-ziada, line 142: Stale comment, references "ctx" but the parameter is now params with an embedded context.Context.

pkg/api/toolsets.go
```diff
@@ -131,7 +135,14 @@ type ResourceContent struct {
...
+
+type ResourceHandlerParams struct {
```
nader-ziada, line 145: Missing godoc on exported ResourceHandlerParams type. Every other exported type in this file has one.

pkg/api/toolsets.go
```diff
@@ -111,10 +111,14 @@ func NewToolCallResultStructured(structured any, err error) *ToolCallResult {
...
-	Description string
-	MIMEType    string
+	URI          string
+	Name         string
+	Description  string
+	MIMEType     string
+	Audience     []string // ["user", "assistant"]
+	Priority     *float64
+	LastModified *string
+	Title        string `json:"title,omitempty"`
```
nader-ziada: Title has json:"title,omitempty" but no other field in Resource or ResourceTemplate has a JSON tag. These structs aren't JSON-serialized directly, they're mapped manually to SDK types. Remove the tag or add tags consistently. Same on line 162.

pkg/mcp/resources_annotations_test.go
```diff
@@ -0,0 +1,518 @@
...
+		s.Require().NoError(err)
+		s.Require().Len(result.Resources, 1)
+		s.Require().NotNil(result.Resources[0].Annotations)
+		s.Equal(0.6, result.Resources[0].Annotations.Priority)
+	})
+
+	s.Run("annotations persist after reload", func() {
+		newConfig := config.Default()
+		newConfig.Toolsets = []string{"resource-test"}
+		newConfig.KubeConfig = s.Cfg.KubeConfig
```
nader-ziada, line 520: Reload test should inherit ReadOnly from s.Cfg per project testing guidelines. 
Add newConfig.ReadOnly = s.Cfg.ReadOnly after setting KubeConfig.

pkg/api/toolsets.go
```diff
@@ -141,17 +152,21 @@ type ServerResource struct {
...
 }
 
 // ResourceTemplateHandler is called when a client reads a resource matching a template.
 // Session state (auth, request context) is available on ctx via sessionInjectionMiddleware.
 // The uri parameter is the actual resource URI that matches the template.
 // Handlers should return a ResourceContent with exactly one of Text or Blob set.
-type ResourceTemplateHandler func(ctx context.Context, uri string) (*ResourceContent, error)
+type ResourceTemplateHandler func(params ResourceHandlerParams) (*ResourceContent, error)
```
nader-ziada, line 172: Stale comment, "The uri parameter" no longer exists. It's now params.URI.

nader-ziada (changes requested):
added some small suggestions

pkg/api/toolsets.go
```diff
@@ -111,10 +111,14 @@ func NewToolCallResultStructured(structured any, err error) *ToolCallResult {
...
-	Name        string
-	Description string
-	MIMEType    string
+	URI          string
+	Name         string
+	Description  string
+	MIMEType     string
+	Audience     []string // ["user", "assistant"]
+	Priority     *float64
+	LastModified *string
```
manusa, line 115: The issue proposes to add these fields nested in Annotations (ResourceAnnotations) to mirror the MCP spec, but here they are flattened.

Is there any reason for that?

pkg/api/toolsets.go
```diff
@@ -141,17 +152,21 @@ type ServerResource struct {
...
-	Name        string
-	Description string
-	MIMEType    string
+	URITemplate  string
+	Name         string
+	Description  string
+	MIMEType     string
+	Audience     []string // ["user", "assistant"]
+	Priority     *float64
+	LastModified *string
```
manusa: The issue proposes to add these fields nested in Annotations (ResourceAnnotations) to mirror the MCP spec, but here they are flattened.

Is there any reason for that?

pkg/mcp/resources_gosdk.go
manusa: Separate from the population question above: `params.URI` is left empty for static resources, while templates get `req.Params.URI`. The issue says "A single handler type covers both static resources and templates (`URI` is populated from the read request)". `res.Resource.URI` is already in scope (it's used in the `ReadResourceResult` below), so this can be `api.ResourceHandlerParams{Context: ctx, URI: res.Resource.URI}` — otherwise a handler shared across both kinds sees a populated URI in one path and `""` in the other.

pkg/mcp/resources_annotations_test.go
```diff
@@ -0,0 +1,518 @@
...
+	s.InitMcpClient()
+
+	s.Run("zero priority is preserved and annotations are not nil", func() {
+		result, err := s.ListResources()
+		s.Require().NoError(err)
+		s.Require().Len(result.Resources, 1)
+
+		resource := result.Resources[0]
+		s.Require().NotNil(resource.Annotations, "annotations should not be nil even for zero priority")
+		s.Equal(0.0, resource.Annotations.Priority, "zero priority should be preserved")
```
manusa, line 341: This assertion passes vacuously: the SDK's `Annotations.Priority` is a non-pointer `float64` with `json:"priority,omitempty"`, so an explicit `0.0` is dropped on the wire and the `0.0` observed here is just the client-side Go zero value — the test can't distinguish "preserved" from "dropped-and-defaulted". Only the `NotNil(resource.Annotations)` check verifies real behavior (the server emits `"annotations":{}`). Same applies to the `minPriority` half of `TestResourceAnnotationsPriorityBoundaries` below. Worth reframing these to assert what's actually observable — and the `*float64` godoc/description shouldn't claim the not-set-vs-zero distinction survives serialization, because it can't.

pkg/mcp/resources_annotations_test.go
```diff
@@ -0,0 +1,518 @@
...
+	priority := 0.7
+
+	testToolset := &mockResourceToolset{
+		resources: []api.ServerResource{
+			{
+				Resource: api.Resource{
+					URI:      "test://example/multi-audience",
+					Name:     "Multiple Audience",
+					MIMEType: "text/plain",
+					Audience: []string{"user", "assistant", "system"},
```
manusa: `"system"` isn't a valid MCP audience role — the spec (and the field's own comment in `api/toolsets.go`) defines only `user` and `assistant`. This test codifies pass-through of an out-of-spec value as the expected contract, so if `buildAnnotations` ever starts validating roles, this test will read as a regression. Suggest proving ordering with valid roles instead (e.g. `["assistant", "user"]`) — and deciding explicitly whether the conversion should validate/filter roles.

manusa (changes requested):
Thanks — the annotations/Title wiring and the handler-signature migration are clean, and the empty→nil annotation behavior is well covered. Extending the existing review comments, three things block this for me:

1. **Item 3 of #1157 is only half-done.** `ResourceHandlerParams` ships with `BaseConfig`/`KubernetesClient` embedded but never populated — the issue's core motivation. The "fundamentally cannot be populated" rationale doesn't hold (the tools' default-target fallback applies; see inline). Also, the description defers this to "#1157 item 4", but the issue has only 3 items — as written, "Closes #1157" would auto-close the issue with the motivating capability missing; please either wire the fields or change to "Refs #1157".
2. **Contract gaps:** static-resource handlers get `URI == ""`; audience values are unvalidated and one test asserts the spec-invalid role `"system"` round-trips.
3. **Test/description accuracy:** the description's Testing section lists test names that don't exist (9 of 12), including a claimed `LastModified: &""` edge case that is genuinely untested; the zero-priority "preservation" assertions pass vacuously (the SDK drops `priority: 0` via `omitempty`); and the new 518-line suite duplicates `ResourceSuite` scaffolding that a table-driven test in the existing file would collapse.

Details inline.

pkg/mcp/resources_gosdk.go
manusa: Two data points to add here:

1. The PR description argues `KubernetesClient` "fundamentally cannot be populated" because resources carry no cluster targeting — but tools already handle the no-argument case via the default-target fallback (`tools_gosdk.go:64-65`: `GetString(s.p.GetTargetParameterName(), s.p.GetDefaultTarget())` → `s.p.GetDerivedKubernetes`). This wrapper can do the same with `s.p.GetDerivedKubernetes(ctx, s.p.GetDefaultTarget())`, plus `cfg := s.configuration.Load()` for `BaseConfig`.
2. The nil-panic is worse than a failed request: go-sdk v1.6.1 doesn't recover handler panics (its jsonrpc2 layer runs handlers in a bare goroutine), and our only `recover()` (`middleware.go:200`) doesn't wrap the handler chain — so the first handler that touches these fields takes down the whole process, all sessions in HTTP mode.

If populating stays deferred, the two embeds should be dropped from `ResourceHandlerParams` until wired.

pkg/api/toolsets.go
manusa: +1 — and given the current call sites, the godoc should also state which fields are actually populated (Context always; URI for templates only; BaseConfig/KubernetesClient currently never), so handler authors don't assume tool-params parity.

pkg/mcp/resources_annotations_test.go
```diff
@@ -0,0 +1,518 @@
...
+	"testing"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	"github.com/containers/kubernetes-mcp-server/pkg/config"
+	"github.com/containers/kubernetes-mcp-server/pkg/toolsets"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"github.com/stretchr/testify/suite"
+)
+
+type ResourceAnnotationsSuite struct {
```
manusa, line 13: This suite duplicates `ResourceSuite`'s scaffolding verbatim (`SetupTest`/`TearDownTest`/`originalToolsets`, same `mockResourceToolset`, same `"resource-test"` toolset name). There's no need for a second suite: define these tests as methods on `ResourceSuite` instead, and keep them in this separate file if you prefer — the project already does this elsewhere, e.g. `PromptsTestSuite` is declared in `pkg/prompts/prompts_test.go` and has additional test methods in `pkg/prompts/prompts_merge_test.go`. That removes the duplicated scaffolding and the extra `suite.Run` entry point entirely.

For the per-test repetition (each of the 12 tests repeats the same `toolsets.Clear()` / `toolsets.Register(...)` / `s.Cfg.Toolsets = ...` / `s.InitMcpClient()` block around the single `api.Resource` literal that varies), follow the parameterized suite-helper pattern from `ElicitationSuite.registerElicitingToolset(handler)` (`elicit_test.go:36-59`) — e.g. a `registerResources(resources ...api.ServerResource)` helper on `ResourceSuite` that both this file's tests and the existing ones can use.

Separately: none of the 12 tests ever reads a resource, so the template handlers written here against `params.URI` are never invoked.

Cali0707:
@nbottari9 mind looking into the failing tests here? It seems related to the error messages you added/changed...

PR #1170, opened 2026-05-27
fix(ci): move to using core instead of kubernetes suite

Cali0707:
I had a mcpchecker run fail when I passed in kubernetes in `/run-mcpchecker`, it is because the suite label is actually `core`.

This works in the periodic jobs since those use the eval.yaml default instead of the override passed into the action step.

This PR fixes everything to align to the toolset/suite name "core" and also cleans up the action a little bit

### DIFF
--- a/.github/workflows/mcpchecker.yaml
+++ b/.github/workflows/mcpchecker.yaml
@@ -13,12 +13,12 @@ on:
   workflow_dispatch:
     inputs:
       suite:
-        description: 'Which task suite to run (kubernetes, kubevirt, kiali, tekton, or all)'
+        description: 'Which task suite to run (core, helm, kubevirt, kiali, tekton, or all)'
         required: false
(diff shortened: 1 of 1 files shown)


.github/workflows/mcpchecker.yaml
```diff
@@ -13,12 +13,12 @@ on:
...
       suite:
-        description: 'Which task suite to run (kubernetes, kubevirt, kiali, tekton, or all)'
+        description: 'Which task suite to run (core, helm, kubevirt, kiali, tekton, or all)'
         required: false
-        default: 'kubernetes'
+        default: 'core'
         type: choice
         options:
-          - kubernetes
+          - core
```
matzew, line 21: so... it would always have failed?

matzew (approved):
LGTM

PR #1172, opened 2026-05-28
ci(workflows): track Go version from go.mod and gate directive bumps

manusa:
Supersedes #1124.

Dependabot's #1124 bumped k8s to 1.36 which requires `go 1.26.0` in go.mod, but the build workflows were pinned to `GO_VERSION: 1.25`. CI was installing Go 1.25.10 and bailing with `go.mod requires go >= 1.26.0`. Rather than chasing that env var every time a dep nudges go.mod, the workflows now read `go-version-file: go.mod`, so they pick up whatever go.mod requires.

To avoid silently bumping Go from under downstream consumers (the openshift/openshift-mcp-server build image takes a while to pick up new Go versions), there's a new `check-go-version.yaml` workflow. It runs on PRs that touch go.mod, compares the `go` directive between base and head, and fails the check if it changed. Maintainers can apply a `downstream-go-verified` label to bypass once they've confirmed the new version is available in the downstream image. Toggling the label re-triggers the workflow.

Go 1.26 has been confirmed available downstream, so this PR is fine to merge once reviewed.

The second commit cherry-picks dependabot's k8s group bump verbatim, preserving the original commit message. Closes #1124 once merged.


matzew:
@manusa in https://github.com/marcnuri-forks/kubernetes-mcp-server/blob/ci/go-version-from-gomod/AGENTS.md#coding-style

it says:

> Go modules target Go **1.25** (see `go.mod`).

Not sure we really must mention the concrete version there? I guess is redundant due to `go.mod` files?

manusa:
> @manusa in https://github.com/marcnuri-forks/kubernetes-mcp-server/blob/ci/go-version-from-gomod/AGENTS.md#coding-style
> 
> it says:
> 
> > Go modules target Go **1.25** (see `go.mod`).
> 
> Not sure we really must mention the concrete version there? I guess is redundant due to `go.mod` files?

Good catch.

That is outdated and missed by the task, not sure if it matters too much the specific version, let me see if I can rephrase it

PR #1176, opened 2026-06-01
feat(kiali): add built-in MCP prompts to the Kiali toolset

hhovsepy:
**Summary**

- Implements GetPrompts() for the Kiali toolset, providing 11 built-in MCP prompts that MCP clients (e.g., Claude Desktop, Cursor) can discover and use for common service mesh tasks
- Adds lightweight resource listing prompts (list_applications, list_namespaces, list_services, list_workloads, list_istio_config) that call list_or_get_resources with structured analysis instructions
- Adds advanced operational prompts (mesh_health_check, mesh_topology, traffic_topology, service_troubleshoot, trace_analysis, istio_config_review) that pre-fetch data from Kiali endpoints and construct detailed analysis guides for the AI model
- Organizes all prompt logic under a new pkg/toolsets/kiali/prompts/ package with reusable handler patterns

Kiali PR: https://github.com/kiali/kiali/pull/9768

### DIFF
--- a/pkg/toolsets/kiali/prompts/istio_config_review.go
+++ b/pkg/toolsets/kiali/prompts/istio_config_review.go
@@ -1,0 +1,109 @@
+package prompts
+
+import (
+	"fmt"
+
+	"k8s.io/klog/v2"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	kialiclient "github.com/containers/kubernetes-mcp-server/pkg/kiali"
+	"github.com/containers/kubernetes-mcp-server/pkg/toolsets/kiali/tools"
+)
+
+func InitIstioConfigReview() []api.ServerPrompt {
+	return []api.ServerPrompt{
+		{
+			Prompt: api.Prompt{
+				Name:        "istio-config-review",
+				Title:       "Review Istio Configuration",
+				Description: "Review and validate Istio configuration in a namespace, checking for misconfigurations and best practice violations",
+				Arguments: []api.PromptArgument{
+					{
+						Name:        "namespace",
+						Description: "Namespace to review Istio configuration for",
+						Required:    true,
+					},
+				},
+			},
+			Handler: istioConfigReviewHandler,
+		},
+	}
+}
+
+func istioConfigReviewHandler(params api.PromptHandlerParams) (*api.PromptCallResult, error) {
+	args := params.GetArguments()
+	namespace := args["namespace"]
+
+	if namespace == "" {
+		return nil, fmt.Errorf("namespace argument is required")
+	}
+
+	klog.Infof("Starting Istio config review for namespace %s...", namespace)
+
+	kiali := kialiclient.NewKiali(params, params.RESTConfig())
+
+	istioContent := fetchKialiData(kiali, params, tools.KialiManageIstioConfigReadEndpoint,
+		map[string]any{"namespace": namespace, "action": "list"})
+
+	resourceContent := fetchKialiData(kiali, params, tools.KialiListOrGetResourcesEndpoint,
+		map[string]any{"namespace": namespace, "resource_type": "services"})
+
+	promptText := buildIstioConfigReviewPrompt(namespace, istioContent, resourceContent)
+
+	return api.NewPromptCallResult(
+		"Istio configuration data gathered successfully",
+		[]api.PromptMessage{
+			{
+				Role: "user",
+				Content: api.PromptContent{
+					Type: "text",
+					Text: promptText,
+				},
+			},
+			{
+				Role: "assistant",
+				Content: api.PromptContent{
+					Type: "text",
+					Text: "I'll review the Istio configuration and check for issues or misconfigurations.",
+				},
+			},
+		},
+		nil,
+	), nil
+}
+
+func buildIstioConfigReviewPrompt(namespace, istioData, resourceData string) string {
+	return fmt.Sprintf(`# Istio Configuration Review
+
+## Namespace: %s
+
+Review the Istio configuration in this namespace for correctness, best practices, and potential issues.
+
+---
+
+## Istio Resources
+
+%s
+
+---
+
+## Services in Namespace
+
+%s
+
+---
+
+## Review Checklist
+
+Analyze the configuration above and check for:
+
+1. **Validation Errors**: Are there any Istio validation warnings or errors reported?
+2. **VirtualService Issues**: Do VirtualServices reference hosts and subsets that actually exist? Are weights valid?
+3. **DestinationRule Consistency**: Do DestinationRules define subsets that match actual workload labels?
+4. **Traffic Policy**: Are there conflicting traffic policies or missing mTLS settings?
+5. **Gateway Configuration**: Are Gateways properly configured with correct hosts and ports?
+6. **Best Practices**: Are there any configuration anti-patterns (e.g., overly broad host matches, missing retries)?
+
+Provide a summary of findings with severity (critical, warning, info) and recommendations.
+`, namespace, istioData, resourceData)
+}
--- a/pkg/toolsets/kiali/prompts/istio_config_review.go
+++ b/pkg/toolsets/kiali/prompts/istio_config_review.go
@@ -46,7 +46,7 @@ func istioConfigReviewHandler(params api.PromptHandlerParams) (*api.PromptCallRe
 		map[string]any{"namespace": namespace, "action": "list"})
 
 	resourceContent := fetchKialiData(kiali, params, tools.KialiListOrGetResourcesEndpoint,
-		map[string]any{"namespace": namespace, "resource_type": "services"})
+		map[string]any{"namespaces": namespace, "resourceType": "service"})
 
 	promptText := buildIstioConfigReviewPrompt(namespace, istioContent, resourceContent)
 
--- a/pkg/toolsets/kiali/prompts/list_resources.go
+++ b/pkg/toolsets/kiali/prompts/list_resources.go
@@ -98,9 +98,60 @@ func InitListIstioConfig() []api.ServerPrompt {
 					},
 				},
 			},
-			Handler: listResourceHandler("istio"),
+			Handler: listIstioConfigHandler,
 		},
 	}
+}
+
+func listIstioConfigHandler(params api.PromptHandlerParams) (*api.PromptCallResult, error) {
+	args := params.GetArguments()
+	namespace := args["namespace"]
+
+	klog.Info("Starting list istio config prompt...")
+
+	reqArgs := map[string]any{"action": "list"}
+	if namespace != "" {
+		reqArgs["namespace"] = namespace
+	}
+
+	kiali := kialiclient.NewKiali(params, params.RESTConfig())
+	content, err := kiali.ExecuteRequest(params.Context, tools.KialiManageIstioConfigReadEndpoint, reqArgs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list istio config: %w", err)
+	}
+
+	scope := "all namespaces"
+	if namespace != "" {
+		scope = fmt.Sprintf("namespace '%s'", namespace)
+	}
+
+	promptText := fmt.Sprintf(`# List Istio Configuration
+
+## Scope
+%s
+
+## Data
+
+%s
+
+## Instructions
+
+Summarize the Istio configuration listed above. Highlight any that need attention.
+`, scope, content)
+
+	return api.NewPromptCallResult(
+		"Istio configuration data retrieved successfully",
+		[]api.PromptMessage{
+			{
+				Role: "user",
+				Content: api.PromptContent{
+					Type: "text",
+					Text: promptText,
+				},
+			},
+		},
+		nil,
+	), nil
 }
 
 func InitMeshTopology() []api.ServerPrompt {
--- a/pkg/toolsets/kiali/prompts/list_resources.go
+++ b/pkg/toolsets/kiali/prompts/list_resources.go
@@ -78,7 +78,7 @@ func InitListWorkloads() []api.ServerPrompt {
 					},
 				},
 			},
-			Handler: listResourceHandler("workloads"),
+			Handler: listResourceHandler("workload"),
 		},
 	}
 }
--- a/pkg/toolsets/kiali/prompts/list_resources.go
+++ b/pkg/toolsets/kiali/prompts/list_resources.go
@@ -58,7 +58,7 @@ func InitListServices() []api.ServerPrompt {
 					},
 				},
 			},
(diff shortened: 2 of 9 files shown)


pkg/toolsets/kiali/prompts/service_troubleshoot.go
```diff
@@ -0,0 +1,127 @@
...
+	if service == "" {
+		return nil, fmt.Errorf("service argument is required")
+	}
+
+	klog.Infof("Starting service troubleshoot prompt for %s/%s...", namespace, service)
+
+	kiali := kialiclient.NewKiali(params, params.RESTConfig())
+
+	logsContent := fetchKialiData(kiali, params, tools.KialiGetLogsEndpoint,
+		map[string]any{"namespace": namespace, "workload": service + "-v1"})
```
josunect: Shouldn't this be using name instead of "workload"? 
```suggestion
		map[string]any{"namespace": namespace, "name": service + "-v1"})
```

pkg/toolsets/kiali/prompts/list_resources.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func listResourceHandler(resourceType string) api.PromptHandlerFunc {
+	return func(params api.PromptHandlerParams) (*api.PromptCallResult, error) {
+		args := params.GetArguments()
+		namespace := args["namespace"]
+
+		klog.Infof("Starting list %s prompt...", resourceType)
+
+		reqArgs := map[string]any{"resource_type": resourceType}
```
josunect: Should the key be resourceType? 
```suggestion
		reqArgs := map[string]any{"resourceType": resourceType}
```

pkg/toolsets/kiali/prompts/list_resources.go
```diff
@@ -0,0 +1,206 @@
...
+				Description: "List applications in the mesh namespaces",
+				Arguments: []api.PromptArgument{
+					{
+						Name:        "namespace",
+						Description: "Optional namespace to filter applications (default: all namespaces)",
+						Required:    false,
+					},
+				},
+			},
+			Handler: listResourceHandler("applications"),
```
josunect: Should this be app?
```suggestion
			Handler: listResourceHandler("app"),
```

pkg/toolsets/kiali/prompts/list_resources.go
```diff
@@ -0,0 +1,206 @@
...
+
+func InitListNamespaces() []api.ServerPrompt {
+	return []api.ServerPrompt{
+		{
+			Prompt: api.Prompt{
+				Name:        "list-namespaces",
+				Title:       "List Namespaces",
+				Description: "List all namespaces with their sidecar injection status and Istio labels",
+			},
+			Handler: listResourceHandler("namespaces"),
```
josunect: Shouldn't be namespace? 
```suggestion
			Handler: listResourceHandler("namespace"),
```

pkg/toolsets/kiali/prompts/list_resources.go
```diff
@@ -0,0 +1,206 @@
...
+				Description: "List services in the mesh namespaces",
+				Arguments: []api.PromptArgument{
+					{
+						Name:        "namespace",
+						Description: "Optional namespace to filter services (default: all namespaces)",
+						Required:    false,
+					},
+				},
+			},
+			Handler: listResourceHandler("services"),
```
josunect: Shouldn't be service?
```suggestion
			Handler: listResourceHandler("service"),
```

pkg/toolsets/kiali/prompts/list_resources.go
```diff
@@ -0,0 +1,206 @@
...
+				Description: "List workloads in the mesh namespaces",
+				Arguments: []api.PromptArgument{
+					{
+						Name:        "namespace",
+						Description: "Optional namespace to filter workloads (default: all namespaces)",
+						Required:    false,
+					},
+				},
+			},
+			Handler: listResourceHandler("workloads"),
```
josunect: Shouldn't be workload?
```suggestion
			Handler: listResourceHandler("workload"),
```

pkg/toolsets/kiali/prompts/list_resources.go
```diff
@@ -0,0 +1,206 @@
...
+func listResourceHandler(resourceType string) api.PromptHandlerFunc {
+	return func(params api.PromptHandlerParams) (*api.PromptCallResult, error) {
+		args := params.GetArguments()
+		namespace := args["namespace"]
+
+		klog.Infof("Starting list %s prompt...", resourceType)
+
+		reqArgs := map[string]any{"resource_type": resourceType}
+		if namespace != "" {
+			reqArgs["namespace"] = namespace
```
josunect: Shouldn't be namespaces? 
```suggestion
			reqArgs["namespaces"] = namespace
```

pkg/toolsets/kiali/prompts/istio_config_review.go
```diff
@@ -0,0 +1,109 @@
...
+
+	klog.Infof("Starting Istio config review for namespace %s...", namespace)
+
+	kiali := kialiclient.NewKiali(params, params.RESTConfig())
+
+	istioContent := fetchKialiData(kiali, params, tools.KialiManageIstioConfigReadEndpoint,
+		map[string]any{"namespace": namespace, "action": "list"})
+
+	resourceContent := fetchKialiData(kiali, params, tools.KialiListOrGetResourcesEndpoint,
+		map[string]any{"namespace": namespace, "resource_type": "services"})
```
josunect: Should the parameter names be updated? 
```suggestion
		map[string]any{"namespaces": namespace, "resourceType": "service"})
```

josunect (commented):
Thanks for the PR, @hhovsepy ! 
I've added some comments.

hhovsepy:
Thank you for the suggestions Josune, all were valid, also I add missing handler for istio config list.

josunect (approved):
LGTM

Tests passed, also I've done some manual testing using the inspector: 

Image

pkg/toolsets/kiali/prompts/traffic_topology.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func trafficTopologyHandler(params api.PromptHandlerParams) (*api.PromptCallResult, error) {
+	args := params.GetArguments()
+	namespaces := args["namespaces"]
+
+	klog.Info("Starting traffic topology analysis prompt...")
+
+	reqArgs := map[string]any{}
+	if namespaces != "" {
```
manusa: `get_mesh_traffic_graph`'s tool schema marks `namespaces` as **required** (`tools/get_mesh_traffic_graph.go:39`), but here it's optional: when the arg is empty, `namespaces` is omitted and the call hard-fails at line 47 — so invoking `traffic-topology` with no arguments (the common case) could error out instead of returning a mesh-wide graph. `mesh-topology` (`list_resources.go:229`) has the same shape but passes `nil` unconditionally and swallows the error via `fetchKialiData`, so its graph section would silently come back empty instead.

Could you confirm the Kiali endpoint defaults to all namespaces when omitted? If it doesn't, please default to the mesh namespaces here.

pkg/toolsets/kiali/prompts/list_resources.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	kialiclient "github.com/containers/kubernetes-mcp-server/pkg/kiali"
+	"github.com/containers/kubernetes-mcp-server/pkg/toolsets/kiali/tools"
+)
+
+func InitListApplications() []api.ServerPrompt {
+	return []api.ServerPrompt{
+		{
+			Prompt: api.Prompt{
+				Name:        "list-applications",
```
manusa: **Prompt naming — please qualify the generic names before merge.** Unlike tool names (`get_logs` → `kiali_get_logs`), prompt names get **no toolset prefix**, so every prompt lands in one flat namespace shared with all other enabled toolsets. Four of these are plain Kubernetes vocabulary with nothing tying them to the mesh: `list-applications`, `list-namespaces`, `list-services`, `list-workloads`. With both `core` and `kiali` enabled, a user reading the prompt list can't tell these apart from core cluster operations by name (the descriptions mention "mesh namespaces", but pickers lead with the name).

There's no clash *today* — `core` only registers `cluster-health-check` — but `core` is expected to grow more prompts, and a future `core` `list-namespaces` / `list-services` would **collide directly** with these. Please add a `mesh-`/`kiali-` qualifier so they're self-identifying and future-proof, consistent with `mesh-health-check`/`mesh-topology` in this same PR and with the other toolsets (`cluster-health-check`, `vm-troubleshoot`). `mesh-` is brand-neutral and survives a downstream rebrand:

- `list-applications` → `mesh-list-applications`
- `list-namespaces` → `mesh-list-namespaces`
- `list-services` → `mesh-list-services`
- `list-workloads` → `mesh-list-workloads`

(`list-istio-config`, `traffic-topology`, `service-troubleshoot`, `trace-analysis`, `istio-config-review` already read as mesh-specific.) A rename will need `make update-readme-tools` to refresh the docs.

pkg/toolsets/kiali/prompts/service_troubleshoot.go
```diff
@@ -0,0 +1,127 @@
...
+	if service == "" {
+		return nil, fmt.Errorf("service argument is required")
+	}
+
+	klog.Infof("Starting service troubleshoot prompt for %s/%s...", namespace, service)
+
+	kiali := kialiclient.NewKiali(params, params.RESTConfig())
+
+	logsContent := fetchKialiData(kiali, params, tools.KialiGetLogsEndpoint,
+		map[string]any{"namespace": namespace, "name": service + "-v1"})
```
manusa: `name: service + "-v1"` assumes Istio's bookinfo-style `<service>-v1` workload naming. `get_logs` treats `name` as a pod/workload name, so for services that don't follow that convention the lookup finds nothing and the "Workload Logs" section degrades to `(data unavailable: …)` — i.e. the prompt's primary signal is missing for most real services. Consider deriving the backing workload (e.g. `list_or_get_resources` with `resourceType=workload`) or taking an explicit `workload`/`pod` argument.

manusa (changes requested):
Thanks for this — the structure is clean (the `listResourceHandler` factory is a nice touch), the README/`docs/configuration.md` regenerate with no drift, and build/vet/the kiali suite all pass.

Reviewing from the kubernetes-mcp-server side — Kiali owns the toolset's deeper behavior and its own tests, so I've kept this to three things worth a look before merge:

1. **`traffic-topology` can fail on its default no-arg invocation** — `get_mesh_traffic_graph` marks `namespaces` required, but the prompt makes it optional and hard-fails when omitted.
2. **Generic prompt names** — `list-namespaces` / `list-services` / `list-workloads` aren't mesh-qualified; with `core` + `kiali` enabled they read like core cluster ops, and this will only get worse as `core` adds its own prompts.
3. **`service-troubleshoot` hardcodes a `-v1` workload suffix** — fine for bookinfo, empty logs for most real services.

Not flagging test coverage — assuming that's handled in the Kiali repo.

hhovsepy:
Thank you very much @manusa for the review, all points were approached.
The 1. point with namespaces, I decided to make "namespaces" required, as comma separated namespaces, or user will need to explicitly put "all" in case of all mesh namespaces, but not default all namespaces when not set.

manusa:
Hi @hhovsepy thanks for taking care of this. Last commit wasn't signed off and we have a DCO failure.

pkg/toolsets/kiali/prompts/list_resources.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+}
+
+func meshTopologyHandler(params api.PromptHandlerParams) (*api.PromptCallResult, error) {
+	klog.Info("Starting mesh topology prompt...")
+
+	kiali := kialiclient.NewKiali(params, params.RESTConfig())
+
+	statusContent := fetchKialiData(kiali, params, tools.KialiGetMeshStatusEndpoint, nil)
+	graphContent := fetchKialiData(kiali, params, tools.KialiGetMeshTrafficGraphEndpoint, nil)
```
manusa: This is the other half of my earlier `traffic-topology` comment and it's still open. `meshTopologyHandler` passes `nil` to `get_mesh_traffic_graph`, but that endpoint requires `namespaces` (`tools/get_mesh_traffic_graph.go:39`) and `ExecuteRequest` doesn't validate client-side. Since the call goes through `fetchKialiData`, the error is swallowed and the **"Traffic Graph" section silently renders as `(data unavailable: …)`** — so this prompt ships with half its content permanently empty.

`traffic-topology` already solved this; please resolve namespaces the same way before the graph call:

```go
resolved, err := resolveNamespaces(kiali, params, "all")
if err != nil {
    return nil, fmt.Errorf("failed to resolve mesh namespaces: %w", err)
}
graphContent := fetchKialiData(kiali, params, tools.KialiGetMeshTrafficGraphEndpoint,
    map[string]any{"namespaces": resolved})
```

(or drop the graph from this prompt if it's redundant with `traffic-topology`). A behavioral test on `mesh-topology` would have caught this — worth adding with the fix.

manusa (changes requested):
Thanks for the iterations, @hhovsepy — the prompt naming and the `-v1` workload suffix from my earlier review are both sorted. josunect's threads are all resolved and the suite/docs are clean.

One thing left before this is good to go — it's the other half of my original traffic-graph comment:

**`mesh-topology` ships with its traffic-graph section permanently empty.** `meshTopologyHandler` calls `get_mesh_traffic_graph` with no `namespaces`, but that endpoint requires them, so the section always degrades to `(data unavailable: …)`. `traffic-topology` already solves this with `resolveNamespaces` — same fix applies here (details inline). That's the only blocker from my side.

Non-blocking polish, take it or leave it:
- "all mesh namespaces" in the `traffic-topology` arg/error/README actually means *all accessible* namespaces — minor wording.
- A one-line comment on `fetchKialiData` to mark the soft-fail intentional (it's what hid the issue above).
- A `mesh-topology` test would've caught this — worth one with the fix, though I'm leaving broader coverage to the Kiali repo.

Fix the `mesh-topology` graph and I'm happy to approve.

pkg/toolsets/kiali/prompts/list_resources.go
hhovsepy: Sorry for the confusion @manusa , I have fixed that as well now.
Also the non-blocking polish is done.

manusa (approved):
Thanks @hhovsepy — all my review points are addressed. `mesh-topology` now resolves namespaces before the traffic-graph call (with `TestMeshTopologyPrompt` covering it), and the wording and `fetchKialiData` doc comment are in too. Verified build/vet/the Kiali suite locally. LGTM.

PR #1178, opened 2026-06-01
fix(tokenexchange): enforce require_tls on token exchange HTTP client

### LINKED ISSUE #1111: Token-exchange HTTP client should honor require_tls
The OAuth state HTTP client wraps its transport in `config.NewTLSEnforcingTransport(transport, cfg.IsRequireTLS)` (`pkg/oauth/state.go:118`). The token-exchange HTTP client does not (`pkg/tokenexchange/config.go:120-125`).

With `require_tls=true` and a misconfigured `http://` token URL, the OAuth client refuses the request, but the strategy exchangers (`rfc8693`, `keycloak-v1`, `entra-obo`) still attempt it via `cfg.HTTPClient()`.

**Suggested fix:** wrap the transport in `pkg/tokenexchange/config.go` `HTTPClient()` with `config.NewTLSEnforcingTransport`, mirroring `pkg/oauth/state.go:118`. To avoid an import cycle (`pkg/tokenexchange` → `pkg/config`), the enforcer can be passed in as a constructor option or via a small interface on `TargetTokenExchangeConfig` that callers populate from `BaseConfig.IsRequireTLS()`.

Pre-existing; not introduced by #1109.

lunarwhite:
### Summary

Wraps the `TargetTokenExchangeConfig.HTTPClient()` transport with TLS enforcement when `require_tls` is enabled, fixing a gap where `rfc8693`, `keycloak-v1`, and `entra-obo` strategy exchangers would bypass the security check.

When `require_tls=true`, the token exchange HTTP client now rejects non-HTTPS URLs at the transport level, matching the enforcement already in place for OAuth and Kiali clients.

### Motivation

Fixes https://github.com/containers/kubernetes-mcp-server/issues/1111

### DIFF
--- a/pkg/kubernetes/token_exchange.go
+++ b/pkg/kubernetes/token_exchange.go
@@ -43,6 +43,8 @@ func ExchangeTokenInContext(
 		return stsExchangeTokenInContext(ctx, baseConfig, oidcProvider, httpClient, subjectToken, stsConfig)
 	}
 
+	exCfg.SetRequireTLS(baseConfig.IsRequireTLS)
+
 	exchanger, ok := tokenexchange.GetTokenExchanger(tep.GetTokenExchangeStrategy())
 	if !ok {
 		klogutil.LogWarn(klog.FromContext(ctx), "token exchange strategy not found in registry, falling back to sts exchange",
@@ -183,6 +185,7 @@ func strategyBasedTokenExchange(
 			return ctx, fmt.Errorf("token exchange config validation: %w", err)
 		}
 	}
+	cfg.SetRequireTLS(baseConfig.IsRequireTLS)
 
 	if httpClient != nil {
 		ctx = context.WithValue(ctx, oauth2.HTTPClient, httpClient)
--- a/pkg/tokenexchange/config.go
+++ b/pkg/tokenexchange/config.go
@@ -8,6 +8,8 @@ import (
 	"os"
 	"sync"
 	"time"
+
+	"k8s.io/klog/v2"
 )
 
 const (
@@ -77,6 +79,45 @@ type TargetTokenExchangeConfig struct {
 	assertionMutex sync.Mutex `toml:"-"`
 	// clientMutex protects HTTP client creation from race conditions
 	clientMutex sync.Mutex `toml:"-"`
+
+	// requireTLS is nil or a function that returns true when TLS is required.
+	// When non-nil, HTTPClient() wraps the transport to reject non-HTTPS URLs at runtime.
+	requireTLS func() bool `toml:"-"`
+}
+
+// SetRequireTLS configures TLS enforcement for the HTTP client.
+// When enforcer is non-nil and returns true, HTTPClient() will reject
+// requests to non-HTTPS URLs. The enforcer is captured when the client
+// is built, so callers must set it before the first HTTPClient() call.
+func (c *TargetTokenExchangeConfig) SetRequireTLS(enforcer func() bool) {
+	c.clientMutex.Lock()
+	defer c.clientMutex.Unlock()
+	c.requireTLS = enforcer
+}
+
+// tlsEnforcingTransport wraps an http.RoundTripper and rejects non-HTTPS requests
+// when RequireTLS returns true. Inlined here to avoid an import cycle with pkg/config.
+type tlsEnforcingTransport struct {
+	Base       http.RoundTripper
+	RequireTLS func() bool
+}
+
+func (t *tlsEnforcingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if t.RequireTLS != nil && t.RequireTLS() && !isSecureHTTPScheme(req.URL.Scheme) {
+		klog.FromContext(req.Context()).V(1).Info("require_tls: blocked request to host", "host", req.URL.Host)
+		return nil, fmt.Errorf("require_tls is enabled but request to %s uses %q scheme (secure scheme required)",
+			req.URL.Host, req.URL.Scheme)
+	}
+	return t.Base.RoundTrip(req)
+}
+
+func isSecureHTTPScheme(scheme string) bool {
+	switch scheme {
+	case "https", "wss":
+		return true
+	default:
+		return false
+	}
 }
 
 // Validate checks that the configuration values are valid
@@ -117,7 +158,7 @@ func (c *TargetTokenExchangeConfig) HTTPClient() (*http.Client, error) {
 		return c.client, nil
 	}
 
-	transport := http.DefaultTransport.(*http.Transport).Clone()
+	baseTransport := http.DefaultTransport.(*http.Transport).Clone()
 
 	// Always set MinVersion for security, regardless of CAFile
 	tlsConfig := &tls.Config{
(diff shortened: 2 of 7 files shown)


lunarwhite:
Kindly

/cc @manusa 

for review when you get a moment

pkg/tokenexchange/config.go
```diff
@@ -75,6 +77,42 @@ type TargetTokenExchangeConfig struct {
...
+// tlsEnforcingTransport wraps an http.RoundTripper and rejects non-HTTPS requests
+// when RequireTLS returns true. Inlined here to avoid an import cycle with pkg/config.
+type tlsEnforcingTransport struct {
+	Base       http.RoundTripper
+	RequireTLS func() bool
+}
+
+func (t *tlsEnforcingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if t.RequireTLS != nil && t.RequireTLS() && !isSecureHTTPScheme(req.URL.Scheme) {
+		klog.V(1).Infof("require_tls: blocked request to %s", req.URL.Host)
```
manusa: Use the contextual logger here to match `pkg/config/tls.go:59` and the project-wide migration in #1202:

```go
klog.FromContext(req.Context()).V(1).Info("require_tls: blocked request to host", "host", req.URL.Host)
```

`klog.V(1).Infof(...)` is the pre-#1202 style and drops the request context.

pkg/tokenexchange/config.go
```diff
@@ -75,6 +77,42 @@ type TargetTokenExchangeConfig struct {
...
+
+	// requireTLS is nil or a function that returns true when TLS is required.
+	// When non-nil, HTTPClient() wraps the transport to reject non-HTTPS URLs at runtime.
+	requireTLS func() bool `toml:"-"`
+}
+
+// SetRequireTLS configures TLS enforcement for the HTTP client.
+// When enforcer is non-nil and returns true, HTTPClient() will reject
+// requests to non-HTTPS URLs.
+func (c *TargetTokenExchangeConfig) SetRequireTLS(enforcer func() bool) {
```
manusa, line 91: `SetRequireTLS` writes `c.requireTLS` without holding `clientMutex`, but `HTTPClient()` reads it under `clientMutex` (line 172). It's race-free in the current call paths, but the type is documented as concurrency-safe and this breaks that contract: it'll fail `-race` as soon as a `GetTokenExchangeConfig` provider returns a config shared across goroutines (the `ExchangeTokenInContext` path this PR patches). Please guard the write with `clientMutex` or set the field at construction.

pkg/tokenexchange/config.go
```diff
@@ -128,7 +166,12 @@ func (c *TargetTokenExchangeConfig) HTTPClient() (*http.Client, error) {
 		tlsConfig.RootCAs = caCertPool
 	}
 
-	transport.TLSClientConfig = tlsConfig
+	baseTransport.TLSClientConfig = tlsConfig
+
+	var transport http.RoundTripper = baseTransport
+	if c.requireTLS != nil {
```
manusa: The wrap-or-not decision is made here at build time and the client is then memoized (cache check at line 116 only invalidates on `CAFile` change). If `HTTPClient()` is ever called before `SetRequireTLS()`, the cached client is unwrapped permanently and enforcement is silently skipped. All current callers set it first, but that ordering isn't documented or enforced. Also, since `stsConfigCacheKey` omits `require_tls` and the captured `IsRequireTLS` is bound to the build-time config, toggling `require_tls` alone on SIGHUP won't re-wrap an otherwise-unchanged cached client. Consider documenting the ordering and/or including `require_tls` in the cache key.

pkg/kubernetes/token_exchange.go
```diff
@@ -179,6 +181,7 @@ func strategyBasedTokenExchange(
 		if err := cfg.Validate(); err != nil {
 			return ctx, fmt.Errorf("token exchange config validation: %w", err)
 		}
+		cfg.SetRequireTLS(baseConfig.IsRequireTLS)
```
manusa: `SetRequireTLS` is only called inside the `if cfg == nil` (fresh-build) branch, so the cached-config path relies entirely on `getOrBuildStsConfig` having set it. That coupling is fragile and undocumented, and because this PR conflicts with `main` (where `getOrBuildStsConfig` was just refactored in #1217/#1223), please make sure the `SetRequireTLS` call is re-applied to the new `getOrBuildStsConfig` after rebasing, or the cached path silently loses enforcement.

pkg/tokenexchange/config_test.go
```diff
@@ -0,0 +1,250 @@
...
+	s.Run("allows HTTPS when requireTLS returns true", func() {
+		cfg := &TargetTokenExchangeConfig{}
+		cfg.SetRequireTLS(func() bool { return true })
+
+		client, err := cfg.HTTPClient()
+		s.Require().NoError(err)
+		s.NotNil(client)
+
+		// HTTPS should not be blocked - use a transport-level check
+		resp, err := client.Get("https://example.com/token")
```
manusa: This makes a real network request to `https://example.com/token`, which is flaky/slow and fails in offline or sandboxed CI. Use `httptest.NewTLSServer()` for a deterministic local HTTPS endpoint instead. While here, a `true→false` toggle test on a cached client would cover the per-request evaluation that's the whole point of the `func() bool`.

manusa (changes requested):
Thanks for taking this on, @lunarwhite, and for following the existing `pkg/oauth/state.go` pattern and the setter approach suggested on the issue. I dug into it independently and the gap is real: all three strategy exchangers (`rfc8693`, `keycloak-v1`, `entra-obo`) go through `cfg.HTTPClient()`, and Layer-1 validation (`ValidateRequireTLS`) only covers `authorization_url`/`server_url`/`sse_base_url`, not the STS/OIDC-discovered token URL, so this runtime check really is the only thing guarding that request. Good catch, and the cross-reference comment you left in `pkg/config/tls.go` about the duplication is appreciated.

A few things I'd like to see before this goes in:

- **It needs a rebase first.** The branch predates #1217/#1223, which refactored these same files, and GitHub currently reports the PR as conflicting. The `pkg/kubernetes/provider_token_exchange.go` hunk targets the old `getOrBuildStsConfig` (the one that read `p.baseConfig` as a field and used `klog.Warningf`); on `main` that function now takes a `baseConfig` parameter and uses `logger.Error`. Please rebase and re-apply the `SetRequireTLS` wiring into the new version, and see my inline note on why the cached-config path depends on that line surviving the rebase.
- **A test for the wiring itself.** Right now all the new tests live in `pkg/tokenexchange` and exercise the config and exchangers directly, and nothing checks that `pkg/kubernetes` actually calls `SetRequireTLS` before an exchange runs, so removing those calls would keep the suite green. Could you add a `pkg/kubernetes`-layer test that an `http://` token URL is rejected end-to-end when `require_tls=true`? That's the part most at risk during the rebase.
- The rest of my notes are inline: the logging style, the unsynchronized `SetRequireTLS`/cached-client hazards, and the flaky network call in the test.

I've left the specifics inline so they're easy to work through one at a time. Thanks again, this is a worthwhile fix and it's close.

pkg/tokenexchange/config.go
lunarwhite: @manusa Appreciate your throughout review! I've incorporated all your suggestions in the latest commit. Please let me know if anything else worth addressing.

manusa (approved):
LGTM, thx!

PR #917, opened 2026-03-12
feat(tekton): add Tekton Pipelines eval tasks

ksimon1:
feat(tekton): add Tekton Pipelines install/uninstall/status Makefile targets and eval tasks

This PR introduces a new tekton eval set to kubernetes-mcp-server, exposing domain-specific Tekton operations.

PR #1179, opened 2026-06-02
test(evals): add argocd eval suite

anandrkskd:
argocd toolset reads Application, AppProject, and ArgoCD instance

Add a new `argocd` toolset providing 6 read-only MCP tools for ArgoCD Operator CRDs: list/get for Application, AppProject, and ArgoCD instances.

Get operations return enriched output with a summary header (project, sync status, health, source, destination) followed by full YAML. List operations return standard table/yaml output.

Uses dynamic client with unstructured field extraction (NestedString/ NestedMap/NestedSlice) — no ArgoCD Go module dependencies.

Toolset is disabled by default; enable with `--toolsets argocd`.

Assisted-by: claudecode


Cali0707:
Hey @anandrkskd would you be able to add some evals to this new toolset?

Also, can you run the evals you create with:
1. only core toolset enabled
2. core + argocd toolsets enabled

For a lot of toolsets we have been finding that models are able to fully accomplish tasks with the generic CRUD tools in the core toolset, without need for specific ones for other APIs.

This means that for any cases where evals identify you do need custom argo tools (i.e. it is not just CRUD), you will be able to create more targeted tools and not have as much token overhead from the CRUD tools

@manusa and I have been discussing how to add a "supported" table to the repo, which will only depend on evals, not necessarily having a toolset as well. So if evals show that we don't need custom argo tools, argo can still be "supported" as part of the server

anandrkskd:
Hey @Cali0707, I have added evals and they are passing without enabling argocd specific toolset, I am planning to remove my implementation as the generic tools sets are able to perform the action we are expecting.
what would you suggest ?

Cali0707:
Hey @anandrkskd - thanks for checking this with evals!

IMO the path forwards here is to:
1. merge your evals (but not the toolset)
2. We can mark Argo as validated/whatever wording is decided in https://github.com/containers/kubernetes-mcp-server/issues/1230 as that lands

That way we can still highlight the argo compatability with this MCP server, without adding tools that the models don't need

anandrkskd:
I will update the PR, remove the toolset and go forward from there, thanks!

evals/tasks/argocd/multi-app-sync-waves/task.yaml
```diff
@@ -0,0 +1,104 @@
...
+
+          for APP in db-migration api-server web-frontend; do
+            kubectl get application "$APP" -n argocd-eval-sync-waves -o name || { echo "ERROR: Application $APP not found"; exit 1; }
+          done
+
+          DB_WAVE=$(kubectl get application db-migration -n argocd-eval-sync-waves -o jsonpath='{.metadata.annotations.argocd\.argoproj\.io/sync-wave}')
+          API_WAVE=$(kubectl get application api-server -n argocd-eval-sync-waves -o jsonpath='{.metadata.annotations.argocd\.argoproj\.io/sync-wave}')
+          WEB_WAVE=$(kubectl get application web-frontend -n argocd-eval-sync-waves -o jsonpath='{.metadata.annotations.argocd\.argoproj\.io/sync-wave}')
+
+          if [ "$DB_WAVE" -ge "$API_WAVE" ] 2>/dev/null; then
```
manusa: If the agent omits the `argocd.argoproj.io/sync-wave` annotations, `$DB_WAVE`/`$API_WAVE`/`$WEB_WAVE` are empty, `[ "" -ge "" ]` exits 2, and `2>/dev/null` plus the `if` swallow it as ordering-OK. So missing annotations pass, while all-0 (present but wrong) correctly fails, which rewards skipping the annotations that are the point of the task. Assert each wave is a non-empty integer (`[[ "$DB_WAVE" =~ ^-?[0-9]+$ ]]`) before comparing, and drop the `2>/dev/null`.

evals/tasks/argocd/create-argocd-instance/task.yaml
```diff
@@ -0,0 +1,58 @@
...
+          name: argocd-eval-create-inst
+
+  verify:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          set -e
+          kubectl get argocd argocd-dev -n argocd-eval-create-inst -o name || { echo "ERROR: ArgoCD instance not found"; exit 1; }
+          ROUTE=$(kubectl get argocd argocd-dev -n argocd-eval-create-inst -o jsonpath='{.spec.server.route}')
+          if [ -z "$ROUTE" ]; then
```
manusa: The prompt asks for the server route enabled, but this reads the whole `{.spec.server.route}` object and only checks `-z`. `spec.server.route.enabled: false` renders `{"enabled":false}` (non-empty), so the opposite of the request passes. Read `{.spec.server.route.enabled}` and assert it equals `true`.

evals/tasks/argocd/get-application/task.yaml
```diff
@@ -0,0 +1,68 @@
...
+          destination:
+            server: https://kubernetes.default.svc
+            namespace: default
+
+  verify:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          set -e
+          kubectl get application app-frontend -n argocd-eval-get-app -o name || exit 1
```
manusa: This verify only re-confirms the setup-created resource still exists (a tautology for a read-only get), and the `llmJudge: contains: "app-frontend"` string is taken straight from the prompt, so the task passes without the agent actually retrieving anything. Same pattern in `get-appproject` and `get-argocd-instance`. Consider asserting on a distinctive field value the agent has to surface.

evals/tasks/argocd/create-applicationset/task.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+  verify:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          set -e
+          kubectl get applicationset env-apps -n argocd-eval-create-appset -o name || { echo "ERROR: ApplicationSet not found"; exit 1; }
+          ELEMENTS=$(kubectl get applicationset env-apps -n argocd-eval-create-appset -o jsonpath='{.spec.generators[0].list.elements}')
+          for ENV in dev staging prod; do
+            echo "$ELEMENTS" | grep -q "$ENV" || { echo "ERROR: Environment $ENV not found in list generator"; exit 1; }
```
manusa, line 35: Verification only checks the dev/staging/prod list-generator elements; the generated-Application template (repoURL, path, project, per-env namespace) required by the prompt isn't verified, so a correct generator with a broken `spec.template` passes.

manusa (changes requested):
Hi @anandrkskd, thanks for this, and for validating with evals first instead of assuming the toolset was needed. Also, long time no see :), nice to see you collaborating in this repo too!

Dropping the toolset and keeping the evals matches the plan from the thread, and the tasks themselves are solid: valid argoproj.io GVKs, good breadth (get/list/create/delete/update plus realistic multi-step scenarios), namespace-scoped idempotent setup/cleanup, no secrets. Things to tackle before merge, in priority order:

1. No install path for the ArgoCD CRDs. Setup `k8s.create`s `argoproj.io/*` CRs and verify `kubectl get`s them, but there's no `build/argocd.mk` (tekton, kubevirt, and kiali each have one) and no CI wiring, so setup fails with "no matches for kind" on any cluster built from this repo. Tekton's eval PR (#917) shipped its install target and tasks together; this suite needs the same to be reproducible.
2. The taskSet is only in `openai-agent/eval.yaml`. `claude-code/eval.yaml` (the recommended keyless path) isn't updated, so `make run-evals SUITE=argocd AGENT=claude-code` silently runs nothing.
3. argocd isn't wired into CI: it's missing from the workflow's suite choices and comment parser and has no install step, so it never runs on normal CI and would fail at setup on the `all` path.
4. Two verify scripts pass on wrong behavior (details inline): `multi-app-sync-waves` treats missing sync-wave annotations as correct, and `create-argocd-instance` checks that the route is configured rather than enabled.
5. README suites list and judge-backed counts aren't updated for argocd.
6. The `get-*` verifies are tautological and their llmJudge string just echoes the prompt, so they pass without the agent retrieving anything (details inline).
7. `create-applicationset` doesn't verify the generated Application template, only the list generator (details inline).

evals/tasks/argocd/create-applicationset/task.yaml
anandrkskd: Fixed with checks for repo_url, destination_server, project and path.

evals/tasks/argocd/get-application/task.yaml
anandrkskd: updated the check for get-application get-appproject and get-argocd-instance, checks app,appproj and instance details intead of infromation passed form prompt.

evals/tasks/argocd/create-argocd-instance/task.yaml
anandrkskd: now the eval checks for route field to be true, and fails if opposite.

evals/tasks/argocd/multi-app-sync-waves/task.yaml
anandrkskd: updated eval.

anandrkskd:
Hi @manusa Thanks! TBH I also went for toolset first, but watching generic toolset working so well I switched to the generic toolset — no point building tools nobody needs :D, And great to work alongside you again, excited to contribute here! 
I have updated the evals and CI related changes. looking forward for some review comments 😉

anandrkskd:
Hi @manusa can I get a review ? I have updated the evals. TIA

evals/core-eval-testing/builtin-openai/eval-all.yaml
```diff
@@ -83,3 +83,13 @@ config:
             toolPattern: "netobserv_.*"
         minToolCalls: 1
         maxToolCalls: 10
+    # ArgoCD tasks
```
Cali0707, line 86: Let's also add this to a dedicated `eval-argo.yaml` file in this dir, and symlink that into each other agent dir under `core-eval-testing`

anandrkskd: updated the PR with seperate `eval-argo.yaml`.

wtam2018:
@manusa Can we get another review from you? TIA.

.github/workflows/mcpchecker.yaml
```diff
@@ -254,6 +255,10 @@ jobs:
             netobserv)
               MATRIX=$(single_matrix netobserv "${EVAL_DIR}/eval-netobserv.yaml" "core,config,netobserv")
               ;;
+            argocd)
+              echo "label-selector=suite=argocd" >> $GITHUB_OUTPUT
+              echo "toolsets=core,config" >> $GITHUB_OUTPUT
+              ;;
```
Cali0707, line 261: You will need to construct/export `MATRIX` the same way other toolsets here do.

Also you should add `argocd` to the list of allowable values in line 183, and to the full matrix constructed above

anandrkskd: updated the PR. added MATRIX changes

anandrkskd (commented):
/run-mcpchecker

anandrkskd:
@Cali0707  @manusa Can we get another review? all the checks seems to be passing. TIA.

Cali0707 (approved):
LGTM

PR #1181, opened 2026-06-03
feat(auth): decouple STS endpoint and make RFC 8693 token-type params configurable

zruchi:
## Summary

Two related changes to `pkg/tokenexchange` / `pkg/kubernetes`, fully backwards-compatible:

- RFC 8693 `subject_token_type` was sent empty when unset (violates [RFC 8693 section 2.1](https://datatracker.ietf.org/doc/html/rfc8693#section-2.1)) and `requested_token_type` was hardcoded. Both are now configurable via `sts_subject_token_type` / `sts_requested_token_type`, defaulting to `access_token`.
- Add `sts_token_url` to point token exchange at a separate STS gateway, decoupled from the IDP at `authorization_url`. Unblocks cross-realm RFC 8693 and `skip_jwt_verification=true` + token exchange (which had no provider to discover an endpoint from).

## New TOML keys

| Key | Default | Purpose |
|---|---|---|
| `sts_subject_token_type` | `urn:ietf:params:oauth:token-type:access_token` | RFC 8693 `subject_token_type` form param |
| `sts_requested_token_type` | `urn:ietf:params:oauth:token-type:access_token` | RFC 8693 `requested_token_type` form param |
| `sts_token_url` | *(falls back to OIDC discovery)* | Explicit STS endpoint |

## Notable behavior changes

- `SecurityTokenService.IsEnabled()` now keys off `TokenURL != ""` instead of `Provider != nil` — "is exchange usable?" is correctly "do I have an endpoint to POST to?". Required for the passthrough + gateway combo.
- New `resolveStsTokenURL` helper: explicit `sts_token_url` wins, else OIDC-discovered endpoint.
- `sts_token_url` validated as `http`/`https`, warns on `http`.

## Example: separate gateway from issuer

```toml
authorization_url       = "https://idp.example.com/realms/users"     # validates user JWT
sts_token_url           = "https://sts-gateway.internal/oauth/token" # mints cluster token
token_exchange_strategy = "rfc8693"
sts_client_id           = "mcp-backend"
sts_client_secret       = "..."
sts_audience            = "kubernetes-api"
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -497,6 +497,7 @@ Configure OAuth/OIDC authentication for HTTP mode deployments.
 | `sts_federated_token_file` | string | `""` | Path to a JWT file from an external identity provider, e.g., SPIRE JWT-SVID (for `federated` auth style). |
 | `sts_subject_token_type` | string | `"urn:ietf:params:oauth:token-type:access_token"` | `subject_token_type` form parameter for RFC 8693 token exchange. Override to `urn:ietf:params:oauth:token-type:jwt` for cross-realm flows. |
 | `sts_requested_token_type` | string | `"urn:ietf:params:oauth:token-type:access_token"` | `requested_token_type` form parameter for RFC 8693 token exchange. Override to `urn:ietf:params:oauth:token-type:jwt` when the STS must mint a fresh JWT rather than echo the subject token shape. |
+| `sts_token_url` | string | `""` | Explicit token endpoint for token exchange. When set, takes precedence over the endpoint discovered from the OIDC provider at `authorization_url`. Use this to point token exchange at a separate STS gateway, or to enable token exchange when no `authorization_url` is configured (e.g., with `skip_jwt_verification=true`). |
 | `cluster_auth_mode` | string | `""` | Cluster auth mode: `passthrough` (forward Authorization header when present, fall back to kubeconfig when absent) or `kubeconfig` (always use kubeconfig credentials). Defaults to `passthrough`. |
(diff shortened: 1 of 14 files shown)


theJC:
We need one piece of this fix as well to properly integrate with Okta and use it as our AS.  How can we move this forward?

It appears this ticket and https://github.com/containers/kubernetes-mcp-server/pull/1173 have overlap in parts of what they changed ( the sts_token_url and sts_subject_token_type part)

Should we we break out to smaller updates instead of being grouped together in order to hopefully get them incrementally approved since each would be better scoped?

zruchi:
> #1173

make sense. Let me break this into two separate pr.

theJC:
Personally, I'd recommend perhaps start with one that is solely focused on fixing the `subject_token_type` - but thats me being selfish because thats the piece I need 😉 

[§2.1](https://datatracker.ietf.org/doc/html/rfc8693#section-2.1) made subject_token_type REQUIRED; the global exchange path sends it empty. access_token is the correct identifier for the token this path actually carries (a validated OAuth access token from the Authorization header), with sts_subject_token_type to override when it isn't.

I feel from a spec compliance standpoint, this is the most important to fix to ensure the REQUIRED field is set.

The OPTIONAL one would be a bit less important IMHO, also it might require a little discussion because the requested_token_type comment in this PRs code says "Per RFC 8693 section 2.1 requested_token_type defaults to access_token when omitted." That one is actually wrong, no?  — §2.1 says "If the requested type is unspecified, the issued token type is at the discretion of the authorization server."

Cali0707:
Hey yes, would be happy to review pieces of this incrementally.

PR #1183, opened 2026-06-04
HTTP: Don't fail for errors during shutdown

2uasimojo:
Errors during shutdown, like "context deadline exceeded", have been causing `Run()` to fail (and print the usage statement). This is only cosmetically annoying. But philosophically, we only want `Run()` to fail if it encounters an error during startup, or dies unexpectedly while running.

With this change, we still log shutdown errors, but they no longer bubble up the stack. Cosmetically, this means the server doesn't print the usage statement when it shuts down "normally" (errors notwithstanding).

### DIFF
--- a/pkg/http/http.go
+++ b/pkg/http/http.go
@@ -184,22 +184,15 @@ func Serve(ctx context.Context, mcpServer *mcp.Server, cfgState *config.StaticCo
 
 	klog.V(0).Infof("Shutting down HTTP server gracefully...")
 
-	// Attempt to shut down both servers, collecting all errors
-	var shutdownErrs []error
-
 	if err := httpServer.Shutdown(shutdownCtx); err != nil {
+		// Don't fail Run() for errors during shutdown
 		klog.Errorf("HTTP server shutdown error: %v", err)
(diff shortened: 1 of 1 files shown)


2uasimojo:
/assign @nader-ziada 
/cc @Cali0707 

This error path was added via https://github.com/containers/kubernetes-mcp-server/pull/590/changes#diff-4bf68c95e63b2d0c3eda7f760fbe68cac1e0502eb4b76a7292720e40df68e576R135-R150 -- please advise if I've missed something that this is going to mess up!

nader-ziada:
> /assign @nader-ziada /cc @Cali0707
> 
> This error path was added via https://github.com/containers/kubernetes-mcp-server/pull/590/changes#diff-4bf68c95e63b2d0c3eda7f760fbe68cac1e0502eb4b76a7292720e40df68e576R135-R150 -- please advise if I've missed something that this is going to mess up!

i don't think it will mess anything, should be fine.

nader-ziada:
lgtm, but i'll wait to see of @Cali0707 has any issues with this change

Cali0707 (approved):
LGTM, thanks @2uasimojo !

PR #1184, opened 2026-06-05
fix(core): clarify SSA full-state semantics in resources_create_or_update description

### LINKED ISSUE #1164: resources_create_or_update: expose fieldManager as optional parameter to prevent SSA field abandonment across sequential tool calls

lunarwhite:
### Summary

Clarify the Server-Side Apply semantics in the `resources_create_or_update` tool description so models treat the manifest as the complete desired state instead of a partial patch.

### Motivation

The tool does a force apply under a single field manager, so when a later call omits a field an earlier call set, SSA removes it by design. The description never spelled that out, so the tool got used as a partial patch and fields were silently dropped.

This PR just updates the tool and `resource` parameter descriptions to say the manifest is the full desired state, and to tell the model to fetch with `resources_get`, modify, then re-apply the full resource. No new parameters and no behavior change. `README.md` and the `pkg/mcp/testdata/toolsets-*.json` snapshots are regenerated to match.

Fixes #1164

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,9 +377,9 @@ In case multi-cluster support is enabled (default) and you have access to multip
   - `name` (`string`) **(required)** - Name of the resource
   - `namespace` (`string`) - Optional Namespace to retrieve the namespaced resource from (ignored in case of cluster scoped resources). If not provided, will get resource from configured namespace
 
-- **resources_create_or_update** - Create or update a Kubernetes resource in the current cluster by providing a YAML or JSON representation of the resource
+- **resources_create_or_update** - Create or update a Kubernetes resource via Server-Side Apply. The manifest is the complete desired state: any field this tool previously set and the new manifest omits is removed. To edit an existing resource, fetch it with resources_get, modify it, then re-apply the full resource.
 (common apiVersion and kind include: v1 Pod, v1 Service, v1 Node, apps/v1 Deployment, networking.k8s.io/v1 Ingress, route.openshift.io/v1 Route)
-  - `resource` (`string`) **(required)** - A JSON or YAML containing a representation of the Kubernetes resource. Should include top-level fields such as apiVersion,kind,metadata, and spec
+  - `resource` (`string`) **(required)** - Complete YAML or JSON representation of the Kubernetes resource (full desired state, not a partial patch). Include apiVersion, kind, metadata, and the full spec.
 
 - **resources_delete** - Delete a Kubernetes resource in the current cluster by providing its apiVersion, kind, optionally the namespace, and its name
 (common apiVersion and kind include: v1 Pod, v1 Service, v1 Node, apps/v1 Deployment, networking.k8s.io/v1 Ingress, route.openshift.io/v1 Route)
--- a/pkg/mcp/testdata/toolsets-core-tools.json
+++ b/pkg/mcp/testdata/toolsets-core-tools.json
@@ -396,11 +396,11 @@
       "openWorldHint": true,
       "title": "Resources: Create or Update"
     },
-    "description": "Create or update a Kubernetes resource in the current cluster by providing a YAML or JSON representation of the resource\n(common apiVersion and kind include: v1 Pod, v1 Service, v1 Node, apps/v1 Deployment, networking.k8s.io/v1 Ingress)",
+    "description": "Create or update a Kubernetes resource via Server-Side Apply. The manifest is the complete desired state: any field this tool previously set and the new manifest omits is removed. To edit an existing resource, fetch it with resources_get, modify it, then re-apply the full resource.\n(common apiVersion and kind include: v1 Pod, v1 Service, v1 Node, apps/v1 Deployment, networking.k8s.io/v1 Ingress)",
     "inputSchema": {
       "properties": {
         "resource": {
-          "description": "A JSON or YAML containing a representation of the Kubernetes resource. Should include top-level fields such as apiVersion,kind,metadata, and spec",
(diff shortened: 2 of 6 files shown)


manusa (changes requested):
Thanks for the thorough write-up and the live-cluster reproduction, your read on the SSA mechanism is correct.

I don't think a `fieldManager` parameter is the right fix though, and I'd like to redirect this before it goes further. What you're hitting is Server-Side Apply working as designed: `resources_create_or_update` does a force apply under a single field manager, the same thing `kubectl apply --server-side --force-conflicts` does, so when the second call omits a field the first call set, SSA removes it. The tool is being used as a partial patch when its contract is "this is the complete desired state."

A `fieldManager` parameter doesn't really solve that. The default stays the same, so the abandonment still happens out of the box unless the model passes distinct names, and reasoning about SSA ownership per call is fragile. It also leaves orphan `managedFields` entries behind and makes it impossible to ever remove a field through the tool. We already have the `ssa-field-preservation` eval from #1166 covering this, and the direction it encodes is that the model should re-fetch and re-apply the full object, not juggle field managers.

The real gap is that the tool definition never tells the model the apply semantics. So instead of adding a parameter, I'd like this PR to fix the wording. To be precise about what I'm asking for, the final PR should contain only these changes:

**Revert back to `main` (drop all the `fieldManager` plumbing):**
- `pkg/kubernetes/pods.go` — back to `c.resourcesCreateOrUpdate(ctx, toCreate)`
- `pkg/kubernetes/resources.go` — restore both signatures (`ResourcesCreateOrUpdate(ctx, resource string)` and `resourcesCreateOrUpdate(ctx, resources []*unstructured.Unstructured)`) and the `FieldManager: version.BinaryName` apply option
- `pkg/toolsets/kubevirt/vm/create/tool.go` — back to `ResourcesCreateOrUpdate(params, vmYaml)`
- `pkg/mcp/resources_test.go` — drop the new `fieldManager` tests (`TestResourcesCreateOrUpdateSSAFieldAbandonment` and the non-string-`fieldManager` case)
- `pkg/toolsets/core/resources.go` — remove the `fieldManager` input property, the handler block that reads it, and the now-unused `version` import

**The only real change, in `pkg/toolsets/core/resources.go`:**
- Tool `Description` becomes (keep the existing `"\n" + commonApiVersion` suffix as is):
  > Create or update a Kubernetes resource via Server-Side Apply. The manifest is the complete desired state: any field this tool previously set and the new manifest omits is removed. To edit an existing resource, fetch it with resources_get, modify it, then re-apply the full resource.
- The `resource` param `Description` becomes:
  > Complete YAML or JSON representation of the Kubernetes resource (full desired state, not a partial patch). Include apiVersion, kind, metadata, and the full spec.

**Then regenerate the generated files:**
- Run `make update-readme-tools` to refresh `README.md` and the `pkg/mcp/testdata/toolsets-*.json` snapshots. After this, the only diff in those four JSON files should be the changed `resources_create_or_update` description and `resource` text, with no `fieldManager` property.

Net result is a two-string change plus regenerated snapshots, no API surface change and no behavior change. Happy to help if anything is unclear, and thanks again for the contribution.

lunarwhite:
Hi @manusa, I'm trying to understand each [eval regression](https://github.com/containers/kubernetes-mcp-server/actions/runs/28033201080) from the log

> - filter-namespace-by-name: PASSED → FAILED - verification failed: verify[0] failed: script execution failed: exit status 1

The agent correctly used `namespaces_list` with `fieldSelector=metadata.name=ns-beta` and got the right answer, but the verify script failed because the agent's output includes table headers and prose that don't contain "ns-beta". I can see same result running it against the main branch.

> - setup-dev-cluster: PASSED → FAILED - Too many tool calls: expected <= 20, got 42

The `maxToolCalls: 20` budget perhaps too tight for a "hard" task creating multiple resources. But I have no idea where should we enhance.

Seemingly neither of them was introduced by this PR's tool description change. Maybe we should we fix them as followup if needed?

manusa:
Hi, @Cali0707 
Could you please look into this? these regressions all seem like false positives, but before changing anything in the core toolset we need to have a valid evals run.

Cali0707 (approved):
LGTM, thanks @lunarwhite

PR #1122, opened 2026-04-30
fix(validation)!: reject kubeconfig auth mode with require_oauth at startup

nader-ziada:
`cluster_auth_mode="kubeconfig"` with `require_oauth=true` is an invalid combination: all users would share a single cluster identity, breaking per-user audit trails. Previously this failed at runtime with a confusing `oauth token required` error. Catch it at config validation with a clear message instead.

PR #1187, opened 2026-06-08
docs: remove obsolete ServiceAccount example from Entra ID setup

cajieh:
The "With ServiceAccount Credentials" section in the Entra ID guide showed `require_oauth = true` with `cluster_auth_mode = "kubeconfig"` — a combination #1122 deliberately rejects at startup (a shared ServiceAccount collapses all authenticated users to one cluster identity, breaking per-user audit trails).

Removing only the OAuth lines would leave an unauthenticated, ServiceAccount-privileged HTTP endpoint in an authentication guide. This PR instead removes the obsolete section entirely (including the "Creating a ServiceAccount Kubeconfig" subsection) and adds a note pointing readers whose clusters don't accept Entra ID tokens directly to the supported On-Behalf-Of token exchange flow, which preserves per-user identity.

Refs #1122

### DIFF
--- a/docs/ENTRA_ID_SETUP.md
+++ b/docs/ENTRA_ID_SETUP.md
@@ -119,67 +119,12 @@ Replace:
 >
 > In `passthrough` mode, if token exchange is configured (`token_exchange_strategy` or `sts_audience`), the token is exchanged before being passed to the cluster.
 
-### With ServiceAccount Credentials
-
-If your Kubernetes cluster doesn't accept Entra ID tokens on the API server, use this configuration:
-
-```toml
-# Use kubeconfig ServiceAccount credentials for cluster access
-cluster_auth_mode = "kubeconfig"
-kubeconfig = "/path/to/sa-kubeconfig"
-```
-
-This setup:
-- **Cluster access uses ServiceAccount token** (from kubeconfig)
-
-> **Note:** `require_oauth = true` is incompatible with `cluster_auth_mode = "kubeconfig"` and is rejected at startup.
-
-#### Creating a ServiceAccount Kubeconfig
-
-Your regular kubeconfig likely uses interactive login. Create a kubeconfig with a static ServiceAccount token:
-
-```bash
-# Create ServiceAccount
-kubectl create sa mcp-server -n default
-
-# Grant permissions (adjust role as needed)
-kubectl create clusterrolebinding mcp-server-reader \
-  --clusterrole=view \
-  --serviceaccount=default:mcp-server
-
-# Create a token (adjust duration to your security requirements)
-kubectl create token mcp-server -n default --duration=720h > sa-token
-
-# Create kubeconfig with the token
-export SA_TOKEN=$(cat sa-token)
-export CLUSTER_URL=$(kubectl config view --minify -o jsonpath='{.clusters[0].cluster.server}')
-export CLUSTER_CA=$(kubectl config view --raw --minify -o jsonpath='{.clusters[0].cluster.certificate-authority-data}')
-
-cat > mcp-kubeconfig << EOF
-apiVersion: v1
-kind: Config
-clusters:
-- cluster:
-    certificate-authority-data: ${CLUSTER_CA}
-    server: ${CLUSTER_URL}
-  name: cluster
-contexts:
-- context:
-    cluster: cluster
-    user: mcp-server
-  name: mcp-context
-current-context: mcp-context
-users:
-- name: mcp-server
-  user:
-    token: ${SA_TOKEN}
-EOF
-```
-
-Then run:
-```bash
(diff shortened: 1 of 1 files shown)


nader-ziada:
Hey @cajieh, thanks for the fix, can you please sign the commit to fix the DCO error

```
There is one commit incorrectly signed off. This means that the author of this commit failed to include a Signed-off-by line in the commit message.

To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).

Here is how to fix the problem so that this code can be merged.
```

cajieh:
> Hey @cajieh, thanks for the fix, can you please sign the commit to fix the DCO error
> 
> ```
> There is one commit incorrectly signed off. This means that the author of this commit failed to include a Signed-off-by line in the commit message.
> 
> To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).
> 
> Here is how to fix the problem so that this code can be merged.
> ```

@nader-ziada Done, thank you.

manusa (changes requested):
Thanks for catching this — the example genuinely was broken, and your diagnosis (rejected at startup) is correct. I'd like to take a different approach to the fix, though, because removing only the OAuth lines leaves the example in a worse state than before. Requesting changes.

### Root cause (for context)
This section was added in #953 with `require_oauth = true` + `cluster_auth_mode = "kubeconfig"`, which was valid at the time. #1122 (`fix(validation)!: reject kubeconfig auth mode with require_oauth at startup` — a breaking change) deliberately made that combination invalid: a single shared ServiceAccount collapses all authenticated users to one cluster identity and breaks per-user audit trails (see the error in `pkg/config/config.go`). The doc was simply never updated afterward, which is the bug this PR is addressing.

### Why removing just the OAuth lines isn't the right fix
With `require_oauth` gone, the example now contains no Entra ID / OAuth configuration at all — yet it lives in the Entra ID setup guide. More importantly, this guide targets HTTP mode throughout (OAuth, MCP Inspector redirect URIs, `/.well-known` discovery), and in HTTP mode `require_oauth = false` leaves the MCP endpoint unauthenticated. From the middleware contract in `pkg/http/authorization.go`: *"requireOAuth is false: the OAuth flow is skipped, and the server is effectively unprotected."*

So the revised example turns a fail-closed misconfiguration (server refuses to start) into a fail-open one — a reader following an authentication guide would deploy an open, ServiceAccount-privileged HTTP endpoint, with nothing in the section warning them.

### The scenario is already covered
This section's premise — "your cluster doesn't accept Entra ID tokens on the API server" — is exactly what the next section, **With Token Exchange (On-Behalf-Of Flow)**, solves in the supported, identity-preserving way (it exchanges the user's Entra token for a downstream token the cluster accepts). The ServiceAccount approach is obsolete post-#1122.

### Requested change
Please **remove the `### With ServiceAccount Credentials` section entirely**, including its `#### Creating a ServiceAccount Kubeconfig` subsection — i.e. everything from `### With ServiceAccount Credentials` up to (but not including) `### With Token Exchange (On-Behalf-Of Flow)`. Optionally, add a one-line pointer telling readers whose clusters don't accept Entra ID tokens to use On-Behalf-Of token exchange.

### PR metadata
- The `Fixes #328` reference is incorrect — #328 is an unrelated, already-merged PR (`test(mcp): refactor events toolset tests`). The change this follows from is **#1122**. Please reference that instead in both the commit message and the PR description (e.g. "Follow-up to #1122" / "Refs #1122" — there's no tracking issue to "Fix").
- Please also update the PR title/description to reflect the new approach (removing the obsolete section, rather than "fixing the unsupported combination").

manusa:
Pushed a maintainer commit to take this over the line. Instead of dropping only the OAuth lines, I removed the `### With ServiceAccount Credentials` section entirely (including "Creating a ServiceAccount Kubeconfig").

Rationale: the example combined `require_oauth = true` with `cluster_auth_mode = "kubeconfig"`, which #1122 rejects at startup (a shared ServiceAccount collapses all authenticated users to one cluster identity). Keeping the section with `require_oauth` removed would have left an unauthenticated, ServiceAccount-privileged HTTP endpoint documented in an authentication guide — trading a fail-closed misconfiguration for a fail-open one. It's replaced with a short note that states the constraint and points readers to the supported On-Behalf-Of token exchange flow, which preserves per-user identity.

I also updated the title/description: the original `Fixes #328` was incorrect (#328 is an unrelated merged PR), so this is now a follow-up to #1122.

Thanks @cajieh for catching the broken example!

manusa (approved):

PR #1188, opened 2026-06-08
[WIP] feat: add SupportsGVKs method to Provider interface

cajieh:
## Summary

Add `SupportsGVKs(gvks []schema.GroupVersionKind) bool` to the cluster provider interface. This method reports whether every GVK is available on every target exposed by the provider.

All existing providers implement a no-op that returns `true`, so current behavior is preserved. This allows opt-in GVK discovery per provider in future tickets.

## Implementations

| Provider | Behavior |
|----------|----------|
| singleClusterProvider | returns true (no-op) |
| kubeConfigClusterProvider | returns true (no-op) |
| acmHubClusterProvider | returns true (no-op) |
| kcpClusterProvider | returns true (no-op) |
| tokenExchangingProvider | delegates to wrapped provider |

## Tests

- `pkg/kubernetes/provider_supports_gvks_test.go` - single, kubeconfig, acm no-op; token exchange delegation
- `pkg/kcp/provider_supports_gvks_test.go` - kcp no-op

## Enables future work

- OCPMCP-325: GVK annotations on tools
- OCPMCP-304: Dynamic tool list filtering


cajieh:
Closing in favor of #1189 which has the correct commit history (single commit based on containers/main)

PR #1189, opened 2026-06-08
feat: add HasGVKs method to Provider interface

cajieh:
## Summary
- Add `HasGVKs(gvks []schema.GroupVersionKind) bool` to the `Provider` interface.
  Reports whether every GVK in `gvks` is available on at least one target exposed by the provider.
- All existing provider implementations are no-ops that return `true`, preserving current tool visibility.
- Providers can opt in to GVK discovery by implementing real logic later.

## Test plan
- [x] All existing tests pass
- [x] Build and lint pass

Implements: OCPMCP-328

Co-Authored-By: claude

### DIFF
--- a/pkg/kubernetes/provider.go
+++ b/pkg/kubernetes/provider.go
@@ -6,6 +6,7 @@ import (
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"github.com/containers/kubernetes-mcp-server/pkg/oauth"
 	"github.com/containers/kubernetes-mcp-server/pkg/tokenexchange"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 // McpReload is a function type that defines a callback for reloading MCP toolsets (including tools, prompts, or other configurations)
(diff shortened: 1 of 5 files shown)


pkg/kcp/provider_supports_gvks_test.go
```diff
@@ -0,0 +1,38 @@
+package kcp
```
Cali0707: let's not add this whole new test file for now, since it is just testing a "default value" from the method

pkg/kubernetes/provider.go
```diff
@@ -31,6 +32,10 @@ type Provider interface {
 	// WatchTargets sets up a watcher for changes in the cluster targets and calls the provided McpReload function when changes are detected
 	WatchTargets(reload McpReload)
 	Close()
+	// SupportsGVKs reports whether every GVK in gvks is available on every target
+	// exposed by this provider. Providers that have not opted in to GVK discovery
+	// should return true so existing tools remain visible.
+	SupportsGVKs(gvks []schema.GroupVersionKind) bool
```
Cali0707: Let's change the semantics here a bit:
```suggestion
	// HasGVKs reports whether every GVK in gvks is available on at least one target target
	// exposed by this provider. Providers that have not opted in to GVK discovery
	// should return true so existing tools remain visible.
	HasGVKs(gvks []schema.GroupVersionKind) bool
```

Cali0707 (approved):
LGTM, thanks @cajieh !

Issue #239, opened 2025-08-05
mcp server not killed correctly when started with npx


feloy:
On Mac:

```
% npx kubernetes-mcp-server@latest --port 8080
```

```
% ps
32354 ttys003    0:00.35 npm exec kubernetes-mcp-server@latest --port 8080    
32370 ttys003    0:00.03 node /Users/phmartin/.npm/_npx/acdc49484db51b02/node_modules/.bin/kubernetes-mcp-server --port 8080
32371 ttys003    0:00.04 /Users/phmartin/.npm/_npx/acdc49484db51b02/node_modules/kubernetes-mcp-server-darwin-arm64/bin/kubernetes-mcp-server-darwin-arm64 --port 8080


% kill 32354
% ps
32371 ttys003    0:00.04 /Users/phmartin/.npm/_npx/acdc49484db51b02/node_modules/kubernetes-mcp-server-darwin-arm64/bin/kubernetes-mcp-server-darwin-arm64 --port 8080

```

Note that, started without any flag, the server is stopped:

```
% npx kubernetes-mcp-server@latest
```

```
% ps
32840 ttys003    0:00.35 npm exec kubernetes-mcp-server@latest  
32856 ttys003    0:00.04 node /Users/phmartin/.npm/_npx/acdc49484db51b02/node_modules/.bin/kubernetes-mcp-server
32857 ttys003    0:00.03 /Users/phmartin/.npm/_npx/acdc49484db51b02/node_modules/kubernetes-mcp-server-darwin-arm64/bin/kubernetes-mcp-server-darwin-arm64

% kill 32840
% ps
```

I changed the command started by execFileSync at https://github.com/containers/kubernetes-mcp-server/blob/main/npm/kubernetes-mcp-server/bin/index.js#L24C14-L24C26 with a shell command (`sh` and `sh -c "sleep 100s"`), and the same problem happens.

PR #1191, opened 2026-06-08
fix(http): SIGHUP reloads rather than terminating

2uasimojo:
HTTP mode was erroneously trapping SIGHUP for shutdown rather than letting the config reload handler do its thang. Add a test case exposing the hole, and fix it.

Co-Authored-By: claude

### DIFF
--- a/pkg/http/http.go
+++ b/pkg/http/http.go
@@ -151,7 +151,7 @@ func Serve(ctx context.Context, mcpServer *mcp.Server, cfgState *config.StaticCo
 	defer cancel()
 
 	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGHUP, syscall.SIGTERM)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
 
 	serverErr := make(chan error, 1)
 	go func() {
--- a/pkg/kubernetes-mcp-server/cmd/root_sighup_test.go
+++ b/pkg/kubernetes-mcp-server/cmd/root_sighup_test.go
@@ -3,6 +3,9 @@
 package cmd
 
 import (
+	"context"
+	"fmt"
+	"net/http"
 	"os"
 	"path/filepath"
 	"slices"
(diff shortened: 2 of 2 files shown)


2uasimojo:
Provenance: #241, https://github.com/containers/kubernetes-mcp-server/pull/524/changes#diff-4097c7955abe165c7962da05d47ae37ea206d89c534fa2539721e87d79e70d65

pkg/kubernetes-mcp-server/cmd/root_sighup_test.go
```diff
@@ -372,3 +394,167 @@ func TestSIGHUPInvokesLogSinkReload(t *testing.T) {
...
+
+func (s *HTTPSIGHUPSuite) SetupTest() {
+	s.baseSIGHUPSetup = setupSIGHUPTest(s.T())
+	s.configPath = filepath.Join(s.tempDir, "config.toml")
+	http.DefaultClient.Timeout = 10 * time.Second
+}
+
+func (s *HTTPSIGHUPSuite) TearDownTest() {
+	if s.stopServer == nil {
+		return
```
nader-ziada, line 420: TearDownTest returns early when stopServer == nil, skipping s.teardown()

maybe dd defer s.teardown() at top.

pkg/kubernetes-mcp-server/cmd/root_sighup_test.go
```diff
@@ -372,3 +394,167 @@ func TestSIGHUPInvokesLogSinkReload(t *testing.T) {
...
+	httpAddress     string
+	timeoutCancel   context.CancelFunc
+	stopServer      context.CancelFunc
+	waitForShutdown func() error
+}
+
+func (s *HTTPSIGHUPSuite) SetupTest() {
+	s.baseSIGHUPSetup = setupSIGHUPTest(s.T())
+	s.configPath = filepath.Join(s.tempDir, "config.toml")
+	http.DefaultClient.Timeout = 10 * time.Second
```
nader-ziada: Mutates global http.DefaultClient.Timeout without restoring it. Use a local &http.Client{Timeout: 10*time.Second} instead.

nader-ziada (commented):
@2uasimojo thanks for fixing! 

a couple of minor comments

pkg/kubernetes-mcp-server/cmd/root_sighup_test.go
```diff
@@ -372,3 +394,167 @@ func TestSIGHUPInvokesLogSinkReload(t *testing.T) {
...
 }
+
+// HTTPSIGHUPSuite tests the SIGHUP configuration reload behavior for the HTTP server
+type HTTPSIGHUPSuite struct {
+	suite.Suite
+	*baseSIGHUPSetup
+	configPath      string
+	httpAddress     string
+	timeoutCancel   context.CancelFunc
+	stopServer      context.CancelFunc
```
Cali0707: later (line 497) this seems to be assigned to a `func()` that sends the SIGTERM signall rather than the cancel func from a context, maybe we can type it as just a `func()` to be clearer?

2uasimojo:
@Cali0707 @nader-ziada Thanks for the reviews! Addressed those issues and rebased. Should be ready for another look.

(Sorry, usually I would do the rebase and the fixups in separate pushes for ease of review. Here's a [link](https://github.com/containers/kubernetes-mcp-server/compare/b1d9b537715fb6cb4a304a221a5f3b3c36c89644..221dd3d9455e64e4f215cb8ec1755699aa854bd5#diff-8c31e94eb37379ecfc7b3801cb7ce3f5f957e34c39aeca3980298e4525e3df9d) to the one file I actually touched, for convenience.)

Cali0707 (approved):
LGTM

manusa:
> Provenance: #241, https://github.com/containers/kubernetes-mcp-server/pull/524/changes#diff-4097c7955abe165c7962da05d47ae37ea206d89c534fa2539721e87d79e70d65

I believe this predates the time when dynamic config support was implemented via SIGHUP (#524)

Good catch. Reviewing, mostly concerned about the intersection with npx/JS wrapper execution)

pkg/http/http.go
```diff
@@ -151,7 +151,7 @@ func Serve(ctx context.Context, mcpServer *mcp.Server, cfgState *config.StaticCo
 	defer cancel()
 
 	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGHUP, syscall.SIGTERM)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
```
manusa, line 154: Note this is the same line #241 added (so an npx-forwarded signal would stop the HTTP server). Removing SIGHUP here is correct for the config-reload case, but it leaves one edge: in HTTP mode started **without** a config file (`--port` alone — exactly the #239 scenario), the reload handler at `root.go:364` is never registered, so nothing catches SIGHUP and Go's default disposition **terminates the process abruptly**, bypassing `httpServer.Shutdown`/`mcpServer.Shutdown`. That still kills the child — so #239's orphan bug does *not* come back — but it's an abrupt-vs-graceful change, and it contradicts `docs/configuration.md:99` ("If neither is specified, SIGHUP signals are ignored"). Suggest either updating that doc line, or draining SIGHUP with a no-op in `Serve` so the no-config case stays truly ignored (the reload handler still gets its own copy when a config is present).

manusa (commented):
Verified the one-line fix against the SIGHUP double-handling bug — it's correct: removing SIGHUP from `Serve`'s `signal.Notify` leaves the config-reload handler (`root.go:385`) as the sole SIGHUP consumer in HTTP+config mode, where the shutdown path previously raced it and won. The new `HTTPSIGHUPSuite` is a real regression guard (fails on the old code, passes on the fix). 👍

Confirmed this does **not** regress the npx orphan fix (#239 / #241): #241's actual fix is the npm wrapper's `spawn` + signal forwarding, which this PR leaves untouched, and the #239 repro uses plain `kill` (SIGTERM) — `Serve` still handles SIGINT/SIGTERM, so the child still shuts down and is never orphaned. Even a SIGHUP forwarded to a no-config HTTP server still terminates the child (Go's default disposition), so no orphan returns.

One edge worth resolving inline (the no-config HTTP case). A few minor test-polish items (dead errgroup timeout context / process-wide SIGTERM teardown, the fixed 100 ms sleep, the `len() >` assertion) are non-blocking — happy to file a follow-up rather than hold the merge.

manusa (approved):
LGTM, filing a follow-up issue for the raised points in the previous comment.

PR #1193, opened 2026-06-09
feat: implement HasGVKs for single cluster provider

shahsahil264:
## Summary
- Implement real `HasGVKs` logic for `singleClusterProvider`, replacing the no-op stub that always returned `true`.
- Uses the cluster's REST mapper to verify that every requested GVK is available on the cluster.
- Includes a nil guard on the manager to safely return `true` (keeping tools visible) if the provider is not yet initialized.

## Test plan
- [x] Unit tests covering nil, empty, single valid, all valid, single invalid, and mixed GVK lists
- [x] All existing tests pass

Implements: OCPMCP-327

### DIFF
--- a/pkg/kubernetes/provider_single.go
+++ b/pkg/kubernetes/provider_single.go
@@ -122,6 +122,19 @@ func (p *singleClusterProvider) Close() {
 	}
 }
 
-func (p *singleClusterProvider) HasGVKs(_ []schema.GroupVersionKind) bool {
+// HasGVKs uses the cluster's REST mapper to verify that every requested GVK is available.
+func (p *singleClusterProvider) HasGVKs(gvks []schema.GroupVersionKind) bool {
+	if len(gvks) == 0 {
+		return true
+	}
+	if p.manager == nil {
+		return true
+	}
+	mapper := p.manager.kubernetes.RESTMapper()
+	for _, gvk := range gvks {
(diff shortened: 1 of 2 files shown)


shahsahil264:
@Cali0707 PTAL
Thanks!

pkg/kubernetes/provider_single.go
```diff
@@ -122,6 +122,19 @@ func (p *singleClusterProvider) Close() {
...
 
-func (p *singleClusterProvider) HasGVKs(_ []schema.GroupVersionKind) bool {
+// HasGVKs uses the cluster's REST mapper to verify that every requested GVK is available.
+func (p *singleClusterProvider) HasGVKs(gvks []schema.GroupVersionKind) bool {
+	if len(gvks) == 0 {
+		return true
+	}
+	if p.manager == nil {
+		return true
+	}
```
Cali0707, line 134: @shahsahil264 can you add a Warning level log here to tell the user that this proceeded assuming tall GVKs are available since the manager was not initialized?

Cali0707 (commented):
Overall this looks great @shahsahil264 !

Just one small comment about needing a log, then lets get this PR in!

pkg/kubernetes/provider_single.go
shahsahil264: Sure! Thanks for the review

Cali0707 (approved):
LGTM

PR #1196, opened 2026-06-10
feat!: target-specific tool filtering

2uasimojo:
- We will now run potentially-expensive per-target queries to decide whether particular tools should be exposed.
- The `Openshift` interface is no more. It was being used at several levels of execution which called for different behaviors/semantics. It is replaced by `HasGVKs()` and `AnyTargetHasGVKs()` -- see below.
- `HasGVKs()` is a new low-level utility function that operates at the scope of a single discovery client (talking to a single cluster). It treats `404` errors appropriately ("does not have the requested GVKs") but bubbles other errors up to the caller. It directly replaces the `IsOpenshift()` (note: lowercase `shift`, not the same as `IsOpenShift()`) utility in the `ClusterState` watcher, and also serves as the workhorse within...
- `AnyTargetHasGVKs()` is a higher-level method on a new `FilteringProvider` interface.
- `FilteringProvider` takes the place of the `Openshift` interface as composed into the various `Provider` implementations. It is named/positioned to be extensible to more toolset filters in the future.
- `ProviderGVKFilter` is the equivalent of an "abstract class": it satisfies `FilteringProvider` by providing a concrete implementation of `AnyTargetHasGVKs()` that iterates over targets retrieved from a `ManagerProvider`, asking `HasGVKs()` of each of them. Importantly, it contains logic for how to handle edge cases (e.g. errors from `HasGVKs()`). Specific `Provider` implementations take advantage of this by embedding `ProviderGVKFilter` and supplying their specific implementation of `ManagerProvider`.
- The erstwhile `Provider` interface is split up. Currently all `Provider`s are "kubernetes Providers"; this division anticipates higher level Providers like ACM. Methods applicable to kubernetes environments remain in `kubernetes.Provider`, while higher-level methods are moved to a new `provider.Provider` interface.
- `ManagerProvider` is a new, thin interface that provides the glue between a `kubernetes.Provider` and the existing implementation of `FilteringProvider`. Its implementors must supply the `GetTargetManagers()` function that allows `ProviderGVKFilter()` to iterate over targets.
- `ServerTool` grows a new field, `TargetCompatibilityFilters`, a slice of (closure, so they can use a `FilteringProvider` without accepting it as a param) functions each returning a `bool`. ~~When `enable_target_compatibility_tool_filters` is on, the high-level `isToolApplicable()` filter invokes those funcs to decide whether to expose the tool.~~ The high-level `isToolApplicable()` filter **always** invokes those funcs to decide whether to expose the tool _(the `enable_target_compatibility_tool_filters` config option was later removed — see commit 6fe306dc; filtering is now unconditional)_. This abstraction is extensible for arbitrary future target-specific filters; but for now we register a specific use of `AnyTargetHasGVKs()` that looks for the Route Kind to determine whether to filter the `projects_list` tool. This replaces the previous special-case `IsOpenShift()` conditional.
- On the testing side, ~~the switch from `IsOpenShift()` to `HasGVKs()` rendered the existing `MockServer` insufficient for certain suites. It is replaced by `envtest` for these cases.~~ the discovery-only provider suites are still served by `MockServer` (its discovery handler now returns a proper `404` for absent GroupVersions); only the `mcp` suites, which exercise real resource CRUD, use `envtest`. The existing `envtest` helpers are refactored so they can be shared by these suites and the existing consumers. _(Updated after commit 3a1370d7 reverted the `pkg/kubernetes` provider suites to `MockServer`, so only one envtest control plane runs; two running concurrently were OOM-killing the macos-15-intel CI job.)_

Behaviorally:
- Whereas previously we would *always* query *one* target for `IsOpenShift()` and filter according to that one target's response, we now properly query all targets for the `Provider`.
- We are permissive by default. We behave as if the target *is* OpenShift if:
  - *any* target reports that it has the Route Kind.
  - we experience any errors attempting to query the targets.

I.e. we will only *hide* a tool if all of the following are true:
- All targets were queried successfully
- All targets reported `HasGVKs(Route): false`

Signed-off-by: Eric Fried <efried@redhat.com>
Co-Authored-By: claude


pkg/toolsets/core/resources.go
```diff
@@ -12,11 +12,27 @@ import (
...
+		for _, m := range managers {
+			if m.IsOpenShift(context.Background()) {
+				isOpenShift = true
+				break
+			}
+		}
+	}
+	if isOpenShift {
 		commonApiVersion += ", route.openshift.io/v1 Route"
 	}
```
Cali0707, line 21: @2uasimojo maybe we can handle this more simply by wiring the `HasGVKs` method through to here instead of your new `ManagerProvider` interface?

Then we can lean on the built in support within the different providers for checking that efficiently, by just looking up the GVK for a Route?

pkg/provider/provider.go
```diff
@@ -0,0 +1,22 @@
...
+	"context"
+)
+
+// TargetManager represents a manager for a single cluster target.
+// It provides target-specific discovery capabilities for things
+// like tool filtering.
+type TargetManager interface {
+	// IsOpenShift reports whether the target is an OpenShift cluster.
+	IsOpenShift(ctx context.Context) bool
+}
```
Cali0707, line 19: IMO we can probably get rid of having `IsOpenShift` checked on any cluster through leveraging the `HasGVKs` method a bit better

So perhaps we don't need this?

pkg/api/toolsets.go
```diff
@@ -14,6 +15,9 @@ type ServerTool struct {
 	Handler            ToolHandlerFunc
 	ClusterAware       *bool
 	TargetListProvider *bool
+	// TargetCompatibilityFilters is a slice of closures, each of which answers whether the ServerTool
+	// should be exposed (true) or not (false).
+	TargetCompatibilityFilters []func() bool
```
Cali0707, line 19: Maybe we can instead go with the annotations mechanism we had discussed instead? We can extend the `ToolAnnotation` struct with a new field that has a list of required GVKs - IMO that would meet our current use cases.

If we need specific callbacks in the future, we could add this back?

Cali0707 (commented):
Thanks for tackling this @2uasimojo 

I really like that we are getting close to getting rid of the `IsOpenShift` checks. Since that is only being used to enable tools/hints for specific APIs, maybe we can lean into the annotations approach we discussed offline rather than having an interface to iterate over all the managers from a provider?

pkg/toolsets/core/resources.go
cajieh: When filtering is off we expose all tools, but this block still queries every target to add Route hints to `resources_* descriptions`. is this intentional , or should description enrichment follow the same gate as tool visibility?

pkg/kubernetes/provider_kubeconfig.go
```diff
@@ -132,20 +133,23 @@ func (p *kubeConfigClusterProvider) managerForContext(context string) (*Manager,
...
-	if m == nil {
-		return false
-	}
-	return m.IsOpenShift(ctx)
+	defer p.mu.RUnlock()
+	return len(p.managers) > 1
 }
 
-func (p *kubeConfigClusterProvider) IsMultiTarget() bool {
+func (p *kubeConfigClusterProvider) getTargetsAndManagers() ([]string, []provider.TargetManager, error) {
```
cajieh: Is this used anywhere?

pkg/kubernetes/provider_kubeconfig.go
```diff
@@ -159,6 +163,23 @@ func (p *kubeConfigClusterProvider) GetTargets(_ context.Context) ([]string, err
 	return contextNames, nil
 }
 
+func (p *kubeConfigClusterProvider) GetTargetManagers(_ context.Context) ([]provider.TargetManager, error) {
+	contextNames, err := p.GetTargets(context.Background())
```
cajieh: Wondering why this uses `context.Background()` instead of the passed-in `ctx` — intentional?

pkg/config/config.go
```diff
@@ -154,6 +154,12 @@ type StaticConfig struct {
 	// Defaults to false.
 	ValidationEnabled bool `toml:"validation_enabled,omitempty"`
 
+	// EnableTargetCompatibilityToolFilters enables filtering of tools based on
+	// cluster target compatibility (e.g., hiding OpenShift-specific tools when
+	// connected to a non-OpenShift cluster).
+	// Defaults to false.
+	EnableTargetCompatibilityToolFilters bool `toml:"enable_target_compatibility_tool_filters,omitempty"`
```
cajieh: Should we add config tests for `EnableTargetCompatibilityToolFilters`?

cajieh (commented):
Added a few comments to help me follow and understand the changes. 
Thank you. 👍

pkg/toolsets/core/resources.go
2uasimojo: To be clear, HasGVKs() is just a slightly higher-level incarnation of IsOpenShift(), right? I.e.
- It's still asking about a single target; so how would we address multiple targets -- or more generally, different Provider abstractions -- in this context?
- It doesn't inherently indicate whether it's "all" or "any"; or how it should behave if some targets are currently unreachable; or how it should respond to errors.

More on this below.

pkg/provider/provider.go
2uasimojo: Yup, I was thinking in the next card/PR/iteration to simply replace IsOpenShift() with HasGVKs(routev1) or similar. But perhaps I'm misunderstanding the intent of HasGVKs...

pkg/toolsets/core/resources.go
Cali0707: @2uasimojo my thinking is that HasGVKs will abstract away actually reaching out to all the different targets into the provider layer (which IMO is the correct place to manage making calls to multiple targets)

pkg/provider/provider.go
Cali0707: I think HasGVKs(routev1) is exactly what I am thinking :)

pkg/api/toolsets.go
2uasimojo: I think you're saying that by wiring in ManagerProvider=>TargetManager at such a high level, I'm restricting e.g. the ACM provider's ability to do something more abstract than querying each of its targets separately? Which it may of course do under the covers by e.g. consulting a local cache; but I'm tying its hands from being able to answer the question more generally?

The part I'm struggling with is that "Is it OpenShift?" or "Does it have the Route kind?" are very single-target-oriented. I pondered for a while how the question could be phrased (i.e. a sane function signature) such that it would be both abstract/generic enough and specific enough. Even if we're *just* looking at "is it OpenShift?" we would have to phrase it like, "Are any of the targets under your purview OpenShift? But also if you can't answer that right now for a specific target, you should assume yes." Zoom out to other potential future questions and that semantic doesn't necessarily hold ("Does it have such-and-such security feature? If you can't answer assume *no*.")

And this is why I wanted to find a way for that logic to live close to where it's used. Like in this case we can only expect the `projects_list` tool to understand how it should filter itself based on openshift-ness. But that same logic doesn't necessarily apply to other characteristics. So if we wanted to do this with an annotation (that wasn't itself a callback) we would be opening the gate to a proliferation of future annotations, all with unique semantics.

Now, if we're talking about DRYing logic like "if any target is openshift; if errors assume yes" that can still totally be done -- the TargetCompatibilityFilters entry can 100% call out to a shared func somewhere.

pkg/toolsets/core/resources.go
2uasimojo: You're right, I need to figure out how to wire the config option into here.

...though I sure would like to understand the point of this section better first (see TODO a few lines up).

pkg/kubernetes/provider_kubeconfig.go
2uasimojo: Nope. I had in fact removed it locally but forgot to check that change in. Good eye :)

pkg/kubernetes/provider_kubeconfig.go
2uasimojo: Yup, good catch, will fix.

pkg/config/config.go
2uasimojo: There are tests that exercise it in context -- i.e. turn it on or off and inspect the resulting tool lists (toolsets_test.go, mcp_watch_test.go). Is that sufficient or do you see value in a narrower test, e.g. in config_test.go? What would it do?

pkg/api/toolsets.go
2uasimojo: Summary of offline discussion:
- s/IsOpenShift()/HasGVKs(routev1)/g
- Iterating over targets should be invisible at the tool(set) level, hidden under Provider (because it'll be n/a for some Providers, e.g. ACM with Search). Possibility: Provider.WhateverWeCallIt() returns a slice of `true`/`false`/`error` for each target, and the tool(set) logics that to decide whether to filter.

pkg/provider/provider.go
2uasimojo: Subcategories of `provider.Provider` (this interface) will need to implement `FilteringProvider` to supply `AnyTargetHasGVKs()`, as demonstrated in `kubernetes.Provider`. While all `kubernetes.Provider`z can share an implementation by embedding `kubernetes.ProviderGVKFilter`, we expect other `provider.Provider`z (like ACM) to have different implementations.

Although that's a lot of pieces flying around, it seemed like the cleanest division of responsibility while allowing the consumers (like tool filtering) to be agnostic of those details.

pkg/toolsets/core/resources.go
2uasimojo: Right, so we've split into `HasGVKs()` which talks to one target and doesn't prescribe what happens on errors; and `AnyTargetHasGVKs()` which is higher level, talks to all the targets (or, in the future, ACM's search rollup, etc.), and has specific behaviors for edge cases. Here we're using the latter. Sane?

internal/test/envtest.go
2uasimojo, line 1: ntr: This should be mostly a lift-and-shift from pkg/mcp, like [common_test.go](https://github.com/containers/kubernetes-mcp-server/pull/1196/changes#diff-064036c629017da64e4638eed7f2ca7be9ae9754f38af1d74cb0d36ba62a7e40). Moved up so it could be shared without import cycles.

While I carefully scrutinized generated changes on the prod side, I did *not* inspect this stuff line by line.

pkg/api/toolsets.go
2uasimojo: > * Iterating over targets should be invisible at the tool(set) level

This is now the case...

> * , hidden under Provider

And this almost is -- it's FilteringProvider, which must be implemented by any Provider that gets here.

> a slice of `true`/`false`/`error` for each target

Didn't end up doing it this way. Instead I split "has GVKs" into two levels as described [here](https://github.com/containers/kubernetes-mcp-server/pull/1196/changes#r3482350688), and embedded the logic for "what do we do about errors" and other edge cases into the higher-level one.

> the annotations mechanism we had discussed

Also didn't do this just yet. See how you feel about the current (more generalized/extensible) approach now. If you still want to move to annotations, we can.

pkg/kubernetes/provider_gvk_filter.go
```diff
@@ -0,0 +1,63 @@
...
+// AnyTargetHasGVKs reports whether every GVK in gvks is available on at least one target
+// exposed by this provider. Returns true if an error occurs during discovery to avoid
+// excluding tools due to transient issues.
+func (f *ProviderGVKFilter) AnyTargetHasGVKs(ctx context.Context, gvks []schema.GroupVersionKind) bool {
+	if len(gvks) == 0 {
+		return true
+	}
+
+	logger := klog.FromContext(ctx)
+	mgrs, err := f.managerProvider.GetTargetManagers(context.Background())
```
cajieh: The incoming `ctx` parameter isn't used here  - `context.Background()` is passed instead. Should this propagate the caller's context for cancellation/timeout support and consistency?

pkg/kubernetes/provider_gvk_filter.go
```diff
@@ -0,0 +1,63 @@
...
+		klogutil.LogWarn(logger, "AnyTargetHasGVKs couldn't retrieve target managers; assuming all GVKs are available", klogutil.Err(err))
+		return true
+	}
+
+	for _, mgr := range mgrs {
+		k, err := mgr.Derived(context.Background())
+		if err != nil {
+			// Can't get discovery client; assume target has the GVKs to avoid
+			// hiding tools due to transient errors
+			klogutil.LogWarn(logger, "AnyTargetHasGVKs couldn't derive a Kubernetes interface for a manager (%s); assuming all GVKs are available", klogutil.Err(err))
```
cajieh: nit: It looks like the `%s` will be printed literally instead of substituting the error. Although the error still lands in `exception.message` via `klogutil.Err(err)`, so this is cosmetic only.

pkg/kubernetes/provider_gvk_filter.go
```diff
@@ -0,0 +1,63 @@
...
+			return true
+		}
+
+		hasGVKs, err := api.HasGVKs(k.DiscoveryClient(), gvks)
+		if err != nil {
+			// Discovery error; assume target has the GVKs to avoid hiding tools
+			klogutil.LogWarn(logger, "AnyTargetHasGVKs couldn't query a client (%s); assuming all GVKs are available", klogutil.Err(err))
```
cajieh: The same literal `%s` here as well.

pkg/kubernetes/provider_gvk_filter.go
```diff
@@ -0,0 +1,63 @@
...
+	logger := klog.FromContext(ctx)
+	mgrs, err := f.managerProvider.GetTargetManagers(context.Background())
+	// If an error occurs, don't exclude tools
+	if err != nil {
+		klogutil.LogWarn(logger, "AnyTargetHasGVKs couldn't retrieve target managers; assuming all GVKs are available", klogutil.Err(err))
+		return true
+	}
+
+	for _, mgr := range mgrs {
+		k, err := mgr.Derived(context.Background())
```
cajieh: The incoming `ctx` is ignored here too in favor of `context.Background()`.

cajieh (commented):
Looks good. Added minor nits. I'll defer to @Cali0707  for the final review.

2uasimojo:
Thanks @cajieh, good eye. I've fixed those locally (and addressed the merge conflicts). Will wait for Calum/Marc to review before re-pushing.

pkg/kubernetes/provider_gvk_filter.go
2uasimojo: (This happened while resolving merge conflict for the recent contextual logging commit -- I `Attr`-ified the arg but forgot to rip out the format macro.)

pkg/provider/provider.go
```diff
@@ -0,0 +1,19 @@
+package provider
```
manusa: Since this is an interface and is actually exposed by `pkg/api`, is there any reason to have this in a separate `pkg/provider`?
Can we move it to `pkg/api`?

This would also prevent us from doing changes to `pkg/api/imports_test.go`

manusa:
**Heads-up (assisted first-pass): two kcp-provider issues in the new target-compatibility filtering, with a repro test**

<details>
<summary>✅ Resolved</summary>

Sharing one finding ahead of a fuller review. When `enable_target_compatibility_tool_filters` is enabled, the kcp provider has a couple of problems around `AnyTargetHasGVKs` / `GetTargetManagers`:

**1. Guaranteed nil-pointer panic (multi-workspace kcp).** `reset()` seeds `p.managers[ws] = nil` for every discovered workspace and only materializes the default workspace's `*Manager`; the rest are created lazily by `managerForWorkspace()`. But `kcpClusterProvider.GetTargetManagers()` ranges the map and returns every value including those nils, and `ProviderGVKFilter.AnyTargetHasGVKs()` then calls `mgr.Derived()` on each (`pkg/kubernetes/provider_gvk_filter.go:43`), which dereferences the receiver (`pkg/kubernetes/manager.go:179`). With kcp + the flag on + >1 workspace, `initResources`/`initNamespaces` call `AnyTargetHasGVKs` at tool-(re)registration time; kcp has neither the Route nor Project GVK, so the loop never short-circuits and is guaranteed to hit a nil entry and panic. (The kubeconfig provider avoids this because its `GetTargetManagers` routes every context through `managerForContext`, which never returns nil.)

**2. Unsynchronized `managers` map.** Unlike `kubeConfigClusterProvider` (which guards `managers` with a `sync.RWMutex`), `kcpClusterProvider` has no lock. `reset()` reassigns the map and `managerForWorkspace()` writes entries, while the new read path (`GetTargetManagers` via the cluster-state watcher's reapply) reads it concurrently. `Server.reloadMu` only serializes `applyToolsets`, not `reset()`, so concurrent map read+write is reachable, which is a Go runtime fatal.

Here is a self-contained, network-free repro you can drop in at `pkg/kcp/provider_filter_test.go` (white-box, `package kcp`). It fails today and doubles as a regression guard:

<details>
<summary><code>pkg/kcp/provider_filter_test.go</code></summary>

```go
package kcp

import (
	"context"
	"fmt"
	"sync"
	"testing"

	"github.com/containers/kubernetes-mcp-server/pkg/kubernetes"
	"github.com/stretchr/testify/suite"
	"k8s.io/apimachinery/pkg/runtime/schema"
)

// ProviderFilterSuite reproduces the target-compatibility filtering problems on the
// kcp provider introduced in this PR.
//
// reset() seeds p.managers[ws] = nil for every discovered workspace and only assigns
// a real *Manager for the default workspace; the others are created lazily by
// managerForWorkspace(). GetTargetManagers() ranges the whole map and returns every
// value, including the not-yet-initialized nils. AnyTargetHasGVKs() (from the embedded
// *kubernetes.ProviderGVKFilter) iterates those managers and calls mgr.Derived(),
// which dereferences the receiver, so a nil *Manager is a guaranteed panic.
type ProviderFilterSuite struct {
	suite.Suite
}

func TestProviderFilterSuite(t *testing.T) {
	suite.Run(t, new(ProviderFilterSuite))
}

// newTestProvider builds a kcpClusterProvider whose managers map contains the given
// entries, wired with a ProviderGVKFilter exactly as newKcpClusterProvider does.
// No network/kubeconfig is required: AnyTargetHasGVKs only reads the managers map.
func newTestProvider(managers map[string]*kubernetes.Manager) *kcpClusterProvider {
	p := &kcpClusterProvider{
		defaultWorkspace: "root",
		managers:         managers,
	}
	p.ProviderGVKFilter = kubernetes.NewProviderGVKFilter(p)
	return p
}

// TestAnyTargetHasGVKsWithUninitializedManager reproduces the panic: a nil (lazily
// initialized) workspace manager must not cause a crash.
//
// BEFORE the fix: FAILS, AnyTargetHasGVKs calls Derived() on a nil *Manager and panics.
// AFTER the fix (skip nil managers, or guard mgr == nil before Derived): PASSES.
func (s *ProviderFilterSuite) TestAnyTargetHasGVKsWithUninitializedManager() {
	routeGVK := []schema.GroupVersionKind{
		{Group: "route.openshift.io", Version: "v1", Kind: "Route"},
	}

	s.Run("does not panic when a single workspace manager is uninitialized", func() {
		p := newTestProvider(map[string]*kubernetes.Manager{
			"root:org:team": nil,
		})
		s.NotPanics(func() {
			_ = p.AnyTargetHasGVKs(context.Background(), routeGVK)
		}, "AnyTargetHasGVKs must skip uninitialized (nil) target managers instead of dereferencing them")
	})

	s.Run("does not panic with multiple uninitialized workspace managers", func() {
		// Mirrors the post-reset state for a multi-workspace kcp tree. All nil here to
		// keep the repro deterministic and network-free; in production the default
		// would be non-nil but, lacking the Route GVK, the loop still advances into nils.
		p := newTestProvider(map[string]*kubernetes.Manager{
			"root":            nil,
			"root:org:team":   nil,
			"root:org:team-2": nil,
		})
		s.NotPanics(func() {
			_ = p.AnyTargetHasGVKs(context.Background(), routeGVK)
		}, "AnyTargetHasGVKs must skip every uninitialized (nil) target manager")
	})
}

// TestManagersConcurrentAccessIsRaceFree reproduces the data race: kcpClusterProvider.managers
// is read and written without synchronization (kubeConfigClusterProvider guards the same
// state with a sync.RWMutex). The new read path (GetTargetManagers, reached from the
// cluster-state watcher's reapply -> AnyTargetHasGVKs) races with the write paths
// (reset() reassigning the map, managerForWorkspace() inserting entries).
//
// Run with the race detector:  go test -race ./pkg/kcp/...
//
// AFTER the fix (add a sync.RWMutex to kcpClusterProvider and guard every access to
// managers, including the writes below), route this test's writes through the now-locked
// write path so reads and writes are serialized by the same lock.
func (s *ProviderFilterSuite) TestManagersConcurrentAccessIsRaceFree() {
	p := newTestProvider(map[string]*kubernetes.Manager{"root": nil})

	const readers = 8
	const writers = 8
	const iterations = 500

	var wg sync.WaitGroup
	wg.Add(readers + writers)

	for i := 0; i < readers; i++ {
		go func() {
			defer wg.Done()
			for j := 0; j < iterations; j++ {
				_, _ = p.GetTargetManagers(context.Background())
				_, _ = p.GetTargets(context.Background())
				_ = p.IsMultiTarget()
			}
		}()
	}

	for i := 0; i < writers; i++ {
		go func(id int) {
			defer wg.Done()
			for j := 0; j < iterations; j++ {
				// Stands in for reset()/managerForWorkspace() mutating the map.
				p.writeManagerForTest(fmt.Sprintf("ws-%d-%d", id, j), nil)
			}
		}(i)
	}

	wg.Wait()
}

// writeManagerForTest mutates the managers map the way reset()/managerForWorkspace()
// do in production. It exists only so the race test can drive a write without a live
// kcp API server. Once kcpClusterProvider gains a mutex, this must take the lock too.
func (p *kcpClusterProvider) writeManagerForTest(workspace string, m *kubernetes.Manager) {
	p.managers[workspace] = m
}
```
</details>

Verified on the current PR head:
- `go test ./pkg/kcp/ -run TestProviderFilterSuite/TestAnyTargetHasGVKsWithUninitializedManager` fails with `nil pointer dereference` (`Manager.Derived(0x0)` at `manager.go:179` <- `AnyTargetHasGVKs` at `provider_gvk_filter.go:43`).
- `go test -race ./pkg/kcp/ -run TestProviderFilterSuite/TestManagersConcurrentAccessIsRaceFree` trips the race detector on `managers`.
</details>

manusa (changes requested):
Thanks for putting this together, this is looking very good already.

I did a very quick first pass and added a few comments, and also added a finding from the review agent and a reproducer. In case you want to take care of those already.

It's late and I need to drop, will continue with the proper review tomorrow.

pkg/provider/provider.go
2uasimojo: I'll look into that.

This PR went through at least half a dozen iterations while I was trying to find the appropriate hierarchy for all these interfaces, and a lot of the churn was because of circular imports. In this case, kubernetes.Provider embeds this provider.Provider. I know we can't combine them because e.g. the ACM provider will want provider.Provider but not kubernetes.Provider; but I don't think that should stop us from moving this one into the API package.

2uasimojo: Yeah, nice, that worked just fine 👍

2uasimojo:
> **1. Guaranteed nil-pointer panic (multi-workspace kcp).**

Mm. I did notice this in the kubernetes provider (as you can see) but missed this one. Fixed :+1: 

(Actually I had to dig a little deeper and fix GetTargetManagers(), which meant splitting some code paths into synchronized/unsynchronized incarnations. This was needed for both the kcp and kubernetes provider.)

> **2. Unsynchronized `managers` map.**

This is latent, but I've fixed it.

> Here is a self-contained, network-free repro

The test setup misses initializing some key fields, so the test panics even when I've fixed the races :innocent:

pkg/toolsets/core/namespaces.go
```diff
@@ -35,23 +36,29 @@ func initNamespaces(o api.Openshift) []api.ServerTool {
...
+				OpenWorldHint:   ptr.To(true),
+			},
+		},
+		Handler: projectsList,
+		TargetCompatibilityFilters: []func() bool{
+			func() bool {
+				return p.AnyTargetHasGVKs(context.TODO(), []schema.GroupVersionKind{
+					{Group: "project.openshift.io", Version: "v1", Kind: "Project"},
+				})
+			},
```
manusa, line 59: `projects_list` is OpenShift-only, but this closure only runs when `enable_target_compatibility_tool_filters` is set (the flag gate in `isToolApplicable`, `pkg/mcp/mcp.go`).

Since that flag defaults to off, a default install now advertises `projects_list` on every cluster, including vanilla Kubernetes, where it fails at call time with `no matches for kind "Project"`.

That's a regression from the old `IsOpenShift()` gate, which hid the tool on non-OpenShift clusters.

Can we restore the old default for single-cluster setups while keeping the new multi-target work opt-in?

- In `isToolApplicable`, always evaluate `TargetCompatibilityFilters` (drop the `if EnableTargetCompatibilityToolFilters` wrapper).
- Here, keep only the multi-target fan-out behind the flag:

  ```go
  func() bool {
      // Single-target: one cheap discovery call, same cost/result as the old IsOpenShift() gate.
      // Multi-target: fanning out discovery over every target is too expensive to do eagerly and
      // can't be known cheaply up front, so stay permissive unless the operator opts in.
      if p.IsMultiTarget() && !p.IsTargetCompatibilityToolFiltersEnabled() {
          return true
      }
      return p.AnyTargetHasGVKs(context.TODO(), []schema.GroupVersionKind{
          {Group: "project.openshift.io", Version: "v1", Kind: "Project"},
      })
  }
  ```
  (needs IsMultiTarget() on api.FilteringProvider; the real provider already implements it.)

Net: single-target vanilla K8s hides projects_list again at zero extra cost, single-target OpenShift shows it, and multi-target stays permissive by default (flag opts into the full "any target" fan-out).

The route.openshift.io/v1 Route description enrichment in
resources.go needs the same treatment.

pkg/toolsets/core/resources.go
```diff
@@ -14,9 +14,18 @@ import (
 	"github.com/containers/kubernetes-mcp-server/pkg/output"
 )
 
-func initResources(o api.Openshift) []api.ServerTool {
+func initResources(p api.FilteringProvider) []api.ServerTool {
+	// TODO: What purpose is this string serving in Description fields?
 	commonApiVersion := "v1 Pod, v1 Service, v1 Node, apps/v1 Deployment, networking.k8s.io/v1 Ingress"
-	if o.IsOpenShift(context.Background()) {
+	// Only check GVKs when target compatibility filtering is enabled
+	if p.IsTargetCompatibilityToolFiltersEnabled() {
```
manusa: Check https://github.com/containers/kubernetes-mcp-server/pull/1196/changes#r3506099389

```
routeGVK := []schema.GroupVersionKind{{Group: "route.openshift.io", Version: "v1", Kind: "Route"}}
includeRoute := true // multi-target with the flag off stays permissive (no eager fan-out)
if !p.IsMultiTarget() || p.IsTargetCompatibilityToolFiltersEnabled() {
    includeRoute = p.AnyTargetHasGVKs(context.TODO(), routeGVK)
}
if includeRoute {
    commonApiVersion += ", route.openshift.io/v1 Route"
}
```

internal/test/envtest.go
manusa: Improvements over prior implementation:

- Parallel-safe isolation: unique etcd data dir (os.MkdirTemp) + explicit ControlPlane (random ports). Needed now that two packages spin up envtest.
- Error handling tightened: Start(), AddUser(), KubeConfig() errors now panic instead of being silently dropped. The old code did adminSystemMasterBaseConfig, _ := envTest.Start() — a Start failure would have surfaced as an obscure nil-deref later.
- allow-all is now actually created. Old restoreAuth only did Update, which is a silent no-op when the role is absent (harmless before only because envtest authz defaults to AlwaysAllow). The shared createTestData now Creates it explicitly — strictly more correct.
- Provider leak fixed: BaseMcpSuite gained a provider field that TearDownTest closes. The old InitMcpClient dropped the provider on the floor.

pkg/mcp/common_test.go
```diff
@@ -3,117 +3,41 @@ package mcp
...
-	//Set("v", "9")
-
-	adminSystemMasterBaseConfig, _ := envTest.Start()
-	au := test.Must(envTest.AddUser(envTestUser, adminSystemMasterBaseConfig))
-	envTestRestConfig = au.Config()
-	envTest.KubeConfig = test.Must(au.KubeConfig())
+	// Initialize shared envtest environment once for all test packages.
+	// This runs in the mcp package but envtest is shared via internal/test
+	// so kubernetes, http, watcher packages all use the same instance.
+	test.EnvTest()
```
manusa: Each go test package is a separate process, so package-level state in internal/test is not shared across them. pkg/kubernetes starts its own instance via its own TestMain; http and watcher don't use envtest at all (I checked: only pkg/mcp and pkg/kubernetes have a TestMain calling EnvTest()). This directly contradicts envtest.go's own accurate doc comment ("Each test package process gets its own envtest instance with isolated etcd data directory"). Recommend rewording to something like "initialize this package's envtest instance once."

pkg/mcp/common_test.go
```diff
@@ -125,13 +49,13 @@ func restoreAuth(ctx context.Context) {
 	}, metav1.UpdateOptions{})
 	_, _ = kubernetesAdmin.RbacV1().ClusterRoleBindings().Update(ctx, &rbacv1.ClusterRoleBinding{
 		ObjectMeta: metav1.ObjectMeta{Name: "allow-all"},
-		Subjects:   []rbacv1.Subject{{Kind: "Group", Name: envTestUser.Groups[0]}},
+		Subjects:   []rbacv1.Subject{{Kind: "Group", Name: "test:users"}},
```
manusa, line 53:
```suggestion
		Subjects:   []rbacv1.Subject{{Kind: "Group", Name: test.EnvTestUser().Groups[0]}},
```

internal/test/envtest.go
```diff
@@ -0,0 +1,239 @@
...
+
+// EnvTestUser returns the test user for the shared envtest environment.
+func EnvTestUser() *envtest.User {
+	EnvTest() // Ensure initialized
+	return &envTestUser
+}
+
+// StopEnvTest stops the shared envtest environment if it was started.
+// This should typically be called from TestMain after all tests complete.
+func StopEnvTest() error {
```
manusa, line 156: StopEnvTest can't re-arm. It nils envTestInstance but leaves envTestOnce consumed, so a subsequent EnvTest() in the same process returns nil. Safe today (only called after m.Run()), but the "reinitialize in another test package process" framing is misleading in-process.

pkg/toolsets/core/resources.go
manusa: Code recommendations here depend on the outcome of the annotations based approach https://github.com/containers/kubernetes-mcp-server/pull/1196#discussion_r3396391142

pkg/toolsets/core/namespaces.go
manusa: Code recommendations here depend on the outcome of the annotations based approach https://github.com/containers/kubernetes-mcp-server/pull/1196#discussion_r3396391142

pkg/toolsets/core/resources.go
manusa: I'm currently working on a prototype for annotation-based approach, will post findings here and a reference PR to this branch soon.

manusa: I prototyped the annotations approach as a reference implementation on top of your branch: https://github.com/2uasimojo/openshift-mcp-server/pull/1

It swaps the per-tool closures for a declarative RequiredGVKs []schema.GroupVersionKind on ServerTool, evaluated centrally in one filter. Turns out projects_list was the only closure user and they were all the same AnyTargetHasGVKs(gvks) shape, so behavior is unchanged and the existing filtering tests pass as-is. The "any target / assume yes on error" semantics stay in the provider method, so a future question with different semantics still lives there (and a closure can sit next to the field if we ever need per-tool logic back).

On your point about future questions needing different semantics (all targets, or assume no on error): those still live in the provider method, same as today. A closure calling AnyTargetHasGVKs has the same fixed semantics, so a new semantic needs a new provider method either way. The declarative field just selects it, and a closure can sit right next to it if we ever want per-tool logic back.

It's a draft/prototype, so no pressure to take it. The idea is for us to decide on an approach since we're going to commit some breaking API changes.

manusa (changes requested):
Thanks for the fast turnaround on the kcp issues. Confirmed fixed on the current head: the
nil-manager panic in `AnyTargetHasGVKs`, the unsynchronized `managers` map, and the ctx
propagation plus the `%s` log placeholders.

A few things still to address in this PR before it merges.

### Changes requested

1. **Default tool surface must match the old `IsOpenShift()` gate.** `projects_list` and the
   `route.openshift.io/v1 Route` hint in the `resources_*` descriptions must not be exposed on a
   single-target non-OpenShift cluster. Today they are, because filtering only runs when
   `enable_target_compatibility_tool_filters` is set and that defaults off, so `projects_list`
   ships on vanilla Kubernetes and fails at call time. Keep single-target compatibility filtering
   always on and gate only the multi-target discovery fan-out behind the flag.
   Inline: https://github.com/containers/kubernetes-mcp-server/pull/1196#discussion_r3506099389
   and https://github.com/containers/kubernetes-mcp-server/pull/1196#discussion_r3506104105.

2. **`HasGVKs` "not found" match.** `err.Error() == "not found"` is the load-bearing path for the
   production memcache discovery client (`ErrCacheNotFound` is a plain error, so
   `errors.IsNotFound` is false). A client-go message change would silently re-enable filtered
   tools. Use `errors.Is(err, memory.ErrCacheNotFound)`.

3. **envtest teardown and comments.** `StopEnvTest` nils the instance but cannot re-arm
   `sync.Once`, and the "shared across packages" comment in `common_test.go` contradicts
   envtest.go's own accurate doc.
   Inline: https://github.com/containers/kubernetes-mcp-server/pull/1196#discussion_r3506256608
   and https://github.com/containers/kubernetes-mcp-server/pull/1196#discussion_r3506227380.

4. **Watcher scope.** The cluster-state watcher only watches the default target, so a non-default
   target changing OpenShift-ness at runtime does not re-fire a reapply. The filter now spans all
   targets, so change detection should too (per-target watching).

5. **kcp fan-out under the exclusive lock.** `kcpClusterProvider.GetTargetManagers` materializes
   every workspace manager while holding `p.mu.Lock()`, on every applyToolsets. A caching or
   RLock fast-path would keep one slow or unreachable target from blocking the whole provider.

6. **Test robustness.** `mcp_watch_test.go` registers overlapping mock handlers without a reset
   (the new 404 path can double-write), and `InitMcpClient` leaks the first server and provider
   when it is called twice in a test.

### Prototype
I put up a draft on top of your branch exploring the annotations approach we discussed
(https://github.com/containers/kubernetes-mcp-server/pull/1196#discussion_r3396391142) plus
several of the above, each as an independent commit you can take or drop:
**https://github.com/2uasimojo/openshift-mcp-server/pull/1**
- declarative `RequiredGVKs` on `ServerTool` replacing the per-tool closures, evaluated centrally
- restore the non-OpenShift default for single-target providers (item 1)
- `errors.Is(err, memory.ErrCacheNotFound)` in `HasGVKs`, with a memcache-backed test (item 2)
- flag docs, drop the dead fake method, and fix the two envtest comments (item 3)

Items 4, 5, and 6 are not in the prototype yet.

pkg/toolsets/core/namespaces.go
2uasimojo: This seems like a weird semantic to me: Enable target compatibility tool filtering, but only on multi-target setups? We're arbitrarily deciding that single-target setups can "afford" the discovery call by default? Looking forward to things like ACM, where the cost of that call might be O(1) rather than O(N) (not sure), would we make the same call there?

Is this only because we're trying to preserve backward compatibility with existing behavior? If so, why is that a thing, given the lifecycle stage of the project?

pkg/toolsets/core/resources.go
2uasimojo: Thanks for doing that. I've [commented](https://github.com/2uasimojo/openshift-mcp-server/pull/1#discussion_r3509162531) with my take on that approach. Let's discuss!

pkg/toolsets/core/namespaces.go
manusa: Not presenting a projects_list to a coding agent who hasn't access to OpenShift is not just about preserving an old behavior but preserving good UX (IMO).

I don't really care about the mechanics, but the main concern is that a user who hasn't access to sth shouldn't be presented with a tool to exercise it (e.g. projects_list in vanilla kubernetes). As long as this is cheap (single cluster, or whatever), this should be preserved.

Opened an internal discussion for this.

2uasimojo: Summary of internal discussion:
- Remove the config option for now.
- Always query & filter.
- Leave TODO to add config option later.

pkg/mcp/common_test.go
2uasimojo: Resolved by pulling in https://github.com/2uasimojo/openshift-mcp-server/pull/1/changes/b3c3affd873564c7b55d33e8d8c5c97f25f50068

pkg/mcp/common_test.go
2uasimojo: Resolved by pulling in https://github.com/2uasimojo/openshift-mcp-server/pull/1/changes/b3c3affd873564c7b55d33e8d8c5c97f25f50068

internal/test/envtest.go
2uasimojo: Resolved by pulling in https://github.com/2uasimojo/openshift-mcp-server/pull/1/changes/b3c3affd873564c7b55d33e8d8c5c97f25f50068

pkg/toolsets/core/namespaces.go
2uasimojo: Done. I did it in an isolated commit to make it easier to reintroduce the config option later if desired. (Though we would still need to address the design questions we've discussed.)

manusa:
Pushed a small commit directly to this branch (`3a1370d7`) to unblock the macOS CI and fold in the two review nits, and I'm merging it directly to keep this moving. Flagging the "what and why" here for the record.

**Fixes the `Build on macos-15-intel` failure.** The provider suites in `pkg/kubernetes` were moved to envtest, which stands up a second kube-apiserver+etcd control plane on top of the one `pkg/mcp` already runs. `go test ./...` runs package binaries in parallel, so both control planes come up at once, and on the memory-constrained macos-15-intel runner kube-apiserver gets OOM-killed (`signal: killed`). Linux and the arm64 macOS runner have enough headroom, which is why only that one job failed (3x, so not a flake). Those suites only exercise discovery and provider plumbing, which `MockServer` already covers (same as the `pkg/api` `HasGVKs` tests), so I reverted them to `MockServer`. Now only `pkg/mcp` needs envtest and `pkg/kubernetes` runs with none.

**Two review nits, folded in:**
- `NewInOpenShiftHandler` now also serves `route.openshift.io/v1`, so the OpenShift toolset golden reflects a real OpenShift cluster and guards the Route hint (regenerated `toolsets-full-tools-openshift.json`).
- Replaced the stale `commonApiVersion` TODO in `initResources` with a doc comment.

Verified locally with `-race`: `pkg/kubernetes` (~3s, no envtest), `pkg/api`, and `pkg/mcp` all green; `go build` / `go vet ./...` clean. Thanks for the quick iteration @2uasimojo.

manusa (approved):
LGTM, thx!

PR #1202, opened 2026-06-12
chore(logs)!: switch to contextual logging

Cali0707:
Currently, our logs are not correlated with the otel traces we export.

A prerequisite to fixing this is to use a contextual logger, instead of the global logger, so that logs can inherit trace_id/span_id from the context.

While switching from the global `klog` to `klog.FromContext()` (following https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/3077-contextual-logging/README.md#integration-with-logslog as closely as possible), I had to make a few switches to the existing logs:
1. the contextual logger is a structured logger, so it does not support `Infof`, just `Info` + k/v pairs
2. The contextual logger does not have a warn level, so some were switched to info and others to error

This also required wiring context through a whole bunch of places that probable should have wired context through before (especially those that were setting their own context from `context.Background()`

### DIFF
--- a/pkg/confirmation/confirmation.go
+++ b/pkg/confirmation/confirmation.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
 	"github.com/google/jsonschema-go/jsonschema"
 	"k8s.io/klog/v2"
 )
@@ -56,7 +57,7 @@ func CheckConfirmation(ctx context.Context, elicitor api.Elicitor, message, fall
 			if fallback == "deny" {
 				return ErrConfirmationDenied
 			}
-			klog.FromContext(ctx).Info("Confirmation rules matched but client does not support elicitation, proceeding with fallback \"allow\"", "message", message)
+			klogutil.Warn(ctx, "Confirmation rules matched but client does not support elicitation, proceeding with fallback \"allow\"", "message", message)
 			return nil
 		}
 		return err
--- a/pkg/confirmation/confirmation.go
+++ b/pkg/confirmation/confirmation.go
@@ -7,7 +7,6 @@ import (
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
 	"github.com/google/jsonschema-go/jsonschema"
-	"k8s.io/klog/v2"
 )
 
 // ErrConfirmationDenied is returned when the user declines a confirmation prompt
--- a/pkg/confirmation/confirmation.go
+++ b/pkg/confirmation/confirmation.go
@@ -56,7 +56,7 @@ func CheckConfirmation(ctx context.Context, elicitor api.Elicitor, message, fall
 			if fallback == "deny" {
 				return ErrConfirmationDenied
 			}
-			klog.Warningf("Confirmation rules matched but client does not support elicitation, proceeding with fallback \"allow\": %s", message)
+			klog.FromContext(ctx).Info("Confirmation rules matched but client does not support elicitation, proceeding with fallback \"allow\"", "message", message)
 			return nil
 		}
 		return err
--- a/pkg/helm/helm.go
+++ b/pkg/helm/helm.go
@@ -127,8 +127,11 @@ func (h *Helm) newAction(ctx context.Context, namespace string, allNamespaces bo
 		return nil, err
 	}
 	cfg.RegistryClient = registryClient
+	logger := klog.FromContext(ctx)
 	return cfg, cfg.Init(h.kubernetes, applicableNamespace, storageDriver, func(format string, v ...any) {
-		klog.FromContext(ctx).Info(fmt.Sprintf(format, v...)) // TODO: revisit if this really needs to be info log, this is the helm "debug logger"
+		if logger.V(5).Enabled() {
+			logger.V(5).Info(fmt.Sprintf(format, v...))
+		}
 	})
 }
 
--- a/pkg/helm/helm.go
+++ b/pkg/helm/helm.go
@@ -37,7 +37,7 @@ func (h *Helm) Install(ctx context.Context, chart string, values map[string]inte
 	if err := validateChartReference(chart, h.config); err != nil {
 		return "", err
 	}
-	cfg, err := h.newAction(h.kubernetes.NamespaceOrDefault(namespace), false)
+	cfg, err := h.newAction(ctx, h.kubernetes.NamespaceOrDefault(namespace), false)
 	if err != nil {
 		return "", err
 	}
@@ -74,8 +74,8 @@ func (h *Helm) Install(ctx context.Context, chart string, values map[string]inte
 }
 
 // List lists all the releases for the specified namespace (or current namespace if). Or allNamespaces is true, it lists all releases across all namespaces.
-func (h *Helm) List(namespace string, allNamespaces bool) (string, error) {
-	cfg, err := h.newAction(namespace, allNamespaces)
+func (h *Helm) List(ctx context.Context, namespace string, allNamespaces bool) (string, error) {
+	cfg, err := h.newAction(ctx, namespace, allNamespaces)
 	if err != nil {
 		return "", err
 	}
@@ -94,8 +94,8 @@ func (h *Helm) List(namespace string, allNamespaces bool) (string, error) {
 	return string(ret), nil
 }
 
-func (h *Helm) Uninstall(name string, namespace string) (string, error) {
-	cfg, err := h.newAction(h.kubernetes.NamespaceOrDefault(namespace), false)
+func (h *Helm) Uninstall(ctx context.Context, name string, namespace string) (string, error) {
+	cfg, err := h.newAction(ctx, h.kubernetes.NamespaceOrDefault(namespace), false)
 	if err != nil {
 		return "", err
 	}
@@ -112,7 +112,7 @@ func (h *Helm) Uninstall(name string, namespace string) (string, error) {
 	return fmt.Sprintf("Uninstalled release %s %s", uninstalledRelease.Release.Name, uninstalledRelease.Info), nil
 }
 
-func (h *Helm) newAction(namespace string, allNamespaces bool) (*action.Configuration, error) {
+func (h *Helm) newAction(ctx context.Context, namespace string, allNamespaces bool) (*action.Configuration, error) {
 	storageDriver := ""
 	if h.config != nil {
 		storageDriver = h.config.StorageDriver
@@ -127,7 +127,9 @@ func (h *Helm) newAction(namespace string, allNamespaces bool) (*action.Configur
 		return nil, err
 	}
 	cfg.RegistryClient = registryClient
-	return cfg, cfg.Init(h.kubernetes, applicableNamespace, storageDriver, klog.V(5).Infof)
+	return cfg, cfg.Init(h.kubernetes, applicableNamespace, storageDriver, func(format string, v ...any) {
+		klog.FromContext(ctx).Info(fmt.Sprintf(format, v...)) // TODO: revisit if this really needs to be info log, this is the helm "debug logger"
+	})
 }
 
 // validateChartReference blocks chart references using dangerous URL schemes.
--- a/pkg/http/authorization.go
+++ b/pkg/http/authorization.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/containers/kubernetes-mcp-server/pkg/config"
+	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
 	internalk8s "github.com/containers/kubernetes-mcp-server/pkg/kubernetes"
 	"github.com/containers/kubernetes-mcp-server/pkg/oauth"
 )
@@ -116,7 +117,7 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
 			// Cluster is the sole authority for validating token (ex. sha256 token with OpenShift)
 			if staticConfig.SkipJWTVerification && staticConfig.AuthorizationURL == "" {
 				skipJWTWarningOnce.Do(func() {
-					logger.Info("Bearer token forwarded without local validation (skip_jwt_verification=true and no authorization_url) - the cluster is the sole authority")
+					klogutil.Warn(r.Context(), "Bearer token forwarded without local validation (skip_jwt_verification=true and no authorization_url) - the cluster is the sole authority")
 				})
 				ctx := context.WithValue(r.Context(), internalk8s.OAuthAuthorizationHeader, authHeader)
 				next.ServeHTTP(w, r.WithContext(ctx))
--- a/pkg/http/authorization.go
+++ b/pkg/http/authorization.go
@@ -117,7 +117,7 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
 			// Cluster is the sole authority for validating token (ex. sha256 token with OpenShift)
 			if staticConfig.SkipJWTVerification && staticConfig.AuthorizationURL == "" {
 				skipJWTWarningOnce.Do(func() {
-					klogutil.Warn(r.Context(), "Bearer token forwarded without local validation (skip_jwt_verification=true and no authorization_url) - the cluster is the sole authority")
+					klogutil.WarnLogger(logger, "Bearer token forwarded without local validation (skip_jwt_verification=true and no authorization_url) - the cluster is the sole authority")
 				})
 				ctx := context.WithValue(r.Context(), internalk8s.OAuthAuthorizationHeader, authHeader)
 				next.ServeHTTP(w, r.WithContext(ctx))
--- a/pkg/http/authorization.go
+++ b/pkg/http/authorization.go
@@ -57,6 +57,7 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
 	var skipJWTWarningOnce sync.Once
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			logger := klog.FromContext(r.Context())
 			// Skip auth for infrastructure endpoints (health, metrics) and well-known endpoints
 			// Use prefix matching per endpoint to handle sub-paths like /.well-known/oauth-protected-resource/sse
 			requestPath := r.URL.EscapedPath()
@@ -89,7 +90,11 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
 
 			authHeader := r.Header.Get("Authorization")
 			if authHeader == "" || !strings.HasPrefix(authHeader, "Bearer ") {
-				klog.V(1).Infof("Authentication failed - missing or invalid bearer token: %s %s from %s", r.Method, r.URL.Path, r.RemoteAddr)
+				logger.V(1).Info("Authentication failed - missing or invalid bearer token",
+					"http.request.method", r.Method,
+					"url.path", r.URL.Path,
+					"client.address", r.RemoteAddr,
+				)
 				write401(w, wwwAuthenticateHeader, "missing_token", "Unauthorized: Bearer token required")
 				return
 			}
@@ -98,7 +103,11 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
 
 			// Empty token check post-trimming
 			if token == "" {
-				klog.V(1).Infof("Authentication failed - empty bearer token: %s %s from %s", r.Method, r.URL.Path, r.RemoteAddr)
+				logger.V(1).Info("Authentication failed - empty bearer token",
+					"http.request.method", r.Method,
+					"url.path", r.URL.Path,
+					"client.address", r.RemoteAddr,
+				)
 				write401(w, wwwAuthenticateHeader, "invalid_token", "Unauthorized: Bearer token is empty")
 				return
 			}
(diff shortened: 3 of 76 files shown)


pkg/helm/helm.go
```diff
@@ -127,7 +127,9 @@ func (h *Helm) newAction(namespace string, allNamespaces bool) (*action.Configur
 		return nil, err
 	}
 	cfg.RegistryClient = registryClient
-	return cfg, cfg.Init(h.kubernetes, applicableNamespace, storageDriver, klog.V(5).Infof)
+	return cfg, cfg.Init(h.kubernetes, applicableNamespace, storageDriver, func(format string, v ...any) {
+		klog.FromContext(ctx).Info(fmt.Sprintf(format, v...)) // TODO: revisit if this really needs to be info log, this is the helm "debug logger"
```
manusa: (Related to TODO) Note that this was V(5) before, matching https://kubernetes.io/docs/reference/kubectl/quick-reference/#kubectl-output-verbosity-and-debugging verbosity thresholds

Cali0707: whoops, let me switch that

pkg/kubernetes-mcp-server/cmd/root_sighup_test.go
```diff
@@ -66,13 +67,14 @@ func (s *SIGHUPSuite) TearDownTest() {
 }
 
 func (s *SIGHUPSuite) InitServer(configPath, configDir string) *MCPServerOptions {
-	cfg, err := config.Read(configPath, configDir)
+	ctx := context.Background()
```
manusa: Not sure why we create a new context here as opposed to using s.T().Context() like we do later on in line 77

pkg/kubernetes-mcp-server/cmd/root_sighup_test.go
```diff
@@ -295,7 +297,7 @@ func TestSIGHUPInvokesLogSinkReload(t *testing.T) {
 	mockServer.Handle(test.NewDiscoveryClientHandler())
 	t.Cleanup(mockServer.Close)
 
-	cfg, err := config.Read(configPath, "")
+	cfg, err := config.Read(context.Background(), configPath, "")
```
manusa: Similar here, I'm observing multiple patterns in this suite t.Context(), s.T().Context(), context.Background().
Is there any reason why we're not following a uniform pattern?

Cali0707: I let claude update this file, looks like it didn't do a consistent job - will look through this one.

In general on the files I did myself, it was s.T.Context() where s was available, and t.Context() where s was not

pkg/helm/helm.go
manusa: Also spotted by review agent:

 1. pkg/helm/helm.go:131 — Helm debug logger downgraded from V(5) to unconditional V(0). The callback was cfg.Init(...,
  klog.V(5).Infof) (silent unless -v=5); it is now func(...){ klog.FromContext(ctx).Info(fmt.Sprintf(...)) }, i.e. always
  emitted at default verbosity. Helm's debug logger is chatty, so every install/list/uninstall will now spew Helm internals
  at the default log level. The author's own inline // TODO: revisit if this really needs to be info log flags it — worth
  resolving before merge (...V(5).Info(...), ideally .V(5).Enabled()-guarded to skip the Sprintf).

pkg/kubernetes-mcp-server/cmd/root.go
```diff
@@ -303,26 +304,26 @@ func (m *MCPServerOptions) Validate() error {
...
 
-	klog.V(1).Infof(" - ClusterProviderStrategy: %s", strategy)
+	klog.FromContext(ctx).V(1).Info("Starting kubernetes-mcp-server",
+		"config.path", m.ConfigPath,
+		"config.toolsets", m.StaticConfig.Toolsets,
+		"config.list_output", m.StaticConfig.ListOutput,
+		"config.read_only", m.StaticConfig.ReadOnly,
+		"config.disable_destructive", m.StaticConfig.DisableDestructive,
+		"config.stateless", m.StaticConfig.Stateless,
+		"config.telemetry.enabled", m.StaticConfig.Telemetry.Enabled,
```
manusa: Flagged by review agent:

  2. pkg/kubernetes-mcp-server/cmd/root.go:324 — startup log prints the raw *bool field instead of the computed value.
  Changed from Telemetry.IsEnabled() to m.StaticConfig.Telemetry.Enabled. Telemetry.Enabled is a *bool
  (telemetry_config.go:15) and IsEnabled() factors in the endpoint + nil-default (telemetry_config.go:41-49). With the
  default (nil) config this now logs config.telemetry.enabled=null instead of the effective boolean, and it ignores the
  endpoint-aware auto-enable. The diagnostic line no longer reflects actual telemetry state. (Flagged independently by two
  reviewers + verified.)

pkg/kubernetes-mcp-server/cmd/root.go
```diff
@@ -128,13 +128,14 @@ func NewMCPServerOptions(streams genericiooptions.IOStreams) *MCPServerOptions {
 
 func NewMCPServer(streams genericiooptions.IOStreams) *cobra.Command {
 	o := NewMCPServerOptions(streams)
+	ctx := context.Background()
```
manusa: Consider deriving this from the cobra command context instead of a fresh `context.Background()`. Inside `RunE` the `*cobra.Command` is in scope, so moving this in as `ctx := c.Context()` is the idiomatic choice. It's functionally identical today (`main.go` uses `Execute()`, so cobra defaults the command context to `context.Background()`), but it lets the entry point later opt into cancellation/signal handling via `ExecuteContext(signal.NotifyContext(...))` — which would then flow to both `Serve` and `ServeStdio` for graceful shutdown — without touching this code.

Related: the HTTP branch in `Run` (~line 371) re-declares `ctx := context.Background()` and passes that to `internalhttp.Serve`, shadowing the `ctx` you threaded through `Run`. So on the HTTP path the server's `BaseContext` (and any logger/cancellation carried on the real ctx) never sees it — only the stdio path uses the threaded context. Worth deriving both serve paths from the same root context.

pkg/metrics/otel_stats_collector.go
```diff
@@ -166,9 +169,9 @@ func NewOtelStatsCollectorWithConfig(cfg CollectorConfig) (*OtelStatsCollector,
 	if err != nil {
 		// Use Warning if telemetry was explicitly configured, V(1) otherwise
 		if cfg.Telemetry != nil && cfg.Telemetry.IsEnabled() {
-			klog.Warningf("Failed to create OTLP metrics exporter, OTLP export disabled: %v", err)
+			logger.Error(err, "Failed to create OTLP metrics exporter, OTLP export disabled")
 		} else {
-			klog.V(1).Infof("Failed to create OTLP metrics exporter, OTLP export disabled: %v", err)
+			logger.V(1).Error(err, "Failed to create OTLP metrics exporter, OTLP export disabled")
```
manusa: `logr.Logger.Error` ignores `.V(n)` — `logger.V(1).Error(...)` logs unconditionally at error level, so the `.V(1)` is a no-op here. The original `else` branch was deliberately quiet (`klog.V(1).Infof`, only at `-v=1`) for the not-explicitly-enabled case; it now always emits at error level, collapsing the distinction with the `if` branch above it. If the intent was to keep this path quiet, use `logger.V(1).Info(..., "exception.message", err.Error())`. (Same `.V(1).Error` pattern at the startup sites in `pkg/telemetry/telemetry.go:161, 173, 236`.)

pkg/metrics/otel_stats_collector.go
```diff
@@ -318,11 +321,11 @@ func (c *OtelStatsCollector) RecordHTTPRequest(ctx context.Context, method, path
 
 // GetStats returns a snapshot of current statistics by reading from OTel metrics.
 // Thread-safety is handled by the OTel SDK's ManualReader.
-func (c *OtelStatsCollector) GetStats() *Statistics {
+func (c *OtelStatsCollector) GetStats(ctx context.Context) *Statistics {
 	// Collect current metrics from the manual reader
 	var rm metricdata.ResourceMetrics
 	if err := c.reader.Collect(context.Background(), &rm); err != nil {
-		klog.V(1).Infof("Failed to collect metrics for stats endpoint: %v", err)
+		klog.FromContext(ctx).V(1).Error(err, "Failed to collect metrics for stats endpoint")
```
manusa: Same `.V(1).Error` no-op as above, but higher impact: `GetStats` runs on every `/stats` request, so a transient collect error now logs at error level on every scrape instead of being V(1)-gated. Consider `klog.FromContext(ctx).V(1).Info("Failed to collect metrics for stats endpoint", "exception.message", err.Error())`.

pkg/kubernetes/watcher/kubeconfig.go
```diff
@@ -75,26 +78,26 @@ func (w *Kubeconfig) Watch(onChange func() error) {
...
+				logger.V(3).Info("Kubeconfig file change detected, scheduling debounced reload")
 				if w.debounceTimer != nil {
 					w.debounceTimer.Stop()
 				}
 				w.debounceTimer = time.AfterFunc(w.debounceWindow, func() {
-					klog.V(2).Info("Kubeconfig debounce window expired, triggering reload")
+					logger.V(2).Info("Kubeconfig debounce window expired, triggering reload")
 					if err := onChange(); err != nil {
-						klog.Errorf("Failed to reload after kubeconfig change: %v", err)
+						logger.Error(err, "Failed to reload after kubeconfig change: %v")
```
manusa: Leftover `%v` from the old `klog.Errorf` — logr does not interpolate the message string, so this prints the literal `: %v`. `err` is already passed as the first arg; drop the `: %v` from the message.

manusa (commented):
**Preserving WARN-vs-INFO intent**

One concern with the `klog.Warningf` → structured `.Info` conversions: `logr` has no warn level, so once this lands there's no way to tell which `.Info` calls were *meant* as warnings vs. genuinely informational — that signal lives only in the git history. When someone later wires real severity back in (e.g. an OTel logs exporter, where severity is a first-class `SeverityText`/`SeverityNumber` field), there'll be nothing in the code marking where WARN belonged.

Suggestion: capture the intent in-code now rather than losing it. A tiny helper keeps everything contextual (still trace-correlatable) while recording severity in a greppable, forward-compatible way:

```go
// Warn logs at info level but tags the record as a warning, so the
// WARN/INFO distinction survives until real log severity is wired up.
func Warn(ctx context.Context, msg string, kv ...any) {
    klog.FromContext(ctx).Info(msg, append([]any{"log.severity", "WARN"}, kv...)...)
}
```

`log.severity=WARN` is both filterable today and a clean mapping target when the logs bridge arrives. The sites that were `klog.Warningf` and are now plain `.Info` (i.e. the ones to tag):
- `pkg/config/config.go:520` — `authorization-url is using http` *(security)*
- `pkg/config/config.go:592` — `skip_jwt_verification …` *(security)*
- `pkg/http/authorization.go` — `Bearer token forwarded without local validation` (the `skipJWTWarningOnce` site) *(security)*
- `pkg/confirmation/confirmation.go` — `Confirmation rules matched but client does not support elicitation, proceeding with fallback "allow"`
- `pkg/kcp/provider.go` — `Failed to discover workspaces via API, falling back to kubeconfig`
- `pkg/kubernetes/accesscontrol_round_tripper.go` — `failed to parse Kubernetes API server host …`
- `pkg/kubernetes/provider_single.go` — `HasGVKs called with nil manager …`
- `pkg/kubernetes/provider_token_exchange.go` — `… OIDC provider returned empty token URL`
- `pkg/kubernetes/token_exchange.go` — `token exchange strategy … not found in registry`
- `pkg/toolsets/core/health_check.go` — `Namespace not found, performing cluster-wide health check`

(Separately, `pkg/kcp/workspace_watcher.go`'s `Failed to reload` went from `klog.Errorf` to `.Info` while its sibling watchers kept `.Error` — that one's arguably just a straight bug to restore to `.Error`.)

Not asking to re-introduce a warn level in this PR — just to not throw away the information about where it belonged.

pkg/http/authorization.go
```diff
@@ -89,7 +90,11 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
 
 			authHeader := r.Header.Get("Authorization")
 			if authHeader == "" || !strings.HasPrefix(authHeader, "Bearer ") {
-				klog.V(1).Infof("Authentication failed - missing or invalid bearer token: %s %s from %s", r.Method, r.URL.Path, r.RemoteAddr)
+				logger.V(1).Info("Authentication failed - missing or invalid bearer token",
+					"http.request.method", r.Method,
+					"url.path", r.URL.Path,
+					"client.address", r.RemoteAddr,
```
2uasimojo, line 97: Does this logger have WithField/WithFields methods? Anything that relies on "an even number of varargs which will be treated alternately as keys and values" has always ooked me out.

Cali0707: not that I know of - you can attach values to the logger itself with logger.WithValues, but I don't think there is a nicer API for setting this on the individual logs - I tried to separate these out onto lines to make the k/v relationship clear

2uasimojo: Mm, confirmed. And gross. Oh well.

Yeah, I dig the formatting; I would just dig compiler validation more :P

pkg/confirmation/confirmation.go
```diff
@@ -5,6 +5,7 @@ import (
 	"errors"
 
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
 	"github.com/google/jsonschema-go/jsonschema"
 	"k8s.io/klog/v2"
```
manusa: `k8s.io/klog/v2` is now imported but unused — the only `klog.` reference (`klog.Warningf`) was replaced by `klogutil.Warn`, so this import needs to be dropped. This is what's failing CI's `update-readme-tools` job:

```
pkg/confirmation/confirmation.go:10:2: "k8s.io/klog/v2" imported as klog and not used
```

Worth a local `make build` after fixing, since the compiler stops at the first error and won't have flagged any later files.

pkg/toolsets/core/health_check.go
```diff
@@ -131,93 +135,95 @@ func gatherClusterDiagnostics(params api.PromptHandlerParams, namespace string,
...
-	klog.Info("Collecting node diagnostics...")
+	logger.Info("Collecting node diagnostics...")
 	nodeDiag, err := gatherNodeDiagnostics(params)
 	if err == nil {
 		diag.Nodes = nodeDiag
-		klog.Info("Node diagnostics collected")
+		logger.Info("Node diagnostics collected")
 	} else {
-		klog.Warningf("Failed to collect node diagnostics: %v", err)
+		logger.Error(err, "Failed to collect node diagnostics")
```
manusa: This and the sibling `logger.Error(...)` calls at lines 157, 167, 176, 185, 195, and 214 escalate what were `klog.Warningf` warnings to **error** severity. These are best-effort, recoverable partial failures — each sits in an `if err == nil { … } else { log; continue }` block and the function still returns success with partial diagnostics (e.g. an RBAC-restricted namespace that can't list pods).

Two concerns: (1) a degraded-but-handled condition now logs at ERROR, which with the planned OTel logs exporter surfaces as error-severity telemetry and inflates error/alert rates; (2) it's inconsistent with the sibling warning at line 66 (`Namespace not found`), which was converted to `klogutil.Warn` — the helper this PR introduces precisely to preserve WARN intent. Consider routing these through `klogutil.Warn` too.

pkg/klogutil/util.go
```diff
@@ -0,0 +1,11 @@
+package klogutil
+
+import (
+	"context"
+
+	"k8s.io/klog/v2"
+)
+
+func Warn(ctx context.Context, msg string, kv ...any) {
```
manusa, line 12: `Warn` is a new exported helper with no godoc comment and no unit test. Its whole purpose — prepending `"log.severity", "WARN"` so the WARN-vs-INFO distinction survives — is asserted by zero tests (nothing in `pkg/` checks for `log.severity`/`"WARN"`), and it's used on security-adjacent paths (skip_jwt, auth, token exchange). Given AGENTS.md's emphasis on testing the public API, a small test asserting the tag is prepended (plus an exported-symbol godoc) would be worthwhile.

manusa (commented):
Re-reviewed after the latest push. My earlier inline points are all addressed (helm V(5) gating, the `.V(1).Error` no-ops, `Telemetry.IsEnabled()`, the `kubeconfig.go` `%v`, the root ctx from cobra, and the `klogutil.Warn` WARN-intent helper). Two things to fix, plus the CI break:

1. **Build break (CI):** `pkg/confirmation/confirmation.go` imports `klog` but no longer uses it — the unused import fails the build.
2. **`health_check.go`:** the best-effort "Failed to collect …" warnings were escalated to `logger.Error`.
3. **`klogutil.Warn`:** new exported helper with no test/godoc.

No other correctness or security issues found.

manusa (changes requested):
DELETED COMMENT; IGNORE

pkg/klogutil/util.go
```diff
@@ -0,0 +1,27 @@
...
+// WarnLogger logs the provided message + kvs with a log.severity=WARN on the
+// provided logger. This is done as klog.Logger interface does not have a native
+// Warn method
+func WarnLogger(logger klog.Logger, msg string, kv ...any) {
+	logger.Info(msg, append([]any{"log.severity", "WARN"}, kv...)...)
+}
+
+// WithError returns a key-value pair suitable for passing to a klog call
+// that records the error message under the "exception.message" key.
+func WithError(err error) []string {
```
manusa: **Regression: `WithError` produces malformed structured logs at every call site.**

It returns a `[]string` (half a key/value pair), but every caller passes it as a *single* argument into klog's `...any` list — e.g. `logger.V(4).Info(msg, "url.path", p, klogutil.WithError(err))`. It can't be spread instead (`[]string` → `...any` doesn't compile, and Go won't allow a spread alongside other args anyway), so klog receives the whole slice as one non-string *key* with no value. Verified against klog's textlogger and the `funcr` logger `util_test.go` uses:

```
this PR:  "RBAC pre-validation failed" "<non-string-key: [\"exception.mess>"="<no-value>"
want:     "RBAC pre-validation failed" exception.message="connection refused"
```

This affects all ~24 `klogutil.WithError(...)` sites and breaks `exception.message` as a queryable attribute on every error path — the opposite of this PR's goal. `TestWithError` passes only because it asserts the *returned slice*, never a rendered log line, so it doesn't catch it.

The root issue is that a *fragment-returning* helper can't compose with logr's variadic. A helper that **doesn't return a half-pair** works fine. Proposed approach — typed, composable attributes (this also resolves @2uasimojo's earlier "relies on an even number of varargs" concern, and is verified to preserve `.V(n)` gating):

```go
// pkg/klogutil
type Attr struct{ K string; V any }

func Err(err error) Attr         { return Attr{"exception.message", err.Error()} }
func Field(k string, v any) Attr { return Attr{k, v} }

func toKV(attrs []Attr) []any {
	kv := make([]any, 0, len(attrs)*2)
	for _, a := range attrs {
		kv = append(kv, a.K, a.V)
	}
	return kv
}

func LogInfo(logger klog.Logger, msg string, attrs ...Attr) { logger.Info(msg, toKV(attrs)...) }
func LogWarn(logger klog.Logger, msg string, attrs ...Attr) {
	logger.Info(msg, append([]any{"log.severity", "WARN"}, toKV(attrs)...)...)
}
```

Call sites become compiler-checked — no half-pairs, no odd-vararg risk:

```go
// rbac_validator.go
- klog.FromContext(ctx).V(4).Info("RBAC pre-validation failed", klogutil.WithError(err))
+ klogutil.LogInfo(klog.FromContext(ctx).V(4), "RBAC pre-validation failed", klogutil.Err(err))

// accesscontrol_round_tripper.go (extra field + error)
- logger.V(4).Info("Validation failed", "validator_name", v.Name(), klogutil.WithError(ve))
+ klogutil.LogInfo(logger.V(4), "Validation failed", klogutil.Field("validator_name", v.Name()), klogutil.Err(ve))

// accesscontrol_round_tripper.go (warn + extra field)
+ klogutil.LogWarn(klog.FromContext(ctx), "failed to parse Kubernetes API server host…", klogutil.Field("url.full", cfg.HostURL), klogutil.Err(err))
```

Notes:
- Passing `logger.V(n)` preserves verbosity gating (verified: a `V(4)` call at verbosity 3 emits nothing).
- `Err` centralizes the OTel `exception.message` key and leaves room for e.g. `Type(err)` → `exception.type` later.
- The existing `Warn(ctx, …)` / `WarnLogger(logger, …)` would fold into this `Attr` style, and `WithError` / `TestWithError` would be removed.

If you'd prefer a single fluent entry point over free functions, a thin facade works too (optional):
```go
klogutil.From(ctx).V(4).Info("Validation failed", klogutil.Field("validator_name", name), klogutil.Err(err))
klogutil.Wrap(logger).Warn("host parse failed", klogutil.Field("url.full", host), klogutil.Err(err))
```

I verified both shapes render correctly and gate properly — happy to send a patch if useful.

manusa (changes requested):
One blocking regression in the new `klogutil` helpers: `WithError` is misused at every call site and emits malformed `exception.message` output. Details and a proposed type-safe approach (which also addresses the even-number-of-varargs concern) are inline on the helper definition.

Cali0707:
@manusa good point with the `WithError` change - I tried to clean things up a bit but it definitely has issues. I think for now I will go back to setting those manually.

I like those improvements you have, but I worry about scope creep on this PR (I just wanted to get the context logging to happen here).

What do you think about us working on some follow up PRs to incrementally add in improvements through the `klogutil` package?

manusa:
> What do you think about us working on some follow up PRs to incrementally add in improvements through the klogutil package?

I you revert to use plain key values as before, I can file a thorough issue with the improvements to klogutil as a follow-up.

manusa:
I also have a pre-implementation ready, so it would be very easy to create the PR too.

Cali0707:
@manusa reverted to the no-helper path here, let's do a follow up with your issue?

manusa (approved):
LGTM, thx!

manusa:
> @manusa reverted to the no-helper path here, let's do a follow up with your issue?

https://github.com/containers/kubernetes-mcp-server/issues/1208

PR #1206, opened 2026-06-15
test(exec): de-flake TestPodsExec SPDY stream reset on slow runners

### LINKED ISSUE #1205: Flaky test: pkg/mcp TestPodsExec exec stream reset on slow runners (macos-latest)
## Summary

`TestPodsExec` in `pkg/mcp` is flaky on slow / contended CI runners. The exec subtest intermittently fails with a SPDY **`Stream reset`** instead of returning the expected command output. It is a test-infrastructure timing race in the shared exec mock — the mock server tears down the SPDY connection (`defer conn.Close()`) the instant its handler returns, racing the client's read of the streamed stdout. The product exec path is not at fault.

Because the race lives in the shared mock-server exec helper, it can fail **any** branch/PR build. This occurrence landed on an unrelated Dependabot PR (`helm.sh/helm/v3` `3.21.0 → 3.21.1`), which does not touch exec code at all.

## Affected subtest

- `TestPodsExec/TestPodsExec/pods_exec(name=pod-to-exec, namespace=default, command=[ls -l])/returns_command_output`

Its two sibling subtests — `namespace=nil` and `container=a-specific-container` — exercise the **identical** handler and code path and **passed** in the same run.

## Evidence

Failing run (Dependabot PR, `macos-latest` only — `ubuntu-latest`, `ubuntu-24.04-arm`, `windows-latest`, and `macos-15-intel` all passed):
https://github.com/containers/kubernetes-mcp-server/actions/runs/27523444430/job/81346084366

```
pods_exec_test.go:92:
  Error:    Should be false
  Messages: call tool failed: [0xc0009ba880]
pods_exec_test.go:93:
  Error:    "failed to exec in pod pod-to-exec in namespace default: Stream reset" does not contain "command:ls -l\n"
  Messages: unexpected result failed to exec in pod pod-to-exec in namespace default: Stream reset
```

## Root cause

The exec test drives the shared SPDY mock (`internal/test/mock_server.go` `CreateHTTPStreams`) through this handler (`pkg/mcp/pods_exec_test.go:38-56`):

```go
ctx, err := test.CreateHTTPStreams(w, req, &test.StreamOptions{Stdin: &stdin, Stdout: &stdout})
...
defer func(conn io.Closer) { _ = conn.Close() }(ctx.Closer)
_, _ = io.WriteString(ctx.StdoutStream, "command:"+strings.Join(req.URL.Query()["command"], " ")+"\n")
_, _ = io.WriteString(ctx.StdoutStream, "container:"+strings.Join(req.URL.Query()["container"], " ")+"\n")
```

1. Production exec (`pkg/kubernetes/pods.go:296-316`) runs a `remotecommand` `FallbackExecutor` (WebSocket → SPDY). Against the mock it lands on the SPDY (`v4.channel.k8s.io`) path and copies stdout via `StreamWithContext`.
2. The mock handler writes the two stdout lines and then **immediately returns**, firing `defer conn.Close()`, which closes the whole SPDY connection. It never closes the individual stdout/error streams and never sends a v4 terminal status on the error stream.
3. Closing the connection out from under the still-draining client is a race: if the client has not finished reading the buffered stdout and observing clean per-stream EOFs before the connection-level teardown reaches it, the underlying spdystream layer surfaces a `RST_STREAM` → `Stream reset` rather than EOF, and the executor returns that error.

The smoking gun is the failure shape: only **one** of three identical subtests failed while the other two passed in the same run; the asserted value is a connection-level `Stream reset` (not a content mismatch — the handler does write exactly `command:ls -l\n`); and it reproduced only on the slowest runner in the matrix. A genuine defect in the exec path would fail deterministically across all three subtests and across platforms.

## Proposed fix

Make the shared exec mock close **gracefully** so the client always observes EOF instead of a reset. Fixing it once in `CreateHTTPStreams` / the exec handler benefits every exec test:

- Close the data streams (and send the v4 terminal status on the error stream) **before** closing the SPDY connection, so each stream delivers a clean EOF; only then tear down the connection.

**Lighter mitigation (test resilience only, does not remove the race):** wrap the exec assertion in an `Eventually` / retry so a transient `Stream reset` re-runs instead of reddening the build — but this leaves the underlying close race in place.

## Acceptance criteria

- `TestPodsExec` no longer flakes with `Stream reset` on `macos-latest` / other slow runners (passes reliably under `-race`).
- The exec mock delivers stdout and a clean stream EOF to the client before the SPDY connection is closed.

---
_Triage/analysis assisted by Claude Code._

manusa:
## Summary

`TestPodsExec` (`pkg/mcp`) intermittently failed with a SPDY `Stream reset`
on slow / contended CI runners (e.g. `macos-latest`), reddening unrelated
branch and PR builds because the race lives in the shared exec mock.

## Root cause

The shared SPDY exec mock (`internal/test/mock_server.go`) closed the
connection immediately after writing stdout. `k8s.io/streaming`
`spdy.connection.Close()` resets (`RST_STREAM`) every still-registered
stream, so the teardown raced the client's stream-creation replies. When a
reset overtook a stream's `SYN_REPLY`, the `remotecommand` executor
surfaced `spdystream.ErrReset` ("Stream reset") from `WaitTimeout` during
stream creation — only on slow runners, and only intermittently.

## Fix

Add a graceful `StreamContext.Close()` that:

- writes the v4 `StatusSuccess` terminal status on the error stream, then
- closes each stream (stdout/stderr/stdin/error) — each `Close()` blocks
  until that stream's reply has been flushed and sends a FIN, giving the
  client a clean EOF, before
- closing the SPDY connection.

Because all frames are serialized by the framer, every reply and the
streamed stdout now reach the client ahead of the connection teardown's
resets. Both exec handlers call `ctx.Close()`, and the underlying
connection field is unexported so this graceful teardown is the only path.

## Verification

- Reproduced the flake in a CPU-constrained container under `-race`: the
  old immediate-close path failed 3/3 (`Stream reset` ×2, `Timeout` ×1);
  the graceful path ran 3000 iterations with zero failures.
- `go test -race` for `internal/test` and `pkg/mcp` `TestPodsExec` pass;
  `make lint` reports 0 issues.

Closes #1205

### DIFF
--- a/internal/test/mock_server.go
+++ b/internal/test/mock_server.go
@@ -112,11 +112,39 @@ type streamAndReply struct {
 }
 
 type StreamContext struct {
-	Closer       io.Closer
+	closer       io.Closer
 	StdinStream  io.ReadCloser
 	StdoutStream io.WriteCloser
 	StderrStream io.WriteCloser
+	errorStream  io.WriteCloser
 	writeStatus  func(status *apierrors.StatusError) error
+}
+
+// Close gracefully tears down the exec streams so the client always observes a
+// clean EOF (and a v4 success terminal status on the error stream) before the
+// SPDY connection is torn down. Closing each stream first blocks until that
+// stream's reply has been flushed and then sends a FIN, so the connection
+// teardown's stream resets reach the client only after it has received every
+// reply and a clean EOF, instead of racing them — the race that otherwise
+// surfaces intermittently as a "Stream reset" on slow runners. The underlying
+// connection is intentionally unexported, so handlers must defer Close to tear
+// the streams down — this graceful teardown is the only path.
+func (c *StreamContext) Close() error {
+	// Send the v4 success terminal status on the error stream.
+	if c.writeStatus != nil {
+		_ = c.writeStatus(&apierrors.StatusError{ErrStatus: metav1.Status{Status: metav1.StatusSuccess}})
+	}
+	// Close every stream so each delivers a clean EOF (and flushes its reply)
+	// before the connection-level teardown.
+	for _, stream := range []io.Closer{c.StdoutStream, c.StderrStream, c.StdinStream, c.errorStream} {
+		if stream != nil {
+			_ = stream.Close()
+		}
+	}
+	if c.closer != nil {
+		return c.closer.Close()
+	}
+	return nil
 }
 
 type StreamOptions struct {
@@ -148,7 +176,7 @@ func CreateHTTPStreams(w http.ResponseWriter, req *http.Request, opts *StreamOpt
 		return nil
 	})
 	ctx := &StreamContext{
-		Closer: connection,
+		closer: connection,
 	}
 
 	// wait for stream
@@ -173,6 +201,7 @@ WaitForStreams:
 			switch streamType {
 			case v1.StreamTypeError:
 				replyChan <- struct{}{}
+				ctx.errorStream = stream
 				ctx.writeStatus = v4WriteStatusFunc(stream)
 			case v1.StreamTypeStdout:
 				replyChan <- struct{}{}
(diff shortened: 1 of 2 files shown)


Cali0707 (approved):
LGTM, thanks for tracking this down @manusa !

PR #1212, opened 2026-06-17
fix(http): ignore SIGHUP in no-config HTTP mode

### LINKED ISSUE #1211: Follow-up to #1191: HTTP SIGHUP no-config edge case + test hardening
Follow-up to #1191 ("HTTP: SIGHUP reloads rather than terminating"), which was merged to keep momentum. During review a few items were deferred so they wouldn't hold up the fix; this issue tracks them. Line references are as of #1191's merge.

### Background

#1191 removed `syscall.SIGHUP` from `Serve`'s `signal.Notify` (`pkg/http/http.go:154`) so that, in HTTP mode **with** a config file, SIGHUP is handled solely by the config-reload handler (`pkg/kubernetes-mcp-server/cmd/root.go:385`) instead of also racing the shutdown path. That line was originally added in #241 (fix for #239, "mcp server not killed correctly when started with npx"). The core #239 fix — the npm wrapper's `spawn` + signal forwarding — is untouched, and the SIGINT/SIGTERM shutdown paths still work, so #239 is **not** regressed.

---

### 1. No-config HTTP mode: SIGHUP now terminates abruptly, contradicting the docs (Important)

HTTP mode can be entered with `--port` alone and **no** config file (`root.go:228`, `root.go:369`). In that path the reload handler is **not** registered (it's gated on `ConfigPath != "" || ConfigDir != ""`, `root.go:364`), so after #1191 nothing in the process catches SIGHUP and Go's default disposition **terminates the process abruptly** — bypassing `httpServer.Shutdown` / `mcpServer.Shutdown` (no graceful drain, no metrics flush).

- Before #1191: SIGHUP → graceful shutdown.
- After #1191 (no config): SIGHUP → abrupt termination.
- This contradicts `docs/configuration.md:99`: *"If neither is specified, SIGHUP signals are ignored."*

The child still dies, so #239's orphan bug does not return — but the behavior is neither "graceful" nor "ignored".

**Options:**
- (a) Update `docs/configuration.md:99` to state SIGHUP terminates the HTTP server when no config is set; or
- (b) Drain SIGHUP with a no-op in `Serve` so the no-config case stays truly "ignored" (the reload handler still receives its own copy when a config is present — `signal.Notify` delivers to all registered channels).

### 2. `HTTPSIGHUPSuite` lifecycle is fragile (Minor)

In `TestSIGHUPReloadsConfigFromFile` (`root_sighup_test.go:485-498`):
- `timeoutCtx` is passed to `errgroup.WithContext` but the derived context is discarded (`group, _ :=`), and the goroutine runs `opts.Run()` → `Serve(context.Background(), …)` (`root.go:370`). So the 10 s timeout governs nothing — it's misleading dead code.
- The server is stopped only by a **process-wide** `SIGTERM` (`syscall.Kill(syscall.Getpid(), …)`, line 498). Safe today (sequential tests; `Serve` still catches SIGTERM) but brittle.
- Consequently `group.Wait()` in teardown has no independent timeout: if SIGTERM handling ever regresses, the test hangs until the global `go test` deadline with no localized diagnostic.

The repo's `pkg/http/http_test.go` already threads a cancelable context into `Serve`; adopting that here (or dropping the errgroup-with-context scaffolding) would be sturdier.

### 3. Fragile / weak assertions (Minor)

- `root_sighup_test.go:551` — the "no shutdown messages in logs" guard relies on a fixed `time.Sleep(100 * time.Millisecond)` then a single buffer read. It works only because the two preceding `Eventually` subtests already give a (hypothetically buggy) shutdown path seconds to emit first; a fixed sleep is fragile to reordering. Prefer sampling right after the earlier `Eventually`s or polling the negative condition.
- `root_sighup_test.go:536` — `len(toolsAfter) > len(toolsBefore)` is a weak second assertion; the preceding `Eventually` already proves a `helm_` tool appeared. A reload that dropped a core tool while adding a helm one would still satisfy `>`.

### 4. npm wrapper SIGHUP log/semantics now mismatched (Minor, separate file)

`npm/kubernetes-mcp-server/bin/index.js` (added in #241, untouched by #1191) forwards SIGHUP to the child and logs `Received SIGHUP, terminating child process...`. With a config file the child now **reloads and keeps running**, so that log is misleading and the wrapper waits on a `'close'` that won't fire (harmless — it's meant to stay alive while the child lives). Conversely this is also a small win: `kill -HUP <npx-parent>` now triggers a real config reload of an npx-launched HTTP server. Worth reconciling the wrapper's SIGHUP handling/logging with the new reload semantics.

---

### Acceptance criteria

- [ ] Resolve item 1 (doc update or no-op SIGHUP drain) so no-config HTTP SIGHUP behavior matches documentation.
- [ ] Harden `HTTPSIGHUPSuite` lifecycle (items 2–3) — optional but recommended.
- [ ] Decide whether to adjust the npm wrapper's SIGHUP handling/logging (item 4).

Refs: #1191, #241, #239

manusa:
Follow-up to #1191, addressing the items deferred during its review (tracked in #1211).

## 1. No-config HTTP mode no longer terminates on SIGHUP (Important)

HTTP mode can be started with `--port` and no config file. In that path the SIGHUP reload handler is not registered (it is gated on `--config`/`--config-dir`), so after #1191 nothing caught SIGHUP and Go's default disposition terminated the process abruptly, skipping graceful shutdown and the metrics flush. That contradicted `docs/configuration.md` ("If neither is specified, SIGHUP signals are ignored").

Registering SIGHUP in `Serve` overrides Go's default disposition, so the no-config case is genuinely ignored. When a config is present, `signal.Notify` multicasts, so the reload handler in `root.go` still receives its own copy and reloads as before. Covered by a unit test (`TestSIGHUPIgnored`) and an end-to-end no-config test through `root.go` (`TestSIGHUPIgnoredWithoutConfig`).

## 2-3. HTTPSIGHUPSuite hardening

- Threads a cancelable context into `Complete`/`Run` and stops the server by canceling it, instead of a process-wide `SIGTERM`. The timeout context is now a real backstop with a localized teardown diagnostic.
- Replaces `len(after) > len(before)` with a per-tool superset check, so a reload that drops a core tool while adding a helm one is caught.
- Replaces a fixed `time.Sleep` plus single buffer read with a polled negative condition.

This also restores test compilation. #1191 was branched before #1202 changed `Complete`/`Run` to take a `ctx`. The two merged with no textual conflict, but the combined test package no longer compiled, so `make test` was red on `main`.

## 4. npm wrapper SIGHUP logging

The wrapper still forwards SIGHUP (so `kill -HUP` on the npx parent reaches the child and can trigger a reload), but no longer logs it as "terminating child process".

## Note on the CI gap

The reason per-PR CI did not catch the breakage is a process gap, not a workflow bug: #1191 was tested against a base that predated #1202. That is best fixed at the repo level (a branch-protection ruleset requiring the build checks and up-to-date branches), handled separately and out of scope for this PR.

## Testing

- `make test` (runs with `-race`): all packages pass (the suite failed to compile before this change), 0 races.
- `TestSIGHUPIgnored` / `TestSIGHUPIgnoredWithoutConfig` verify SIGHUP does not shut down the server; verified red before the fix (the process is killed by SIGHUP) and green after.
- `make lint`: clean.

Closes #1211

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,13 +13,9 @@ on:
       - '.gitignore'
       - 'LICENSE'
       - '*.md'
-  # Run on merge-queue entries so build + test execute against the actual
-  # combination of the PR and the current main, not the (possibly stale) base
-  # the PR was last tested against. This catches semantic merge skew — e.g. a
-  # test that compiles on the PR's base but not after another PR changes a
-  # function signature — which per-PR checks against an out-of-date base miss.
-  # (paths-ignore is not applied to merge_group events, so build + test always
-  # run in the queue.)
+  # Re-run build + test on merge-queue entries against the PR combined with
+  # current main (not the stale base it was last tested on), catching semantic
+  # merge skew. paths-ignore does not apply to merge_group, so this always runs.
   merge_group:
 
 concurrency:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,6 +13,14 @@ on:
       - '.gitignore'
       - 'LICENSE'
       - '*.md'
+  # Run on merge-queue entries so build + test execute against the actual
+  # combination of the PR and the current main, not the (possibly stale) base
+  # the PR was last tested against. This catches semantic merge skew — e.g. a
+  # test that compiles on the PR's base but not after another PR changes a
+  # function signature — which per-PR checks against an out-of-date base miss.
+  # (paths-ignore is not applied to merge_group events, so build + test always
+  # run in the queue.)
+  merge_group:
 
 concurrency:
   # Only run once for latest commit per ref and cancel other (previous) runs.
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,10 +13,6 @@ on:
       - '.gitignore'
       - 'LICENSE'
       - '*.md'
-  # Re-run build + test on merge-queue entries against the PR combined with
-  # current main (not the stale base it was last tested on), catching semantic
-  # merge skew. paths-ignore does not apply to merge_group, so this always runs.
-  merge_group:
 
 concurrency:
   # Only run once for latest commit per ref and cancel other (previous) runs.
--- a/.github/workflows/check-go-version.yaml
+++ b/.github/workflows/check-go-version.yaml
@@ -5,10 +5,6 @@ on:
     types: [opened, synchronize, reopened, labeled, unlabeled]
     paths:
       - go.mod
-  # Required checks must report on merge-queue entries or the queue stalls. This
-  # one needs PR context (a merge_group event has none), so the job below skips
-  # and reports as a passing skip there.
-  merge_group:
 
 permissions:
   contents: read
@@ -24,8 +20,6 @@ defaults:
 jobs:
   check-go-directive:
     name: Check Go directive bump
-    # Needs the PR base ref; skip (and report green) on merge_group.
-    if: ${{ github.event_name == 'pull_request' }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/.github/workflows/check-go-version.yaml
+++ b/.github/workflows/check-go-version.yaml
@@ -5,12 +5,9 @@ on:
     types: [opened, synchronize, reopened, labeled, unlabeled]
     paths:
       - go.mod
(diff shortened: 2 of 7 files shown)


manusa:
Merging since main is currently degraded, this PR addresses that.
Please do review the PR in case there is anything wrong and I'll address in follow up.

PR #1215, opened 2026-06-18
feat(api): add output schema support for MCP tools

andreasgerstmayr:
Allow tools to declare an OutputSchema that describes the structure of their StructuredContent response, forwarded to the Go SDK's mcp.Tool.

### DIFF
--- a/docs/specs/structured-output.md
+++ b/docs/specs/structured-output.md
@@ -177,12 +177,6 @@ This repo takes an **opportunistic** stance:
 When in doubt, leave it off and add it later in a follow-up PR — declaring
 nothing is preferable to declaring something inaccurate.
 
-> **Wiring note:** `api.Tool` (see
-> [`pkg/api/toolsets.go`](../../pkg/api/toolsets.go)) carries `InputSchema` but
-> does not yet expose an `OutputSchema` field. Adding the field, surfacing it
-> through the MCP go-sdk transport, and generating it from typed structs is a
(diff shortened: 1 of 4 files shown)


docs/specs/structured-output.md
```diff
@@ -177,12 +177,6 @@ This repo takes an **opportunistic** stance:
 When in doubt, leave it off and add it later in a follow-up PR — declaring
 nothing is preferable to declaring something inaccurate.
 
-> **Wiring note:** `api.Tool` (see
-> [`pkg/api/toolsets.go`](../../pkg/api/toolsets.go)) carries `InputSchema` but
-> does not yet expose an `OutputSchema` field. Adding the field, surfacing it
-> through the MCP go-sdk transport, and generating it from typed structs is a
```
andreasgerstmayr, line 183: > generating it from typed structs

The `Tool` struct doesn't have access to the return type of the tool handler.

I guess we could provide some separate utility function like this: https://github.com/rhobs/obs-mcp/blob/9a8e6331904b38d627172ea7d9dd0130612ec018/pkg/tools/tooldef.go#L90-L93

manusa: We can create that as a follow-up issue.
AFAIR the go-sdk also contained some tooling with regards to the schemas (at least in the dependent library they created).

manusa (approved):
LGTM, thx!

PR #1217, opened 2026-06-19
fix(auth): refresh STS config after reload

### LINKED ISSUE #1110: STS token exchange: stale cache on SIGHUP reload + missing test coverage (follow-up to #1109)
Follow-up to #1109 / #1108. Now that `certificate_authority` propagates to `TargetTokenExchangeConfig.CAFile`, two related gaps in the same code path are user-visible.

### 1. Stale `*TargetTokenExchangeConfig` cached across SIGHUP

`tokenExchangingProvider.getOrBuildStsConfig` (`pkg/kubernetes/provider_token_exchange.go:71-102`) keys its cache on `stsTokenURL` only. Since the IdP's token URL is typically stable across config changes, rotating any other STS-related field returns the previously cached config:

- `certificate_authority`
- `sts_client_cert_file`, `sts_client_key_file`, `sts_auth_style`
- `sts_client_id`, `sts_client_secret`, `sts_audience`, `sts_scopes`

`Snapshot.HasProviderConfigChanged` (`pkg/oauth/state.go:34-36`) only watches `AuthorizationURL` and `CertificateAuthority`, so STS-only field rotations don't even reach `getOrBuildStsConfig` with a different snapshot.

Worse, `TargetTokenExchangeConfig` memoizes its `*http.Client` on first `HTTPClient()` call (`pkg/tokenexchange/config.go:62, 95-96`), so even if the cached struct were swapped out, an in-flight reference would keep serving the old TLS config (old CA, old cert/key).

`reloadToolsets` (`pkg/mcp/mcp.go:143-190`) does not rebuild `s.p`, so the same `tokenExchangingProvider` instance — with the same cached config — survives reloads.

**Suggested fix:**
- Cache key: a fingerprint of the relevant fields (CA path + cert/key paths + auth style + scopes + client-id/secret + audience), not just the token URL.
- `TargetTokenExchangeConfig`: drop the `*http.Client` memoization (build fresh per `Exchange()`), or expose an explicit invalidation hook. The assertion cache (`cachedAssertion`, `cachedAssertionExpiry`) is intentional and must stay.

### 2. No regression test for `CAFile` propagation

`pkg/tokenexchange/*_test.go` has zero `CAFile` / `certificate_authority` references. The CA-loading branches at `pkg/tokenexchange/config.go:106-117` are uncovered.

**Suggested coverage** (testify/suite, behavior-based, per `CLAUDE.md`):

New `pkg/tokenexchange/config_test.go`:
- `CAFile == ""` → transport's `TLSClientConfig.RootCAs == nil`.
- `CAFile = <self-signed PEM>` against `httptest.NewTLSServer` using that cert → request succeeds; request to a server with a different cert → fails with `x509: certificate signed by unknown authority`.
- `CAFile = <invalid PEM>` → error contains `failed to parse CA certificate`.
- `CAFile = <nonexistent path>` → error contains `failed to read CA file`.

End-to-end behavior test: `StaticConfig` with `CertificateAuthority` set + `httptest.NewTLSServer` exercising a strategy exchanger so propagation is observable through the public surface.

### 3. Interface placement of `GetCertificateAuthority`

`GetCertificateAuthority()` was added to `StsConfigProvider` (`pkg/api/config.go:75`), but `certificate_authority` is a top-level OAuth/TLS setting — also consumed by `pkg/oauth/state.go`. Mixing a non-STS accessor into an `Sts*` interface dilutes its role.

**Suggested fix:** extract `CertificateAuthorityProvider { GetCertificateAuthority() string }`, embed it in `BaseConfig` next to `StsConfigProvider`, remove from `StsConfigProvider`. Only `*StaticConfig` implements these methods (`pkg/config/config.go:427`), so the change is mechanical.

jstar0:
Fixes #1110

## Summary

STS token exchange kept using stale settings after a SIGHUP reload when the IdP token URL stayed the same. That meant rotated client credentials, audience, scopes, auth style, cert/key paths, or CA path could be ignored by later exchange attempts.

## Changes

- Read token exchange settings from the live config state when the wrapper handles a derived Kubernetes request.
- Rebuild the cached STS config when any token exchange or TLS input changes, not only when the token URL changes.
- Invalidate the token exchange HTTP client when `CAFile` changes and add CAFile regression coverage.

## Verification

```bash
go test ./pkg/kubernetes ./pkg/tokenexchange ./pkg/oauth ./pkg/kubernetes-mcp-server/cmd -count=1
go test ./pkg/http -run 'Test.*Token|Test.*OAuth|Test.*Authorization|Test.*WellKnown|Test.*STS|Test.*MCP' -count=1
go test -race ./pkg/kubernetes ./pkg/tokenexchange ./pkg/oauth ./pkg/kubernetes-mcp-server/cmd -count=1
make lint
```

Note: `go test ./pkg/http -run 'TestHealthCheck|TestMiddlewareLogging' -count=1` currently fails on my checkout with health/logging assertions caused by the mock discovery handler returning `unexpected end of JSON input`. I reproduced the same failure on an unmodified `origin/main` checkout, so I did not count it as part of this patch's passing verification.

### DIFF
--- a/pkg/kubernetes-mcp-server/cmd/root.go
+++ b/pkg/kubernetes-mcp-server/cmd/root.go
@@ -335,8 +335,16 @@ func (m *MCPServerOptions) Run(ctx context.Context) error {
 		return err
 	}
 	oauthState := internaloauth.NewState(internaloauth.SnapshotFromConfig(m.StaticConfig, oidcProvider, httpClient))
+	cfgState := config.NewStaticConfigState(m.StaticConfig)
 
-	provider, err := kubernetes.NewProvider(ctx, m.StaticConfig, kubernetes.WithTokenExchange(oauthState))
+	provider, err := kubernetes.NewProvider(
+		ctx,
+		m.StaticConfig,
+		kubernetes.WithTokenExchange(oauthState),
+		kubernetes.WithBaseConfigProvider(func() api.BaseConfig {
+			return cfgState.Load()
+		}),
+	)
 	if err != nil {
 		return fmt.Errorf("unable to create kubernetes target provider: %w", err)
 	}
@@ -355,8 +363,6 @@ func (m *MCPServerOptions) Run(ctx context.Context) error {
 			klog.Errorf("MCP server shutdown error: %v", err)
 		}
 	}()
-
-	cfgState := config.NewStaticConfigState(m.StaticConfig)
 
 	// Set up SIGHUP handler for configuration reload. The returned stop
 	// function unregisters the signal handler and waits for the goroutine
--- a/pkg/kubernetes/provider.go
+++ b/pkg/kubernetes/provider.go
@@ -58,12 +58,19 @@ type TokenExchangeProvider interface {
 type ProviderOption func(*providerOptions)
 
 type providerOptions struct {
-	oauthState *oauth.State
+	oauthState         *oauth.State
+	baseConfigProvider func() api.BaseConfig
 }
 
 func WithTokenExchange(oauthState *oauth.State) ProviderOption {
 	return func(opts *providerOptions) {
 		opts.oauthState = oauthState
+	}
+}
+
+func WithBaseConfigProvider(baseConfigProvider func() api.BaseConfig) ProviderOption {
+	return func(opts *providerOptions) {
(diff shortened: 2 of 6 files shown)


manusa (approved):
LGTM, thx!

PR #1219, opened 2026-06-20
fix(mcp): mark target parameter as required in multi-target mode

### LINKED ISSUE #1162: [multi-cluster] context parameter should be marked required when multiple clusters detected
When multi-cluster is enabled and kubeconfig has multiple contexts, the `context`
parameter is injected into all tool schemas but registered WITHOUT mcp.Required().

This means LLMs correctly read the JSON Schema and treat it as optional, causing
tool calls to fall back to the default context instead of the user-specified cluster.

Fix: add mcp.Required() when registering the context parameter in multi-cluster mode.

One-line fix in provider_kubeconfig.go:
  mcp.WithString(KubeConfigTargetParameterName,
      mcp.Description("..."),
      mcp.Required(),   ← add this
  )

fuleinist:
Fixes containers/kubernetes-mcp-server#1162

### DIFF
--- a/pkg/mcp/testdata/toolsets-full-tools-multicluster.json
+++ b/pkg/mcp/testdata/toolsets-full-tools-multicluster.json
@@ -46,7 +46,7 @@
     "inputSchema": {
       "properties": {
         "context": {
-          "description": "Optional parameter selecting which context to run the tool in. Defaults to fake-context if not set",
+          "description": "Required parameter selecting which context to run the tool in. Defaults to fake-context if not set",
           "type": "string"
         },
         "fieldSelector": {
@@ -59,6 +59,9 @@
           "type": "string"
         }
       },
+      "required": [
+        "context"
+      ],
       "type": "object"
     },
     "name": "events_list",
@@ -75,7 +78,7 @@
     "inputSchema": {
       "properties": {
         "context": {
-          "description": "Optional parameter selecting which context to run the tool in. Defaults to fake-context if not set",
+          "description": "Required parameter selecting which context to run the tool in. Defaults to fake-context if not set",
           "type": "string"
         },
         "fieldSelector": {
@@ -84,6 +87,9 @@
           "type": "string"
         }
       },
+      "required": [
+        "context"
+      ],
       "type": "object"
     },
     "name": "namespaces_list",
@@ -100,7 +106,7 @@
     "inputSchema": {
       "properties": {
         "context": {
-          "description": "Optional parameter selecting which context to run the tool in. Defaults to fake-context if not set",
+          "description": "Required parameter selecting which context to run the tool in. Defaults to fake-context if not set",
           "type": "string"
         },
         "name": {
@@ -120,7 +126,8 @@
       },
       "required": [
         "name",
-        "query"
+        "query",
+        "context"
       ],
       "type": "object"
     },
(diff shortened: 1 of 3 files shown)


manusa:
Before moving on with this, we need actual use cases about why this is needed since our initial analysis pushed the needle towards making it an optional parameter:

https://github.com/containers/kubernetes-mcp-server/issues/1162#issuecomment-4652965859

fuleinist:
Sure — three concrete scenarios where requiring the parameter is the right call. Happy to encode any of these as test cases if helpful.

**1. AI agent deploying to the wrong cluster in a multi-env setup.**
A user tells the agent: "deploy the canary to staging". With the parameter optional, the model reads JSON Schema, sees `context` as optional, and falls back to the kubeconfig default. If the user's `current-context` is `prod` (a common situation — devs set their shell to prod for kubectl convenience, then ask the agent to do staging work), the agent deploys to prod instead of staging. Marking it required forces the agent to think about target. This is the failure mode the original issue is reporting.

**2. Multi-tenant clusters with strict isolation guarantees.**
When the same MCP server is exposed to multiple tenants, each user has a kubeconfig with several contexts. A user with `dev`, `staging`, `prod` asks "list secrets in my dev namespace" — without `Required`, the model can pick the wrong context. In a regulated environment (PCI/HIPAA), the wrong-cluster fallback isn't recoverable: the audit log records the action against the wrong tenant. Forcing `context` means the agent must surface which cluster it's targeting before acting.

**3. CI/CD automation that needs explicit, traceable cluster selection.**
A test harness drives the MCP server across N clusters in a fan-out test matrix. Today, with the parameter optional, the harness has to set `current-context` per invocation via env var and trust the model doesn't override. Required means the harness passes `target=cluster-A` as a tool argument and the server's input validation rejects calls without it — a much cleaner contract for automation than relying on kubeconfig state.

**On your original analysis ("the default is likely correct a lot of the time"):** that's true for the *interactive* single-cluster case, where the user genuinely has one context. The fix in #1219 only adds `Required` in the `isMultiTarget` branch — single-cluster deployments are completely untouched, so the default-still-works ergonomics are preserved for the common case. The change is scoped to the situation where the user has explicitly opted into multi-cluster mode and the model has multiple valid targets to pick from, which is exactly when the wrong-default risk is highest.

If the concern is that requiring it makes the LLM produce too many redundant `context="X"` calls in trivial interactions, an alternative would be to keep the parameter optional but have the server log a warning + return the resolved target in the tool result (so the model self-corrects). Happy to refactor toward that instead if you'd prefer — just let me know which direction matches the design intent.

Cali0707:
@fuleinist thanks for the detailed response. After thinking about your comment, I don't think marking the paramater as `required` in the JSON schema actually addresses the failure modes you've described.

The core issue is that `required` only forces the model to provide _a_ value, it can't ensure that the _correct_ value is provided. In your scenarios, the consequence of error is high. But a model that would silently fall back to the wrong default is equally capable of hallucinating the wrong cluster explicitly. This change only moves the failure mode from wrong implicit default to wrong explicit selection.

Additionally, the server doesn't enfore the required from the JSON schema, omitted parameters still silently fall back to the default target regardless of this change (e.g. if a client ignores the required field). 

For the scenarios you've described the right approach would be to remove the decision from the LLM entirely, through something like:

- Running separate MCP server instances per env, so that the wrong cluster is not reachable
- Use the confirmation_rules we already support to require human approval before executing against a specific cluster (would need an extension to the current work to support setting rules on the cluster target)
- Scope the kubeconfig provided to the MCP server to only expose the contexts appropriate for the current setup

Cali0707:
I'm going to close this PR, as I don't think this is the correct approach to the problem. If there is a concrete proposal for controls around multi-cluster safety, happy to discuss that on the original issue

PR #1220, opened 2026-06-21
feat(kubevirt): add HCO awareness using MCP tooling

codingben:
Add HyperConverged Cluster Operator (HCO) support to the KubeVirt toolset. The hco-status MCP prompt auto-discovers the HyperConverged CR and reports on conditions, component health, configuration, and events.

Includes build/hco.mk for deploying HCO on Kind clusters and eval tasks that exercise HCO workflows to identify gaps in current tooling.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,29 @@ local-env-setup-kubevirt: ## Setup complete local development environment with K
 	@echo "KubeVirt is now available!"
 	@echo "Check status with: make kubevirt-status"
 
+.PHONY: local-env-setup-hco
+local-env-setup-hco: ## Setup complete local development environment with Kind cluster and HCO
+	@echo "========================================="
+	@echo "Kubernetes MCP Server - Local Setup"
+	@echo "            with HCO"
+	@echo "========================================="
+	$(MAKE) kind-create-cluster
+	$(MAKE) hco-install
+	$(MAKE) build
+	@echo ""
+	@echo "========================================="
+	@echo "Local environment ready!"
+	@echo "========================================="
+	@echo ""
+	@echo "Run the MCP server with:"
+	@echo "  ./$(BINARY_NAME)"
+	@echo ""
+	@echo "Or run with MCP inspector:"
+	@echo "  npx @modelcontextprotocol/inspector@latest \$$(pwd)/$(BINARY_NAME)"
+	@echo ""
+	@echo "HCO is now available!"
+	@echo "Check status with: make hco-status"
+
 .PHONY: local-env-setup-tekton
 local-env-setup-tekton: ## Setup complete local development environment with Kind cluster and Tekton Pipelines
 	@echo "========================================="
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,29 @@ local-env-setup-kubevirt: ## Setup complete local development environment with M
 	@echo "KubeVirt is now available!"
 	@echo "Check status with: make kubevirt-status"
 
+.PHONY: local-env-setup-hco
+local-env-setup-hco: ## Setup complete local development environment with Kind cluster and HCO
+	@echo "========================================="
+	@echo "Kubernetes MCP Server - Local Setup"
+	@echo "            with HCO"
+	@echo "========================================="
+	$(MAKE) kind-create-cluster
+	$(MAKE) hco-install
+	$(MAKE) build
+	@echo ""
+	@echo "========================================="
+	@echo "Local environment ready!"
+	@echo "========================================="
+	@echo ""
+	@echo "Run the MCP server with:"
+	@echo "  ./$(BINARY_NAME)"
+	@echo ""
+	@echo "Or run with MCP inspector:"
+	@echo "  npx @modelcontextprotocol/inspector@latest \$$(pwd)/$(BINARY_NAME)"
+	@echo ""
+	@echo "HCO is now available!"
+	@echo "Check status with: make hco-status"
+
 .PHONY: local-env-setup-tekton
 local-env-setup-tekton: ## Setup complete local development environment with Minikube cluster and Tekton Pipelines
 	@echo "========================================="
--- a/build/hco.mk
+++ b/build/hco.mk
@@ -1,0 +1,191 @@
+# HyperConverged Cluster Operator (HCO) installation and management
+#
+# Mirrors the official deploy flow from:
+#   https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/hack/deploy.sh
+#
+# HCO deploys and manages KubeVirt, CDI, Cluster Network Addons, and other
+# components as a single operator. This is an ALTERNATIVE to the standalone
+# KubeVirt install (kubevirt-install) — do not use both together.
+
+# HCO version configuration
+HCO_VERSION ?= v1.18.1
+HCO_NAMESPACE = kubevirt-hyperconverged
+HCO_RAW_URL = https://raw.githubusercontent.com/kubevirt/hyperconverged-cluster-operator/$(HCO_VERSION)/deploy
+HCO_ARCHIVE_URL = https://github.com/kubevirt/hyperconverged-cluster-operator/archive/refs/tags/$(HCO_VERSION).tar.gz
+HCO_CRDS = $(eval HCO_CRDS := $(shell curl -sL $(HCO_ARCHIVE_URL) | tar -tz | grep '/deploy/crds/.*\.crd\.yaml$$' | sed 's#.*/##'))$(HCO_CRDS)
+HCO_CRD_NAME = hyperconvergeds.hco.kubevirt.io
+
+# cert-manager configuration
+CERT_MANAGER_VERSION ?= v1.18.2
+CERT_MANAGER_TIMEOUT ?= 120s
+
+##@ HCO (HyperConverged Cluster Operator)
+
+.PHONY: cert-manager-install
+cert-manager-install: ## Install cert-manager (required by HCO webhooks)
+	@if kubectl get namespace cert-manager > /dev/null 2>&1 && \
+		kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager --timeout=5s > /dev/null 2>&1; then \
+		echo "✅ cert-manager is already installed and running, skipping installation"; \
+	else \
+		echo "========================================="; \
+		echo "Installing cert-manager $(CERT_MANAGER_VERSION)"; \
+		echo "========================================="; \
+		echo ""; \
+		kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MANAGER_VERSION)/cert-manager.yaml; \
+		echo ""; \
+		echo "Waiting for cert-manager deployments to become ready..."; \
+		kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager --timeout=$(CERT_MANAGER_TIMEOUT); \
+		kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-cainjector --timeout=$(CERT_MANAGER_TIMEOUT); \
+		kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=$(CERT_MANAGER_TIMEOUT); \
+		echo ""; \
+		echo "Waiting for cert-manager webhook CA bundle to become valid..."; \
+		for i in $$(seq 1 60); do \
+			if kubectl get validatingwebhookconfigurations cert-manager-webhook -o jsonpath='{.webhooks[0].clientConfig.caBundle}' 2>/dev/null | base64 -d | openssl x509 -noout 2>/dev/null; then \
+				echo "cert-manager webhook CA bundle is valid"; \
+				break; \
+			fi; \
+			if [ $$i -eq 60 ]; then \
+				echo "Warning: cert-manager webhook CA bundle still not valid after 60 attempts"; \
+			fi; \
+			sleep 1; \
+		done; \
+		echo "✅ cert-manager $(CERT_MANAGER_VERSION) is ready"; \
+	fi
+
+.PHONY: cert-manager-uninstall
+cert-manager-uninstall: ## Uninstall cert-manager
+	@echo "Uninstalling cert-manager..."
+	@kubectl delete -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MANAGER_VERSION)/cert-manager.yaml --ignore-not-found 2>/dev/null || true
+	@echo "✅ cert-manager uninstalled"
+
+.PHONY: hco-install
+hco-install: cert-manager-install ## Install HCO on the cluster (manages KubeVirt, CDI, network addons)
+	@echo ""
+	@echo "========================================="
+	@echo "Installing HCO $(HCO_VERSION)"
+	@echo "========================================="
+	@echo ""
+	@echo "Creating namespace..."
+	@kubectl create namespace $(HCO_NAMESPACE) --dry-run=client -o yaml | kubectl apply -f -
+	@echo ""
+	@echo "Installing HCO CRDs..."
+	@for crd in $(HCO_CRDS); do \
+		kubectl apply --server-side -f $(HCO_RAW_URL)/crds/$${crd}; \
+	done
+	@echo ""
+	@echo "Waiting for CRDs to become established..."
+	@for crd in $(HCO_CRDS); do \
+		kubectl wait --for=condition=Established crd/$$(basename $${crd} .crd.yaml) --timeout=60s 2>/dev/null || true; \
+	done
+	@if [ "$$(kubectl get crd $(HCO_CRD_NAME) -o=jsonpath='{.status.conditions[?(@.type=="NonStructuralSchema")].status}' 2>/dev/null)" = "True" ]; then \
+		echo "Warning: HCO CRD reports NonStructuralSchema condition"; \
+		kubectl get crd $(HCO_CRD_NAME) -o go-template='{{ range .status.conditions }}{{ .type }}{{ "\t" }}{{ .status }}{{ "\t" }}{{ .message }}{{ "\n" }}{{ end }}'; \
+	fi
+	@echo ""
+	@echo "Installing HCO RBAC..."
+	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/cluster_role.yaml
+	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/service_account.yaml
+	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/cluster_role_binding.yaml
+	@echo ""
+	@echo "Installing HCO webhooks..."
+	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/webhooks.yaml
+	@echo ""
+	@echo "Installing HCO operator (excluding OpenShift-only components)..."
+	@kubectl apply -n $(HCO_NAMESPACE) -l 'name!=ssp-operator,name!=hyperconverged-cluster-cli-download' -f $(HCO_RAW_URL)/operator.yaml
+	@echo ""
+	@echo "Enabling software emulation for Kind (no hardware virtualization)..."
+	@kubectl set env deployment/hyperconverged-cluster-operator -n $(HCO_NAMESPACE) KVM_EMULATION=true
+	@echo ""
+	@echo "Waiting for HCO operator to become ready..."
+	@kubectl wait deployment/hyperconverged-cluster-operator -n $(HCO_NAMESPACE) --for=condition=Available --timeout=10m
+	@echo "✅ HCO operator is ready"
+	@echo ""
+	@echo "Waiting for HCO webhook to become ready..."
+	@kubectl wait deployment/hyperconverged-cluster-webhook -n $(HCO_NAMESPACE) --for=condition=Available --timeout=5m
+	@echo "✅ HCO webhook is ready"
+	@echo ""
+	@echo "Waiting for sub-operators to become ready..."
+	@kubectl wait deployment/cdi-operator -n $(HCO_NAMESPACE) --for=condition=Available --timeout=9m
+	@kubectl wait deployment/cluster-network-addons-operator -n $(HCO_NAMESPACE) --for=condition=Available --timeout=9m
+	@kubectl wait deployment/kubevirt-migration-operator -n $(HCO_NAMESPACE) --for=condition=Available --timeout=9m 2>/dev/null || true
+	@echo "✅ Sub-operators are ready"
+	@echo ""
+	@echo "Creating HyperConverged CR..."
+	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/hco.cr.yaml
+	@echo ""
+	@echo "Waiting for HCO to become available (this can take several minutes)..."
+	@kubectl wait hyperconverged kubevirt-hyperconverged -n $(HCO_NAMESPACE) --for condition=Available --timeout=30m
+	@echo "✅ HCO is ready"
+	@echo ""
+	@echo "Waiting for component deployments..."
+	@kubectl wait deployment/cdi-apiserver -n $(HCO_NAMESPACE) --for=condition=Available --timeout=6m
+	@kubectl wait deployment/cdi-deployment -n $(HCO_NAMESPACE) --for=condition=Available --timeout=6m
+	@kubectl wait deployment/cdi-uploadproxy -n $(HCO_NAMESPACE) --for=condition=Available --timeout=6m
+	@kubectl wait deployment/virt-api -n $(HCO_NAMESPACE) --for=condition=Available --timeout=6m
+	@kubectl wait deployment/virt-controller -n $(HCO_NAMESPACE) --for=condition=Available --timeout=6m
+	@echo "✅ All component deployments are ready"
+	@echo ""
+	@echo "========================================="
+	@echo "HCO Installation Complete"
+	@echo "========================================="
+	@echo ""
+	@echo "HCO version: $(HCO_VERSION)"
+	@echo "Namespace: $(HCO_NAMESPACE)"
+	@echo ""
+	@echo "Managed components:"
+	@echo "  - KubeVirt"
+	@echo "  - CDI (Containerized Data Importer)"
+	@echo "  - Cluster Network Addons"
+	@echo ""
+	@echo "Verify installation with:"
+	@echo "  make hco-status"
+	@echo ""
+
+.PHONY: hco-uninstall
+hco-uninstall: ## Uninstall HCO and all managed components from the cluster
+	@echo "Uninstalling HCO..."
+	@kubectl delete hyperconverged kubevirt-hyperconverged -n $(HCO_NAMESPACE) --ignore-not-found --timeout=5m
+	@kubectl delete -n $(HCO_NAMESPACE) -l 'name!=ssp-operator,name!=hyperconverged-cluster-cli-download' -f $(HCO_RAW_URL)/operator.yaml --ignore-not-found 2>/dev/null || true
+	@kubectl delete -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/webhooks.yaml --ignore-not-found 2>/dev/null || true
+	@kubectl delete -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/cluster_role_binding.yaml --ignore-not-found 2>/dev/null || true
+	@kubectl delete -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/service_account.yaml --ignore-not-found 2>/dev/null || true
+	@kubectl delete -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/cluster_role.yaml --ignore-not-found 2>/dev/null || true
+	@for crd in $(HCO_CRDS); do \
+		kubectl delete -f $(HCO_RAW_URL)/crds/$${crd} --ignore-not-found 2>/dev/null || true; \
+	done
+	@kubectl delete namespace $(HCO_NAMESPACE) --ignore-not-found || true
+	@echo "✅ HCO uninstalled"
+
+.PHONY: hco-status
+hco-status: ## Show HCO and managed component status
+	@echo "========================================="
+	@echo "HCO Status"
+	@echo "========================================="
+	@echo ""
+	@echo "HyperConverged CR:"
+	@kubectl get hyperconverged -n $(HCO_NAMESPACE) -o wide || echo "HCO not installed"
+	@echo ""
+	@echo "HyperConverged Conditions:"
+	@kubectl get hyperconverged kubevirt-hyperconverged -n $(HCO_NAMESPACE) -o jsonpath='{range .status.conditions[*]}{.type}{"\t"}{.status}{"\t"}{.message}{"\n"}{end}' 2>/dev/null || echo "HCO not installed"
+	@echo ""
+	@echo "KubeVirt (managed by HCO):"
+	@kubectl get kubevirt -A -o wide 2>/dev/null || echo "KubeVirt not found"
+	@echo ""
+	@echo "CDI (managed by HCO):"
+	@kubectl get cdi -o wide 2>/dev/null || echo "CDI not found"
+	@echo ""
+	@echo "Network Addons (managed by HCO):"
+	@kubectl get networkaddonsconfig -o wide 2>/dev/null || echo "Network Addons not found"
+	@echo ""
+	@echo "HCO Operator Pods:"
+	@kubectl get pods -n $(HCO_NAMESPACE) -l name=hyperconverged-cluster-operator
+	@echo ""
+	@echo "KubeVirt Pods:"
+	@kubectl get pods -n $(HCO_NAMESPACE) -l kubevirt.io 2>/dev/null || echo "No KubeVirt pods found"
+	@echo ""
+	@echo "CDI Pods:"
+	@kubectl get pods -n $(HCO_NAMESPACE) -l cdi.kubevirt.io 2>/dev/null || echo "No CDI pods found"
+	@echo ""
+	@echo "Component Versions:"
+	@kubectl get hyperconverged kubevirt-hyperconverged -n $(HCO_NAMESPACE) -o jsonpath='{range .status.versions[*]}  {.name}: {.version}{"\n"}{end}' 2>/dev/null || echo "  N/A"
+	@echo ""
--- a/build/hco.mk
+++ b/build/hco.mk
@@ -1,0 +1,191 @@
+# HyperConverged Cluster Operator (HCO) installation and management
+#
+# Mirrors the official deploy flow from:
+#   https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/hack/deploy.sh
+#
+# HCO deploys and manages KubeVirt, CDI, Cluster Network Addons, and other
+# components as a single operator. This is an ALTERNATIVE to the standalone
+# KubeVirt install (kubevirt-install) — do not use both together.
+
+# HCO version configuration
+HCO_VERSION ?= v1.18.1
+HCO_NAMESPACE = kubevirt-hyperconverged
+HCO_RAW_URL = https://raw.githubusercontent.com/kubevirt/hyperconverged-cluster-operator/$(HCO_VERSION)/deploy
+HCO_API_URL = https://api.github.com/repos/kubevirt/hyperconverged-cluster-operator/contents/deploy/crds?ref=$(HCO_VERSION)
+HCO_CRDS = $(eval HCO_CRDS := $(shell curl -sL $(HCO_API_URL) | grep -o '"name": *"[^"]*\.crd\.yaml"' | sed 's/"name": *"//;s/"//'))$(HCO_CRDS)
+HCO_CRD_NAME = hyperconvergeds.hco.kubevirt.io
+
+# cert-manager configuration
+CERT_MANAGER_VERSION ?= v1.18.2
+CERT_MANAGER_TIMEOUT ?= 120s
+
+##@ HCO (HyperConverged Cluster Operator)
+
+.PHONY: cert-manager-install
+cert-manager-install: ## Install cert-manager (required by HCO webhooks)
+	@if kubectl get namespace cert-manager > /dev/null 2>&1 && \
+		kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager --timeout=5s > /dev/null 2>&1; then \
+		echo "✅ cert-manager is already installed and running, skipping installation"; \
+	else \
+		echo "========================================="; \
+		echo "Installing cert-manager $(CERT_MANAGER_VERSION)"; \
+		echo "========================================="; \
+		echo ""; \
+		kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MANAGER_VERSION)/cert-manager.yaml; \
+		echo ""; \
+		echo "Waiting for cert-manager deployments to become ready..."; \
+		kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager --timeout=$(CERT_MANAGER_TIMEOUT); \
+		kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-cainjector --timeout=$(CERT_MANAGER_TIMEOUT); \
+		kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=$(CERT_MANAGER_TIMEOUT); \
+		echo ""; \
+		echo "Waiting for cert-manager webhook CA bundle to become valid..."; \
+		for i in $$(seq 1 60); do \
+			if kubectl get validatingwebhookconfigurations cert-manager-webhook -o jsonpath='{.webhooks[0].clientConfig.caBundle}' 2>/dev/null | base64 -d | openssl x509 -noout 2>/dev/null; then \
+				echo "cert-manager webhook CA bundle is valid"; \
+				break; \
+			fi; \
+			if [ $$i -eq 60 ]; then \
+				echo "Warning: cert-manager webhook CA bundle still not valid after 60 attempts"; \
+			fi; \
+			sleep 1; \
+		done; \
+		echo "✅ cert-manager $(CERT_MANAGER_VERSION) is ready"; \
+	fi
+
+.PHONY: cert-manager-uninstall
+cert-manager-uninstall: ## Uninstall cert-manager
+	@echo "Uninstalling cert-manager..."
+	@kubectl delete -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MANAGER_VERSION)/cert-manager.yaml --ignore-not-found 2>/dev/null || true
+	@echo "✅ cert-manager uninstalled"
+
+.PHONY: hco-install
+hco-install: cert-manager-install ## Install HCO on the cluster (manages KubeVirt, CDI, network addons)
+	@echo ""
+	@echo "========================================="
+	@echo "Installing HCO $(HCO_VERSION)"
+	@echo "========================================="
+	@echo ""
+	@echo "Creating namespace..."
+	@kubectl create namespace $(HCO_NAMESPACE) --dry-run=client -o yaml | kubectl apply -f -
+	@echo ""
+	@echo "Installing HCO CRDs..."
+	@for crd in $(HCO_CRDS); do \
+		kubectl apply --server-side -f $(HCO_RAW_URL)/crds/$${crd}; \
+	done
+	@echo ""
+	@echo "Waiting for CRDs to become established..."
+	@for crd in $(HCO_CRDS); do \
+		kubectl wait --for=condition=Established crd/$$(basename $${crd} .crd.yaml) --timeout=60s 2>/dev/null || true; \
+	done
+	@if [ "$$(kubectl get crd $(HCO_CRD_NAME) -o=jsonpath='{.status.conditions[?(@.type=="NonStructuralSchema")].status}' 2>/dev/null)" = "True" ]; then \
+		echo "Warning: HCO CRD reports NonStructuralSchema condition"; \
+		kubectl get crd $(HCO_CRD_NAME) -o go-template='{{ range .status.conditions }}{{ .type }}{{ "\t" }}{{ .status }}{{ "\t" }}{{ .message }}{{ "\n" }}{{ end }}'; \
+	fi
+	@echo ""
+	@echo "Installing HCO RBAC..."
+	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/cluster_role.yaml
+	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/service_account.yaml
+	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/cluster_role_binding.yaml
+	@echo ""
+	@echo "Installing HCO webhooks..."
+	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/webhooks.yaml
+	@echo ""
+	@echo "Installing HCO operator (excluding OpenShift-only components)..."
+	@kubectl apply -n $(HCO_NAMESPACE) -l 'name!=ssp-operator,name!=hyperconverged-cluster-cli-download' -f $(HCO_RAW_URL)/operator.yaml
+	@echo ""
+	@echo "Enabling software emulation for Kind (no hardware virtualization)..."
+	@kubectl set env deployment/hyperconverged-cluster-operator -n $(HCO_NAMESPACE) KVM_EMULATION=true
+	@echo ""
+	@echo "Waiting for HCO operator to become ready..."
+	@kubectl wait deployment/hyperconverged-cluster-operator -n $(HCO_NAMESPACE) --for=condition=Available --timeout=10m
+	@echo "✅ HCO operator is ready"
+	@echo ""
+	@echo "Waiting for HCO webhook to become ready..."
+	@kubectl wait deployment/hyperconverged-cluster-webhook -n $(HCO_NAMESPACE) --for=condition=Available --timeout=5m
+	@echo "✅ HCO webhook is ready"
+	@echo ""
+	@echo "Waiting for sub-operators to become ready..."
+	@kubectl wait deployment/cdi-operator -n $(HCO_NAMESPACE) --for=condition=Available --timeout=9m
+	@kubectl wait deployment/cluster-network-addons-operator -n $(HCO_NAMESPACE) --for=condition=Available --timeout=9m
+	@kubectl wait deployment/kubevirt-migration-operator -n $(HCO_NAMESPACE) --for=condition=Available --timeout=9m 2>/dev/null || true
+	@echo "✅ Sub-operators are ready"
+	@echo ""
+	@echo "Creating HyperConverged CR..."
+	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/hco.cr.yaml
+	@echo ""
+	@echo "Waiting for HCO to become available (this can take several minutes)..."
+	@kubectl wait hyperconverged kubevirt-hyperconverged -n $(HCO_NAMESPACE) --for condition=Available --timeout=30m
+	@echo "✅ HCO is ready"
+	@echo ""
+	@echo "Waiting for component deployments..."
+	@kubectl wait deployment/cdi-apiserver -n $(HCO_NAMESPACE) --for=condition=Available --timeout=6m
+	@kubectl wait deployment/cdi-deployment -n $(HCO_NAMESPACE) --for=condition=Available --timeout=6m
+	@kubectl wait deployment/cdi-uploadproxy -n $(HCO_NAMESPACE) --for=condition=Available --timeout=6m
+	@kubectl wait deployment/virt-api -n $(HCO_NAMESPACE) --for=condition=Available --timeout=6m
+	@kubectl wait deployment/virt-controller -n $(HCO_NAMESPACE) --for=condition=Available --timeout=6m
+	@echo "✅ All component deployments are ready"
+	@echo ""
+	@echo "========================================="
+	@echo "HCO Installation Complete"
+	@echo "========================================="
+	@echo ""
+	@echo "HCO version: $(HCO_VERSION)"
+	@echo "Namespace: $(HCO_NAMESPACE)"
+	@echo ""
+	@echo "Managed components:"
+	@echo "  - KubeVirt"
+	@echo "  - CDI (Containerized Data Importer)"
+	@echo "  - Cluster Network Addons"
+	@echo ""
+	@echo "Verify installation with:"
+	@echo "  make hco-status"
+	@echo ""
+
+.PHONY: hco-uninstall
+hco-uninstall: ## Uninstall HCO and all managed components from the cluster
+	@echo "Uninstalling HCO..."
+	@kubectl delete hyperconverged kubevirt-hyperconverged -n $(HCO_NAMESPACE) --ignore-not-found --timeout=5m
+	@kubectl delete -n $(HCO_NAMESPACE) -l 'name!=ssp-operator,name!=hyperconverged-cluster-cli-download' -f $(HCO_RAW_URL)/operator.yaml --ignore-not-found 2>/dev/null || true
+	@kubectl delete -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/webhooks.yaml --ignore-not-found 2>/dev/null || true
+	@kubectl delete -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/cluster_role_binding.yaml --ignore-not-found 2>/dev/null || true
+	@kubectl delete -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/service_account.yaml --ignore-not-found 2>/dev/null || true
+	@kubectl delete -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/cluster_role.yaml --ignore-not-found 2>/dev/null || true
+	@for crd in $(HCO_CRDS); do \
+		kubectl delete -f $(HCO_RAW_URL)/crds/$${crd} --ignore-not-found 2>/dev/null || true; \
+	done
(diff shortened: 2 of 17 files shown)


pkg/toolsets/kubevirt/hco_status.go
```diff
@@ -0,0 +1,293 @@
...
+	result.WriteString(fetchComponentCR(ctx, dynamicClient, "NetworkAddonsConfig", kubevirt.NetworkAddonsConfigGVR, ""))
+
+	return result.String()
+}
+
+func fetchComponentCR(ctx context.Context, dynamicClient dynamic.Interface, name string, gvr schema.GroupVersionResource, namespace string) string {
+	var list *unstructured.UnstructuredList
+	var err error
+
+	if namespace != "" {
```
ksimon1: is this check necessary?

evals/tasks/kubevirt/hco-check-component-health/task.yaml
```diff
@@ -0,0 +1,45 @@
...
+          exit 0
+    - llmJudge:
+        contains: "KubeVirt"
+    - llmJudge:
+        contains: "CDI"
+  prompt:
+    inline: |-
+      The HyperConverged Cluster Operator (HCO) manages several components.
+
+      Use the hco-status prompt to get a comprehensive status report, then analyze the managed component health.
```
lyarwood: Prompts shouldn't request the use of the prompts directly, if possible keep the prompts generic as a user would write, for example:

> Describe the status of HCO

pkg/toolsets/kubevirt/hco_status.go
codingben: No it was a mistake. Removed it.

evals/tasks/kubevirt/hco-check-component-health/task.yaml
codingben: Done.

build/hco.mk
```diff
@@ -0,0 +1,133 @@
...
+	@echo ""
+	@echo "Installing HCO webhooks..."
+	@kubectl apply -f $(HCO_RAW_URL)/webhooks.yaml
+	@echo ""
+	@echo "Installing HCO operator (excluding OpenShift-only components)..."
+	@kubectl apply -n $(HCO_NAMESPACE) -l 'name!=ssp-operator,name!=hyperconverged-cluster-cli-download' -f $(HCO_RAW_URL)/operator.yaml
+	@echo ""
+	@echo "Enabling software emulation for Kind (no hardware virtualization)..."
+	@kubectl set env deployment/hyperconverged-cluster-operator -n $(HCO_NAMESPACE) KVM_EMULATION=true
+	@echo ""
```
nunnatsa, line 98: For test environment only. emulation is not for production

build/hco.mk
```diff
@@ -0,0 +1,133 @@
...
+	@echo "========================================="
+	@echo ""
+	@echo "HCO version: $(HCO_VERSION)"
+	@echo "Namespace: $(HCO_NAMESPACE)"
+	@echo ""
+	@echo "Managed components:"
+	@echo "  - KubeVirt"
+	@echo "  - CDI (Containerized Data Importer)"
+	@echo "  - Cluster Network Addons"
+	@echo "  - HostPath Provisioner"
```
nunnatsa: This is installed separatly.

build/hco.mk
```diff
@@ -0,0 +1,133 @@
...
+HCO_RAW_URL = https://raw.githubusercontent.com/kubevirt/hyperconverged-cluster-operator/$(HCO_VERSION)/deploy
+
+HCO_CRDS = \
+	cluster-network-addons00 \
+	containerized-data-importer00 \
+	hco00 \
+	kubevirt00 \
+	hostpath-provisioner00 \
+	scheduling-scale-performance00 \
+	application-aware-quota00
```
nunnatsa: This list must somehow be dynamic. for example, to add the migController CRD

build/hco.mk
```diff
@@ -0,0 +1,133 @@
...
+	@kubectl get hyperconverged -n $(HCO_NAMESPACE) -o wide || echo "HCO not installed"
+	@echo ""
+	@echo "HyperConverged Conditions:"
+	@kubectl get hyperconverged kubevirt-hyperconverged -n $(HCO_NAMESPACE) -o jsonpath='{range .status.conditions[*]}{.type}{"\t"}{.status}{"\t"}{.message}{"\n"}{end}' 2>/dev/null || echo "HCO not installed"
+	@echo ""
+	@echo "KubeVirt (managed by HCO):"
+	@kubectl get kubevirt -A -o wide 2>/dev/null || echo "KubeVirt not found"
+	@echo ""
+	@echo "CDI (managed by HCO):"
+	@kubectl get cdi -A -o wide 2>/dev/null || echo "CDI not found"
```
nunnatsa: no need for `-A` - this is a cluster scoped CR.

build/hco.mk
```diff
@@ -0,0 +1,133 @@
...
+	@kubectl get pods -n $(HCO_NAMESPACE) -l kubevirt.io 2>/dev/null || echo "No KubeVirt pods found"
+	@echo ""
+	@echo "CDI Pods:"
+	@kubectl get pods -n $(HCO_NAMESPACE) -l cdi.kubevirt.io 2>/dev/null || echo "No CDI pods found"
+	@echo ""
+	@echo "Component Versions:"
+	@echo -n "  KubeVirt: " && kubectl get kubevirt -n $(HCO_NAMESPACE) -o jsonpath='{.items[0].status.observedKubeVirtVersion}' 2>/dev/null || echo "N/A"
+	@echo ""
+	@echo -n "  CDI: " && kubectl get cdi -o jsonpath='{.items[0].status.observedVersion}' 2>/dev/null || echo "N/A"
+	@echo ""
```
nunnatsa, line 191: There are so many more resources...

build/hco.mk
```diff
@@ -0,0 +1,133 @@
...
+
+.PHONY: hco-install
+hco-install: ## Install HCO on the cluster (manages KubeVirt, CDI, network addons)
+	@echo "========================================="
+	@echo "Installing HCO $(HCO_VERSION)"
+	@echo "========================================="
+	@echo ""
+	@echo "Creating namespaces..."
+	@kubectl create namespace $(HCO_NAMESPACE) --dry-run=client -o yaml | kubectl apply -f -
+	@kubectl create namespace openshift --dry-run=client -o yaml | kubectl apply -f - 2>/dev/null || true
```
nunnatsa: why do we need this namespace?

evals/tasks/kubevirt/hco-get-feature-gates/task.yaml
```diff
@@ -0,0 +1,62 @@
...
+        inline: |-
+          #!/usr/bin/env bash
+          set -euo pipefail
+          ACTUAL=$(kubectl get hyperconverged kubevirt-hyperconverged -n kubevirt-hyperconverged \
+              -o jsonpath='{.spec.featureGates.withHostPassthroughCPU}' 2>/dev/null)
+          [[ "$ACTUAL" == "true" ]] || { echo "✗ Feature gate 'withHostPassthroughCPU' is '$ACTUAL', expected 'true'"; exit 1; }
+          echo "All validations passed"
+          exit 0
+    - llmJudge:
+        contains: "withHostPassthroughCPU"
```
nunnatsa: Consider moving to v1 API (from HCO v1.18)

evals/tasks/kubevirt/hco-get-feature-gates/task.yaml
```diff
@@ -0,0 +1,62 @@
...
+  cleanup:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          NS="kubevirt-hyperconverged"
+          HCO_NAME="kubevirt-hyperconverged"
+
+          # Revert the feature gate
+          kubectl patch hyperconverged "$HCO_NAME" -n "$NS" --type=merge \
+              -p '{"spec":{"featureGates":{"withHostPassthroughCPU":false}}}' 2>/dev/null || true
```
nunnatsa: This FG is deprecated.

nunnatsa (commented):
Thanks @codingben 

Added a few comments inline. But my main comment is that we're about to release HCO 1.18 in a day or two. This version introduces the v1 API version (in parallel to the existing v1beta1). The new API version is **not backward compatible**. All the examples here will break and won't work.

I would wait a few days to the release of 1.18, and change everything to v1.

codingben:
> Thanks @codingben
> 
> Added a few comments inline. But my main comment is that we're about to release HCO 1.18 in a day or two. This version introduces the v1 API version (in parallel to the existing v1beta1). The new API version is **not backward compatible**. All the examples here will break and won't work.
> 
> I would wait a few days to the release of 1.18, and change everything to v1.

Ack.

build/hco.mk
codingben: Right, this setup will be never used in production (Kind cluster is a local setup).

build/hco.mk
codingben: Okay removed it.

build/hco.mk
codingben: Okay, let's fetch the CRD list directly from the repo itself.

build/hco.mk
codingben: Fixed.

build/hco.mk
codingben: I've adjusted it.

build/hco.mk
codingben: It was added by a mistake, let me remove it.

evals/tasks/kubevirt/hco-get-feature-gates/task.yaml
codingben: I've replaced it.

codingben:
Can you please run MCP checker? I have locally issues with setting up Kind cluster.

/run mcp-checker

nunnatsa:
Can we use, or mimic the HCO's hack/deploy.sh script?

For example, HCO requires cert-manager. This is not part of the installation script in the PR.

And again, please move to v1 API version.

build/hco.mk
codingben: @nunnatsa I saw that in your official script you're creating this [namespace](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/hack/deploy.sh#L70-L74). Why?

codingben:
> Can we use, or mimic the HCO's hack/deploy.sh script?
> 
> For example, HCO requires cert-manager. This is not part of the installation script in the PR.
> 
> And again, please move to v1 API version.

Use no, because it relies on other internal scripts. We can make the flow similar to that official script. I've added the requirement for this.

codingben:
@nunnatsa When do you plan to release v1.18? I don't want get this PR blocked more than a few days because of it, unless there is a specific reason for it. I make a follow-up PR to upgrade the version to v1.18 later.

nunnatsa:
> @nunnatsa When do you plan to release v1.18? I don't want get this PR blocked more than a few days because of it, unless there is a specific reason for it. I make a follow-up PR to upgrade the version to v1.18 later.

The version is actually released

We still have an issue with the catalaog. Hopefully it will be solved today or early next week. But v1.18.0 is usable and can be peeked already.

pkg/mcp/kubevirt_test.go
```diff
@@ -1043,6 +1047,104 @@ func (s *KubevirtSuite) TestVMGuestInfo() {
...
+		ctx := s.T().Context()
+
+		// Create a HyperConverged CR
+		hcoCR := &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "hco.kubevirt.io/v1",
+				"kind":       "HyperConverged",
+				"metadata": map[string]interface{}{
+					"name":      "kubevirt-hyperconverged",
+					"namespace": "default",
```
nunnatsa: This is not supported. the namespace must be `kubevirt-hyperconverged`.

pkg/mcp/kubevirt_test.go
```diff
@@ -1043,6 +1047,104 @@ func (s *KubevirtSuite) TestVMGuestInfo() {
...
+		})
+	})
+
+	s.Run("hco-status prompt returns report with HCO CR present", func() {
+		dynamicClient := dynamic.NewForConfigOrDie(envTestRestConfig)
+		ctx := s.T().Context()
+
+		// Create a HyperConverged CR
+		hcoCR := &unstructured.Unstructured{
+			Object: map[string]interface{}{
```
nunnatsa: Please replace all the instances of `interface{}` with `any`.

pkg/mcp/kubevirt_test.go
```diff
@@ -1043,6 +1047,104 @@ func (s *KubevirtSuite) TestVMGuestInfo() {
...
+				"apiVersion": "hco.kubevirt.io/v1",
+				"kind":       "HyperConverged",
+				"metadata": map[string]interface{}{
+					"name":      "kubevirt-hyperconverged",
+					"namespace": "default",
+				},
+				"spec": map[string]interface{}{
+					"featureGates": map[string]interface{}{
+						"downwardMetrics": false,
+					},
```
nunnatsa: In the `v1` API version, the structure of the feature gate field was changed.

It should be:


```suggestion
					"featureGates": []any{
						map[string]string{
							"name": "downwardMetrics",
							"state": "Disabled",
						},
					},
```

pkg/mcp/kubevirt_test.go
```diff
@@ -1043,6 +1047,104 @@ func (s *KubevirtSuite) TestVMGuestInfo() {
...
+					"versions": []interface{}{
+						map[string]interface{}{
+							"name":    "operator",
+							"version": "1.17.1",
+						},
+						map[string]interface{}{
+							"name":    "kubevirt",
+							"version": "1.5.2",
+						},
+					},
```
nunnatsa: Please drop the status field. In K8s, users does not set the status field, but only the controller (HCO in this case)

pkg/mcp/kubevirt_test.go
```diff
@@ -1043,6 +1047,104 @@ func (s *KubevirtSuite) TestVMGuestInfo() {
...
+				"spec": map[string]interface{}{
+					"featureGates": map[string]interface{}{
+						"downwardMetrics": false,
+					},
+					"liveMigrationConfig": map[string]interface{}{
+						"completionTimeoutPerGiB":           int64(150),
+						"parallelMigrationsPerCluster":      int64(5),
+						"parallelOutboundMigrationsPerNode": int64(2),
+						"progressTimeout":                   int64(150),
+					},
```
nunnatsa: All the structure of HCO API was changed. See here for details: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/4bd6652450567f5f23d136030ca8b93787ca25c9/docs/enhancements/api-version-v1.md

Also this can help: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/e357bf7cfa06dfdd1d213259f7a52a4eac0ab9f7/docs/cluster-configuration-v1-api.md

pkg/toolsets/kubevirt/hco_status.go
```diff
@@ -0,0 +1,285 @@
...
+	} else {
+		yamlStr, err := output.MarshalYaml(featureGates)
+		if err != nil {
+			fmt.Fprintf(&result, "### Feature Gates\n\n*Error marshaling feature gates: %v*\n\n", err)
+		} else {
+			fmt.Fprintf(&result, "### Feature Gates\n\n```yaml\n%s```\n\n", yamlStr)
+		}
+	}
+
+	liveMigration, found, err := unstructured.NestedMap(hco.Object, "spec", "liveMigrationConfig")
```
nunnatsa: This should be updated to the v1 format

nunnatsa (commented):
Added some inline comments

nunnatsa:
> > @nunnatsa When do you plan to release v1.18? I don't want get this PR blocked more than a few days because of it, unless there is a specific reason for it. I make a follow-up PR to upgrade the version to v1.18 later.
> 
> The version is actually released
> 
> We still have an issue with the catalaog. Hopefully it will be solved today or early next week. But v1.18.0 is usable and can be peeked already.

Released v1.18.1 yesterday. Now it's fully released, and included in the community catalog.

pkg/mcp/kubevirt_test.go
codingben: Fixed.

pkg/mcp/kubevirt_test.go
codingben: Fixed.

pkg/mcp/kubevirt_test.go
codingben: Right.

pkg/mcp/kubevirt_test.go
codingben: Dropped.

pkg/mcp/kubevirt_test.go
codingben: Thanks for pointing to this. I think i've addressed this now.

pkg/toolsets/kubevirt/hco_status.go
codingben: Done.

codingben:
> > > @nunnatsa When do you plan to release v1.18? I don't want get this PR blocked more than a few days because of it, unless there is a specific reason for it. I make a follow-up PR to upgrade the version to v1.18 later.
> > 
> > 
> > The version is actually released
> > We still have an issue with the catalaog. Hopefully it will be solved today or early next week. But v1.18.0 is usable and can be peeked already.
> 
> Released v1.18.1 yesterday. Now it's fully released, and included in the community catalog.

I've updated it to that version.

ksimon1:
@codingben unit tests are failing, can you please check it?

pkg/mcp/common_test.go
```diff
@@ -80,6 +80,11 @@ func TestMain(m *testing.M) {
 			CRD("instancetype.kubevirt.io", "v1beta1", "virtualmachineinstancetypes", "VirtualMachineInstancetype", "virtualmachineinstancetype", true),
 			CRD("instancetype.kubevirt.io", "v1beta1", "virtualmachineclusterpreferences", "VirtualMachineClusterPreference", "virtualmachineclusterpreference", false),
 			CRD("instancetype.kubevirt.io", "v1beta1", "virtualmachinepreferences", "VirtualMachinePreference", "virtualmachinepreference", true),
+			// HCO
+			CRD("hco.kubevirt.io", "v1beta1", "hyperconvergeds", "HyperConverged", "hyperconverged", true),
```
ksimon1: Here it should be v1 too no?

lyarwood: +1 to @ksimon1's comment — this registers the HCO CRD as `v1beta1` but `HyperConvergedGVR` in `pkg/kubevirt/gvr.go` uses `v1`. The dynamic client will try to list `v1` resources against a `v1beta1` CRD, which will fail. This is likely the root cause of the test failures @ksimon1 flagged.

Should be:
```go
CRD("hco.kubevirt.io", "v1", "hyperconvergeds", "HyperConverged", "hyperconverged", true),
```

evals/tasks/kubevirt/hco-status-prompt/task.yaml
```diff
@@ -0,0 +1,45 @@
...
+          exit 0
+    - llmJudge:
+        contains: "Available"
+    - llmJudge:
+        contains: "KubeVirt"
+    - llmJudge:
+        contains: "CDI"
+  prompt:
+    inline: |-
+      Use the hco-status prompt to generate a status report for the HyperConverged Cluster Operator.
```
lyarwood: Same anti-pattern discussed on PR #1239 — the prompt explicitly names the tool/prompt (`Use the hco-status prompt`). This tests instruction-following rather than whether the agent can discover and select the right prompt on its own.

Consider rephrasing to present the problem without naming the prompt, e.g.:

```
Generate a comprehensive status report for the HyperConverged Cluster Operator.
Summarize:
- The overall health of HCO (conditions)
- The status of each managed component (KubeVirt, CDI, Network Addons)
- Any warnings or non-default configuration
```

evals/tasks/kubevirt/hco-check-version/task.yaml
```diff
@@ -0,0 +1,43 @@
...
+          kubectl get hyperconverged kubevirt-hyperconverged -n kubevirt-hyperconverged > /dev/null 2>&1 || { echo "✗ HyperConverged CR not found"; exit 1; }
+          OPERATOR_VERSION=$(kubectl get hyperconverged kubevirt-hyperconverged -n kubevirt-hyperconverged \
+              -o jsonpath='{.status.versions[?(@.name=="operator")].version}' 2>/dev/null)
+          [[ -n "$OPERATOR_VERSION" ]] || { echo "✗ Could not determine HCO operator version"; exit 1; }
+          echo "✓ HCO operator version: $OPERATOR_VERSION"
+          KUBEVIRT_VERSION=$(kubectl get hyperconverged kubevirt-hyperconverged -n kubevirt-hyperconverged \
+              -o jsonpath='{.status.versions[?(@.name=="kubevirt")].version}' 2>/dev/null)
+          [[ -n "$KUBEVIRT_VERSION" ]] || { echo "✗ Could not determine KubeVirt version"; exit 1; }
+          echo "✓ KubeVirt version: $KUBEVIRT_VERSION"
+          echo "All validations passed"
```
lyarwood, line 49: nit: `contains: "version"` is very weak — the word "version" appears in almost any response about software. Consider checking for the actual version string or a more specific term like `contains: "operator"` to verify the agent found the HCO operator version specifically.

lyarwood (changes requested):
## Review: feat(kubevirt): add HCO awareness using MCP tooling

### Summary

Adds a `hco-status` MCP Prompt that auto-discovers the HyperConverged CR and generates a status report covering conditions, component health, configuration (feature gates, virtualization settings), and events. Includes `build/hco.mk` for deploying HCO on Kind clusters and 8 eval tasks exercising HCO workflows.

### Positive Observations

- **Auto-discovery** via `discoverHyperConvergedCR` listing across all namespaces is the right approach — no hardcoded namespace.
- **Well-structured report** — conditions, versions, configuration, components, and events are logically separated.
- **Eval coverage** is comprehensive — 8 tasks covering easy/medium/hard scenarios including a TaintedConfiguration degradation test.
- **HCO v1.18.1 upgrade** was completed as requested by @nunnatsa.
- **Integration test** in `pkg/mcp/kubevirt_test.go` covers both missing-HCO and present-HCO scenarios.

### Critical Issues

1. **CRD version mismatch** (pkg/mcp/common_test.go:84) — The HCO CRD is registered as `v1beta1` in the test environment but `HyperConvergedGVR` in gvr.go uses `v1`. This was already flagged by @ksimon1 and likely causes the test failures.

### Suggestions

See inline comments.

### Existing Discussion Summary

- @nunnatsa's feedback on v1 API migration — addressed (v1.18.1 adopted)
- @nunnatsa's feedback on feature gate structure, status field, namespace — addressed in latest push
- @ksimon1 flagged CRD version mismatch (common_test.go:84) — **unresolved, likely causing test failures**
- @lyarwood's feedback on eval prompt naming — addressed (prompts no longer reference specific tools/prompts, except hco-status-prompt task)
- Various build/hco.mk comments from @nunnatsa — addressed (dynamic CRD list, emulation note, namespace cleanup)

### Commit History

Single squashed commit with clear conventional commit message. Good.

### Verdict

**Request Changes** — The CRD version mismatch needs fixing (tests are failing per @ksimon1's comment). The hco-status-prompt eval explicitly names the prompt, which is the same anti-pattern discussed on PR #1239. Other suggestions are minor.

build/hco.mk
```diff
@@ -0,0 +1,184 @@
...
+		kubectl get crd $(HCO_CRD_NAME) -o go-template='{{ range .status.conditions }}{{ .type }}{{ "\t" }}{{ .status }}{{ "\t" }}{{ .message }}{{ "\n" }}{{ end }}'; \
+	fi
+	@echo ""
+	@echo "Installing HCO RBAC..."
+	@kubectl apply -f $(HCO_RAW_URL)/cluster_role.yaml
+	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/service_account.yaml
+	@kubectl apply -f $(HCO_RAW_URL)/cluster_role_binding.yaml
+	@echo ""
+	@echo "Installing HCO webhooks..."
+	@kubectl apply -f $(HCO_RAW_URL)/webhooks.yaml
```
lyarwood: **Bug: webhooks applied without namespace — certificates and secrets land in `default` instead of `kubevirt-hyperconverged`**

The `webhooks.yaml` manifest contains namespaced resources (Issuer, Certificate) but this `kubectl apply` doesn't include `-n $(HCO_NAMESPACE)`. As a result, cert-manager creates the TLS secrets in the `default` namespace, while the webhook pods in `kubevirt-hyperconverged` can't find them — causing the webhook deployment to hang in `ContainerCreating` with:

```
MountVolume.SetUp failed for volume "apiservice-cert" : secret "hyperconverged-cluster-webhook-service-cert" not found
```

Fix:
```makefile
	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/webhooks.yaml
```

Note: the ValidatingWebhookConfigurations and MutatingWebhookConfigurations in the same file are cluster-scoped and will be applied correctly regardless of `-n`.

build/hco.mk
```diff
@@ -0,0 +1,184 @@
...
+# HCO deploys and manages KubeVirt, CDI, Cluster Network Addons, and other
+# components as a single operator. This is an ALTERNATIVE to the standalone
+# KubeVirt install (kubevirt-install) — do not use both together.
+
+# HCO version configuration
+HCO_VERSION ?= v1.18.1
+HCO_NAMESPACE = kubevirt-hyperconverged
+HCO_RAW_URL = https://raw.githubusercontent.com/kubevirt/hyperconverged-cluster-operator/$(HCO_VERSION)/deploy
+HCO_API_URL = https://api.github.com/repos/kubevirt/hyperconverged-cluster-operator/contents/deploy/crds?ref=$(HCO_VERSION)
+HCO_CRDS = $(shell curl -sL $(HCO_API_URL) | grep -o '"name": *"[^"]*\.crd\.yaml"' | sed 's/"name": *"//;s/"//')
```
ksimon1: `HCO_CRDS` uses recursive expansion (`=`) which re-executes the `curl` API call every time the variable is referenced in a recipe. Both the install and uninstall targets iterate over it, so the network call fires multiple times. Use simply-expanded assignment instead:

```makefile
HCO_CRDS := $(shell curl -sL $(HCO_API_URL) | grep -o '"name": *"[^"]*\.crd\.yaml"' | sed 's/"name": *"//;s/"//')
```

Generated by AI Claude Sonnet 4.6

build/hco.mk
```diff
@@ -0,0 +1,184 @@
...
+	@echo "Creating namespace..."
+	@kubectl create namespace $(HCO_NAMESPACE) --dry-run=client -o yaml | kubectl apply -f -
+	@echo ""
+	@echo "Installing HCO CRDs..."
+	@for crd in $(HCO_CRDS); do \
+		kubectl apply --server-side -f $(HCO_RAW_URL)/crds/$${crd}; \
+	done
+	@echo ""
+	@echo "Checking CRD status..."
+	@sleep 20
```
ksimon1: Fixed `sleep 20` is fragile — if CRDs establish faster or slower than expected, this either wastes time or races ahead. Prefer polling with `kubectl wait` on each CRD:

```makefile
@for crd in $(HCO_CRDS); do \
    kubectl wait --for=condition=Established crd/$$(basename $${crd} .crd.yaml) --timeout=60s 2>/dev/null || true; \
done
```

Generated by AI Claude Sonnet 4.6

pkg/mcp/common_test.go
codingben: Ah good catch, resolved just now

codingben: Thanks Lee, i've addressed it.

evals/tasks/kubevirt/hco-status-prompt/task.yaml
codingben: Done.

evals/tasks/kubevirt/hco-check-version/task.yaml
codingben: Done.

build/hco.mk
codingben: Thanks Lee. Fixed it.

build/hco.mk
codingben: Done.

build/hco.mk
codingben: Done.

build/hco.mk
```diff
@@ -0,0 +1,186 @@
...
+	@echo "  - KubeVirt"
+	@echo "  - CDI (Containerized Data Importer)"
+	@echo "  - Cluster Network Addons"
+	@echo ""
+	@echo "Verify installation with:"
+	@echo "  make hco-status"
+	@echo ""
+
+.PHONY: hco-uninstall
+hco-uninstall: ## Uninstall HCO and all managed components from the cluster
```
ksimon1, line 145: we should call here `cert-manager-uninstall` too

codingben: No. The `cert-manager` should be already installed by `kind-create-cluster` and be deleted by `kind-delete-cluster` if needed. The `hco-install` should skip the installation of it.

codingben: @lyarwood I think this is what caused the issue during your installation locally, `hco-install` was overrding the installation of the `cert-manager` (certificates, etc) causing other errors.

build/hco.mk
```diff
@@ -0,0 +1,191 @@
...
+	@for crd in $(HCO_CRDS); do \
+		kubectl wait --for=condition=Established crd/$$(basename $${crd} .crd.yaml) --timeout=60s 2>/dev/null || true; \
+	done
+	@if [ "$$(kubectl get crd $(HCO_CRD_NAME) -o=jsonpath='{.status.conditions[?(@.type=="NonStructuralSchema")].status}' 2>/dev/null)" = "True" ]; then \
+		echo "Warning: HCO CRD reports NonStructuralSchema condition"; \
+		kubectl get crd $(HCO_CRD_NAME) -o go-template='{{ range .status.conditions }}{{ .type }}{{ "\t" }}{{ .status }}{{ "\t" }}{{ .message }}{{ "\n" }}{{ end }}'; \
+	fi
+	@echo ""
+	@echo "Installing HCO RBAC..."
+	@kubectl apply -f $(HCO_RAW_URL)/cluster_role.yaml
```
lyarwood: **Bug: same namespace issue as the webhooks — `cluster_role.yaml` contains namespaced Roles that land in `default`**

`cluster_role.yaml` contains both ClusterRoles (cluster-scoped) and Roles (namespace-scoped). Without `-n $(HCO_NAMESPACE)`, the Roles are created in `default` instead of `kubevirt-hyperconverged`. This causes the `cluster-network-addons-operator` to fail with RBAC errors:

```
RBAC: role.rbac.authorization.k8s.io "cluster-network-addons-operator" not found
```

Which cascades into HCO never reaching `Available=True`.

Fix:
```makefile
	@kubectl apply -n $(HCO_NAMESPACE) -f $(HCO_RAW_URL)/cluster_role.yaml
```

This is safe — `-n` only affects namespaced resources (Roles, RoleBindings); the ClusterRoles in the same file are cluster-scoped and ignore `-n`.

codingben: Fixed. Thanks Lee.

Makefile
```diff
@@ -144,6 +144,29 @@ local-env-setup-kubevirt: ## Setup complete local development environment with K
 	@echo "KubeVirt is now available!"
 	@echo "Check status with: make kubevirt-status"
 
+.PHONY: local-env-setup-hco
+local-env-setup-hco: ## Setup complete local development environment with Kind cluster and HCO
+	@echo "========================================="
+	@echo "Kubernetes MCP Server - Local Setup"
+	@echo "            with HCO"
+	@echo "========================================="
+	$(MAKE) kind-create-cluster
```
codingben, line 153: This `kind-create-cluster` doesn't work, it's failing on cert-manager step. I'm not sure if it's local issue with Podman or Kind, or with the my laptop's operating system (Fedora 44).

@lyarwood @ksimon1 @nunnatsa Can you please try to run `make kind-create-cluster` locally to see if you're getting any errors/timeouts from the cert-manager?

ksimon1:
@codingben 
```
[ksimon:09:03:03~/go/src/github.com/containers/kubernetes-mcp-server]$ make kind-create-cluster
Installing kind to /home/ksimon/go/src/github.com/containers/kubernetes-mcp-server/_output/tools/bin/kind...
go: downloading sigs.k8s.io/kind v0.30.0
go: downloading al.essio.dev/pkg/shellescape v1.5.1
go: downloading github.com/BurntSushi/toml v1.4.0
go: downloading golang.org/x/sys v0.6.0
Creating placeholder CA certificate for bind mount...
make: *** [build/kind.mk:23: kind-create-certs] error 127

```

Makefile
codingben: I had cache'd issue locally. Running `podman system prune --all --volumes --force` resolved this issue.

codingben:
> @codingben
> 
> ```
> [ksimon:09:03:03~/go/src/github.com/containers/kubernetes-mcp-server]$ make kind-create-cluster
> Installing kind to /home/ksimon/go/src/github.com/containers/kubernetes-mcp-server/_output/tools/bin/kind...
> go: downloading sigs.k8s.io/kind v0.30.0
> go: downloading al.essio.dev/pkg/shellescape v1.5.1
> go: downloading github.com/BurntSushi/toml v1.4.0
> go: downloading golang.org/x/sys v0.6.0
> Creating placeholder CA certificate for bind mount...
> make: *** [build/kind.mk:23: kind-create-certs] error 127
> ```

Thanks @ksimon1, it works on my end now. I had to run `podman system prune --all --volumes --force` to cleanup the environment locally.

codingben:
It's now stuck on waiting for HCO to become available after creating the CR. @lyarwood The cert-manager and whatever certificates error you had should be no more an issue.

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
```diff
@@ -0,0 +1,65 @@
...
+              echo "HCO is not deployed — skipping eval"
+              exit 1
+          fi
+
+          # Add a pause annotation to the HCO CR to trigger an unsafe modification condition.
+          # This annotation tells HCO that an admin is making direct changes to managed CRs,
+          # which causes HCO to report a TaintedConfiguration condition.
+          echo "Adding jsonpatch annotation to trigger TaintedConfiguration..."
+          kubectl annotate hyperconverged "$HCO_NAME" -n "$NS" \
+              hco.kubevirt.io/jsonpatch='[{"op":"add","path":"/spec/configuration/developerConfiguration","value":{"logVerbosity":{"virtLauncher":5}}}]' \
```
nunnatsa: Please avoid using the json patch annotation. This is only for emergency. HCO supports an API for the log verbosity in KubeVirt and CDI: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/docs/cluster-configuration.md#log-verbosity

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
```diff
@@ -0,0 +1,65 @@
...
+  cleanup:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          NS="kubevirt-hyperconverged"
+          HCO_NAME="kubevirt-hyperconverged"
+
+          # Remove the jsonpatch annotation
+          kubectl annotate hyperconverged "$HCO_NAME" -n "$NS" \
+              hco.kubevirt.io/jsonpatch- 2>/dev/null || true
```
nunnatsa: same

evals/tasks/kubevirt/hco-get-feature-gates/task.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+          if ! kubectl get hyperconverged "$HCO_NAME" -n "$NS" > /dev/null 2>&1; then
+              echo "HCO is not deployed — skipping eval"
+              exit 1
+          fi
+
+          # Enable a feature gate so the agent has something specific to find
+          echo "Enabling deployKubeSecondaryDNS feature gate..."
+          kubectl patch hyperconverged "$HCO_NAME" -n "$NS" --type=json \
+              -p '[{"op":"add","path":"/spec/featureGates/-","value":{"name":"deployKubeSecondaryDNS","state":"Enabled"}}]'
```
nunnatsa: This patch will fail if /spec/featureGates is nil.

in this case, you'll need something like:
```
'[{"op":"add","path":"/spec/featureGates","value":[{"name":"deployKubeSecondaryDNS","state":"Enabled"}]}]'
```

Also, the `state` is optional if it need to be enabled. It's ok to do: `{"name": "deployKubeSecondaryDNS"}`

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
codingben: Done. Thanks!

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
codingben: Done.

evals/tasks/kubevirt/hco-get-feature-gates/task.yaml
codingben: Interesting. I've addressed it now.

evals/tasks/kubevirt/hco-get-feature-gates/task.yaml
```diff
@@ -0,0 +1,64 @@
...
+              echo "HCO is not deployed — skipping eval"
+              exit 1
+          fi
+
+          # Enable a feature gate so the agent has something specific to find.
+          # Use merge patch to handle the case where featureGates is nil.
+          # The state field is omitted since Enabled is the default.
+          echo "Enabling deployKubeSecondaryDNS feature gate..."
+          kubectl patch hyperconverged "$HCO_NAME" -n "$NS" --type=merge \
+              -p '{"spec":{"featureGates":[{"name":"deployKubeSecondaryDNS"}]}}'
```
nunnatsa: I think this will override any existing feature gate. Please check that

codingben: It should be okay now. It's adding to the array and won't replace it.

codingben:
@lyarwood @nunnatsa I ran yesterday `podman system prune --all --volumes --force` and HCO webhook installation worked, but now it fails again:
```
Installing HCO webhooks...
validatingwebhookconfiguration.admissionregistration.k8s.io/validate-hco.kubevirt.io unchanged
validatingwebhookconfiguration.admissionregistration.k8s.io/nodemaintenance-validation.kubevirt.io unchanged
validatingwebhookconfiguration.admissionregistration.k8s.io/vssp.kb.io configured
validatingwebhookconfiguration.admissionregistration.k8s.io/hostpathprovisioner.kubevirt.io unchanged
mutatingwebhookconfiguration.admissionregistration.k8s.io/mutate-hco.kubevirt.io configured
Error from server (InternalError): error when creating "https://raw.githubusercontent.com/kubevirt/hyperconverged-cluster-operator/v1.18.1/deploy/webhooks.yaml": Internal error occurred: failed calling webhook "webhook.cert-manager.io": failed to call webhook: Post "https://cert-manager-webhook.cert-manager.svc:443/validate?timeout=30s": dial tcp 10.96.91.234:443: connect: connection refused
```
Do you have idea how to debug it? No AI tool can help here even to analyze the proejct's code or logs.

evals/tasks/kubevirt/hco-get-feature-gates/task.yaml
```diff
@@ -0,0 +1,76 @@
...
+              echo "HCO is not deployed — skipping eval"
+              exit 1
+          fi
+
+          # Enable a feature gate so the agent has something specific to find.
+          # Append to the existing array so we don't clobber pre-existing gates.
+          # JSON patch "add" with "/-" appends; if the array is nil it fails,
+          # so fall back to merge patch to create it.
+          echo "Enabling deployKubeSecondaryDNS feature gate..."
+          if ! kubectl patch hyperconverged "$HCO_NAME" -n "$NS" --type=json \
```
nunnatsa: The patch may be failed for any reason. WDYT about something like this?

```shell
if [[ $(kubectl get hyperconverged "$HCO_NAME" -n "$NS" -o jsonpath='{.spec.featureGates}') != "" ]]; then
    kubectl patch hyperconverged "$HCO_NAME" -n "$NS" --type=json \
    -p '[{"op":"add","path":"/spec/featureGates/-","value":{"name":"deployKubeSecondaryDNS"}}]' 2>/dev/null
else
    # featureGates is nil — create the array with our entry
    kubectl patch hyperconverged "$HCO_NAME" -n "$NS" --type=json \
    -p '[{"op":"add","path":"/spec/featureGates","value":[{"name":"deployKubeSecondaryDNS"}]}]'
fi
```

codingben: Looks good.

codingben:
@nunnatsa Can you please review this PR?

codingben:
@lyarwood It seems like the failure of the webhook is a local Kind/Podman issue, in the [CI job](https://github.com/containers/kubernetes-mcp-server/actions/runs/29914289761/job/88904369166?pr=1220) of this PR it works, but we're not testing the HCO installation in the CI, and I'm not able to test it locally with this environment, can we add the HCO installation to the PR's CI?

codingben:
@lyarwood @ksimon1 @nunnatsa Can you please review this PR?

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
```diff
@@ -0,0 +1,63 @@
...
+          fi
+
+          # Set an elevated log verbosity for virtLauncher using the supported HCO API
+          # (spec.deployment.logVerbosityConfig) instead of the emergency-only jsonpatch annotation.
+          echo "Setting elevated virtLauncher log verbosity via HCO API..."
+          kubectl patch hyperconverged "$HCO_NAME" -n "$NS" --type merge -p \
+              '{"spec":{"deployment":{"logVerbosityConfig":{"kubevirt":{"virtLauncher":5}}}}}'
+
+          echo "Waiting for HCO to reconcile..."
+          kubectl wait hyperconverged "$HCO_NAME" -n "$NS" --for condition=Available --timeout=60s
```
ksimon1: This test should test degraded state of HCO, but instead this is setting valid configuration and waiting for condition available. From this test I would expect causing real degraded state on HCO. @nunnatsa please correct me if I am wrong

codingben: I've improved it, can you check again?

nunnatsa: Not sure it's so easy to make thing deprecated from HCO. I would say it will be a bug in HCO if we'll have something like this. Maybe using a jsonpatch?

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
```diff
@@ -0,0 +1,78 @@
...
+          HCO_NAME="kubevirt-hyperconverged"
+
+          if ! kubectl get hyperconverged "$HCO_NAME" -n "$NS" > /dev/null 2>&1; then
+              echo "HCO is not deployed — skipping eval"
+              exit 1
+          fi
+
+          echo "Injecting bad infra nodePlacement to trigger degraded state..."
+          kubectl patch hyperconverged "$HCO_NAME" -n "$NS" --type merge -p \
+              '{"spec":{"infra":{"nodePlacement":{"nodeSelector":{"non-existent-node/test":"true"}}}}}'
```
nunnatsa: This patch will fail. It's trying to patch HyperConverged with v1beta1 format, while the HC is in v1 format.

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
```diff
@@ -0,0 +1,78 @@
...
+        contains: "non-existent-node"
+  cleanup:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          NS="kubevirt-hyperconverged"
+          HCO_NAME="kubevirt-hyperconverged"
+
+          kubectl patch hyperconverged "$HCO_NAME" -n "$NS" --type merge \
+              -p '{"spec":{"infra":{"nodePlacement":null}}}' 2>/dev/null || true
```
nunnatsa: same

nunnatsa (commented):
The latest change won't work. details in the inline comments.

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
nunnatsa: This is the right format:
```shell
kubectl patch hco kubevirt-hyperconverged --type merge \
    -p '{"spec":{"deployment":{"nodePlacement":{"infra":{"nodeSelector":{"non-existent-node/test":"true"}}}}}}'
```

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
nunnatsa: This is the right format:
```shell
kubectl patch hco kubevirt-hyperconverged --type merge \
    -p '{"spec":{"deployment":{"nodePlacements":{"infra":null}}}}'
```

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
codingben: Thanks. Resolved this issue.

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
codingben: Cool thanks. Fixed.

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
codingben: @nunnatsa I've applied your suggestions below, Thanks for looking on this.

codingben:
@nunnatsa Can you please review again?

codingben:
@nunnatsa Thanks. @lyarwood @ksimon1 @manusa @Cali0707 Can you please approve if all good?

pkg/toolsets/kubevirt/hco_status.go
```diff
@@ -0,0 +1,285 @@
...
+					Type: "text",
+					Text: "I'll analyze the HyperConverged Cluster Operator status report to assess the overall health and configuration.",
+				},
+			},
+		},
+		nil,
+	), nil
+}
+
+func discoverHyperConvergedCR(ctx context.Context, dynamicClient dynamic.Interface) (*unstructured.Unstructured, string, error) {
```
lyarwood, line 130: Tested this against a live OpenShift 4.22 cluster with OpenShift Virtualization deployed in `openshift-cnv`. Cross-namespace discovery (`Namespace("").List`) worked smoothly out-of-the-box!

evals/tasks/kubevirt/hco-status-prompt/task.yaml
```diff
@@ -0,0 +1,45 @@
...
+  description: "Generate a full HCO status report and interpret the results"
+spec:
+  requires:
+    - extension: kubernetes
+      as: k8s
+  setup:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          if ! kubectl get hyperconverged kubevirt-hyperconverged -n kubevirt-hyperconverged > /dev/null 2>&1; then
```
lyarwood: To allow this and the other eval tasks to run seamlessly against both Kind (where HCO defaults to `kubevirt-hyperconverged`) and OpenShift (where HCO defaults to `openshift-cnv`), consider dynamically discovering the HCO namespace in the `setup` and `verify` scripts:

```bash
NS=$(kubectl get hyperconverged -A -o jsonpath='{.items[0].metadata.namespace}' 2>/dev/null || echo "kubevirt-hyperconverged")
HCO_NAME=$(kubectl get hyperconverged -A -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || echo "kubevirt-hyperconverged")
```

evals/tasks/kubevirt/hco-check-version/task.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          set -euo pipefail
+          kubectl get hyperconverged kubevirt-hyperconverged -n kubevirt-hyperconverged > /dev/null 2>&1 || { echo "✗ HyperConverged CR not found"; exit 1; }
+          OPERATOR_VERSION=$(kubectl get hyperconverged kubevirt-hyperconverged -n kubevirt-hyperconverged \
+              -o jsonpath='{.status.versions[?(@.name=="operator")].version}' 2>/dev/null)
+          [[ -n "$OPERATOR_VERSION" ]] || { echo "✗ Could not determine HCO operator version"; exit 1; }
+          echo "✓ HCO operator version: $OPERATOR_VERSION"
+          KUBEVIRT_VERSION=$(kubectl get hyperconverged kubevirt-hyperconverged -n kubevirt-hyperconverged \
```
lyarwood: On OpenShift / HCO v1, `HyperConverged.status.versions` only contains `[{"name":"operator","version":"..."}]` (the KubeVirt engine version is reported in `KubeVirt.status` instead). Querying `.status.versions[?(@.name=="kubevirt")]` on the `HyperConverged` CR fails when tested against live clusters. We should check the operator version from HCO status and/or query the KubeVirt CR directly for the component version.

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
```diff
@@ -0,0 +1,78 @@
...
+          HCO_NAME="kubevirt-hyperconverged"
+
+          if ! kubectl get hyperconverged "$HCO_NAME" -n "$NS" > /dev/null 2>&1; then
+              echo "HCO is not deployed — skipping eval"
+              exit 1
+          fi
+
+          echo "Injecting bad infra nodePlacement to trigger degraded state..."
+          kubectl patch hco "$HCO_NAME" -n "$NS" --type merge \
+              -p '{"spec":{"deployment":{"nodePlacement":{"infra":{"nodeSelector":{"non-existent-node/test":"true"}}}}}}'
```
lyarwood: Small schema fix here: the field in `HyperConvergedSpec.Deployment` is `nodePlacements` (plural with an `s`), not `nodePlacement`. Because Kubernetes merge-patch silently drops unknown fields, `spec.deployment.nodePlacement` was ignored and didn't trigger the expected scheduling failure until changed to `spec.deployment.nodePlacements`.

lyarwood (commented):
Tested this PR against a live OpenShift 4.22 nightly cluster with OpenShift Virtualization (`kubevirt-hyperconverged-operator.v4.22.6`) installed in `openshift-cnv`.

### Testing & Eval Results
- Invoked `hco-status` prompt over MCP directly: generated a clean, accurate, and comprehensive diagnostic report covering all sub-components (KubeVirt, CDI, CNAO, SSP) and active configuration.
- Executed all 8 eval tasks with `mcpchecker check`: **8/8 tasks passed (100% success rate, 24/24 assertions passed)**.

Left a few minor inline comments regarding eval task portability (Kind vs OpenShift namespace discovery, `HyperConverged.status.versions` on OpenShift, and the `nodePlacements` plural schema name in `hco-diagnose-degraded`).

The core MCP prompt and Go implementation look great!

evals/tasks/kubevirt/hco-status-prompt/task.yaml
codingben: Ah good catch. I've addressed it.

evals/tasks/kubevirt/hco-check-version/task.yaml
codingben: Done.

evals/tasks/kubevirt/hco-diagnose-degraded/task.yaml
codingben: Done.

lyarwood (approved):
Looks good to me now, thanks @codingben!

pkg/toolsets/kubevirt/hco_status.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	cr := &list.Items[0]
+
+	conditions, found, err := unstructured.NestedSlice(cr.Object, "status", "conditions")
+	if err != nil || !found {
+		status, sFound, sErr := unstructured.NestedMap(cr.Object, "status")
+		if sErr != nil || !sFound {
+			return fmt.Sprintf("### %s: %s/%s\n\n*No status available*", name, cr.GetNamespace(), cr.GetName())
+		}
```
Cali0707: This error handling is different from the similar call on lines 146-159, is this intentional for some reason?

This one seems to swallow errors the other call doesn't, and it is not clear why the diff

pkg/toolsets/kubevirt/hco_status.go
```diff
@@ -0,0 +1,285 @@
...
+%s
+
+---
+
+## Analysis
+
+Based on the data above, determine:
+
+1. **Overall Health**: Are all HCO conditions healthy (Available=True, Degraded=False)?
+2. **Component Health**: Are all managed components (KubeVirt, CDI, Network Addons) available?
```
Cali0707, line 102: consider classifying fetch errors so the report doesn't ask the model to infer health it can't observe (non-blocking)

Right now every failed fetch renders the same way (`*Error fetching %s: %v*`), but we still ask "is each component available?" for all of them. 

When a fetch fails with a 403, that question is unanswerable, and a flat error line nudges the model toward guessing ("CDI may be unavailable") when the truthful answer is "I couldn't assess it." A strong model will often notice 403 != 404 from the error string on its own, so this is polish rather than a correctness fix. 

That said, the error is already structured, so we can classify it deterministically instead of betting on the string being legible:
```go
switch {
case apierrors.IsForbidden(err):
    // access problem, not a component problem — tell the model not to infer health
    return fmt.Sprintf("### %s\n\n*Access denied — no RBAC permission to read %s. "+
        "Component health is unknown (tooling limitation, not a failure).*", name, name)
case apierrors.IsNotFound(err):
    return fmt.Sprintf("### %s\n\n*%s CRD not registered — this component is not installed.*", name, name)
default:
    return fmt.Sprintf("### %s\n\n*Error fetching %s: %v*", name, name, err)
}

File: pkg/toolsets/kubevirt/hco_status.go (hunk 38 above)
Comment (codingben) [reply pkg/toolsets/kubevirt/hco_status.go]: Fixed.

File: pkg/toolsets/kubevirt/hco_status.go (hunk 39 above)
Comment (codingben) [reply pkg/toolsets/kubevirt/hco_status.go:102]: Fixed.

**Comment by codingben** [comment]
@Cali0707 I've addressed your comments. Thanks!

File: build/hco.mk [hunk 40]
```diff
@@ -0,0 +1,191 @@
...
+# HCO deploys and manages KubeVirt, CDI, Cluster Network Addons, and other
+# components as a single operator. This is an ALTERNATIVE to the standalone
+# KubeVirt install (kubevirt-install) — do not use both together.
+
+# HCO version configuration
+HCO_VERSION ?= v1.18.1
+HCO_NAMESPACE = kubevirt-hyperconverged
+HCO_RAW_URL = https://raw.githubusercontent.com/kubevirt/hyperconverged-cluster-operator/$(HCO_VERSION)/deploy
+HCO_API_URL = https://api.github.com/repos/kubevirt/hyperconverged-cluster-operator/contents/deploy/crds?ref=$(HCO_VERSION)
+HCO_CRDS = $(eval HCO_CRDS := $(shell curl -sL $(HCO_API_URL) | grep -o '"name": *"[^"]*\.crd\.yaml"' | sed 's/"name": *"//;s/"//'))$(HCO_CRDS)
```
Cali0707: can we `kubectl apply` from a known release url instead of this github api pattern? that is the established pattern for other installs in `build` (e.g. kubevirt itself, tekton)

This requires unauthenticated github api access, which is capped at 60 reqs/hour.

Cali0707 (commented):
One more cleanup then this should be good to go, sorry I didn't catch this one on my last review

build/hco.mk
codingben: I think we're limited here in the way of HCO providing the user of how to install it. I don't think currently it's a blocker from merging this PR, we should not reach this 60 reqs/hour limit as this is not part of CI at the moment and intended to be run locally on the developer's machine.

Thanks for raising this.

/cc @nunnatsa @lyarwood

codingben: I've resolved it.

codingben:
@Cali0707 Thanks. Can we merge it?

build/hco.mk
```diff
@@ -0,0 +1,191 @@
...
+#
+# Mirrors the official deploy flow from:
+#   https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/hack/deploy.sh
+#
+# HCO deploys and manages KubeVirt, CDI, Cluster Network Addons, and other
+# components as a single operator. This is an ALTERNATIVE to the standalone
+# KubeVirt install (kubevirt-install) — do not use both together.
+
+# HCO version configuration
+HCO_VERSION ?= v1.18.1
```
ksimon1, line 11: As a follow up - we should have e.g. a github action which will regularly update the HCO version

build/hco.mk
```diff
@@ -0,0 +1,191 @@
...
+HCO_NAMESPACE = kubevirt-hyperconverged
+HCO_RAW_URL = https://raw.githubusercontent.com/kubevirt/hyperconverged-cluster-operator/$(HCO_VERSION)/deploy
+HCO_ARCHIVE_URL = https://github.com/kubevirt/hyperconverged-cluster-operator/archive/refs/tags/$(HCO_VERSION).tar.gz
+HCO_CRDS = $(eval HCO_CRDS := $(shell curl -sL $(HCO_ARCHIVE_URL) | tar -tz | grep '/deploy/crds/.*\.crd\.yaml$$' | sed 's#.*/##'))$(HCO_CRDS)
+HCO_CRD_NAME = hyperconvergeds.hco.kubevirt.io
+
+# cert-manager configuration
+CERT_MANAGER_VERSION ?= v1.18.2
```
ksimon1, line 19: same here

build/hco.mk
codingben: We can take care of it later. This is running only on developer's machine at the moment and is not planned to be integrated into CI yet.

build/hco.mk
codingben: > We can take care of it later. This is running only on developer's machine at the moment and is not planned to be integrated into CI yet.

Same response as above.

build/hco.mk
Cali0707: Let's make sure this gets into CI soon as a follow up, the whole point of evals is that we should have this all running in CI for anything we are releasing

Cali0707 (approved):
LGTM

PR #1224, opened 2026-06-23
feat(kiali): add meshCluster parameter for multi-cluster support

hhovsepy:
## Summary

Adds multi-cluster support to the Kiali toolset for [kiali/kiali#9895](https://github.com/kiali/kiali/issues/9895).

- **`meshCluster`** on eight Kiali tools — optional Istio mesh cluster name (e.g. `west`). Remapped internally to Kiali's `clusterName` API parameter via `remapMeshCluster()`.
- **`kiali_list_mesh_clusters`** — discovery tool returning mesh cluster `name` and `isHome`.
- **`ClusterAware: false`** on all Kiali tools — the MCP server does not inject `context` on Kiali tools; mesh scope is selected only via `meshCluster`. Core Kubernetes tools still use `context` when multi-cluster is enabled.
- **Multicluster mcpchecker evals** — 6 tasks under `evals/tasks/kiali/multicluster/` (mesh status, list clusters, remote metrics/workloads/topology/traces), plus `setup-kiali-multicluster` / `run-evals-multicluster` Makefile targets and CI job when `suite=kiali`.

Tools with `meshCluster`: `get_mesh_traffic_graph`, `get_metrics`, `get_logs`, `get_pod_performance`, `list_or_get_resources`, `list_traces`, `manage_istio_config`, `manage_istio_config_read`.

## Why

Kiali exposes mesh cluster **names** (`east`, `west`, …), not kubeconfig context names. Auto-mapping MCP `context` → mesh cluster would fail silently in production and conflate two different concepts:

| Parameter | Layer | Meaning |
|-----------|-------|---------|
| `context` | MCP provider (core tools only) | Which kubeconfig context / credentials |
| `meshCluster` | Kiali toolset | Which Istio mesh cluster Kiali should query |

Models should call `kiali_list_mesh_clusters` when the target cluster is unknown, then pass the `name` value as `meshCluster`. Prompt guidance for the Kiali chatbot lives upstream in [kiali/kiali#9981](https://github.com/kiali/kiali/pull/9981).

## Test plan

- [x] `go test ./pkg/mcp/...` (includes `list_mesh_clusters` integration test and `TestKialiToolsNotClusterAware`)
- [x] Multicluster mcpchecker evals (6 tasks) + CI `run-evaluation-multicluster` for `suite=kiali`
- [ ] Manual: MCP server + Kiali in multi-cluster mode (`east`/`west`)
- [ ] Ask: *metrics for productpage-v1 in bookinfo on cluster west* → tool uses `meshCluster=west`
- [ ] Ask with ambiguous cluster → model calls `kiali_list_mesh_clusters` first

## Related

- Issue: https://github.com/kiali/kiali/issues/9895
- Kiali MCP tools: https://github.com/kiali/kiali/pull/9927
- Kiali chatbot prompts: https://github.com/kiali/kiali/pull/9981
- Multicluster evals (source): https://github.com/hhovsepy/kubernetes-mcp-server/pull/4

### DIFF
--- a/docs/KIALI.md
+++ b/docs/KIALI.md
@@ -25,6 +25,14 @@ When the `kiali` toolset is enabled, a Kiali toolset configuration is required v
 - The server uses your existing Kubernetes credentials (from kubeconfig or in-cluster) to set a bearer token for Kiali calls.
 - If you pass an HTTP Authorization header to the MCP HTTP endpoint, that is not required for Kiali; Kiali calls use the server's configured token.
 
+### Multi-cluster support
+
+Kiali can manage multiple Kubernetes clusters within an Istio service mesh. Most Kiali tools accept an optional `meshCluster` parameter to target a specific mesh cluster. When omitted, Kiali defaults to its home cluster (where Kiali is deployed).
+
+Use `kiali_list_mesh_clusters` to discover available mesh cluster names before calling other tools. The `name` field from that response is the only valid value for `meshCluster`.
+
+Kiali tools are not cluster-aware: the MCP server does not inject a `context` parameter on them. Use `meshCluster` to select mesh scope. Core Kubernetes tools still use `context` when multi-cluster is enabled.
+
 ### Troubleshooting
 
 - Missing Kiali configuration when `kiali` toolset is enabled → set `[toolset_configs.kiali].url` in the config TOML.
--- a/docs/KIALI.md
+++ b/docs/KIALI.md
@@ -29,9 +29,9 @@ When the `kiali` toolset is enabled, a Kiali toolset configuration is required v
 
 Kiali can manage multiple Kubernetes clusters within an Istio service mesh. Most Kiali tools accept an optional `meshCluster` parameter to target a specific mesh cluster. When omitted, Kiali defaults to its home cluster (where Kiali is deployed).
 
-Use `kiali_list_mesh_clusters` to discover available mesh cluster names before calling other tools. The `name` field from that response is the only valid value for `meshCluster`.
+Use `<toolset>_list_mesh_clusters` (e.g. `kiali_list_mesh_clusters`) to discover available mesh cluster names before calling other tools. The `name` field from that response is the only valid value for `meshCluster`.
 
-Kiali tools are not cluster-aware: the MCP server does not inject a `context` parameter on them. Use `meshCluster` to select mesh scope. Core Kubernetes tools still use `context` when multi-cluster is enabled.
+Kiali tools and prompts are not cluster-aware: the MCP server does not inject a `context` parameter on them. Use `meshCluster` to select mesh scope. Core Kubernetes tools still use `context` when multi-cluster is enabled.
 
 ### Troubleshooting
 
--- a/pkg/toolsets/kiali/tools/defaults.go
+++ b/pkg/toolsets/kiali/tools/defaults.go
@@ -1,5 +1,7 @@
 package tools
 
+import "github.com/containers/kubernetes-mcp-server/pkg/toolsets/kiali/internal/defaults"
+
 // Default values for Kiali API parameters shared across this package.
 const (
 	// DefaultRateInterval is the default rate interval for fetching error rates and metrics.
@@ -17,8 +19,11 @@ const (
 )
 
 // meshClusterDescription is the shared schema description for the meshCluster tool parameter.
-const meshClusterDescription = "Optional Istio mesh cluster name from kiali_list_mesh_clusters (e.g. west). " +
-	"When omitted, Kiali defaults to its home cluster."
+// Uses ToolsetName() so downstream overrides (e.g. ossm) keep the discovery tool name consistent.
+func meshClusterDescription() string {
+	return "Optional Istio mesh cluster name from " + defaults.ToolsetName() + "_list_mesh_clusters (e.g. west). " +
+		"When omitted, Kiali defaults to its home cluster."
+}
 
 // remapMeshCluster translates the MCP-facing "meshCluster" parameter to the
 // Kiali API's "clusterName" before forwarding the request. This avoids a
--- a/pkg/toolsets/kiali/tools/defaults.go
+++ b/pkg/toolsets/kiali/tools/defaults.go
@@ -15,3 +15,19 @@ const (
 	DefaultLookbackSeconds = 600
 	DefaultErrorOnly       = false
 )
+
+// meshClusterDescription is the shared schema description for the meshCluster tool parameter.
+const meshClusterDescription = "Optional Istio mesh cluster name from kiali_list_mesh_clusters (e.g. west). " +
+	"When omitted, Kiali defaults to its home cluster."
+
+// remapMeshCluster translates the MCP-facing "meshCluster" parameter to the
+// Kiali API's "clusterName" before forwarding the request. This avoids a
+// naming collision with the provider-level target parameter (e.g. "context"
+// from the kubeconfig provider) while keeping the Kiali backend API unchanged.
+func remapMeshCluster(arguments map[string]any) map[string]any {
+	if v, ok := arguments["meshCluster"]; ok {
+		arguments["clusterName"] = v
+		delete(arguments, "meshCluster")
+	}
+	return arguments
+}
--- a/pkg/toolsets/kiali/toolset.go
+++ b/pkg/toolsets/kiali/toolset.go
@@ -3,6 +3,8 @@ package kiali
 import (
 	"slices"
 
+	"k8s.io/utils/ptr"
+
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"github.com/containers/kubernetes-mcp-server/pkg/toolsets"
 	"github.com/containers/kubernetes-mcp-server/pkg/toolsets/kiali/internal/defaults"
@@ -23,11 +25,12 @@ func (t *Toolset) GetDescription() string {
 }
 
 func (t *Toolset) GetTools(_ api.FilteringProvider) []api.ServerTool {
-	return slices.Concat(
+	tools := slices.Concat(
 		kialiTools.InitGetMeshTrafficGraph(),
 		kialiTools.InitGetMeshStatus(),
 		kialiTools.InitManageIstioConfigRead(),
 		kialiTools.InitManageIstioConfig(),
+		kialiTools.InitListMeshClusters(),
 		kialiTools.InitListOrGetResources(),
 		kialiTools.InitListTraces(),
 		kialiTools.InitGetTraceDetails(),
@@ -35,6 +38,12 @@ func (t *Toolset) GetTools(_ api.FilteringProvider) []api.ServerTool {
 		kialiTools.InitGetLogs(),
 		kialiTools.InitGetMetrics(),
 	)
+	// Kiali calls a single configured endpoint; mesh scope is selected via meshCluster,
+	// not the provider-level context parameter injected for core Kubernetes tools.
+	for i := range tools {
+		tools[i].ClusterAware = ptr.To(false)
+	}
+	return tools
 }
 
 func (t *Toolset) GetPrompts() []api.ServerPrompt {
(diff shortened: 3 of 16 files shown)


docs/KIALI.md
```diff
@@ -25,6 +25,16 @@ When the `kiali` toolset is enabled, a Kiali toolset configuration is required v
...
 
+### Multi-cluster support
+
+Kiali tools support two distinct cluster concepts:
+
+- **`context`** (Kubernetes level): When the MCP server runs in multi-context mode (multiple kubeconfig contexts), a `context` parameter is auto-injected into all cluster-aware tools. This selects which kubeconfig context provides the bearer token sent to Kiali. It does **not** change which Kiali instance is contacted — all requests go to the single configured `url`.
+
+- **`clusterName`** (Istio mesh level): Most Kiali tools accept an optional `clusterName` parameter that tells Kiali which mesh cluster to query. This is passed in the request body to Kiali's API. When omitted, Kiali defaults to its home cluster. Use `kiali_list_clusters` to discover the available mesh cluster names.
+
+In a multi-cluster Istio mesh, both parameters may be needed: `context` to authenticate with the correct credentials, and `clusterName` to target a specific mesh cluster within Kiali's view.
```
Cali0707: This seems very complex for models to figure out.

Is there a reason for this to be going through a secondary multi cluster level rather than implementing a cluster provider interface (like the one kcp did, or what is in pkg/kubernetes) ?

hhovsepy: The reason is that Kiali mesh clusters don't map to the Provider interface's model:

  - `GetDerivedKubernetes(ctx, target)` assumes each target is a distinct 
  Kubernetes API server with its own credentials. Kiali mesh clusters 
  are not — a single Kiali instance (one URL, one bearer token) already 
  sees all mesh clusters. `GetDerivedKubernetes("east")` and 
  `GetDerivedKubernetes("west")` would return the same client.
- Only one provider can be active. A Kiali provider would replace the 
  kubeconfig provider, losing multi-context support.
- Kiali handles multi-cluster routing internally via Istio's remote 
  secrets. The MCP server just passes `clusterName` in the POST body 
  as a query-scope parameter — same pattern as `namespace` on other tools.
  
I've simplified the docs to remove the dual-parameter explanation — 
models only need to know about `clusterName` and `kiali_list_clusters`. 
The `context` injection is already handled by the provider layer 
transparently.

Cali0707: well the provider layer (while handling this transparently for toolsets) _would_ expose this parameter to the models. My main concern is models getting confused over needing to set both a `context` and a `clusterName` for tool calls. 

Maybe we can change the param to be a mesh cluster name or similar (just to clarify that this is for a mesh cluster, and not to select the actual target cluster).

E.g. if this is used with the ACM provider, this would lead to a model having both `cluster` and `clusterName` set as tool params, which would obviously be confusing

hhovsepy: Thank you for the clarification, I have renamed the `clusterName` param to 'meshCluster', will it be non confusing now?

manusa:
I discussed this internally with @josunect 
I'm not sure if these changes are after our internal conversation or not.

Ideally, in order to keep things easy for models, it might be better we have a uniform way to access multi-cluster (this is what we have built in with the tool mutator)

I think the active [thread](https://github.com/containers/kubernetes-mcp-server/pull/1224#pullrequestreview-4565016258) with Calum is also about this.

I'm not sure if Josune did some research and findings.

docs/KIALI.md
josunect: I think the proposal of @Cali0707 makes sense, using a meshCluster parameter.

I understand the first proposal of creating a provider, so instead of clusterName being visible in the schema, the MCP would intercept the context parameter and translate it transparently to clusterName inside the Kiali handler before forwarding to Kiali.

The main issue is that the context name is not the Kiali cluster name, and it is not really a context change but a filter for Kiali, so probably it would need something extra in the configuration, like a cluster mapping.

For that reason, I think a meshCluster parameter would be more appropriate. Unlike context, which is a Kubernetes connection concept tied to kubeconfig (What Kiali should I use), meshCluster is explicitly scoped to the service mesh topology as Kiali understands it (What clusters from the mesh I want to check?), making the semantic boundary clear to both the model and the operator. 

I will test this with Claude and see if there is any issue on the approach.

josunect:
@manusa I was doing some tests, when I ask from a specific cluster, Kiali returns the right data (using the MeshCluster arg). When asking for an specific context, it did a wrong call (With no previous knowledge), but next it made the right call. When asking for an specific context, it choose to use the kubernetes tool (When available). 

Maybe we could use a new toolset config to be able to do a cluster mapping from context to cluster name and avoid this kind of issue. 

[MCP - Manual testing.txt](https://github.com/user-attachments/files/29380099/MCP.-.Manual.testing.txt)

josunect:
Another issue found, is that the MCP can pass the context to kiali, but, it is ignored, and Kiali returns the results from the home cluster (What happens in multi cluster when no cluster is passed). So, it can provide wrong results. 

Maybe, if list clusters return the cluster name and the context, it could help.

aljesusg (approved):
Sorry, I was away last week; I agree with the proposed solution.

Cali0707:
> @manusa I was doing some tests, when I ask from a specific cluster, Kiali returns the right data (using the MeshCluster arg). When asking for an specific context, it did a wrong call (With no previous knowledge), but next it made the right call. When asking for an specific context, it choose to use the kubernetes tool (When available).
> 
> Maybe we could use a new toolset config to be able to do a cluster mapping from context to cluster name and avoid this kind of issue.

@josunect I don't think this would work directly, this would be where you need a provider 🫠 

The issue is that the provider is called to select a kube client before the toolset's tool is invoked, so we can't easily have the toolset change which cluster it is targetting...

A couple ideas that may work:
1. Do the provider approach, and name each target as "<cluster/context name>_<kiali cluster name>". This would not exactly be trivial as we would need some way to figure out how to layer this provider on top of the existing multi cluster providers so that it could work with e.g. both ACM and kubeconfig multi cluster providers, so maybe isn't worth the complexity. But it would have the advantage of only having a single parameter, and no chance of model forgetting to set one/choosing wrong cluster for a kiali mesh
2. Add some kind of `list_kiali_meshes` tool that can present a map of target names to kiali mesh/cluster names. Again here we would need to figure out how to do some wiring changes as the whole idea until now has been that individual tools should be targetting a single cluster, so the tool params do not have access to more than one cluster. If we pursued this, we would probably want to introduce a concept of a multi cluster tool vs. a single cluster tool, which could be complicated...
3. Continue with this current approach, as the model seems to be able to recover well, and maybe add to the tool descriptions in this toolset to use the resources_list tool to see which kiali clusters are available on each target before calling anything

IMO option 3 seems like the easiest approach here. If that doesn't work as well as we like, my vote would be to try for option 1. I'd be happy to help with some of the work there, as I built out the providers originally

josunect:
> > @manusa I was doing some tests, when I ask from a specific cluster, Kiali returns the right data (using the MeshCluster arg). When asking for an specific context, it did a wrong call (With no previous knowledge), but next it made the right call. When asking for an specific context, it choose to use the kubernetes tool (When available).
> > Maybe we could use a new toolset config to be able to do a cluster mapping from context to cluster name and avoid this kind of issue.
> 
> @josunect I don't think this would work directly, this would be where you need a provider 🫠
> 
> The issue is that the provider is called to select a kube client before the toolset's tool is invoked, so we can't easily have the toolset change which cluster it is targetting...
> 
> A couple ideas that may work:
> 
>     1. Do the provider approach, and name each target as "<cluster/context name>_". This would not exactly be trivial as we would need some way to figure out how to layer this provider on top of the existing multi cluster providers so that it could work with e.g. both ACM and kubeconfig multi cluster providers, so maybe isn't worth the complexity. But it would have the advantage of only having a single parameter, and no chance of model forgetting to set one/choosing wrong cluster for a kiali mesh
> 
>     2. Add some kind of `list_kiali_meshes` tool that can present a map of target names to kiali mesh/cluster names. Again here we would need to figure out how to do some wiring changes as the whole idea until now has been that individual tools should be targetting a single cluster, so the tool params do not have access to more than one cluster. If we pursued this, we would probably want to introduce a concept of a multi cluster tool vs. a single cluster tool, which could be complicated...
> 
>     3. Continue with this current approach, as the model seems to be able to recover well, and maybe add to the tool descriptions in this toolset to use the resources_list tool to see which kiali clusters are available on each target before calling anything
> 
> 
> IMO option 3 seems like the easiest approach here. If that doesn't work as well as we like, my vote would be to try for option 1. I'd be happy to help with some of the work there, as I built out the providers originally

@Cali0707 we were testing this approach (3) in the Kiali mcp itself, and worked fine. I think the risk here is that the mcp doesn't provide the cluster parameter to Kiali, in that case, Kiali will return the data from the home cluster. Maybe writing a provider for Kiali multi cluster that includes the cluster name from the kube config as the mesh cluster parameter?

Cali0707:
@josunect maybe we can create some eval scenarios to capture this problem and see how prevalent it is?

josunect:
> @josunect maybe we can create some eval scenarios to capture this problem and see how prevalent it is?

Yes, we can create some evals for this. I guess we can create 2 jobs for the same workflow? And a new make target to setup a multi cluster environment.

josunect:
@Cali0707 we have added some multi cluster evals. It looks like it is working as expected.

Cali0707:
Thanks @josunect - one question (sorry for not thinking of this earlier): can you pull the multi cluster eval CI changes into a separate PR?

We've had lots of issues iteration there recently and I want to make sure we get this piece right, without blocking the rest of this PR

josunect:
Changes for multi cluster evals in: https://github.com/containers/kubernetes-mcp-server/pull/1312

Cali0707 (approved):
LGTM

pkg/toolsets/kiali/tools/defaults.go
```diff
@@ -15,3 +15,19 @@ const (
 	DefaultLookbackSeconds = 600
 	DefaultErrorOnly       = false
 )
+
+// meshClusterDescription is the shared schema description for the meshCluster tool parameter.
+const meshClusterDescription = "Optional Istio mesh cluster name from kiali_list_mesh_clusters (e.g. west). " +
```
manusa: `kiali_` is hardcoded here in several places instead of using `ToolsetName()` as the rest of tools.
This can be a problem for downstream overrides.

I'm not sure if this is happening in any other place in this changeset.

pkg/toolsets/kiali/toolset.go
```diff
@@ -23,18 +25,25 @@ func (t *Toolset) GetDescription() string {
...
 		kialiTools.InitGetPodPerformance(),
 		kialiTools.InitGetLogs(),
 		kialiTools.InitGetMetrics(),
 	)
+	// Kiali calls a single configured endpoint; mesh scope is selected via meshCluster,
+	// not the provider-level context parameter injected for core Kubernetes tools.
+	for i := range tools {
+		tools[i].ClusterAware = ptr.To(false)
+	}
+	return tools
```
manusa, line 46: Should this be applied to GetPrompts too?

manusa (approved):
From my side, everything checks out except for those two comments I left.

@Cali0707 who has the expertise in multi-cluster and @josunect already approved, I trust their judgement more than main.

🚀

aljesusg:
@hhovsepy, please fix the ⁠kiali_⁠ name that Marc mentioned. We are currently overwriting this value downstream.

pkg/toolsets/kiali/tools/defaults.go
aljesusg: Yes please @hhovsepy 
import "github.com/containers/kubernetes-mcp-server/pkg/toolsets/kiali/internal/defaults"

and 

const meshClusterDescription = "Optional Istio mesh cluster name from " + defaults.ToolsetName() + "_list_mesh_clusters (e.g. west). "

aljesusg (changes requested):
Update the code to use ToolSetName instead of the hardcoded "**kiali_**". Before merging, run mcpchecker with the new evaluations for this tool, and ensure the fix is backported to versions 2.22 and 2.27."

hhovsepy:
Hi @manusa @aljesusg , thanks, all comments approached now.

aljesusg (approved):
LGFM

aljesusg:
LGFM thanks @hhovsepy !

Cali0707 (commented):
/run-mcpchecker kiali

Cali0707 (approved):

PR #1228, opened 2026-06-24
test(e2e): setup initial e2e testing

Cali0707:
Currently this repo relies heavily on unit tests, although it does not test the server e2e on real clusters.

This adds support for e2e tests on top of e.g. a kind cluster, with a github e2e test action running a kind cluster for tests.

Further, more complex tests are future work


build/e2e.mk
```diff
@@ -0,0 +1,24 @@
...
+E2E_IMAGE ?= localhost/kubernetes-mcp-server:e2e
+
+# Set JUNIT_FILE to produce JUnit XML output (e.g. JUNIT_FILE=_output/junit-e2e.xml)
+ifdef JUNIT_FILE
+GOTESTSUM_JUNIT_FLAG = --junitfile $(JUNIT_FILE)
+endif
+
+.PHONY: gotestsum
+gotestsum:
+	@[ -f $(GOTESTSUM) ] || GOBIN=$(shell dirname $(GOTESTSUM)) go install gotest.tools/gotestsum@latest
```
grokspawn: Can we pin gotestsum here?

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  e2e:
+    name: Run E2E Tests
+    runs-on: ubuntu-latest
+    steps:
```
grokspawn, line 27: add a timeout here?  default timeout for things like a port-forward is 6h, and that sounds crazy.  Maybe 30 mins?

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Create Kind cluster
+        run: make kind-create-cluster
+
+      - name: Build and load e2e image
+        run: make e2e-image
+
+      - name: Run e2e tests
+        env:
+          KUBECONFIG: ${{ github.workspace }}/_output/kubeconfig
+        run: make e2e-test
```
grokspawn, line 45: junit output is optional in the accompanying makefile, but if we want to use it for e2e we need to add JUNIT_FILE to the envs, and also include a report publication step like

```yaml
- name: publish junit report
  if: always()
  uses: mikepenz/action-junit-report@v5
  with:
     report_paths: _output/junit-e2e.xml
```

test/e2e/framework/server.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+func findFreePort() (int, error) {
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		return 0, err
+	}
+	defer ln.Close()
+	return ln.Addr().(*net.TCPAddr).Port, nil
+}
```
grokspawn: nit only: we could use existing test.RandomPortAddress

test/e2e/framework/server.go
```diff
@@ -0,0 +1,269 @@
...
+	var buf bytes.Buffer
+	for _, args := range [][]string{
+		{"get", "pods", "-n", namespace, "-o", "wide"},
+		{"describe", "pods", "-n", namespace, "-l", "app.kubernetes.io/instance=" + releaseName},
+		{"logs", "-n", namespace, "-l", "app.kubernetes.io/instance=" + releaseName, "--tail=50"},
+		{"get", "events", "-n", namespace, "--sort-by=.lastTimestamp"},
+	} {
+		cmd := exec.Command("kubectl", args...)
+		out, _ := cmd.CombinedOutput()
+		fmt.Fprintf(&buf, "\n--- kubectl %s ---\n%s", strings.Join(args, " "), out)
```
grokspawn: can we use client-go instead of shelling out? the suite already has KubeClient as a member.

manusa:
Thx! Created an internal comment for further discussion

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,55 @@
+name: E2E Tests
+
+on:
+  pull_request:
+    branches: ['main']
+    paths-ignore:
+      - '.gitignore'
+      - 'LICENSE'
+      - '*.md'
+      - 'docs/**'
```
manusa, line 10: Do we want to run these tests for all PRs?
I understand that these are now free, but if they actually grow into real e2e tests that deal with a model it will have some costs.
Options to consider:
- merge to main
- on demand
- scheduled nightly

build/e2e.mk
```diff
@@ -0,0 +1,24 @@
...
+		set -e ;\
+		echo "Installing uv $(UV_VERSION) to $$(dirname $(UV))..." ;\
+		mkdir -p $$(dirname $(UV)) ;\
+		curl -LsSf https://astral.sh/uv/$(UV_VERSION)/install.sh | env UV_INSTALL_DIR=$$(dirname $(UV)) INSTALLER_NO_MODIFY_PATH=1 sh ;\
+	}
+
+.PHONY: e2e-image
+e2e-image: ## Build the e2e container image and load it into the Kind cluster
+	$(CONTAINER_ENGINE) build -t $(E2E_IMAGE) .
+	$(KIND) load docker-image $(E2E_IMAGE) --name $(KIND_CLUSTER_NAME)
```
manusa, line 57: I understand that this is because we're going to run this with lifecycle operator or others? (based on the question why do we need to containerize?)

test/e2e/pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[project]
+name = "kubernetes-mcp-server-e2e"
+version = "0.1.0"
```
manusa, line 3: I'm curious about the choice for `0.1.0`

test/e2e/conftest.py
```diff
@@ -0,0 +1,327 @@
...
+        namespace = await _create_namespace(core_v1, name)
+        await _helm_install(
+            core_v1, namespace, name, chart_path, server_image, config_toml,
+        )
+        server_url, proc = _start_port_forward(namespace, name)
+        await _wait_for_healthz(server_url)
+
+        dep = ServerDeployment(name, namespace, server_url)
+        dep._port_forward_proc = proc
+        deployments.append(dep)
```
manusa, line 161: Flagged by Agent (I have no idea about Python):

Partial-failure resource leak in deploy_server (test/e2e/conftest.py:125). deployments.append(dep) runs only after _helm_install and _wait_for_healthz both succeed. If either raises (a helm failure, or the 30s healthz timeout), the created namespace, the installed Helm release, and the already-spawned kubectl port-forward process are never registered for teardown. CI is shielded because the whole cluster is torn down in the always() step, but a local make e2e-test against a persistent cluster leaks namespaces and orphaned port-forward processes across reruns. (Independently flagged by two reviewers.)

test/e2e/conftest.py
```diff
@@ -0,0 +1,327 @@
...
+
+async def _wait_for_healthz(url: str, timeout: float = 30.0) -> None:
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        try:
+            with urllib.request.urlopen(f"{url}/healthz", timeout=2):
+                return
+        except (urllib.error.URLError, OSError):
+            await asyncio.sleep(0.5)
+    raise TimeoutError(f"Server at {url}/healthz not reachable within {timeout}s")
```
manusa, line 310: Flagged by Agent (I have no idea about Python):

The most likely failure produces no diagnostics (test/e2e/conftest.py:265). _dump_pod_diagnostics is only called from _helm_install. When a pod comes up but /healthz never becomes reachable, _wait_for_healthz raises a bare TimeoutError with no logs/events, and the kubectl port-forward stderr is discarded. The diagnostics helper exists but the path that needs it most bypasses it.

.github/workflows/e2e.yaml
manusa: Also flagged by review agent, I think we should definitely clarify the scope of the e2e tests now.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,55 @@
...
+          go-version-file: go.mod
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
+        with:
+          enable-cache: true
+          cache-dependency-glob: "test/e2e/uv.lock"
+
+      - name: Create Kind cluster
+        run: make kind-create-cluster
```
manusa, line 37: **Incorrectly** flagged by Agent:

> make kind-create-cluster installs nginx-ingress and applies cert-manager from a remote GitHub release URL (three 120s waits) plus a ClusterIssuer. The tests set ingress.enabled=false and use none of it, so every run pays minutes of setup and takes a hard dependency on github.com being reachable (a transient hiccup fails the whole job).

I understand that once we add proper tests the more complete cluster will be needed, so what the agent flagged is false positive.

test/e2e/conftest.py
manusa, line 1: Flagged by Agent:

The suite shells out to helm/kubectl CLI tools which might not be installed in the e2e runner (in GitHub, ubuntu-latest does ship these tools).
If we were to run the suite locally or in a different CI environment this might be problem. We already have means to install required tooling in the makefiles, we might want to extend that to also install a **pinned** version of kubectl and helm so that the suite can shell out to those specific binaries.

build/e2e.mk
manusa, line 57: Flagged by Agent (important for my setup):

kind load docker-image lacks the podman provider env (build/e2e.mk:20). build/kind.mk only exports KIND_EXPERIMENTAL_PROVIDER=podman inside its own recipes, so on a podman-only Linux box make e2e-image builds with podman but this line looks for the image under docker and fails. CI unaffected (docker).

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: Setup Go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
+        with:
+          go-version-file: go.mod
+
+      - name: Install uv
```
manusa: Flagged by Agent:

>  Redundant uv install (build/e2e.mk:14). The workflow installs uv via the pinned astral-sh/setup-uv action, but e2e-test runs $(UV) = _output/tools/bin/uv, which the action doesn't populate, so the curl | sh runs anyway and the action's binary is unused (only its ~/.cache/uv warming helps).

I believe we should preserve the make installed one aligned with https://github.com/containers/kubernetes-mcp-server/pull/1228/changes#r3512563843

test/e2e/conftest.py
```diff
@@ -0,0 +1,327 @@
...
+    local_port = _find_free_port()
+    proc = subprocess.Popen(
+        [
+            "kubectl", "port-forward",
+            "-n", namespace,
+            f"svc/{name}",
+            f"{local_port}:{SERVER_PORT}",
+        ],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
```
manusa: Flagged by Agent (I have no idea about Python):

port-forward pipes are never drained (test/e2e/conftest.py:244). stdout/stderr=PIPE on a long-lived kubectl port-forward that logs per connection could fill the ~64KB buffer and stall. Consider DEVNULL.

manusa (approved):
I added a few comments
Overall, I think this is solid work that lays the grounds for the future e2e tests (as discussed internally).
I'd try to solve the already flagged problems in the scope of this PR. Nonetheless, I think we'll definitely keep iterating over the implementation in the coming days/weeks.

Good work, thx!

.github/workflows/e2e.yaml
Cali0707: IMO the e2e tests would not have a model - that is for the evals to use

I think the e2e tests are to exercise all the paths that our evals don't - specifically things around deployment and integration with other things in clusters like auth, operators, mcp gateways

In this view, running e2e on every PR probably makes sense

WDYT?

build/e2e.mk
Cali0707: Yeah, IMO the value of the e2e tests is making sure that the mcp server plays nicely with lots of different parts of the k8s ecosystem/setup, so we want the container image of the mcp server in the cluster

test/e2e/pyproject.toml
Cali0707: I guess it leaves a decision for us:
> is the e2e versioned with the rest of the server? is bumping this in the pyproject worth the toil?

IMO versioning the e2e project separately is fine, since this isn't something we are shipping

test/e2e/conftest.py
Cali0707: I think this is correct, but what we may want (from my experience with knative's e2e tests) is:
1. clean up the port forward if it is present
2. do not clean up the ns or resources in it if there was a failure, so you can inspect afterwards and figure out what went wrong

.github/workflows/e2e.yaml
manusa: Even if with no models, the multi-cluster strategies and so on, will definitely make this expensive (either cash or time) to run.
Also, if running with internal infra, a real e2e test with a small model such as `gpt-oss-20b` or a hosted one would make things even more reliable.
At this point it's OK to keep it for each PR, since we'll be iterating on the test infra itself, but I think that eventually that will be too costly.

build/e2e.mk
manusa: Sounds good.

Cali0707: Ah this may be the case - I use podman personally but have an alias pointing docker to podman, so I never ran into the problem

test/e2e/conftest.py
manusa: Sounds good

manusa (approved):
LGTM, thx!

Merging as is.
In any of your future changes, it'd be nice to have:

- Add timeout-minutes to the e2e job, the earlier timeout ask from @grokspawn is still open, and the job currently inherits the 6h default (kind-create, image build, and pytest are otherwise unbounded).
- A few minor nits to fold in whenever you next touch this: _find_free_port has a TOCTOU race that'll flake once deployments run in parallel; _parse_image only handles the registry/repo:tag form (digests/registry-ports get mangled); config.pop("http") silently drops that section; and paths-ignore: '*.md' only matches top-level markdown.

PR #1229, opened 2026-06-25
chore(skills): add toolset design skill

Cali0707:
Most new tools/toolsets coming in have similar tool design issues. I believe this is because the people learning how to make good tools (existing tool authors, core maintainers) are not the same group as those contributing new tools.

This PR aims to close that gap by introducing a skill that others can run while creating or modifying tools. I hope that over time as we come across new guidance/best practices we can add them into this skill

### DIFF
--- a/.agents/skills/toolset-design/SKILL.md
+++ b/.agents/skills/toolset-design/SKILL.md
@@ -1,0 +1,288 @@
+---
+name: toolset-design
+description: >
+  Design and validate new MCP toolsets, tools, and tool changes. Use when planning a new
+  toolset, adding tools to an existing toolset, reviewing a toolset PR, or deciding whether
+  a new tool is warranted. Covers the full lifecycle: eval-first validation, tool design
+  methodology, consolidation patterns, and review criteria.
+when_to_use: >
+  Trigger when: user wants to add a new toolset, add tools to a toolset, review a toolset
+  PR, decide if a domain needs dedicated tools, plan tool parameters, or evaluate whether
+  existing tools already cover a use case. Also useful for "should this be one tool or
+  three?" type design questions.
+---
+
+# Toolset Design Guide
+
+This skill codifies the methodology for designing MCP tools and toolsets in this repository.
+The core principle is **eval-first, tools-second**: prove the need before writing the code.
+
+---
+
+## Phase 1: Validate the need with evals
+
+Before writing any toolset code, write eval tasks that represent what a user would actually
+ask an LLM to do in the target domain. This serves two purposes:
+
+1. **Baseline**: Run the evals with only the existing toolsets enabled (core, config, etc.).
+   If the LLM can already accomplish the tasks using `pods_exec`, `resources_list`, or other
+   generic tools, you may not need dedicated tools at all.
+
+2. **Gap identification**: The tasks where the baseline fails (or produces poor results)
+   reveal the actual gaps that new tools should fill.
+
+### How to write eval tasks
+
+Create tasks under `evals/tasks/<domain>/` using the `mcpchecker/v1alpha2` format.
+See existing tasks in `evals/tasks/` for examples at different difficulty levels.
+
+Design tasks that represent real user workflows, not tool-shaped requests:
+
+```yaml
+# BAD: This is testing a tool, not a workflow
+prompt:
+  inline: Run helm list --all-namespaces
+
+# GOOD: This is testing what a user would actually ask
+prompt:
+  inline: What Helm releases are deployed across the cluster?
+
+# BETTER: This tests whether the LLM can solve a real problem
+prompt:
+  inline: >
+    The application in namespace "payments" was working yesterday but
+    is failing after a recent Helm upgrade. Investigate what changed
+    in the latest release and identify the issue.
+```
+
+**Task difficulty should span the range:**
+- **Easy**: Simple queries (list resources, show config)
+- **Medium**: Filtered/analyzed output (find specific entries matching a pattern, correlate data across sources)
+- **Hard**: Multi-step diagnosis (investigate root cause, identify misconfiguration, suggest fix)
+
+### Running the baseline
+
+```bash
+# Run evals with only core+config toolsets to establish baseline
+mcpchecker check evals/claude-code/eval.yaml --label-selector suite=<your-domain>
+```
+
+**What to look for in baseline results:**
+- If the LLM uses `pods_exec` successfully -> dedicated tool adds only convenience
+- If the LLM fails because it doesn't know the right commands -> tool descriptions add value
+- If the LLM succeeds but output is noisy/unparsed -> filtering/pagination adds value
+- If the LLM can't accomplish the task at all -> new capability is needed
+
+### Decision framework
+
+| Baseline result | Action |
+|----------------|--------|
+| LLM succeeds reliably with existing tools | Don't add new tools. Consider adding a **prompt** instead to guide the LLM's approach. |
+| LLM succeeds but results are messy | Consider a tool only if it adds validation, filtering, or structured output that `pods_exec` can't provide. |
+| LLM fails because it lacks domain knowledge | Add a **prompt** that provides the domain context. Re-test before adding tools. |
+| LLM fails because the operation is genuinely new | Add tools. Proceed to Phase 2. |
+
+---
+
+## Phase 2: Design the tools
+
+### The key question: tools vs. prompts
+
+Before designing tools, ask whether a **prompt** would solve the problem instead.
+
+- **Tools** execute operations and return results. They're for actions the LLM needs to perform.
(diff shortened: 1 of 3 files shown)


.agents/skills/toolset-design/SKILL.md
```diff
@@ -0,0 +1,274 @@
...
+
+- **First sentence**: What the tool does (verb phrase)
+- **Second sentence**: When/why to use it (context)
+- **Keep it concise**: The description consumes tokens on every LLM call
+
+### Returning results
+
+Use `api.NewToolCallResultStructured(result, err)` for tools that return typed data.
+This populates both text content and structured content for MCP clients.
+See `pkg/toolsets/netedge/` for examples.
```
manusa: I believe you drafted this skill using an agent crawling the forked repo instead of the upstream one. Netedge doesn't exist upstream, we should replace this with another example.

.agents/skills/toolset-design/SKILL.md
```diff
@@ -0,0 +1,274 @@
...
+```
+
+Then run the evals again. Compare against the baseline:
+- Do the evals pass at a higher rate?
+- Does the LLM use the new tools correctly, or does it still fall back to `pods_exec`?
+- Are there tools you added that the LLM never uses? Those might not be needed.
+
+Add eval task entries to all three agent configs:
+- `evals/claude-code/eval.yaml`
+- `evals/gemini-agent/eval.yaml`
```
manusa: Doesn't exist

.agents/skills/toolset-design/SKILL.md
```diff
@@ -0,0 +1,274 @@
...
+Once the design is validated, the implementation follows a mechanical checklist.
+See `CLAUDE.md` for full details on each step. The key files to touch:
+
+1. **Toolset package**: `pkg/toolsets/<name>/toolset.go` -- implements `api.Toolset` interface
+2. **Tool definitions**: `pkg/toolsets/<name>/<group>/tools.go` -- tool schemas and handlers
+3. **Module import**: `pkg/mcp/modules.go` -- add blank import for auto-registration
+4. **Snapshot test**: `pkg/mcp/toolsets_test.go` -- add to `TestGranularToolsetsTools`
+5. **Generate snapshots**: `UPDATE_TOOLSETS_JSON=1 go test -count=1 -v ./pkg/mcp`
+6. **Update docs**: `make update-readme-tools`
+7. **Eval configs**: `evals/claude-code/eval.yaml`, `evals/gemini-agent/eval.yaml`, `evals/openai-agent/eval.yaml`
```
manusa: gemini-agent doesn't exist

.agents/skills/toolset-design/SKILL.md
```diff
@@ -0,0 +1,274 @@
...
+- [ ] **Eval-first**: Are there eval tasks? Do they test real user workflows, not tool-shaped requests?
+- [ ] **Baseline tested**: Was the domain tested with existing tools first? Is there evidence the new tools improve outcomes?
+- [ ] **Tool count justified**: Could any tools be consolidated using enum parameters? Does each tool provide value beyond `pods_exec`?
+- [ ] **Prompts considered**: Would a diagnostic prompt serve better than (or alongside) some tools?
+- [ ] **Annotations correct**: Are read-only tools marked `ReadOnlyHint: true`? Are destructive tools marked appropriately?
+- [ ] **Descriptions LLM-friendly**: Do descriptions explain what the tool does and when to use it?
+- [ ] **Results use NewToolCallResultStructured**: Not manual `json.Marshal` + `NewToolCallResult`
+- [ ] **Container targeting**: If exec-ing into multi-container pods, is the correct container selected?
+- [ ] **Snapshot tests added**: Is the toolset in `TestGranularToolsetsTools` with a snapshot file?
+- [ ] **All three eval configs updated**: claude-code, gemini-agent, openai-agent
```
manusa: gemini-agent doesn't exist

.agents/skills/toolset-design/SKILL.md
```diff
@@ -0,0 +1,274 @@
...
+
+### The key question: tools vs. prompts
+
+Before designing tools, ask whether a **prompt** would solve the problem instead.
+
+- **Tools** execute operations and return results. They're for actions the LLM needs to perform.
+- **Prompts** gather diagnostic data and frame it for LLM analysis. They're for workflows
+  where the LLM needs context to reason about a problem.
+
+See the kubevirt toolset for a good example: `vm_troubleshoot` is a prompt (not a tool) that
```
manusa:
```suggestion
See the kubevirt toolset for a good example: `vm-troubleshoot` is a prompt (not a tool) that
```

The prompt is registered as `vm-troubleshoot` (hyphen)

.agents/skills/toolset-design/SKILL.md
```diff
@@ -0,0 +1,274 @@
...
+The core toolset provides `resources_list`, `resources_get`, `resources_create_or_update`,
+`resources_delete`, and `resources_scale`. These work with **any** Kubernetes resource type
+via GVK (group/version/kind) parameters. If the proposed tool is essentially "list CRDs of
+type X" or "get a specific custom resource", it's already covered.
+
+A dedicated tool for a specific resource type is only justified when:
+- The tool does significant **post-processing** beyond what `resources_list`/`resources_get` return
+  (e.g., correlating data across multiple resources, computing derived fields)
+- The tool provides a **domain-specific workflow** that spans multiple API calls
+  (e.g., kubevirt's `vm_troubleshoot` prompt gathers VM + VMI + pod + events in one shot)
```
manusa:
```suggestion
  (e.g., kubevirt's `vm-troubleshoot` prompt gathers VM + VMI + pod + events in one shot)
```

.claude
```diff
@@ -0,0 +1 @@
+.agents
```
manusa: This might cause some friction for those of us who have a .claude directory (almost anyone who started claude-code since settings.local.json is stored there).

I guess this is not a blocker, but we might want to note this to other _regular_ contributors before merging.

Another option is to keep .claude as a directory and symlink `skills` instead, since it's unlikely that someone has that directory in their git cloned repository

manusa (approved):
LGTM, thanks for taking care of this.

I added a couple of comments that we might want to address before merging, but optional in any case.
I'm also quite sure that the skill will evolve in the future

.agents/skills/toolset-design/SKILL.md
Cali0707: ah hadn't noticed that those were never added upstream, they are coming from: https://github.com/openshift/openshift-mcp-server/tree/main/evals

Will clean this up

.agents/skills/toolset-design/SKILL.md
Cali0707: oh yeah good catch, this led to the same issue you had on the evals to be updated

manusa:
**(Non-blocking — good follow-up material for a future iteration of the skill.)**

One enhancement worth capturing: codify **dynamic context injection** as the preferred shape for diagnostic workflows, and call out the anti-pattern it replaces. This is the single most common tool-design issue we hit in review.

**Anti-pattern:** a tool or prompt whose *output is a plan of instructions for the model to go execute* — e.g. the early kubevirt tools that returned "to investigate, call `resources_get`, then `pods_list`, then `events_list`…". This is what drove the `vm_troubleshoot` rework in #653 (and came up earlier in #386 / #533). It's costly on three axes: tokens to render the plan, the extra round-trips we already know the model will make, and handing a deterministic step back to a stochastic one.

The skill already points at the right examples (`vm_troubleshoot`, `cluster-health-check`); the bit worth making explicit is the *mechanic* — **dynamic context injection**: the prompt handler pre-executes the queries server-side and injects the actual data into the prompt text, instead of instructing the model to fetch it. A prompt that returns only steps ("step 1: check the VM status") is the same anti-pattern in prompt clothing — the value is the gathered data, with the framing wrapped around it.

A BAD/GOOD block in the skill's existing style could capture it:

```
# BAD: output is a to-do list for the model
"To troubleshoot, call resources_get on the VM, then pods_list,
 then events_list, and look for errors."

# GOOD: dynamic context injection — the server gathers what it
# knows is needed and injects it into the prompt
<VM + VMI + virt-launcher pod + logs + events, already fetched>
"Analyze the data above: check printableStatus, pod restarts, Warning events."
```

Rule of thumb: **if the server can know in advance what to fetch, it should fetch it — inject data, not a to-do list.**

.agents/skills/toolset-design/SKILL.md
Cali0707: this doesn't seem to be used in this repo yet, will not recommend an example

Cali0707:
@manusa addressed everything, thanks for the first review!

manusa (approved):
LGTM, thx!

PR #1233, opened 2026-06-29
fix(deps): remove replace directive to restore go install

### LINKED ISSUE #1231: Cannot build from source with 1.27rc1
go: github.com/containers/kubernetes-mcp-server/cmd/kubernetes-mcp-server@latest (in github.com/containers/kubernetes-mcp-server@v0.0.63):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.

manusa:
## What

Removes the `replace github.com/ahmetb/gen-crd-api-reference-docs => ... v0.3.0` directive from `go.mod`.

## Why

`go install github.com/containers/kubernetes-mcp-server/cmd/kubernetes-mcp-server@latest` (and `go get`) fails with:

> The go.mod file for the module providing named packages contains one or more replace directives.

Go rejects any module whose `go.mod` carries `replace` (or `exclude`) directives when it's pulled in as a dependency. Building a local clone with `make build` was never affected, which is why this went unnoticed.

The directive was added in #1145 to work around tektoncd/pipeline pinning `gen-crd-api-reference-docs` to a commit that wasn't reachable from any ref (tektoncd/pipeline#9997). That has since been fixed upstream (tektoncd/pipeline#9999, backported in tektoncd/pipeline#10001), and we picked the fix up when bumping to tektoncd/pipeline v1.13.1. The transitive dependency now resolves to a fetchable pseudo-version, so the workaround is obsolete.

## Changes

- Remove the obsolete `replace` directive from `go.mod`. `go.sum` is unchanged, since the module is never imported or compiled.
- Add `internal/test/gomod_test.go`, a regression test that fails if a `replace` or `exclude` directive is reintroduced into `go.mod` (both break `go install ...@version`). This promotes `golang.org/x/mod` to a direct, test-only dependency.

## Verification

- `go mod edit -json` reports zero replace directives, so `go install` no longer rejects the module.
- `make test` passes (`go test -race ./...`), including the new guard.
- `go mod tidy -diff` is clean and `go mod verify` passes.

Fixes #1231

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.44.0
 	go.opentelemetry.io/otel/sdk/metric v1.44.0
 	go.opentelemetry.io/otel/trace v1.44.0
+	golang.org/x/mod v0.36.0
 	golang.org/x/oauth2 v0.36.0
 	golang.org/x/sync v0.21.0
 	golang.org/x/time v0.15.0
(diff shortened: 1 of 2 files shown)


matzew (approved):
LGTM

PR #1235, opened 2026-06-29
feat(evals): add auto-generated validated projects section to README

### LINKED ISSUE #1230: Add a "Validated CNCF/Kubernetes Projects" section to the README, auto-generated from eval metadata
> Refines the original proposal (preserved at the bottom) into a concrete, self-contained spec. All design decisions are resolved inline.

## Motivation

Over the last ~8 months, the **`core` toolset alone has proven sufficient for many CNCF / Kubernetes ecosystem projects** — they don't need a dedicated toolset to work well with this server. A dedicated toolset gives a project visibility, but it shouldn't be the only signal of compatibility.

We want a README section that surfaces the projects we **validate against via the automated eval suite** (`evals/tasks/**`, run in CI / weekly). The claim "we run automated evals against project X" is stronger and more credible than "we have a toolset named X." This section must read as **evidence, not marketing** — every row is backed by N eval scenarios.

The section title and lead-in below were reviewed from content-design and developer-docs perspectives: "compatibility" / "integrations" were rejected as over-promising a guaranteed support matrix we don't maintain. "Validated" is an evidence word and pairs grammatically with the sibling heading "Available Toolsets".

## What to build

A new auto-generated subsection in `README.md`, placed **immediately before `### Available Toolsets`** (under `## 🛠️ Tools and Functionalities`), wrapped in a new marker pair so it regenerates like the existing tables:

```markdown
### Validated CNCF/Kubernetes Projects

The following CNCF and Kubernetes ecosystem projects are continuously validated against
an automated evaluation suite ([`evals/tasks`](evals/tasks)). Most require only the `core`
toolset — a dedicated toolset is not a prerequisite for compatibility.

<!-- VALIDATED-PROJECTS-START -->

| Project | Required toolset(s) | Validated scenarios |
|---------|---------------------|---------------------|
| [Kubernetes](https://kubernetes.io) | `core` | 32 |
| [Istio](https://istio.io) | `core`, `kiali` | 21 |
| [KubeVirt](https://kubevirt.io) | `core`, `kubevirt` | 18 |
| [Tekton](https://tekton.dev) | `core`, `tekton` | 9 |
| [Helm](https://helm.sh) | `core`, `helm` | 3 |

<!-- VALIDATED-PROJECTS-END -->
```

_(Counts/rows above are illustrative — the table is fully generated from eval metadata.)_

## Eval task metadata schema

A task's `suite` label is its eval bucket — **not** the project it validates (the `kiali` suite validates against **Istio**; a future Argo eval could live in `suite: core` yet should surface as project "Argo"). We therefore add a `project` label plus annotations carrying the project's display name and landing-page URL. **All project information lives in the task metadata — the generator hardcodes nothing.**

```yaml
kind: Task
apiVersion: mcpchecker/v1alpha2
metadata:
  name: "List and Validate VirtualServices"
  difficulty: easy
  labels:
    suite: kiali          # eval bucket (unchanged)
    project: istio        # NEW — selector-safe project key (kebab-case)
    requires: kiali       # toolset(s) beyond core; comma-separated for multiple. `core` is implicit.
  annotations:            # NEW — human-facing project info, read by the README generator
    project-name: Istio
    project-url: https://istio.io
spec:
  ...
```

Rules (all locked — no follow-up decisions required):

- **`project` is a label** (kebab-case, selector-safe) so tasks can also be filtered with `--label-selector project=istio`.
- **`project-name` / `project-url` are annotations**, not labels — annotations are the correct home for free-form values like URLs and are never used as selectors. mcpchecker parses task metadata leniently (it already accepts the non-k8s value `requires: kubevirt,tekton`, a comma that k8s label values forbid), so these added fields are tolerated and ignored by the eval runner. _(Implementation check: run `mcpchecker check` against one annotated task to confirm; if a strict-decode error ever appears, encode the same two values as additional `labels` instead — the generator change is a one-line key lookup.)_
- **`core` is implicit.** `requires` lists only the *additional* toolsets beyond `core`, matching today's `requires: kubevirt` / `requires: kubevirt,tekton`.
- **Tasks without a `project` label are skipped** by the generator, so labels can be **backfilled incrementally** — the first PR seeds an initial set and never blocks on full coverage.

## Generator (extend the existing tool — no new GitHub Action)

Extend `internal/tools/update-readme/main.go` (already run by `make update-readme-tools` for both `README.md` and `docs/configuration.md`). Add a step that:

1. Walks `evals/tasks/` recursively for `*.yaml`. **Skip any path containing `/artifacts/`.** Parse each file with `sigs.k8s.io/yaml` (already a direct dependency) into a minimal struct:
   ```go
   type evalTask struct {
       Kind     string `json:"kind"`
       Metadata struct {
           Labels      map[string]string `json:"labels"`
           Annotations map[string]string `json:"annotations"`
       } `json:"metadata"`
   }
   ```
2. **Skip unless the root `kind` is `Task` or `Prompt`** (artifact manifests and embedded specs such as `kind: Deployment` must not be counted — filter on the *root* kind, never a substring grep).
3. Skip tasks whose `labels["project"]` is empty.
4. Aggregate per `project` key:
   - `scenarios` += 1
   - toolset set = `{"core"}` ∪ each comma-split, trimmed, non-empty entry of `labels["requires"]`
   - resolve `project-name` / `project-url` from annotations
   - **Consistency guard:** every project key must resolve to exactly one non-empty `(name, url)`; if two tasks of the same project disagree, or none supply them, `panic` so the build fails loudly (mirrors how the rest of this tool panics on bad input).
5. Sort projects alphabetically by display name (deterministic). For each, emit one table row: the **Project** cell is a markdown link `[name](url)`; the **Required toolset(s)** cell lists the toolsets as inline-code tokens, `core` first then the sorted extras; the **Validated scenarios** cell is the count. Inject the table between `<!-- VALIDATED-PROJECTS-START -->` and `<!-- VALIDATED-PROJECTS-END -->` using the existing `replaceBetweenMarkers` helper.

Because `replaceBetweenMarkers` no-ops when the markers are absent, this only touches `README.md` (markers are not added to `docs/configuration.md`). The evals path is relative to the repo root, which is the working directory `make update-readme-tools` runs from.

## CI verification

**Already covered.** The `check-readme` job in `.github/workflows/build.yaml` runs `make update-readme-tools` and fails on any `git diff` in `README.md`. Once the generator is folded in, a missing or stale Validated-Projects table fails CI automatically — no new workflow needed.

## Initial seeding (first PR)

Label the existing suites so the section isn't empty. `core` / `config` get `project` only (core is implicit); `kiali` and `helm` also need a `requires` backfill (they carry none today); `kubevirt` / `tekton` already carry `requires`:

| Existing `suite` | Add `project` | `project-name` / `project-url` | `requires` to ensure |
|---|---|---|---|
| `core`     | `kubernetes` | Kubernetes / https://kubernetes.io | _(none — core only)_ |
| `config`   | `kubernetes` | Kubernetes / https://kubernetes.io | _(none)_ |
| `kiali`    | `istio`      | Istio / https://istio.io           | `kiali` _(backfill)_ |
| `kubevirt` | `kubevirt`   | KubeVirt / https://kubevirt.io     | `kubevirt` _(present)_ |
| `tekton`   | `tekton`     | Tekton / https://tekton.dev        | `tekton` _(present)_ |
| `helm`     | `helm`       | Helm / https://helm.sh             | `helm` _(backfill)_ |

Rationale for `kiali → istio`: those scenarios operate on Istio mesh resources (VirtualServices, DestinationRules, Gateways) surfaced via the Kiali toolset, so the validated ecosystem project is Istio. The mapping is only a label and can be adjusted per task.

## Implementation checklist

1. Add the `### Validated CNCF/Kubernetes Projects` subsection + marker pair to `README.md`, before `### Available Toolsets`.
2. Extend `internal/tools/update-readme/main.go` with the evals scan + table generator described above.
3. Seed `project` labels + `project-name` / `project-url` annotations (and the `kiali` / `helm` `requires` backfill) per the table above.
4. Run `make update-readme-tools`; commit the regenerated `README.md`. `check-readme` enforces it thereafter.

Remaining `project` labels across other tasks can be added as follow-up work — the generator skips unlabeled tasks, so coverage grows incrementally.

---

<details>
<summary>Original proposal (Calum Murray)</summary>

One of the interesting developments we have seen over the last ~8 months is that the core toolset is all that is needed for many CNCF projects, rather than their own toolsets.

However, having a toolset can give visibility to a project's inclusion in this MCP server. We need some way to give visibility to projects which we have evaluated different combos of toolsets against, to highlight that this works correctly.

To do this we should:
1. Before the toolsets section in the readme add a new section to capture all the validated projects (or maybe better wording if someone has an idea)
2. Build a tool in `internal` that parses through all the eval tasks in `evals` and based on some label, generates which projects we test against (and can call supported). We would also likely need a label to correlate which toolset(s) are enabled when this eval is run. For example, Argo may work well with only the core toolset, while kubevirt will need the core + kubevirt toolsets. We need to capture that somehow
3. Add labels to each eval to support 2. IMO this can mostly be done as follow up work
4. Add a github action to verify the README contents

</details>

lunarwhite:
### Summary

Adds a new auto-generated section to README.md that surfaces which CNCF/Kubernetes ecosystem projects are continuously validated by the eval suite. The table is maintained by the same `make update-readme-tools` pipeline (and `check-readme` CI job) that already keeps the toolsets documentation in sync.

Changes:

- Seeded all current Task/Prompt YAML files with `project` label, `project-name`/`project-url` annotations, and `requires` label backfills where missing.
- Extended `internal/tools/update-readme/main.go` with a `generateValidatedProjects()` function that walks `evals/tasks/`, aggregates tasks by `project` label, and emits a markdown table between pre-defined markers.
- Updated `evals/tasks/README.md` with contributor guidance on the metadata schema.

Additional decisions undocumented in the issue:

- KubeVirt row includes `tekton` in toolsets, while the issue's illustrative table didn't account for this.
  - Fact: Two pre-existing tasks (`windows-golden-image-no-eula`, `windows-golden-image-success`) carry `requires: kubevirt,tekton`, so the generator produces `core`, `kubevirt`, `tekton`.
  - Current implementation: The generator unions all requires values across a project's tasks, surfacing `tekton`.
- Added one sentence documenting the schema in `evals/tasks/README.md` for future contributors.
- YAML parse errors would panic and fail the build loudly.

### Motivation

Resolves #1230

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,26 @@ The Kubernetes MCP server supports enabling or disabling specific groups of tool
 This allows you to control which Kubernetes functionalities are available to your AI tools.
 Enabling only the toolsets you need can help reduce the context size and improve the LLM's tool selection accuracy.
 
+### Validated Kubernetes Ecosystem Projects
+
+The following CNCF and Kubernetes ecosystem projects are covered by
+automated evaluation scenarios in [`evals/tasks`](evals/tasks). Most scenarios
+work with just the `core` toolset. The dedicated toolsets below are optional
+and only needed for the project-specific scenarios noted.
+
+<!-- VALIDATED-PROJECTS-START -->
+
+| Project | Optional toolset(s) | Eval scenarios |
+|---------|---------------------|----------------|
+| [Helm](https://helm.sh) | `helm` | 3 |
+| [Istio](https://istio.io) | `kiali` | 5 |
+| [Kiali](https://kiali.io) | `kiali` | 16 |
+| [Kubernetes](https://kubernetes.io) | - | 32 |
+| [KubeVirt](https://kubevirt.io) | `kubevirt`, `tekton` | 19 |
+| [Tekton](https://tekton.dev) | `tekton` | 9 |
+
+<!-- VALIDATED-PROJECTS-END -->
+
 ### Available Toolsets
 
 The following sets of tools are available (toolsets marked with ✓ in the Default column are enabled by default):
--- a/evals/tasks/README.md
+++ b/evals/tasks/README.md
@@ -24,7 +24,11 @@ Most subdirectories under `core/`, `config/`, `helm/`, `kiali/`, `kubevirt/`, or
 1. Pick the closest family and create a new subfolder.
 2. Author the task YAML referencing MCP tools, expected observations, and any artifacts.
 3. Provide helper scripts if the scenario needs deterministic setup or verification.
-4. Document nuances in a local `README.md` so future contributors and eval authors can replay the scenario manually.
+4. Add project metadata so the task appears in the README validated-projects table (unlabeled tasks are skipped).
+   - The toolset column comes from the `requires` label, not `project`; set `requires` when a scenario needs a non-core toolset.
+   - `project` is a label; `project-name` and `project-url` are annotations. The generator panics if `project` is set without both.
+   - Example: [helm/install-chart/install-chart.yaml](helm/install-chart/install-chart.yaml).
+5. Document nuances in a local `README.md` so future contributors and eval authors can replay the scenario manually.
 
 Well-scoped, deterministic tasks make it easier to compare agents and regressions over time, so keep inputs minimal, outputs explicit, and always clean up what you create.
 
--- a/internal/tools/update-readme/main.go
+++ b/internal/tools/update-readme/main.go
@@ -9,6 +9,8 @@ import (
 	"slices"
 	"strings"
 
+	"sigs.k8s.io/yaml"
+
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"github.com/containers/kubernetes-mcp-server/pkg/config"
 	"github.com/containers/kubernetes-mcp-server/pkg/toolsets"
@@ -31,6 +33,21 @@ func (p *FilterProvider) AnyTargetHasGVKs(_ context.Context, _ []schema.GroupVer
 
 var _ api.FilteringProvider = (*FilterProvider)(nil)
 
+type evalTask struct {
+	Kind     string `json:"kind"`
+	Metadata struct {
+		Labels      map[string]string `json:"labels"`
+		Annotations map[string]string `json:"annotations"`
+	} `json:"metadata"`
+}
+
+type projectInfo struct {
+	Name     string
+	URL      string
+	Toolsets map[string]bool
+	Count    int
+}
+
 func main() {
 	// Snyk reports false positive unless we flow the args through filepath.Clean and filepath.Localize in this specific order
 	var err error
@@ -43,6 +60,10 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
+
+	// Validated Projects
+	updated := generateValidatedProjects(string(readme))
+
 	// Available Toolsets
 	toolsetsList := toolsets.Toolsets()
 
@@ -75,8 +96,8 @@ func main() {
 		}
 		fmt.Fprintf(&availableToolsets, "| %-*s | %-*s | %-*s |\n", maxNameLen, toolset.GetName(), maxDescLen, toolset.GetDescription(), defaultHeaderLen, defaultIndicator)
 	}
-	updated := replaceBetweenMarkers(
-		string(readme),
+	updated = replaceBetweenMarkers(
+		updated,
 		"<!-- AVAILABLE-TOOLSETS-START -->",
 		"<!-- AVAILABLE-TOOLSETS-END -->",
 		availableToolsets.String(),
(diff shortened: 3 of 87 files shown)


README.md
```diff
@@ -258,6 +258,25 @@ The Kubernetes MCP server supports enabling or disabling specific groups of tool
...
+an automated evaluation suite ([`evals/tasks`](evals/tasks)). Most require only the `core`
+toolset — a dedicated toolset is not a prerequisite for compatibility.
+
+<!-- VALIDATED-PROJECTS-START -->
+
+| Project | Required toolset(s) | Validated scenarios |
+|---------|---------------------|---------------------|
+| [Helm](https://helm.sh) | `core`, `helm` | 3 |
+| [Istio](https://istio.io) | `core`, `kiali` | 5 |
+| [Kiali](https://kiali.io) | `core`, `kiali` | 16 |
```
lunarwhite: `Istio` and `Kiali` are separated into two rows as per suggestion: https://github.com/containers/kubernetes-mcp-server/issues/1230#issuecomment-4808191331, https://github.com/containers/kubernetes-mcp-server/issues/1230#issuecomment-4808837421

README.md
```diff
@@ -258,6 +258,25 @@ The Kubernetes MCP server supports enabling or disabling specific groups of tool
...
+| [Kubernetes](https://kubernetes.io) | `core` | 32 |
+| [KubeVirt](https://kubevirt.io) | `core`, `kubevirt`, `tekton` | 18 |
```
lunarwhite: This (`core`, `kubevirt`, `tekton`) diverges from the spec's expected output (`core`, `kubevirt`). Not sure if it's acceptable from a maintainer's view.

Cali0707: @lyarwood any thoughts here from the kubevirt perspective?

lunarwhite: Hi @lyarwood, do you have any preference for above question?

In short, we're adding an auto-generated `Validated CNCF/Kubernetes Projects` table in the README. In current table, `KubeVirt` row displays `core, kubevirt, tekton` in its toolsets because: 1) the generator would union all requires values across a project's tasks, while 2) two `windows-golden-image` prompt tasks carry requires: `kubevirt,tekton`.

Table preview: https://github.com/containers/kubernetes-mcp-server/pull/1235/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

Does this look good to you? Any inputs would help.

lyarwood: From a kubevirt toolset perspective I think this lgtm, @ksimon1 can you also sign off on this?

README.md
```diff
@@ -258,6 +258,25 @@ The Kubernetes MCP server supports enabling or disabling specific groups of tool
 This allows you to control which Kubernetes functionalities are available to your AI tools.
 Enabling only the toolsets you need can help reduce the context size and improve the LLM's tool selection accuracy.
 
+### Validated CNCF/Kubernetes Projects
```
manusa: Can we rename this to "Validated Kubernetes Ecosystem Projects" or similar? Tekton is a CD Foundation project and Kiali isn't a CNCF project, so "CNCF/Kubernetes Projects" isn't quite right for every row. Keeping "CNCF and Kubernetes ecosystem projects" in the body below is fine as a general umbrella.

README.md
```diff
@@ -258,6 +258,25 @@ The Kubernetes MCP server supports enabling or disabling specific groups of tool
...
+
+The following CNCF and Kubernetes ecosystem projects are continuously validated against
+an automated evaluation suite ([`evals/tasks`](evals/tasks)). Most require only the `core`
```
manusa: I'd reword this so it lands the intended message: the `core` toolset is usually all you need, and the dedicated toolsets are optional extras you add only when a specific scenario calls for them. As it reads now, "Most require only the `core` toolset ... a dedicated toolset is not a prerequisite" clashes with the table, where 5 of the 6 rows list an extra toolset. Something like: "Most scenarios work with just the `core` toolset. The dedicated toolsets below are optional and only needed for the project-specific scenarios noted." Also "continuously validated" is a bit strong for a CI and weekly run, and let's drop the em-dash on the next line.

README.md
```diff
@@ -258,6 +258,25 @@ The Kubernetes MCP server supports enabling or disabling specific groups of tool
...
+toolset — a dedicated toolset is not a prerequisite for compatibility.
+
+<!-- VALIDATED-PROJECTS-START -->
+
+| Project | Required toolset(s) | Validated scenarios |
```
manusa: Same idea as the intro: "Required toolset(s)" makes these look mandatory. Could we frame it as optional instead, either by renaming the column (something like "Optional toolset(s)" or "Toolset(s) used") or by dropping the implicit `core` and only listing the extras? That also settles the KubeVirt `tekton` question in the thread below, since tekton is only used by the two windows-golden-image scenarios, not KubeVirt in general. Minor: "Validated scenarios" is really the number of scenarios defined in the suite rather than ones currently passing, so "Eval scenarios" might read better.

internal/tools/update-readme/main.go
```diff
@@ -195,3 +216,107 @@ func replaceBetweenMarkers(content, startMarker, endMarker, replacement string)
...
+			p = &projectInfo{
+				Name:     projectName,
+				URL:      projectURL,
+				Toolsets: map[string]bool{"core": true},
+			}
+			projects[projectKey] = p
+		} else {
+			if p.Name != projectName {
+				panic(fmt.Sprintf("project %q has conflicting names: %q vs %q", projectKey, p.Name, projectName))
+			}
```
manusa, line 282: Nice catch guarding against the same key with a different name or url. One gap: two different `project` keys that resolve to the same display name aren't caught, so a typo like `kubevrit` would silently add a second "KubeVirt" row with a split count and CI would still pass. Could we also assert the display name is unique across keys?

evals/tasks/README.md
```diff
@@ -21,7 +21,8 @@ Every subdirectory under `core/`, `kiali/`, or `kubevirt/` defines a single scen
 1. Pick the closest family and create a new subfolder.
 2. Author the task YAML referencing MCP tools, expected observations, and any artifacts.
 3. Provide helper scripts if the scenario needs deterministic setup or verification.
-4. Document nuances in a local `README.md` so future contributors and eval authors can replay the scenario manually.
+4. Add `project` / `project-name` / `project-url` metadata so the task appears in the README's validated-projects table (unlabeled tasks are skipped).
+5. Document nuances in a local `README.md` so future contributors and eval authors can replay the scenario manually.
 
 Well-scoped, deterministic tasks make it easier to compare agents and regressions over time, so keep inputs minimal, outputs explicit, and always clean up what you create.
```
manusa, line 33: Two things that would help a contributor adding a brand new project here. The toolset column comes from the `requires` label, not `project`, so worth mentioning they should set `requires` when a scenario needs a non-core toolset. And it's worth noting `project` is a label while `project-name` and `project-url` are annotations, since the generator panics if the label is there without both annotations. Linking an existing task as an example would make it obvious.

manusa (changes requested):
Thanks for putting this together. The generator approach is clean and I like that it's all driven from the eval metadata with nothing hardcoded. I ran it locally and it reproduces the committed table exactly, so the readme check is happy.

Most of my feedback is about wording. The message I want the section to land is that the `core` toolset is usually all you need, and the dedicated toolsets are optional extras you only reach for when a specific scenario needs them. Right now both the intro and the "Required toolset(s)" column read as if those toolsets are hard requirements, which is the opposite of what we want. I left the specifics inline, including the heading rename and a small generator gap around duplicate project names.

One heads up that isn't inline: `evals/tasks/kubevirt/update-vm-instancetype/task.yaml` landed on main after you branched, so once you rebase it'll need the `project` label too, otherwise KubeVirt will show 18 when it should be 19.

lunarwhite:
@manusa Thanks for your review. I incorporated them after a rebase.

manusa (approved):
LGTM, thx!

README.md
ksimon1: Fine by me :+1:

PR #365, opened 2025-10-08
refactor: move provider initialization into factory

Cali0707:
This PR enables adding other `ManagerProvider` implementations more easily in downstream repos such as the `openshift-mcp-server`, without introducing any git conflicts between the two

Relates/follows up to #348
Relates to #83

PR #1239, opened 2026-06-29
feat(kubevirt): add vm_troubleshoot tool for automated VM diagnostics

sradco:
## Summary

- Adds a new `vm_troubleshoot` MCP Tool that collects comprehensive diagnostic data for KubeVirt VirtualMachines (VM/VMI status, DataVolume/PVC state, cloud-init config, virt-launcher pod state, pod logs, and events)
- **Includes automated heuristic issue detection** that produces pre-digested "Detected Issues" and "Suggested Fixes" sections, enabling small LLMs (e.g. gpt-4o-mini) to relay actionable findings without synthesizing root causes from raw YAML
- Complements the existing `vm-troubleshoot` Prompt with a proper Tool that LLMs can invoke automatically based on user intent
- Includes targeted event collection (no namespace-wide scan), sensitive field redaction for cloud-init, and correct NotFound-only PVC fallback logic

## Motivation

The existing `vm-troubleshoot` is implemented as a Prompt, which requires explicit user selection and cannot be auto-invoked by LLMs during troubleshooting conversations. This PR adds a parallel Tool implementation that:

1. **Enables proactive invocation** — LLMs can call it automatically when users describe VM issues
2. **Enriches diagnostics** — adds DataVolume/PVC status (with StorageClass) and cloud-init content extraction
3. **Improves small-model performance** — heuristic checks pre-analyze collected data and surface root causes directly, so even small models produce accurate answers
4. **Addresses security** — redacts sensitive cloud-init fields (passwords, SSH keys, tokens) while preserving diagnostic content (runcmd, packages, hostname)
5. **Improves performance** — uses targeted field selectors for events; extracts only diagnostic-relevant pod fields (phase, conditions, restartCount, nodeSelector) instead of full pod YAML
6. **Handles edge cases** — multi-line block scalar redaction, NotFound-only PVC fallback, float64-safe numeric conversion for real cluster data

### Automated Issue Detection

The tool runs heuristic checks on collected diagnostic data and produces a structured report:

```
## Detected Issues

- **CRITICAL**: StorageClass "fast-ssd" referenced by DataVolume "rootdisk" does not exist on this cluster. Available StorageClasses: ocs-storagecluster-ceph-rbd (default).
- **WARNING**: VM is stuck in Provisioning state.

## Suggested Fixes

1. Change the DataVolume storageClassName from "fast-ssd" to an existing StorageClass (e.g., ocs-storagecluster-ceph-rbd (default)).
```

Heuristic checks implemented:
- Missing StorageClass detection (lists available alternatives)
- DataVolume/PVC errors with StorageClass deduplication
- Dangerous cloud-init commands (shutdown, poweroff, halt, reboot, systemctl variants, init 0/6)
- Hostname nodeSelector migration blockers
- Pod crashloops and OOMKill detection
- Failed VirtualMachineInstanceMigration detection
- VM condition analysis (Ready=False reasons, printableStatus)

### Prompt vs. Tool coexistence

The existing `vm-troubleshoot` **Prompt** and this new `vm_troubleshoot` **Tool** serve different purposes:

* **Prompt**: Provides a guided step-by-step troubleshooting workflow template. Must be explicitly selected by the user or UI. Useful for structured walkthroughs.
* **Tool**: Returns diagnostic data with automated analysis. Can be auto-invoked by LLMs when users ask questions like "why is my VM not starting?" without requiring explicit tool selection.

Both are kept because they serve different interaction patterns. The Tool enables the agentic/autonomous troubleshooting use case that the Prompt cannot support.

## Testing

Tested manually against 3 broken VM scenarios on OpenShift 4.22:

| Scenario | Root cause | Tool provides |
|----------|-----------|---------------|
| VM stuck Provisioning | Missing StorageClass + no accessModes | CRITICAL issue with available SC list + fix suggestion |
| VM crashloop (30-60s restarts) | cloud-init runcmd: shutdown -h now | CRITICAL cloud-init detection + crashloop explanation |
| Live migration failure | nodeSelector pinning VM to single node | WARNING nodeSelector + failed migration detection |

Tested with both gpt-4o-mini (troubleshooting mode) and gpt-5.2 (simple questions) — tool is auto-invoked and produces accurate root-cause identification. Small models correctly relay the pre-analyzed findings without needing to interpret raw YAML.

## Key Design Decisions

- **float64-safe numeric conversion**: Kubernetes dynamic client deserializes JSON numbers as `float64`; the `toInt64()` helper ensures restartCount detection works on real clusters (not just in tests using `int64` directly)
- **False-positive prevention in cloud-init**: Skips comments, echo/print lines, and YAML config keys to avoid flagging innocuous text containing "shutdown" or "halt"
- **Deduplication**: When a missing StorageClass is detected, downstream PVC "not found" errors are suppressed to avoid reporting symptoms alongside root causes
- **API limiting**: Migration list capped at 50 objects to avoid unbounded queries in busy namespaces

## Related

* Coexists with the existing `vm-troubleshoot` Prompt (different interaction patterns)
* Related to #653 (ksimon1's troubleshoot action in vm_lifecycle — that adds troubleshooting as a lifecycle action; this PR provides a dedicated read-only diagnostic collector)
* Moves work from downstream `openshift/openshift-mcp-server` PR #365 to upstream per reviewer guidance

## Test plan

* Unit tests for all helper functions (fetchVMStatus, fetchVMIStatus, fetchVolumes, fetchDataVolumeStatus, extractCloudInit, fetchVirtLauncherPod, fetchVirtLauncherPodLogs)
* Unit tests for all heuristic check functions (checkVMConditions, checkStorageClass, checkDataVolumeErrors, checkCloudInitCommands, checkNodeSelector, checkPodHealth, checkMigrationStatus)
* Cloud-init redaction tests (inline values, SSH keys, block scalars, false-positive resistance, token fields)
* Cloud-init dangerous command detection tests (shutdown, halt, poweroff, systemctl variants, array format, false-positive resistance)
* Pod diagnostics extraction test (verifies env vars, images, volumeMounts are excluded)
* float64 restartCount test (simulates real cluster JSON deserialization)
* Tool registration, schema, and annotation validation tests
* Snapshot test updated
* `go build ./...` and `go test` pass locally

Signed-off-by: Shirly Radco <sradco@redhat.com>
Co-authored-by: AI Assistant <noreply@cursor.com>

### DIFF
--- a/evals/tasks/kubevirt/troubleshoot-vm-cloudinit-shutdown/task.yaml
+++ b/evals/tasks/kubevirt/troubleshoot-vm-cloudinit-shutdown/task.yaml
@@ -1,0 +1,100 @@
+kind: Task
+apiVersion: mcpchecker/v1alpha2
+metadata:
+  labels:
+    suite: kubevirt
+    requires: kubevirt
+  name: "troubleshoot-vm-cloudinit-shutdown"
+  difficulty: medium
+  description: "Diagnose a VM crashlooping due to a shutdown command in cloud-init"
+spec:
+  requires:
+    - extension: kubernetes
+      as: k8s
+  setup:
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-troubleshoot-ci
+        ignoreNotFound: true
+    - k8s.create:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-troubleshoot-ci
+    - k8s.create:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachine
+        metadata:
+          name: crashloop-vm
+          namespace: vm-test-troubleshoot-ci
+        spec:
+          runStrategy: Always
+          template:
+            spec:
+              domain:
+                devices:
+                  disks:
+                    - name: containerdisk
+                      disk:
+                        bus: virtio
+                    - name: cloudinit
+                      disk:
+                        bus: virtio
+                resources:
+                  requests:
+                    memory: 2Gi
+              terminationGracePeriodSeconds: 0
+              volumes:
+                - name: containerdisk
+                  containerDisk:
+                    image: quay.io/containerdisks/fedora:latest
+                - name: cloudinit
+                  cloudInitNoCloud:
+                    userData: |
+                      #cloud-config
+                      runcmd:
+                        - shutdown -h now
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          echo "Waiting for VM to attempt boot and crash..."
+          sleep 60
+  verify:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          NS="vm-test-troubleshoot-ci"
+
+          echo "=== Verification: Agent identified cloud-init issue ==="
+
+          if ! kubectl get virtualmachine crashloop-vm -n "$NS" > /dev/null 2>&1; then
+              echo "✗ VirtualMachine crashloop-vm no longer exists"
+              exit 1
+          fi
+          echo "✓ VirtualMachine crashloop-vm still exists"
+          exit 0
+    - llmJudge:
+        contains: "shutdown -h now"
+  cleanup:
+    - k8s.delete:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachine
+        metadata:
+          name: crashloop-vm
+          namespace: vm-test-troubleshoot-ci
+        ignoreNotFound: true
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-troubleshoot-ci
+        ignoreNotFound: true
+  prompt:
+    inline: |-
+      A VirtualMachine named "crashloop-vm" in the vm-test-troubleshoot-ci namespace keeps restarting.
+      Diagnose why this VM is crashlooping. Report:
+      - The root cause of the crashloop
+      - Which cloud-init command is causing the problem
+      - How to fix it
--- a/evals/tasks/kubevirt/troubleshoot-vm-missing-storageclass/task.yaml
+++ b/evals/tasks/kubevirt/troubleshoot-vm-missing-storageclass/task.yaml
@@ -1,0 +1,105 @@
+kind: Task
+apiVersion: mcpchecker/v1alpha2
+metadata:
+  labels:
+    suite: kubevirt
+    requires: kubevirt
+  name: "troubleshoot-vm-missing-storageclass"
+  difficulty: medium
+  description: "Diagnose a VM stuck in Provisioning due to a non-existent StorageClass"
+spec:
+  requires:
+    - extension: kubernetes
+      as: k8s
+  setup:
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-troubleshoot-sc
+        ignoreNotFound: true
+    - k8s.create:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-troubleshoot-sc
+    - k8s.create:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachine
+        metadata:
+          name: broken-sc-vm
+          namespace: vm-test-troubleshoot-sc
+        spec:
+          runStrategy: Always
+          dataVolumeTemplates:
+            - metadata:
+                name: broken-sc-vm-rootdisk
+              spec:
+                storage:
+                  storageClassName: non-existent-sc-xyz
+                  accessModes:
+                    - ReadWriteOnce
+                  resources:
+                    requests:
+                      storage: 10Gi
+                source:
+                  registry:
+                    url: docker://quay.io/containerdisks/fedora:latest
+          template:
+            spec:
+              domain:
+                devices:
+                  disks:
+                    - name: rootdisk
+                      disk:
+                        bus: virtio
+                resources:
+                  requests:
+                    memory: 2Gi
+              terminationGracePeriodSeconds: 0
+              volumes:
+                - name: rootdisk
+                  dataVolume:
+                    name: broken-sc-vm-rootdisk
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          echo "Waiting for VM to enter Provisioning state..."
+          sleep 15
+  verify:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          NS="vm-test-troubleshoot-sc"
+
+          echo "=== Verification: Agent identified missing StorageClass ==="
+
+          if ! kubectl get virtualmachine broken-sc-vm -n "$NS" > /dev/null 2>&1; then
+              echo "✗ VirtualMachine broken-sc-vm no longer exists"
+              exit 1
+          fi
+          echo "✓ VirtualMachine broken-sc-vm still exists"
+          exit 0
+    - llmJudge:
+        contains: "non-existent-sc-xyz"
+  cleanup:
+    - k8s.delete:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachine
+        metadata:
+          name: broken-sc-vm
+          namespace: vm-test-troubleshoot-sc
+        ignoreNotFound: true
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-troubleshoot-sc
+        ignoreNotFound: true
+  prompt:
+    inline: |-
+      A VirtualMachine named "broken-sc-vm" in the vm-test-troubleshoot-sc namespace is not starting.
+      Diagnose why this VM is stuck. Report:
+      - The root cause of the issue
+      - What StorageClass is missing
+      - What alternative StorageClasses are available on the cluster
--- a/pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
+++ b/pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
@@ -1,0 +1,656 @@
+package troubleshoot
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/kubevirt"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/dynamic"
+)
+
+type severity int
+
+const (
+	severityCritical severity = iota
+	severityWarning
+	severityInfo
+)
+
+func (s severity) String() string {
+	switch s {
+	case severityCritical:
+		return "CRITICAL"
+	case severityWarning:
+		return "WARNING"
+	default:
+		return "INFO"
+	}
+}
+
+type issue struct {
+	Severity severity
+	Message  string
+	Fix      string
+}
+
+func analyzeIssues(ctx context.Context, dynamicClient dynamic.Interface, namespace, name string, vm, vmi *unstructured.Unstructured, podDiagnostics []map[string]interface{}) string {
+	var issues []issue
+
+	issues = append(issues, checkVMConditions(vm)...)
+
+	scIssues := checkStorageClass(ctx, dynamicClient, vm)
+	issues = append(issues, scIssues...)
+	issues = append(issues, checkDataVolumeErrors(vm, scIssues)...)
+
+	issues = append(issues, checkCloudInitCommands(vm, vmi)...)
+	issues = append(issues, checkNodeSelector(vm, vmi)...)
+	issues = append(issues, checkPodHealth(podDiagnostics)...)
+	issues = append(issues, checkMigrationStatus(ctx, dynamicClient, namespace, name)...)
+
+	if len(issues) == 0 {
+		return "## Detected Issues\n\n*No issues automatically detected. Review the raw diagnostic data below for manual analysis.*"
+	}
+
+	var result strings.Builder
+	result.WriteString("## Detected Issues\n\n")
+	for _, iss := range issues {
+		fmt.Fprintf(&result, "- **%s**: %s\n", iss.Severity, iss.Message)
+	}
+
+	var fixes []string
+	for _, iss := range issues {
+		if iss.Fix != "" {
+			fixes = append(fixes, iss.Fix)
+		}
+	}
+	if len(fixes) > 0 {
+		result.WriteString("\n## Suggested Fixes\n\n")
+		for i, fix := range fixes {
+			fmt.Fprintf(&result, "%d. %s\n", i+1, fix)
+		}
+	}
+
+	return result.String()
+}
+
+func checkVMConditions(vm *unstructured.Unstructured) []issue {
+	if vm == nil {
+		return nil
+	}
+
+	var issues []issue
+
+	conditions, _, _ := unstructured.NestedSlice(vm.Object, "status", "conditions")
+	for _, cond := range conditions {
+		condMap, ok := cond.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		condType, _ := condMap["type"].(string)
+		status, _ := condMap["status"].(string)
+		reason, _ := condMap["reason"].(string)
+		message, _ := condMap["message"].(string)
+
+		if condType != "Ready" || status != "False" {
+			continue
+		}
+
+		switch reason {
+		case "VMINotExists":
+			issues = append(issues, issue{
+				Severity: severityWarning,
+				Message:  "VM condition Ready=False with reason VMINotExists. The VirtualMachineInstance has not been created.",
+			})
+		default:
+			if message != "" && (strings.Contains(message, "Guest") || strings.Contains(strings.ToLower(message), "error")) {
+				issues = append(issues, issue{
+					Severity: severityWarning,
+					Message:  fmt.Sprintf("VM condition Ready=False: %s", message),
+				})
+			}
+		}
+	}
+
+	printableStatus, _, _ := unstructured.NestedString(vm.Object, "status", "printableStatus")
+	switch printableStatus {
+	case "Provisioning":
+		issues = append(issues, issue{
+			Severity: severityWarning,
+			Message:  "VM is stuck in Provisioning state. This typically means a DataVolume or PVC cannot be created.",
+		})
+	case "CrashLoopBackOff":
+		issues = append(issues, issue{
+			Severity: severityCritical,
+			Message:  "VM is in CrashLoopBackOff state. The guest OS or cloud-init is causing repeated failures.",
+			Fix:      "Check the cloud-init userData for commands that shut down or crash the guest (e.g., shutdown -h now, exit 1).",
+		})
+	case "ErrImagePull", "ImagePullBackOff":
+		issues = append(issues, issue{
+			Severity: severityCritical,
+			Message:  fmt.Sprintf("VM is in %s state. The container disk image cannot be pulled.", printableStatus),
+			Fix:      "Verify the containerDisk image reference exists and is accessible from the cluster.",
+		})
+	}
+
+	return issues
+}
+
+func checkDataVolumeErrors(vm *unstructured.Unstructured, scIssues []issue) []issue {
+	if vm == nil {
+		return nil
+	}
+
+	hasMissingSC := false
+	for _, sci := range scIssues {
+		if sci.Severity == severityCritical {
+			hasMissingSC = true
+			break
+		}
+	}
+
+	var issues []issue
+
+	volumeStatuses, _, _ := unstructured.NestedSlice(vm.Object, "status", "volumeSnapshotStatuses")
+	for _, vs := range volumeStatuses {
+		vsMap, ok := vs.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		volName, _ := vsMap["name"].(string)
+		reason, _ := vsMap["reason"].(string)
+		if reason != "" && strings.Contains(reason, "not found") {
+			if hasMissingSC {
+				continue
+			}
+			issues = append(issues, issue{
+				Severity: severityCritical,
+				Message:  fmt.Sprintf("Volume %q: PVC not found. The backing storage does not exist.", volName),
+				Fix:      fmt.Sprintf("Ensure the PVC or DataVolume for volume %q is created and in Bound state.", volName),
+			})
+		}
+	}
+
+	return issues
+}
+
+func checkStorageClass(ctx context.Context, dynamicClient dynamic.Interface, vm *unstructured.Unstructured) []issue {
+	if vm == nil || dynamicClient == nil {
+		return nil
+	}
+
+	dvTemplates, found, err := unstructured.NestedSlice(vm.Object, "spec", "dataVolumeTemplates")
+	if err != nil || !found || len(dvTemplates) == 0 {
+		return nil
+	}
+
+	var issues []issue
+	for _, dvTemplate := range dvTemplates {
+		dvMap, ok := dvTemplate.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		metadata, _ := dvMap["metadata"].(map[string]interface{})
+		dvName, _ := metadata["name"].(string)
+
+		spec, _ := dvMap["spec"].(map[string]interface{})
+		if spec == nil {
+			continue
+		}
+
+		scName := extractStorageClassName(spec)
+		if scName == "" {
+			continue
+		}
+
+		_, err := dynamicClient.Resource(kubevirt.StorageClassGVR).Get(ctx, scName, metav1.GetOptions{})
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				issues = append(issues, issue{
+					Severity: severityWarning,
+					Message:  fmt.Sprintf("Unable to verify StorageClass %q for DataVolume %q: %v", scName, dvName, err),
+				})
+				continue
+			}
+			availableSCs := listAvailableStorageClasses(ctx, dynamicClient)
+			issues = append(issues, issue{
+				Severity: severityCritical,
+				Message:  fmt.Sprintf("StorageClass %q referenced by DataVolume %q does not exist on this cluster. Available StorageClasses: %s.", scName, dvName, availableSCs),
+				Fix:      fmt.Sprintf("Change the DataVolume storageClassName from %q to an existing StorageClass (e.g., %s).", scName, availableSCs),
+			})
+		}
+	}
+
+	return issues
+}
+
+func extractStorageClassName(dvSpec map[string]interface{}) string {
+	if storage, ok := dvSpec["storage"].(map[string]interface{}); ok {
+		if sc, ok := storage["storageClassName"].(string); ok && sc != "" {
+			return sc
+		}
+	}
+	if pvc, ok := dvSpec["pvc"].(map[string]interface{}); ok {
+		if sc, ok := pvc["storageClassName"].(string); ok && sc != "" {
+			return sc
+		}
+	}
+	return ""
+}
+
+func listAvailableStorageClasses(ctx context.Context, dynamicClient dynamic.Interface) string {
+	scList, err := dynamicClient.Resource(kubevirt.StorageClassGVR).List(ctx, metav1.ListOptions{})
+	if err != nil || len(scList.Items) == 0 {
+		return "(none found)"
+	}
+
+	var names []string
+	for _, sc := range scList.Items {
+		name := sc.GetName()
+		annotations := sc.GetAnnotations()
+		if annotations["storageclass.kubernetes.io/is-default-class"] == "true" {
+			name += " (default)"
+		}
+		names = append(names, name)
+	}
+	return strings.Join(names, ", ")
+}
+
+func checkCloudInitCommands(vm, vmi *unstructured.Unstructured) []issue {
+	var volumes []interface{}
+	if vm != nil {
+		volumes, _, _ = unstructured.NestedSlice(vm.Object, "spec", "template", "spec", "volumes")
+	}
+	if len(volumes) == 0 && vmi != nil {
+		volumes, _, _ = unstructured.NestedSlice(vmi.Object, "spec", "volumes")
+	}
+	if len(volumes) == 0 {
+		return nil
+	}
+
+	runStrategy := ""
+	if vm != nil {
+		runStrategy, _, _ = unstructured.NestedString(vm.Object, "spec", "runStrategy")
+	}
+
+	var issues []issue
+
+	for _, vol := range volumes {
+		volMap, ok := vol.(map[string]interface{})
+		if !ok {
+			continue
+		}
+
+		for _, ciKey := range []string{"cloudInitNoCloud", "cloudInitConfigDrive"} {
+			ciData, exists := volMap[ciKey].(map[string]interface{})
+			if !exists {
+				continue
+			}
+
+			userData, _ := ciData["userData"].(string)
+			if userData == "" {
+				continue
+			}
+
+			for _, cmd := range detectDangerousCloudInitCommands(userData) {
+				msg := fmt.Sprintf("Cloud-init userData contains %q command that will shut down or restart the guest immediately after boot.", cmd)
+				if runStrategy == "Always" || runStrategy == "" {
+					msg += " Combined with runStrategy Always, this causes a crashloop."
+				}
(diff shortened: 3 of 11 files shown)


pkg/toolsets/kubevirt/vm/troubleshoot/tool.go
```diff
@@ -0,0 +1,540 @@
...
+				continue
+			}
+			fmt.Fprintf(&result, "### %s\n\n*DataVolume not found*\n\n", dvName)
+			pvc, pvcErr := dynamicClient.Resource(kubevirt.PersistentVolumeClaimGVR).Namespace(namespace).Get(ctx, dvName, metav1.GetOptions{})
+			if pvcErr != nil {
+				fmt.Fprintf(&result, "*PVC also not found: %v*\n\n", pvcErr)
+			} else {
+				pvcStatus, _, _ := unstructured.NestedMap(pvc.Object, "status")
+				if pvcStatus != nil {
+					yamlStr, _ := output.MarshalYaml(pvcStatus)
```
ksimon1: I think here we should check it the fn returned error

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
```diff
@@ -0,0 +1,551 @@
...
+		if scName == "" {
+			continue
+		}
+
+		_, err := dynamicClient.Resource(kubevirt.StorageClassGVR).Get(ctx, scName, metav1.GetOptions{})
+		if err != nil {
+			availableSCs := listAvailableStorageClasses(ctx, dynamicClient)
+			issues = append(issues, issue{
+				Severity: severityCritical,
+				Message:  fmt.Sprintf("StorageClass %q referenced by DataVolume %q does not exist on this cluster. Available StorageClasses: %s.", scName, dvName, availableSCs),
```
ksimon1, line 223: I think this is incorrect, as the error doesn't have to be only that SC is not found

pkg/toolsets/kubevirt/vm/troubleshoot/tool.go
```diff
@@ -0,0 +1,540 @@
...
+	objectNames := []string{vmName}
+	objectNames = append(objectNames, podNames...)
+
+	var relatedEvents []map[string]any
+	for _, objName := range objectNames {
+		events, err := core.EventsList(ctx, namespace, api.ListOptions{
+			ListOptions: metav1.ListOptions{FieldSelector: "involvedObject.name=" + objName},
+		})
+		if err != nil {
+			continue
```
ksimon1, line 357: Do we want to log here any error?

pkg/toolsets/kubevirt/vm/troubleshoot/tool.go
```diff
@@ -0,0 +1,540 @@
...
+	}
+
+	var result strings.Builder
+	var podNames []string
+	var latestDiag map[string]interface{}
+	result.WriteString("## virt-launcher Pod\n\n")
+	for _, pod := range podList.Items {
+		podNames = append(podNames, pod.GetName())
+		summary := extractPodDiagnostics(&pod)
+		latestDiag = summary
```
ksimon1: This will return latestDiag only for the last pod

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
```diff
@@ -0,0 +1,551 @@
...
+		stripped == "systemctl reboot":
+		return "reboot"
+	case stripped == "init 0" || stripped == "/sbin/init 0":
+		return "init 0"
+	case stripped == "init 6" || stripped == "/sbin/init 6":
+		return "init 6"
+	}
+
+	if strings.Contains(lower, "[") {
+		if strings.Contains(lower, "\"shutdown\"") || strings.Contains(lower, "'shutdown'") {
```
ksimon1: this might miss case with `[shutdown, -h, now]`

lyarwood (commented):
Thanks @sradco, I'll review in more detail tomorrow but initially it looks  promising. We do however mandate that each new tool is accompanied by one or preferably more evals under https://github.com/containers/kubernetes-mcp-server/tree/main/evals/tasks/kubevirt in seperate commits that both show the need for this tooling and exercise it in a way that can be asserted over time to ensure correctness. Would you mind having a look into these next week, happy to help get you up and running if you need help.

pkg/toolsets/kubevirt/vm/troubleshoot/tool.go
sradco: fixed. Now checks the error and reports it in the output instead of silently producing empty YAML.

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
sradco: +1. Fixed by checking apierrors.IsNotFound(err) explicitly. Non-NotFound errors now produce a WARNING with the actual error message instead of assuming the SC is missing.

pkg/toolsets/kubevirt/vm/troubleshoot/tool.go
sradco: +1. Event fetch errors are now collected and displayed in the Events section so they're visible in the report, rather than silently continuing.

pkg/toolsets/kubevirt/vm/troubleshoot/tool.go
sradco: +1. I changed from a single `map[string]interface{}` to `[]map[string]interface{}`. All pod diagnostics are now collected and analyzed by `checkPodHealth`, not just the last one.

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
sradco: +1. I replaced the `Contains`-based matching with `extractFirstArrayElement()` that properly parses the first element of the YAML array. 
Also handles the no-space variant `[systemctl,reboot]` and avoids false positives when dangerous words appear as arguments (e.g., [logger, halt requested by user]).

sradco:
> Thanks @sradco, I'll review in more detail tomorrow but initially it looks promising. We do however mandate that each new tool is accompanied by one or preferably more evals under https://github.com/containers/kubernetes-mcp-server/tree/main/evals/tasks/kubevirt in seperate commits that both show the need for this tooling and exercise it in a way that can be asserted over time to ensure correctness. Would you mind having a look into these next week, happy to help get you up and running if you need help.

Added 2 evals

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
```diff
@@ -0,0 +1,602 @@
...
+
+	vmimList, err := dynamicClient.Resource(kubevirt.VirtualMachineInstanceMigrationGVR).Namespace(namespace).List(ctx, metav1.ListOptions{
+		Limit: 50,
+	})
+	if err != nil || len(vmimList.Items) == 0 {
+		return nil
+	}
+
+	var issues []issue
+	for _, vmim := range vmimList.Items {
```
ksimon1, line 556: Would it make sense to filter out old migrations (e.g. older than 2 days)?

sradco: Instead of a hard time cutoff (which is somewhat arbitrary), I now report only the 3 most recent failed migrations sorted by creation time, with age included in the message (e.g., "failed (3h ago)"). 
If more exist, an INFO note says how many older ones were skipped. This keeps output focused without losing potentially relevant failures.

pkg/toolsets/kubevirt/vm/troubleshoot/tool.go
```diff
@@ -0,0 +1,559 @@
+package troubleshoot
```
lyarwood, line 1: **Suggestion: Extract shared data-fetching logic with the existing prompt handler**

This file reimplements `fetchVMStatus`, `fetchVMIStatus`, `fetchVolumes`, `fetchVirtLauncherPod`, `fetchVirtLauncherPodLogs`, and `fetchEvents` — all of which have near-identical versions in `pkg/toolsets/kubevirt/vm_troubleshoot.go` (the prompt handler).

The tool versions are enhanced (DataVolume/PVC status, cloud-init extraction with redaction, pod diagnostics summary, targeted events), but the base fetching logic is the same. As these evolve independently, bug fixes in one will need manual replication in the other.

Consider extracting the shared data-fetching logic into a common internal package (e.g., `pkg/toolsets/kubevirt/internal/diagnostics/`) that both the prompt and tool import, with the tool adding its analysis layer on top. This could be a follow-up PR if you'd prefer to keep this one focused.

lyarwood (commented):
Review of the vm_troubleshoot tool implementation. All prior review comments from @ksimon1 have been addressed in the current code. Two eval tasks added as requested. Overall the code is solid — main concern is maintainability around prompt/tool duplication.

**Summary**: The heuristic analysis architecture is well-designed, cloud-init redaction is thorough, and the float64-safe numeric handling shows good attention to real-cluster behavior. See inline comments for specific suggestions.

pkg/toolsets/kubevirt/vm/troubleshoot/tool.go
```diff
@@ -0,0 +1,559 @@
...
+					ReadOnlyHint:    ptr.To(true),
+					DestructiveHint: ptr.To(false),
+					IdempotentHint:  ptr.To(true),
+					OpenWorldHint:   ptr.To(true),
+				},
+			},
+			Handler: troubleshoot,
+		},
+	}
+}
```
lyarwood, line 69: **Suggestion: Add an MCP-level integration test**

`pkg/mcp/kubevirt_test.go` tests the `vm-troubleshoot` *prompt* at the MCP layer but doesn't have a `CallTool("vm_troubleshoot", ...)` test for the new *tool*. An MCP-level test would exercise the full handler path (parameter extraction via the MCP server, tool registration, report assembly) and match the testing pattern used by other kubevirt tools in that file.

The unit tests here are thorough for the internal logic, but an integration test would catch wiring issues (e.g., if the tool handler signature or `ToolHandlerParams` fields change upstream).

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
```diff
@@ -0,0 +1,602 @@
...
+
+	return issues
+}
+
+func checkMigrationStatus(ctx context.Context, dynamicClient dynamic.Interface, namespace, name string) []issue {
+	if dynamicClient == nil {
+		return nil
+	}
+
+	vmimList, err := dynamicClient.Resource(kubevirt.VirtualMachineInstanceMigrationGVR).Namespace(namespace).List(ctx, metav1.ListOptions{
```
lyarwood, line 548: **Suggestion: Consider label-based filtering for migrations**

This lists up to 50 migrations namespace-wide and filters by `vmiName` client-side. In namespaces with heavy migration activity, the relevant migration could fall outside the first 50 results.

If KubeVirt's VMIM resources carry a label like `kubevirt.io/migration-target-vmi=<name>` (or similar), a `LabelSelector` would be more efficient and avoid the cap concern. Worth checking the KubeVirt API — if no such label exists, the current approach with `Limit: 50` is a reasonable pragmatic bound.

pkg/toolsets/kubevirt/vm/troubleshoot/tool.go
sradco: +1. Agreed — the duplication is a maintenance concern. The tool versions are enhanced (DataVolume/PVC status, cloud-init extraction with redaction, pod diagnostics summary, targeted events per object), but the base fetching logic is shared.

I'll open a follow-up PR to extract the common data-fetching logic into `pkg/toolsets/kubevirt/internal/diagnostics/` that both the prompt and tool import, with the tool adding its analysis layer on top. This keeps the current PR focused on the new tool functionality.

pkg/toolsets/kubevirt/vm/troubleshoot/tool.go
sradco: +1. Added a `TestVMTroubleshoot` MCP-level integration test in `pkg/mcp/kubevirt_test.go` that exercises the full handler path through `CallTool("vm_troubleshoot", ...)`. It covers:

1. **Missing required params** — verifies `namespace` and `name` are required (matches the pattern used by other kubevirt tools)
2. **Existing VM** — creates a VM with a containerDisk volume, calls the tool, verifies the diagnostic report header, section structure, and VM name/namespace presence
3. **Non-existent VM** — verifies the tool handles gracefully (returns report with error info, no `IsError`)

This catches wiring issues like `ToolHandlerParams` field changes or parameter extraction regressions.

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
sradco: **Correction to my earlier reply** — I was wrong. The `kubevirt.io/vmi-name` label (`MigrationSelectorLabel`) is still present on VMIM resources. [PR #17382](https://github.com/kubevirt/kubevirt/pull/17382) proposed removing it but was closed without merging.

I'll switch `checkMigrationStatus` to use a `LabelSelector` with `kubevirt.io/vmi-name=<name>` for server-side filtering. This removes the `Limit: 50` cap concern entirely and is more efficient.

sradco: fixed

evals/tasks/kubevirt/troubleshoot-vm-cloudinit-shutdown/task.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+          echo "=== Verification: Agent used vm_troubleshoot and identified cloud-init issue ==="
+
+          if ! kubectl get virtualmachine crashloop-vm -n "$NS" > /dev/null 2>&1; then
+              echo "✗ VirtualMachine crashloop-vm no longer exists"
+              exit 1
+          fi
+          echo "✓ VirtualMachine crashloop-vm still exists"
+          exit 0
+    - llmJudge:
```
lyarwood, line 78: nit: `contains: "shutdown"` is a fairly weak check — the agent could pass by saying something generic like "the VM appears to shut down" without actually diagnosing the cloud-init cause.

Consider tightening to `contains: "shutdown -h now"` or adding a second `contains` for `"cloud-init"` to confirm the agent identified the specific command and its source.

For comparison, the missing-storageclass task's `contains: "non-existent-sc-xyz"` is much stronger since it requires the exact StorageClass name.

evals/tasks/kubevirt/troubleshoot-vm-cloudinit-shutdown/task.yaml
```diff
@@ -0,0 +1,100 @@
...
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachine
+        metadata:
+          name: crashloop-vm
+          namespace: vm-test-troubleshoot-ci
+        ignoreNotFound: true
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
```
lyarwood, line 91: The prompt explicitly names the tool to use (`Use the vm_troubleshoot tool`). This is an anti-pattern for evals — it tests whether the agent can follow an instruction rather than whether it can select the right tool for the job.

Consider rephrasing to present the problem and let the agent figure out which tools to use, e.g.:

```
A VirtualMachine named "crashloop-vm" in the vm-test-troubleshoot-ci namespace keeps restarting.
Diagnose why this VM is crashlooping. Report:
- The root cause of the crashloop
- Which cloud-init command is causing the problem
- How to fix it
```

evals/tasks/kubevirt/troubleshoot-vm-missing-storageclass/task.yaml
```diff
@@ -0,0 +1,105 @@
...
+              echo "✗ VirtualMachine broken-sc-vm no longer exists"
+              exit 1
+          fi
+          echo "✓ VirtualMachine broken-sc-vm still exists"
+          exit 0
+    - llmJudge:
+        contains: "non-existent-sc-xyz"
+  cleanup:
+    - k8s.delete:
+        apiVersion: kubevirt.io/v1
```
lyarwood, line 87: Same comment as the cloud-init task — the prompt explicitly names `vm_troubleshoot` which makes the eval test instruction-following rather than tool selection. Consider removing the tool name and letting the agent discover the right tool on its own, e.g.:

```
A VirtualMachine named "broken-sc-vm" in the vm-test-troubleshoot-sc namespace is not starting.
Diagnose why this VM is stuck. Report:
- The root cause of the issue
- What StorageClass is missing
- What alternative StorageClasses are available on the cluster
```

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
```diff
@@ -0,0 +1,649 @@
...
+	}
+	return false
+}
+
+func isConfigKeyLine(line string) bool {
+	lower := strings.ToLower(line)
+	if strings.Contains(lower, ":") && !strings.HasPrefix(lower, "-") && !strings.HasPrefix(lower, "[") {
+		parts := strings.SplitN(lower, ":", 2)
+		key := strings.TrimSpace(parts[0])
+		if strings.Contains(key, "_") || strings.Contains(key, "-") {
```
lyarwood: **Bug: `isConfigKeyLine` misses YAML keys without underscores or hyphens**

This function only returns `true` if the key part of a YAML line contains `_` or `-`. Keys like `hostname`, `packages`, or `bootcmd` would not be recognized, so a line like:

```yaml
hostname: shutdown
```

would fall through to `matchDangerousCommand` and trigger a false positive (the stripped value `shutdown` matches exactly).

The intent is to skip lines that are YAML key-value pairs (not shell commands), but the current heuristic is too narrow. Consider using a broader check — e.g., if the line has the form `word: value` and doesn't start with `- `, it's likely a config key regardless of whether the key name contains `_` or `-`.

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
```diff
@@ -0,0 +1,649 @@
...
+	trimmed := strings.TrimSpace(line)
+	lower := strings.ToLower(trimmed)
+
+	stripped := strings.TrimPrefix(lower, "- ")
+	stripped = strings.Trim(stripped, "[]\"' ")
+
+	switch {
+	case stripped == "shutdown" || stripped == "shutdown -h now" ||
+		stripped == "/sbin/shutdown" || strings.HasPrefix(stripped, "/sbin/shutdown ") ||
+		strings.HasPrefix(stripped, "shutdown -"):
```
lyarwood: nit: `strings.HasPrefix(stripped, "shutdown -")` is quite broad — it matches `shutdown -v`, `shutdown --help`, `shutdown -c` (cancel a pending shutdown), etc. Only `-h`, `-P`, and `-r` (and bare `shutdown`) are actually dangerous power-off/reboot variants.

Not a high-priority fix since false positives here are relatively harmless (they'd surface a warning about a non-dangerous shutdown invocation), but worth noting for future refinement.

lyarwood (commented):
## Review: feat(kubevirt): add vm_troubleshoot tool for automated VM diagnostics

### Summary

This PR adds a well-designed `vm_troubleshoot` MCP tool with heuristic analysis for automated VM diagnostics, plus two eval tasks. The core implementation (analysis engine, cloud-init redaction, diagnostic data collection) is solid. All prior review feedback from @ksimon1 has been addressed, the MCP-level integration test was added as requested, and the migration label selector fix is in place.

### Positive Observations

- **Heuristic analysis architecture** is clean — `analyzeIssues` orchestrates individual checkers that each return typed `issue` structs with severity + fix suggestions. Easy to extend with new checks.
- **Cloud-init redaction** is thorough — handles inline values, block scalars, SSH key patterns, and token fields while preserving diagnostic content.
- **float64-safe numeric conversion** via `toInt64()` shows good attention to real-cluster behavior where JSON numbers deserialize as float64.
- **Deduplication** between StorageClass and DataVolume errors avoids reporting symptoms alongside root causes.
- **Pod diagnostics extraction** strips env vars, images, and volumeMounts to avoid leaking sensitive data.
- **Test coverage** is comprehensive — unit tests for all heuristic checks, redaction, false-positive resistance, and an MCP integration test.

### Issues

See inline comments. Summary:

1. **`isConfigKeyLine` false-negative** (analysis.go:349) — YAML keys without underscores or hyphens (e.g. `hostname: shutdown`) are not recognized as config keys, which could cause a false positive in dangerous command detection.
2. **`matchDangerousCommand` over-broad prefix match** (analysis.go:371) — `shutdown -v` or `shutdown --help` would match as dangerous when they're not power-off commands.
3. **Eval prompts name the tool explicitly** (already commented) — anti-pattern for evals.
4. **Eval llmJudge `contains: "shutdown"` is weak** (already commented) — should be tightened.

### Existing Discussion Summary

- @ksimon1's feedback (error handling, array format parsing, multi-pod diagnostics, migration filtering) — all addressed in current code.
- @lyarwood's suggestions (shared data-fetching extraction, MCP integration test, label-based migration filtering) — integration test added, migration label selector fixed, data-fetching extraction deferred to follow-up PR.

### Verdict

Close to ready. The two eval issues (already commented) and the `isConfigKeyLine` false-negative are the main items to address. The `shutdown -` prefix match is minor.

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
ksimon1: @sradco this is not fixed yet

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
ksimon1: @sradco this is not fixed yet

sradco: +1. Fixed by broadening `isConfigKeyLine` to recognize any single-word YAML key (not just those containing `_` or `-`). The check now treats any line matching the pattern `word: value` (that does not start with `- ` or `[`) as a config key. This handles `hostname: shutdown`, `packages: [halt]`, `bootcmd: []`, etc. without false positives.

Added `hostname: shutdown` and `packages: [halt]` to the existing false-positive test to cover this case.

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
sradco: +1. Tightened the shutdown flag matching to only match actually dangerous flags: `-h` (halt), `-P` (poweroff), and `-r` (reboot). Non-dangerous flags like `-v`, `--help`, and `-c` (cancel) no longer trigger a warning.

Added a test confirming `shutdown -c` does not produce a false positive.

evals/tasks/kubevirt/troubleshoot-vm-cloudinit-shutdown/task.yaml
sradco: fixed

evals/tasks/kubevirt/troubleshoot-vm-cloudinit-shutdown/task.yaml
sradco: fixed

evals/tasks/kubevirt/troubleshoot-vm-missing-storageclass/task.yaml
sradco: fixed

Cali0707:
@sradco can you run `make update-readme-tools`

sradco:
@Cali0707, @lyarwood, @ksimon1 I run `make update-readme-tools`, PR is ready.

lyarwood (approved):
All review feedback has been addressed:

- `isConfigKeyLine` broadened to recognize any single-word YAML key — fixes the `hostname: shutdown` false positive
- `matchDangerousCommand` tightened to only match `-h`, `-P`, `-r` flags — `shutdown -c` no longer triggers
- Eval `llmJudge` tightened from `contains: "shutdown"` to `contains: "shutdown -h now"`
- Eval prompts no longer name `vm_troubleshoot` — both present the problem and let the agent choose the tool
- New test cases added for both fixes

LGTM — nice work on the heuristic analysis engine and the thorough test coverage.

lyarwood (approved):

sradco:
@Cali0707 Hi, can we merge this pr?

Cali0707 (approved):
LGTM

lyarwood (approved):

sradco:
The code needed a rebase.

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
```diff
@@ -0,0 +1,651 @@
...
+		if strings.HasPrefix(trimmed, "#") {
+			continue
+		}
+		if isEchoOrPrintLine(trimmed) {
+			continue
+		}
+		if isConfigKeyLine(trimmed) {
+			continue
+		}
+		if cmd := matchDangerousCommand(trimmed); cmd != "" {
```
Cali0707: Do we want to only surface the first match, or is it worth accumulating all dangerous matches?

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
```diff
@@ -0,0 +1,651 @@
...
+
+	stripped := strings.TrimPrefix(lower, "- ")
+	stripped = strings.Trim(stripped, "[]\"' ")
+
+	switch {
+	case stripped == "shutdown" || stripped == "shutdown -h now" ||
+		stripped == "/sbin/shutdown" || strings.HasPrefix(stripped, "/sbin/shutdown ") ||
+		strings.HasPrefix(stripped, "shutdown -h") ||
+		strings.HasPrefix(stripped, "shutdown -P") || strings.HasPrefix(stripped, "shutdown -p") ||
+		strings.HasPrefix(stripped, "shutdown -r"):
```
Cali0707: Isn't `shutdown -r` a reboot?

pkg/toolsets/kubevirt/vm/troubleshoot/tool.go
```diff
@@ -0,0 +1,562 @@
...
+
+	name, err := api.RequiredString(params, "name")
+	if err != nil {
+		return api.NewToolCallResult("", err), nil
+	}
+
+	ctx := params.Context
+	if ctx == nil {
+		ctx = context.Background()
+	}
```
Cali0707: no need for this nil check - this should be reliably populated by the core of the mcp server

pkg/toolsets/kubevirt/vm/troubleshoot/tool_test.go
```diff
@@ -0,0 +1,773 @@
...
+
+	s.Run("description mentions priority and scope", func() {
+		tools := Tools(&fakeProvider{})
+		desc := tools[0].Tool.Description
+		s.Contains(desc, "FIRST")
+		s.Contains(desc, "root-cause")
+		s.Contains(desc, "StorageClasses")
+		s.Contains(desc, "cloud-init")
+	})
+}
```
Cali0707: Generally speaking we do this verification through the pkg/mcp/testdata/<name>.json, and by checking once that that is what we expect. Any future changes would fail those tests.

pkg/toolsets/kubevirt/vm/troubleshoot/tool_test.go
```diff
@@ -0,0 +1,773 @@
...
+
+func (s *TroubleshootToolSuite) TestFetchVMStatus() {
```
Cali0707: Maybe for these tests we can just test the public API of the tools in pkg/mcp/kubevirt_test.go ?

We generally try to prefer testing the public api of the tools

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
sradco: Updated

pkg/toolsets/kubevirt/vm/troubleshoot/analysis.go
sradco: Updated

pkg/toolsets/kubevirt/vm/troubleshoot/tool.go
sradco: Updated

pkg/toolsets/kubevirt/vm/troubleshoot/tool_test.go
sradco: Updated

pkg/toolsets/kubevirt/vm/troubleshoot/tool_test.go
sradco: Updated

Cali0707:
@lyarwood @ksimon1 @sradco is the plan to keep the existing troubleshoot prompt as well as the tool?

If yes, can we try and consolidate the logic between the two to a package function of some kind that they reuse? IMO this implementation seems better than the one in the prompt, but we do have essentially two versions of really similar logic (especially with fetching the data like `fetchVirtLauncherPodLogs`, `fetchVirtLauncherPod`, etc.)

sradco:
The tool subsumes the prompt.
It collects the same data plus DataVolumes/cloud-init/StorageClass, adds automated heuristic checks, and can be auto-invoked by the LLM without explicit user selection.
Critically, the pre-digested "Detected Issues" output works significantly better with smaller models (tested with gpt-4o-mini) that struggle to synthesize root causes from the raw YAML the prompt dumps.

Rather than consolidating into a shared package, would it make sense to deprecate the prompt in a follow-up PR once this merges? 
@ksimon1 since you authored the original prompt, what's your take?

manusa (commented):
Thanks for working on this.

The format check is failing (pkg/toolsets/kubevirt/vm/troubleshoot/analysis_test.go:593)

My coding agent captured the following as a critical issue that could prevent merging the PR as is:

- Cloud-init secrets (passwords, SSH keys, tokens) are dumped unredacted in the "## Volumes" section, the redaction feature is fully bypassed.
See `pkg/toolsets/kubevirt/vm/troubleshoot/tool.go:184` (fetchVolumes, wired at tool.go:80, emitted at tool.go:108)
- `pkg/toolsets/kubevirt/vm/troubleshoot/tool.go:479-549 `
  It's a line-prefix/substring denylist over a hand-picked key set. It misses common real secret shapes, several of which are the inverse of correct (it redacts the low-sensitivity item and leaks the high-sensitivity one):
  - ssh_keys: host private keys (rsa_private/ed25519_private) leak while adjacent rsa_public is redacted.
  - chpasswd: / nested list: plaintext user:password pairs leak.
  - plain_text_passwd: leaks (doesn't match passwd: prefix) while passwd: is caught; hashed_passwd: (crackable) also leaks.
  - write_files: content: (base64 creds, TLS keys, kubeconfigs) leaks.
  - Folded/indent-indicator scalars (>-, >+, |2) on a sensitive key: value lines leak because redactIndent is only armed for |/>/|+/|-.
  - Secrets inside runcmd (bearer tokens, curl -u user:pass) leak.
  - SSH formats beyond the 3 patterns (ssh-dss, sk-ssh-ed25519@openssh.com) leak.
  - Pod logs (fetchVirtLauncherPodLogs) are surfaced with no redaction at all.

This is highlighted because the PR description claims that "redacts sensitive cloud-init fields … while preserving diagnostic content"


The agent suggests the following for the first point:

- Either redact cloud-init within fetchVolumes too, or drop the raw userData/networkData from the volumes dump (strip cloud-init bodies and rely solely on the redacted section).

ksimon1:
> Rather than consolidating into a shared package, would it make sense to deprecate the prompt in a follow-up PR once this merges? @ksimon1 since you authored the original prompt, what's your take?

Agreed on deprecating the prompt given the overlap — but note the prompt's 'Fix the Issue' step (https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/toolsets/kubevirt/vm_troubleshoot.go#L166) has no equivalent here.

sradco:
> > Rather than consolidating into a shared package, would it make sense to deprecate the prompt in a follow-up PR once this merges? @ksimon1 since you authored the original prompt, what's your take?
> 
> Agreed on deprecating the prompt given the overlap — but note the prompt's 'Fix the Issue' step (https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/toolsets/kubevirt/vm_troubleshoot.go#L166) has no equivalent here.

Thank you for pointing this out.
The tool is `ReadOnlyHint: true` so I moved the remediation guidance into the tool description instead of the report output.
It now references the specific tools available (vm_lifecycle, resources_create_or_update) and only kicks in when the user explicitly asks to fix the issue.
I believe it would be a safer default for a Tool that LLMs can auto-invoke vs a Prompt that requires explicit user selection. Do you agree?

sradco:
Hi @Cali0707 , @ksimon1 , @lyarwood Can we merge this PR?

ksimon1:
/lgtm
The troubleshoot prompt will be removed in follow up PR?

manusa (approved):
LGTM, thx!

PR #1240, opened 2026-06-29
cleanup(logs): more contextual logging

Cali0707:
It looks like we missed a few spots of contextual logging, this cleans that up


manusa (approved):
LGTM, thx!

PR #1243, opened 2026-06-30
cleanup(logs): more contextual logging in kubevirt and mcp

manusa:
Follow-up to #1240, cleaning up the last couple of spots that weren't going through the contextual logging framework.

## Changes

**`pkg/kubevirt/resources.go`** — the resource discovery helpers logged their "failed to list" branches through Go's global `slog` default logger (the DataSource lines live in the unexported `collectDataSources` helper that backs `SearchDataSources`; the preference and instancetype lines in `SearchPreferences` and `SearchInstancetypes`). Nothing wires that logger to klog, so those lines bypassed the log sink entirely (no `-v` verbosity, no file output, no secret redaction, no OTEL), and since slog's default minimum level is `Info` the `Debug` lines were silently dropped at runtime anyway. All six now go through `klog.FromContext(ctx)` at `V(4)` using the `klogutil` helpers, with `kubernetes.namespace.name` / `exception.message` keys to match the rest of the codebase.

**`pkg/mcp/mcp.go`** — the SDK logger fallback in `NewServer` now derives from `klog.FromContext(ctx)` instead of `klog.Background()`, so it honors the contextual logger when no `SDKLogger` is supplied (the normal path already passes one from the logging sink).

## Testing

- `go build ./...`, `go vet`, `gofmt` all clean
- `go test ./pkg/kubevirt/...` passes
- `make lint` reports 0 issues

### DIFF
--- a/pkg/kubevirt/resources.go
+++ b/pkg/kubevirt/resources.go
@@ -3,12 +3,14 @@ package kubevirt
 import (
 	"context"
 	"fmt"
-	"log/slog"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/klog/v2"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/kubevirt/resources.go
```diff
@@ -72,7 +74,7 @@ func collectDataSources(ctx context.Context, dynamicClient dynamic.Interface) ma
 	for _, ns := range wellKnownNamespaces {
 		list, err := dynamicClient.Resource(DataSourceGVR).Namespace(ns).List(ctx, metav1.ListOptions{})
 		if err != nil {
-			slog.Debug("failed to list DataSources in well-known namespace", "namespace", ns, "error", err)
+			klogutil.LogInfo(klog.FromContext(ctx).V(4), "failed to list DataSources in well-known namespace", klogutil.Field("kubernetes.namespace.name", ns), klogutil.Err(err))
```
Cali0707, line 77: Whoops didn't consider grepping for slog - nice catch!

Cali0707 (approved):
LGTM

pkg/kubevirt/resources.go
manusa: Thank you Opus 4.8 :)

PR #1245, opened 2026-06-30
feat(tekton): add PipelineRun troubleshooting tools

waveywaves:
## What

- Add `tekton_pipelinerun_cancel` to cancel a PipelineRun by setting `spec.status=Cancelled`.
- Add `tekton_pipelinerun_logs` to collect logs from TaskRuns owned by a PipelineRun.
- Add the `pipeline-troubleshoot` prompt to gather PipelineRun status, related TaskRuns, logs, events, Pipeline-as-Code Repository resources, and TektonConfig resources.
- Document how to use generic `resources_list` / `resources_get` for PAC Repository and TektonConfig visibility.
- Update generated README/configuration tables, Tekton docs, eval tasks, and MCP tests.

## Why

PipelineRun cancellation and log collection are Tekton workflows that require domain-specific behavior beyond plain resource get/list. PAC Repository and TektonConfig visibility is handled without thin wrapper tools: generic resource tools cover direct reads, and the troubleshooting prompt pre-fetches those resources when diagnosing PipelineRuns.

## Testing

- `make update-readme-tools`
- `go test ./pkg/toolsets/tekton -count=1`
- `go test ./pkg/mcp -run 'TestTektonMcp|TestToolsets' -count=1`
- `go test ./...` *(fails locally: linker reports `no space left on device` while building several test binaries)*

## Notes

No repository or organization PR template is configured; this body follows the repository contribution guidance: background, user-facing changes, tests, and docs.

### DIFF
--- a/pkg/toolsets/tekton/pipeline_troubleshoot.go
+++ b/pkg/toolsets/tekton/pipeline_troubleshoot.go
@@ -1,0 +1,253 @@
+package tekton
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	"github.com/containers/kubernetes-mcp-server/pkg/kubernetes"
+	"github.com/containers/kubernetes-mcp-server/pkg/output"
+	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func pipelineTroubleshootPrompts() []api.ServerPrompt {
+	return []api.ServerPrompt{
+		{
+			Prompt: api.Prompt{
+				Name:        "pipeline-troubleshoot",
+				Title:       "Tekton PipelineRun Troubleshoot",
+				Description: "Gather PipelineRun status, TaskRuns, logs, events, Pipeline-as-Code Repository, and TektonConfig context for Tekton troubleshooting",
+				Arguments: []api.PromptArgument{
+					{
+						Name:        "namespace",
+						Description: "Namespace of the PipelineRun to troubleshoot",
+						Required:    true,
+					},
+					{
+						Name:        "name",
+						Description: "Name of the PipelineRun to troubleshoot",
+						Required:    true,
+					},
+				},
+			},
+			Handler: pipelineTroubleshootHandler,
+		},
+	}
+}
+
+func pipelineTroubleshootHandler(params api.PromptHandlerParams) (*api.PromptCallResult, error) {
+	args := params.GetArguments()
+	namespace := args["namespace"]
+	name := args["name"]
+	if namespace == "" {
+		return nil, fmt.Errorf("namespace argument is required")
+	}
+	if name == "" {
+		return nil, fmt.Errorf("name argument is required")
+	}
+
+	pipelineRun, pipelineRunText := fetchPipelineRunForPrompt(params, namespace, name)
+	taskRuns, taskRunsText := fetchPipelineRunTaskRunsForPrompt(params, namespace, name)
+	logsText := fetchPipelineRunLogsForPrompt(params, namespace, taskRuns)
+	eventsText := fetchPipelineRunEventsForPrompt(params, namespace, name, taskRuns)
+	pacText := fetchPipelineRunPACRepositoriesForPrompt(params, namespace)
+	tektonConfigText := fetchTektonConfigsForPrompt(params)
+
+	promptText := buildPipelineTroubleshootPrompt(namespace, name, pipelineRun, pipelineRunText, taskRunsText, logsText, eventsText, pacText, tektonConfigText)
+	return api.NewPromptCallResult(
+		"PipelineRun troubleshooting data gathered successfully",
+		[]api.PromptMessage{
+			{
+				Role: "user",
+				Content: api.PromptContent{
+					Type: "text",
+					Text: promptText,
+				},
+			},
+			{
+				Role: "assistant",
+				Content: api.PromptContent{
+					Type: "text",
+					Text: "I'll analyze the collected Tekton data to identify the PipelineRun issue.",
+				},
+			},
+		},
+		nil,
+	), nil
+}
+
+func fetchPipelineRunForPrompt(params api.PromptHandlerParams, namespace, name string) (*unstructured.Unstructured, string) {
+	pipelineRun, err := params.DynamicClient().Resource(pipelineRunGVR).Namespace(namespace).Get(params.Context, name, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Sprintf("*Error fetching PipelineRun: %v*", err)
+	}
+	return pipelineRun, yamlBlock("PipelineRun", pipelineRun)
+}
+
+func fetchPipelineRunTaskRunsForPrompt(params api.PromptHandlerParams, namespace, pipelineRunName string) ([]tektonv1.TaskRun, string) {
+	taskRuns, err := pipelineRunTaskRuns(params.Context, params.DynamicClient(), namespace, pipelineRunName)
+	if err != nil {
+		return nil, fmt.Sprintf("*Error listing TaskRuns: %v*", err)
+	}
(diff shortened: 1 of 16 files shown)


pkg/toolsets/tekton/pipelinerun.go
```diff
@@ -96,3 +154,74 @@ func restartPipelineRun(params api.ToolHandlerParams) (*api.ToolCallResult, erro
...
+	}
+
+	var sb strings.Builder
+	for _, taskRun := range taskRuns {
+		fmt.Fprintf(&sb, "# TaskRun: %s\n", taskRun.Name)
+		collectTaskRunLogs(params, &sb, namespace, &taskRun, tailLines)
+	}
+	if sb.Len() == 0 {
+		return api.NewToolCallResult(fmt.Sprintf("No logs available for PipelineRun '%s' in namespace '%s'", name, namespace), nil), nil
+	}
```
Cali0707, line 206: I think this is incorrect, since you write the `# TaskRun: TaskRunName` for each taskRun, and you validate that `len(taskRuns)` is greater than zero

So I think to check if there are no logs available, you would need to only write the headers to the string builder if there are logs available for that task run

pkg/toolsets/tekton/pipeline_troubleshoot.go
```diff
@@ -0,0 +1,253 @@
...
+	for _, taskRun := range taskRuns {
+		fmt.Fprintf(&sb, "### TaskRun: %s\n\n", taskRun.Name)
+		collectTaskRunLogsWithClient(params.Context, params.KubernetesClient, &sb, namespace, &taskRun, kubernetes.DefaultTailLines)
+		sb.WriteString("\n")
+	}
+	return sb.String()
+}
+
+func fetchPipelineRunEventsForPrompt(params api.PromptHandlerParams, namespace, pipelineRunName string, taskRuns []tektonv1.TaskRun) string {
+	events, err := params.CoreV1().Events(namespace).List(params.Context, metav1.ListOptions{})
```
Cali0707: can we use a field selector here on `involvedObject.name` somehow to reduce the potential API hit? listing all the events in a namespace can be quite expensive

pkg/toolsets/tekton/pipelinerun.go
```diff
@@ -17,31 +22,84 @@ func pipelineRunTools() []api.ServerTool {
...
+				Annotations: api.ToolAnnotations{
+					Title:           "Tekton: Cancel PipelineRun",
+					ReadOnlyHint:    ptr.To(false),
+					DestructiveHint: ptr.To(true),
+					IdempotentHint:  ptr.To(true),
+					OpenWorldHint:   ptr.To(true),
+				},
+			},
+			Handler: cancelPipelineRun,
+		},
```
Cali0707, line 43: is it possible to combine these two tools and make the action (cancel, restart) an enum? This is the pattern kubevirt uses in a few tools

pkg/toolsets/tekton/pipeline_troubleshoot.go
```diff
@@ -0,0 +1,253 @@
...
+		return "*No TektonConfig resources found*"
+	}
+	yaml, err := output.MarshalYaml(list)
+	if err != nil {
+		return fmt.Sprintf("*Error formatting TektonConfig resources: %v*", err)
+	}
+	return fmt.Sprintf("```yaml\n%s```", yaml)
+}
+
+func buildPipelineTroubleshootPrompt(namespace, name string, pipelineRun *unstructured.Unstructured, pipelineRunText, taskRunsText, logsText, eventsText, pacText, tektonConfigText string) string {
```
ksimon1: Nit: maybe use struct which would hold text params?

pkg/toolsets/tekton/pipelinerun.go
waveywaves: PipelineRun log collection now collects each TaskRun into a temporary builder first, then writes the `# TaskRun:` header only when that TaskRun has collected output. The `No logs available` branch should now work correctly. I applied the same pattern to the troubleshooting prompt log section too.

pkg/toolsets/tekton/pipeline_troubleshoot.go
waveywaves: Replaced the namespace-wide event list with targeted event queries using field selectors for related objects: PipelineRun, TaskRuns, and TaskRun pods. This now filters by `involvedObject.kind` and `involvedObject.name`, then de-dupes returned events. Added coverage with both a related and unrelated event.

pkg/toolsets/tekton/pipelinerun.go
waveywaves: Combined cancel/restart into `tekton_pipelinerun_lifecycle` with an `action` enum: `restart` or `cancel`, following the KubeVirt lifecycle pattern. Updated tests, docs, eval README, and generated tool snapshot.

pkg/toolsets/tekton/pipeline_troubleshoot.go
waveywaves: Added `pipelineTroubleshootData` and changed `buildPipelineTroubleshootPrompt` to take that struct instead of a long positional argument list.

pkg/toolsets/tekton/pipeline_troubleshoot.go
```diff
@@ -0,0 +1,305 @@
...
+		if target.name == "" {
+			continue
+		}
+		selector := fields.SelectorFromSet(fields.Set{
+			"involvedObject.kind": target.kind,
+			"involvedObject.name": target.name,
+		}).String()
+		events, err := params.CoreV1().Events(namespace).List(params.Context, metav1.ListOptions{FieldSelector: selector})
+		if err != nil {
+			return fmt.Sprintf("*Error listing events for %s %s: %v*", target.kind, target.name, err)
```
ksimon1: Would it make sense to just log this error instead of returning and continue gathering other events? Because like this, it will ignore other targets if first target fails e.g. due to RBAC issues.

pkg/toolsets/tekton/pipelinerun.go
```diff
@@ -88,11 +153,84 @@ func restartPipelineRun(params api.ToolHandlerParams) (*api.ToolCallResult, erro
...
+			sb.WriteString("\n")
+		}
+	}
+	if sb.Len() == 0 {
+		return api.NewToolCallResult(fmt.Sprintf("No logs available for PipelineRun '%s' in namespace '%s'", name, namespace), nil), nil
+	}
+	return api.NewToolCallResult(sb.String(), nil), nil
+}
+
+func listPipelineRunTaskRuns(params api.ToolHandlerParams, namespace, pipelineRunName string) ([]tektonv1.TaskRun, error) {
```
ksimon1: Is this FN necessary? It just calls another FN.

ksimon1:
@waveywaves can you please resolve merge conflicts?

pkg/toolsets/tekton/pipeline_troubleshoot.go
```diff
@@ -0,0 +1,308 @@
...
+		return nil, fmt.Sprintf("*Error listing TaskRuns: %v*", err)
+	}
+	if len(taskRuns) == 0 {
+		return nil, "*No TaskRuns found for this PipelineRun*"
+	}
+
+	var sb strings.Builder
+	for _, taskRun := range taskRuns {
+		fmt.Fprintf(&sb, "### TaskRun: %s\n\n", taskRun.Name)
+		if status, err := output.MarshalYaml(taskRun.Status); err == nil {
```
ksimon1, line 124: Nit: if marshalling fails, then there is no record of it. It would be probably good to log it

ksimon1:
@manusa can you please check this PR?

Cali0707 (approved):
@waveywaves are you able to fix the merge conflicts?

Beyond those, LGTM

Cali0707 (commented):
/run-mcpchecker tekton

waveywaves:
@Cali0707 @manusa  the PR has been rebased and ready for another review

Cali0707 (approved):

Cali0707:
@waveywaves @ksimon1 the troubleshoot-pipelinerun eval is failing on this PR. As the other new eval is passing and this is still overall above the threshold, I am still fine with merging this.

That said, you may want to iterate on improving the tooling so that the eval passes (IMO this failing shows that it is a strong task, and a good stress test of the tools we have here).

PR #1248, opened 2026-06-30
chore(ci): drive toolset labels and owner notifications from CODEOWNERS

manusa:
## What

Makes **`.github/CODEOWNERS` the single source of truth** for both toolset ownership and toolset labels, and adds one workflow that acts on it.

- **`.github/CODEOWNERS`** — ownership map, with a `# toolset: <name>` marker per block.
- **`.github/workflows/toolset-triage.yaml`** — on each PR, parses CODEOWNERS to (a) apply/remove the `toolset/<name>` label and (b) @-mention the owners of changed paths.
- **Retires `actions/labeler`** (`.github/labeler.yaml` + its workflow) — labels are now derived from CODEOWNERS, so there is no second path list to drift.

## Ownership

| Area | Owners |
|------|--------|
| Default (everything else) | @manusa @Cali0707 |
| kiali | @aljesusg @josunect |
| kubevirt | @ksimon1 @lyarwood |
| tekton | @ksimon1 |
| kcp | @mjudeikis |

Each toolset covers its `pkg/toolsets/<t>`, `pkg/<t>`, `pkg/mcp/<t>*`, `build/<t>.mk`, `evals/tasks/<t>` and docs. Toolset lines list the toolset owners only; core maintainers retain admin/maintain override and own everything else via the default rule.

## Why a workflow at all

GitHub only auto-requests review from code owners that have **write** access ([docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)). Most toolset maintainers are external contributors who can't be granted write, so GitHub silently skips them — no request, no notification. The workflow closes that gap by @-mentioning them in a sticky comment.

It also takes over **labeling**: reconciling `toolset/*` against the *current* changed files (add **and** remove) fixes the over-labeling the old `actions/labeler` produced (its default `sync-labels: false` never removed stale labels). It also closes coverage gaps — the old labeler config didn't label `evals/tasks/kiali|kubevirt`, the toolset docs, or **kcp** at all.

Runs as `pull_request_target` (so it can label/comment on fork PRs) and never checks out or runs PR code.

## ✅ kubevirt owners — confirmed

@lyarwood [confirmed](https://github.com/containers/kubernetes-mcp-server/pull/1248#issuecomment-4844192956) the listed set, so **@ksimon1** and **@lyarwood** stay as the kubevirt toolset owners.

<details>
<summary>Original request (resolved)</summary>

@ksimon1 @lyarwood @codingben @Ruclo — I've listed **@ksimon1** and **@lyarwood** as the kubevirt toolset owners based on commit history. Since the four of you are the main kubevirt contributors, could you confirm that's the right set, or whether it should be refined (add/remove owners)?

</details>

## Notes

- `main` currently enforces only status checks (no "require code owner review" rule), so this does not block any PRs.
- External owners show a GitHub "no write access" warning on their CODEOWNERS lines — expected, and exactly why the workflow exists.


lyarwood:
> ## kubevirt owners — please confirm 🙏
> @ksimon1 @lyarwood @codingben @Ruclo — I've listed **@ksimon1** and **@lyarwood** as the kubevirt toolset owners based on commit history. Since the four of you are the main kubevirt contributors, could you confirm that's the right set, or whether it should be refined (add/remove owners)?

This works for me thanks!

Cali0707:
@manusa as a follow up let's see if we can get the mcpchecker action to accept triggers from the CODEOWNERS for their respective toolsets

Cali0707 (approved):
LGTM, thanks for this improvement @manusa !

PR #1250, opened 2026-06-30
ci(release): scope publish workflows to first-party tag-push releases

manusa:
## What

The MCPB bundle (`release-mcpb.yaml`) and MCP Registry (`release-mcp-registry.yaml`) publish workflows trigger on `workflow_run` completions of the `Release` workflow. This scopes their job `if` conditions so they only run for this repository's own tag-push releases (or an explicit `workflow_dispatch`), rather than for any `Release` workflow completion.

## Why

Both jobs read `github.event.workflow_run.*`. Constraining the trigger to a `push` event originating from this repository keeps publishing deterministic and avoids acting on unrelated `Release` completions.

## Notes

No change to the normal release path — a `v*` tag push still produces the bundles and the registry entry exactly as before, and `workflow_dispatch` is unchanged.

### DIFF
--- a/.github/workflows/release-mcp-registry.yaml
+++ b/.github/workflows/release-mcp-registry.yaml
@@ -22,7 +22,13 @@ permissions:
 jobs:
   publish:
     name: Publish to MCP Registry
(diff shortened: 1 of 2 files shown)


Cali0707 (approved):
LGTM

Issue #1238, opened 2026-06-29
Core eval task-definition bugs: brittle prose-grep verifies, nginx 8080 vs 80, resize-pvc storage


manusa:
> ## Status — covered by PR #1253
>
> The agent-independent task bugs are fixed in #1253 and confirmed by a green mcpchecker `core` run (26/29, up from the 25/29 baseline). The four target tasks flip **FAILED → PASSED**:
>
> - ✅ **`multi-container-pod-communication`** (item 1) — verify probes `:8080` (nginx-unprivileged does not listen on `:80`); `setup.sh`/`cleanup.sh` namespace aligned to `multi-container-logging`.
> - ✅ **`resize-pvc`** (item 2) — the eval cluster now installs a trimmed hostPath CSI driver (`make csi-hostpath-install`) that provides a resize-capable `csi-hostpath-sc` StorageClass; CI installs it for the core suite and the PVC points at it.
> - ✅ **`filter-events-by-type`** (item 3) — verify anchors on the event content (`ErrImagePull` / `ImagePullBackOff` / `Failed to pull image` / the bogus image name) instead of grepping the agent's prose.
> - ✅ **`filter-namespace-by-name`** (item 4) — verify checks `ns-beta` present and `ns-alpha` / `ns-gamma` absent, robust to surrounding prose.
>
> **Deferred (not in #1253):**
> - **`ssa-field-preservation`** (item 5) — separate slice, needs mcpchecker/kubernetes-extension#55.
> - **`maxToolCalls` assertion audit** (item 6, expanded below) — analyze *why* gpt-5 is so chatty before touching the threshold.
>
> **No change, working as intended:** `fix-service-routing` (item 7), `deployment-traffic-switch` (item 8).

## Context

While working on #1234 (PR #1237, improving the local `mcpchecker` eval workflow), I ran the full `core` suite locally with the `claude-code` agent on Sonnet, keyless, against a bare `kind` cluster plus the MCP server. 6 of 29 tasks failed and I filed this issue.

I then re-verified every point, because some of the original conclusions were wrong. Two framing errors mattered:

1. **The original 6 failures are a single Sonnet run. CI runs a different agent (`openai:gpt-5`).** `.github/workflows/mcpchecker.yaml` drives `evals/openai-agent/eval.yaml` and commits `evals/results/openai-agent-latest.json`. "Does this affect CI?" has to be read off that file, not inferred from a local Sonnet run. Different agents fail different tasks.
[...]

PR #1253, opened 2026-07-02
fix(evals): correct core task-definition bugs

manusa:
## What

Fixes agent-independent bugs in several `core` mcpchecker eval task
definitions (reported in #1238). These tasks failed for every agent
because the task scripts and checks were wrong, not because of agent
capability.

## Changes

**multi-container-pod-communication**
- `verify.sh` curled the nginx-unprivileged container on `localhost:80`,
  but that image only listens on `8080` (it cannot bind `:80` as a
  non-root user), so curl exited 7 and the task never passed. Now probes
  `:8080`.
- `setup.sh` and `cleanup.sh` used the namespace `multi-container-test`
  while the prompt and `verify.sh` use `multi-container-logging`.
  Everything now uses `multi-container-logging`.

**filter-events-by-type and filter-namespace-by-name**
- Both verifies grep'd the agent's prose for a specific wording and word
  order, so correct answers failed depending on how the model phrased
  them. Checked against six weekly gpt-5 runs from the results history,
  two of which never write the words "Warning" or "bad-pod" at all.
- filter-events-by-type now anchors on the event content itself
  (kubelet's verbatim `ErrImagePull` / `ImagePullBackOff` / `Failed to
  pull image` / the bogus image name) and still excludes `good-pod`.
- filter-namespace-by-name now checks that `ns-beta` is present and
  `ns-alpha` / `ns-gamma` are absent, regardless of surrounding prose.

**resize-pvc**
- kind's default `standard` StorageClass (`rancher.io/local-path`) has
  `allowVolumeExpansion: false`, so the resize never took effect and
  verify timed out.
- Adds a trimmed hostPath CSI driver (`dev/config/csi-hostpath/`,
  installed via `make csi-hostpath-install`) that provides a
  `csi-hostpath-sc` StorageClass with `allowVolumeExpansion: true`. The
  manifest is trimmed from csi-driver-host-path v1.17.1 to just the
  plugin, node-driver-registrar, external-provisioner, and
  external-resizer (no attacher, snapshotter, or health-monitor).
- CI installs it for the core suite (and the `all` run) before starting
  the server, and the task PVC points at the new StorageClass.

## Not included

Per the issue, `ssa-field-preservation` (needs an upstream
mcpchecker/kubernetes-extension change) and the `fix-probes`
`maxToolCalls` assertion are left for follow-ups. `fix-service-routing`
and `deployment-traffic-switch` are working as intended and are not
touched.

## Verification

- Ran the two rewritten verifies against six real gpt-5 runs per task
  (mined from the weekly results history) plus synthetic wrong answers:
  all correct answers pass, all wrong answers fail. The old checks failed
  5/6 and 6/6 of the real correct answers.
- Stood up a kind cluster (k8s v1.34) and ran the real `setup.sh` /
  `verify.sh` end to end: `resize-pvc` reaches 3Gi and passes,
  `multi-container-pod-communication` passes on `:8080`, and the CSI
  plugin runs with zero restarts.

Refs #1238

### DIFF
--- a/.github/workflows/mcpchecker.yaml
+++ b/.github/workflows/mcpchecker.yaml
@@ -179,6 +179,12 @@ jobs:
       - name: Setup Kind cluster
         run: make kind-create-cluster KIND_CLUSTER_NAME="$KIND_CLUSTER_NAME"
 
+      # Only the core suite needs a resize-capable StorageClass (for resize-pvc).
+      # Runs for suite=core and for the "all" run (empty label-selector).
+      - name: Install CSI hostPath driver (resize-capable StorageClass)
+        if: needs.check-trigger.outputs.label-selector == 'suite=core' || needs.check-trigger.outputs.label-selector == ''
+        run: make csi-hostpath-install
+
       - name: Install Istio/Kiali and bookinfo demo
         if: contains(needs.check-trigger.outputs.toolsets, 'kiali')
         run: make setup-kiali
--- a/build/storage.mk
+++ b/build/storage.mk
@@ -1,0 +1,41 @@
+# CSI hostPath driver installation (resize-capable StorageClass for evals)
+#
+# Provides the `csi-hostpath-sc` StorageClass with allowVolumeExpansion: true so
+# the core `resize-pvc` eval task can run. kind's default `standard` StorageClass
+# (rancher.io/local-path) does not support volume expansion.
+
+# Reference-only: the upstream release the vendored manifest was trimmed from
+# (kubernetes-csi/csi-driver-host-path). Used only in the log line below; the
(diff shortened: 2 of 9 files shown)


manusa:
mcpchecker `core` validated the fixes end to end: the run flips all four target tasks **FAILED → PASSED** (`filter-events-by-type`, `filter-namespace-by-name`, `multi-container-pod-communication`, `resize-pvc`) — 26/29 vs the 25/29 baseline on main. The rest of the diff is gpt-5 run-to-run noise (mainly the tight suite-wide `maxToolCalls: 20`), and none of those tasks are touched by this PR.

Cali0707 (approved):
LGTM

PR #1254, opened 2026-07-02
feat!(mcp): add 2026-07-28 spec support

Cali0707:
This PR bumps to the first RC for the 1.7.0 MCP SDK, which includes support for the new spec version and some deprecations. We mostly seem to be affected by logging so far.

If you test this out with a client and see any issues, please leave comments here so we can investigate and fix/report upstream.


manusa (approved):
LGTM, thx!

Issue #1129, opened 2026-05-04
cluster-health-check prompt times out on real (Open)Shift clusters


manusa:
## Summary

The `cluster-health-check` prompt (`pkg/toolsets/core/health_check.go`) times out under MCP clients (Claude Desktop / Inspector / Cursor / etc.) when invoked against real Kubernetes — and especially OpenShift — clusters. The handler does a large number of **sequential** Kubernetes API calls with **no per-call timeout, no parallelism, no result paging, and no early termination**, and on cluster scope it fans out across every `openshift-*` namespace.

There is no per-request timeout imposed anywhere in our stack (handler, middleware, HTTP server, go-sdk), so the timeout is being raised by the MCP client (commonly 30–60s for `prompts/get`). The handler must finish within that window.

## Where the time goes

`gatherClusterDiagnostics` (`pkg/toolsets/core/health_check.go:127-222`) runs every section in series. On a cluster-wide invocation against a real OpenShift cluster, the API calls are roughly:

| # | Call | Notes |
|---|---|---|
| 1 | `Nodes().List()` | full node objects |
| 2 | `Pods(\"\").List()` | **all pods cluster-wide**; no `Limit`, no `FieldSelector` — can be MBs |
| 3 | `Deployments(\"\").List()` | all deployments |
| 4 | `StatefulSets(\"\").List()` | all stateful sets |
| 5 | `DaemonSets(\"\").List()` | all daemon sets |
| 6 | `PersistentVolumeClaims(\"\").List()` | all PVCs |
| 7 | `ResourcesList(ClusterOperator)` (`health_check.go:475`) | on a fresh derived client this triggers full discovery (`/apis/...` enumeration) before the list |
| 8 | `Namespaces().List()` (inside `gatherEventDiagnostics`, line 552) | only on cluster scope, used to find `openshift-*` |
| 9 | `Events(\"default\").List()` and `Events(\"kube-system\").List()` | 2 calls |
| 10 | `Events(\"openshift-X\").List()` — **N calls in series**, where N ≈ 30–60+ on real OpenShift | dominates wall time |
| 11 | `Namespaces().List()` (line 214, again, just to count total) | **redundant** — already listed at step 8 |

That is ~40–70 sequential round-trips on a typical OpenShift cluster. Each `Events().List()` returns the full event store for that namespace; the 1-hour and severity filters are applied **client-side** at `health_check.go:575-586`.

### Aggravating factors

[...]

PR #1263, opened 2026-07-06
feat(prompts): Rework core health check prompts

difrost:
## Summary

Closes the performance and reliability [issues](https://github.com/containers/kubernetes-mcp-server/issues/1129) for the `cluster-health-check` prompt.

The previous implementation did all diagnostic work sequentially with no per-call timeout, no parallelism, no paging, and no prompt-size guardrails. On real clusters (especially OpenShift with many `openshift-*` namespaces) this caused client-side timeouts, excessive API-server load, and prompts large enough to be silently truncated or prohibitively expensive.

This PR reworks the health-check prompts from scratch, introduces a small task-runner library for time-bounded concurrency, splits cluster-wide and namespace-scoped checks into separate prompts, and caps the rendered prompt size while preserving partial-failure behavior.

> **Status:** This is a WIP / design proposal. The implementation has been tested in practice against clusters ranging from small to thousands of pods, but the configuration surface and default rate limits are intentionally conservative and will be refined in follow-up PRs.

### 1. New `pkg/tasks` concurrency library

- Added a lightweight `Task` runner backed by `errgroup.WithContext`.
- Supports per-task timeouts and optional concurrency limits (`New` / `NewWithLimit`).
- Isolates per-task errors and panics so one failing task does not abort unrelated work.
- Included comprehensive unit tests (`pkg/tasks/tasks_test.go`).

This directly addresses the issue's findings that the prompt had:
- no parallelism,
- no per-section timeout, and
- no fail-fast isolation.

### 2. Reworked health-check architecture (`pkg/toolsets/core/prompts/`)

The old `pkg/toolsets/core/health_check.go` monolith was replaced by a split, three-stage pipeline:

| Stage | Responsibility |
|---|---|
| **Gather** | Fetch nodes, pods, deployments, statefulsets, daemonsets, PVCs, and optionally events / cluster operators. Runs in parallel with per-check timeouts. |
| **Analyze** | Transform raw lists into typed health reports (pod, workload, PVC, node, event, cluster-operator). Also runs in parallel. |
| **Render** | Build markdown output from analysis reports with explicit line budgets. |

Files introduced:
- `health_check.go` — cluster-wide prompt orchestration
- `namespace_health_check.go` — namespace-scoped prompt orchestration
- `gather.go` — all gather tasks
- `analyze.go` — all analysis tasks
- `render.go` — markdown rendering with budgets
- `extract.go` / `names.go` — shared helpers
- `*_test.go` — unit tests for analysis and rendering

### 3. Split prompts: `cluster-health-check` vs `namespace-health-check`

Because many MCP clients silently truncate large prompts and token cost grows with prompt size, cluster-wide checks must be heavily condensed. Namespace-scoped checks can afford more detail.

| Prompt | Scope | Argument changes |
|---|---|---|
| `cluster-health-check` | All namespaces for workloads; default / `*-system` / `openshift-*` namespaces for events | Removed `namespace` argument; kept optional `check_events` |
| `namespace-health-check` | Single namespace | New required `namespace` argument; optional `check_events` |

Updated `README.md`, `docs/prompts.md` to reflect the new prompt definitions.

### 4. Prompt-size budget and data shaping

- Introduced `reportLineLimit = 200` shared across namespace-health and event sections.
- Events are prioritized because they carry higher diagnostic value.
- Cluster-wide rendering uses compact counts (`deployments: 3 unhealthy (5 replicas missing)`).
- Namespace-scoped rendering uses detailed names (`my-deploy (2 replicas missing)`).
- Event messages are truncated to 150 runes, multi-line messages collapsed, and event groups truncated at object boundaries.
- Namespaces are sorted by issue count descending for triage priority.

The exact output shape is the result of several design-and-test iterations aimed at fitting the most useful signal into a small token budget.

### 5. Performance improvements

- **Parallel gather and analyze phases** using `pkg/tasks`.
- **Cached namespace list** reused for both event filtering and the namespace count, removing the duplicate list call.
- **OpenShift gating**: `ClusterOperator` discovery/list only happens when `openshift.IsOpenshift()` returns true.
- **Event collection capped**:
  - `FieldSelector: type!=Normal` applied server-side.
  - `Limit: 100` per namespace event list.
  - Event fan-out limited to relevant namespaces with rate-aware concurrency (`maxEventConcurrency = 20`).
- **Soft per-check deadline** (`perCheckTimeout = 15s`) and per-event-gather deadline (`eventCheckTimeout = 13s`) so one slow API cannot sink the whole prompt.

### 6. client-go rate limits raised

`pkg/kubernetes/manager.go` now defaults QPS/Burst to `25/50` when the REST config still carries zero values (the client-go default is `5/10`). This is a pragmatic default for prompt fan-out but is **hardcoded for now** and will be exposed via configuration in a follow-up PR.

### 7. Behavioral MCP-layer tests

Added `pkg/mcp/mcp_prompt_cluster_health_check_test.go` and `pkg/mcp/mcp_prompt_namespace_health_check_test.go`. Both are full end-to-end `envtest` suites (no mocks) that cover:

- prompt registration and metadata
- happy path completing under 5 seconds
- `check_events=false` and invalid `check_events` defaulting to `true`
- non-existent namespace handling
- non-OpenShift cluster behavior
- cluster-wide workload and event detection
- namespace-scoped pod, workload, PVC, and event detection
- section ordering and graceful empty sections

Also added a shared `requireMessageText` helper in `pkg/mcp/common_test.go`.

## Known limitations / follow-up work

1. **Partial failure testing**  
   The MCP tests cover empty/no-data sections but not a true gather/analysis task failure while other tasks succeed. Inducing this reliably against the shared `envtest` control plane without cross-test interference would require complex RBAC setup. We deferred that coverage until the prompt internals expose a safer injection point.

2. **Hardcoded rate limits**  
   The new `QPS=25 / Burst=50` defaults in `pkg/kubernetes/manager.go` were chosen empirically on several clusters, but they should be configurable. A follow-up PR will expose them through `StaticConfig`.

3. **Paging for very large lists**  
   Pods and deployments are still listed in a single call. For clusters much larger than those already tested, server-side paging (`Limit`/`Continue`) may be needed. This is left as future optimization.

4. **Design is not final**  
   The output format, budgets, and timeout values are the result of practical iteration, but we expect further tuning based on feedback and additional cluster profiles.

## Verification

The reworked prompts are already in regular use for health-check assessments against clusters of varying sizes, including those running thousands of pods.


Cali0707:
Thanks @difrost for tackling this! The prompt split, budget system, event filtering improvements, and behavioral test coverage are all solid. The functional requirements from #1129 all seem to be addressed as well.

My main ask is to reduce the implementation weight. This is ~5k lines of change for what is a errgroup + timeouts + capping fix, I'm not convinced the problem needs such a complex solution. A few specifics:

1. `pkg/tasks`: use `errgroup` directly. There's no second consumer for this library, and the library + tests can be replaced with errgroup usage directly within the handlers. If a reusable task runner is needed later, we can extract one once we see the concrete use cases (and build on the current impl here if that makes sense then)
2. Drop the analyze phase parallelism unless there is measurable improvements on large clusters. I benchmarked these funcs locally and with 6000 pods, 500 deployments, 3000 events and 100 nodes doing them all sequentially only took ~0.24 milliseconds.
3. Drop `eventCheckTimeout`. The inner 13s timeout doesn't necessarily with the effective deadline. In go, context trees inherit whichever parent deadline is shorter, so ther outer (minus elapsed time) can often win, given the timing of calls between the two timeouts being set. The 2s of headroom isn't really needed as `wt.Complete()` after a cancellation returns near instantly.

PR #1267, opened 2026-07-06
chore(ci): set 30m timeout on e2e job

Cali0707:
Follow up on #1228 that adds a timeout to the overall e2e action


manusa (approved):
LGTM, thx!

PR #1268, opened 2026-07-06
chore(e2e): cleanup e2e framework

Cali0707:
Follow up on #1228 

This addresses some final comments left by @manusa :
> A few minor nits to fold in whenever you next touch this: _find_free_port has a TOCTOU race that'll flake once deployments run in parallel; _parse_image only handles the registry/repo:tag form (digests/registry-ports get mangled); config.pop("http") silently drops that section; and paths-ignore: '*.md' only matches top-level markdown.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -6,7 +6,7 @@ on:
     paths-ignore:
       - '.gitignore'
       - 'LICENSE'
-      - '*.md'
+      - '**/*.md'
       - 'docs/**'
 
 permissions:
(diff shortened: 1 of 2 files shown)


manusa (approved):
LGTM, thx!

I also pushed a small fix directly to this branch (rather than a follow-up PR) to expedite merging, these are the changes (feel free to rollback if deemed unnecessary):

- Port-forward stdout now goes to a temp file polled with a 30s deadline, instead of an undrained PIPE read with a single readline().
  kubectl port-forward writes Handling connection for <port> to stdout on every connection, so the old approach could deadlock once the ~64 KB pipe buffer filled (and readline() could hang unbounded if kubectl wedged).
  It also makes local-port parsing dual-stack-safe.
- proc.wait() after each kill() so the terminated port-forward is reaped (also fixes the "exit code None" diagnostic).

matzew (approved):
LGTM

PR #1270, opened 2026-07-07
feat(tls): add TLS_MIN_VERSION and TLS_CIPHER_SUITES env var support

### LINKED ISSUE #1266: Support `TLS_MIN_VERSION` and `TLS_CIPHER_SUITES` env vars/config values
In https://github.com/kubernetes-sigs/mcp-lifecycle-operator/pull/257 the kubernetes mcp lifecycle operator is adding support for setting these env variables on all MCP servers deployed to a cluster by the operator.

The goal here is to support consistency of TLS configuration across an entire cluster.

We should, everywhere we use TLS, support for reading the TLS min version and supported cipher suites from config and env vars, with env vars taking precedence over the config values.

cajieh:
Add support for configuring TLS minimum version and cipher suites via environment variables and global TOML, enabling operator-level control over TLS settings.

Jira Ticket: [Add TLS min version and cipher suite configuration via env vars and global TOML](https://redhat.atlassian.net/browse/OCPMCP-354)

## Changes

- Add `pkg/tlsutil` for parsing TLS settings and building `crypto/tls.Config` (no env resolution or logging)
- Add `TLS_MIN_VERSION` env var (values: `"1.0"`, `"1.1"`, `"1.2"`, `"1.3"`)
- Add `TLS_CIPHER_SUITES` env var (comma-separated cipher suite names)
- Add `tls_min_version` and `tls_cipher_suites` global TOML options
- Resolve env/TOML precedence in `GetTLSMinVersionConfig()` / `GetTLSCipherSuitesConfig()` on `StaticConfig`
- Log env overrides once at startup/reload via `Validate()`
- Wire TLS config into inbound HTTPS and outbound clients (Kiali, NetObserv, OAuth, token exchange, well-known metadata)
- Track TLS settings in the OAuth snapshot so SIGHUP reload rebuilds the OIDC HTTP client when they change
- Invalidate STS/token-exchange client caches when TLS settings change
- Add tests for parsing, validation, env precedence, OAuth snapshot comparison, client cache invalidation, and NetObserv TLS application

## Precedence

`TLS_MIN_VERSION` / `TLS_CIPHER_SUITES` override TOML when set. When neither is set, both inbound and outbound default to TLS 1.2 with Go's default cipher suites. TOML values apply as fallbacks to **both** inbound and outbound paths.

## Reload behavior

- **Inbound HTTPS:** applied at server start; changing TLS settings requires a **process restart**
- **Outbound clients:** OAuth and token exchange pick up TOML/env changes on SIGHUP reload; Kiali and NetObserv re-read settings on each tool invocation

Closes https://github.com/containers/kubernetes-mcp-server/issues/1266

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -17,12 +17,17 @@ import (
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
 	"github.com/containers/kubernetes-mcp-server/pkg/output"
+	"github.com/containers/kubernetes-mcp-server/pkg/tlsutil"
 	"github.com/containers/kubernetes-mcp-server/pkg/tokenexchange"
 	"github.com/containers/kubernetes-mcp-server/pkg/toolsets"
 )
 
 const (
 	DefaultDropInConfigDir = "conf.d"
+
+	// Environment variable names for TLS configuration.
+	EnvTLSMinVersion   = "TLS_MIN_VERSION"
+	EnvTLSCipherSuites = "TLS_CIPHER_SUITES"
 )
 
 // ToolOverride contains per-tool configuration overrides.
@@ -126,6 +131,12 @@ type StaticConfig struct {
 	// When true, the server will refuse to start without TLS certificates,
 	// and outbound connections to non-HTTPS endpoints will be rejected.
 	RequireTLS bool `toml:"require_tls,omitempty"`
+	// TLSMinVersion is the minimum TLS version to accept (e.g., "1.2", "1.3").
+	// Defaults to TLS 1.2 if not set. Can be overridden by TLS_MIN_VERSION env var.
+	TLSMinVersion string `toml:"tls_min_version,omitempty"`
+	// TLSCipherSuites is a list of supported cipher suites for TLS connections.
+	// If empty, Go's default cipher suites are used. Can be overridden by TLS_CIPHER_SUITES env var.
+	TLSCipherSuites []string `toml:"tls_cipher_suites,omitempty"`
 
 	// HTTP server configuration (timeouts, size limits)
 	HTTP HTTPConfig `toml:"http,omitempty"`
@@ -471,6 +482,28 @@ func (c *StaticConfig) IsRequireTLS() bool {
 	return c.RequireTLS
 }
 
+// GetTLSMinVersionConfig returns the effective tls_min_version, with TLS_MIN_VERSION
+// env var taking precedence over the TOML/CLI value.
+func (c *StaticConfig) GetTLSMinVersionConfig() string {
+	if envValue := os.Getenv(EnvTLSMinVersion); envValue != "" {
+		return envValue
+	}
+	return c.TLSMinVersion
+}
+
+// GetTLSCipherSuitesConfig returns the effective tls_cipher_suites, with
+// TLS_CIPHER_SUITES env var taking precedence over the TOML/CLI value.
+func (c *StaticConfig) GetTLSCipherSuitesConfig() []string {
+	if envValue := os.Getenv(EnvTLSCipherSuites); envValue != "" {
+		suites := strings.Split(envValue, ",")
+		for i, suite := range suites {
+			suites[i] = strings.TrimSpace(suite)
+		}
+		return suites
+	}
+	return c.TLSCipherSuites
+}
+
 func (c *StaticConfig) IsRequireOAuth() bool {
 	return c.RequireOAuth
 }
@@ -556,6 +589,9 @@ func (c *StaticConfig) Validate(ctx context.Context) error {
 		if _, err := os.Stat(c.TLSKey); err != nil {
 			return fmt.Errorf("tls-key must be a valid file path: %w", err)
 		}
+	}
+	if err := c.validateTLSSettings(ctx); err != nil {
+		return err
 	}
 	if err := c.ValidateRequireTLS(); err != nil {
 		return err
@@ -651,6 +687,33 @@ func (c *StaticConfig) validateTokenExchange() error {
 		}
 	default:
 		return fmt.Errorf("invalid sts_auth_style %q: must be %q, %q, %q, or %q", c.StsAuthStyle, tokenexchange.AuthStyleParams, tokenexchange.AuthStyleHeader, tokenexchange.AuthStyleAssertion, tokenexchange.AuthStyleFederated)
+	}
+	return nil
+}
+
+// validateTLSSettings validates TLS settings via the getters so env overrides are included,
+// and logs once at startup/reload when TLS env vars override TOML.
+func (c *StaticConfig) validateTLSSettings(ctx context.Context) error {
+	logger := klogutil.FromContext(ctx).V(1)
+	if os.Getenv(EnvTLSMinVersion) != "" {
+		klogutil.LogInfo(logger, "TLS min version overridden by environment variable",
+			klogutil.Field("env", EnvTLSMinVersion),
+			klogutil.Field("value", c.GetTLSMinVersionConfig()),
+		)
+	}
+	if os.Getenv(EnvTLSCipherSuites) != "" {
+		klogutil.LogInfo(logger, "TLS cipher suites overridden by environment variable",
+			klogutil.Field("env", EnvTLSCipherSuites),
+		)
+	}
+
+	minVersion := c.GetTLSMinVersionConfig()
+	if _, err := tlsutil.ParseTLSVersion(minVersion); err != nil {
+		return err
+	}
+	cipherSuites := c.GetTLSCipherSuitesConfig()
+	if _, err := tlsutil.ParseTLSCipherSuites(cipherSuites); err != nil {
+		return err
 	}
 	return nil
 }
--- a/pkg/kiali/kiali.go
+++ b/pkg/kiali/kiali.go
@@ -2,7 +2,6 @@ package kiali
 
 import (
 	"context"
-	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
@@ -17,6 +16,7 @@ import (
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"github.com/containers/kubernetes-mcp-server/pkg/config"
 	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
+	"github.com/containers/kubernetes-mcp-server/pkg/tlsutil"
 )
 
 type Kiali struct {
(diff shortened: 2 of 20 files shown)


pkg/tlsutil/tlsutil.go
```diff
@@ -0,0 +1,184 @@
+// Package tlsutil provides TLS configuration utilities for outbound HTTPS clients.
```
cajieh: Added to break an import cycle: pkg/config imports pkg/tokenexchange for AuthStyle constants, so tokenexchange can’t import TLS helpersfrom pkg/config.                                                                                                                          
Could refactor that dependency instead (e.g. move AuthStyle elsewhere and keep TLS in pkg/config), but that looks like scope creep for this PR. Would rather do that in a follow-up.

cajieh:
~~The [issue ](https://github.com/containers/kubernetes-mcp-server/issues/1266) says config + env everywhere TLS is used, with env taking precedence. Added TOML+env for the HTTP server, but outbound clients are env-only today (TOML ignored). Should outbound also fall back to global `tls_min_version` / `tls_cipher_suites` from TOML, or is env-only outbound acceptable?~~

pkg/config/config.go
```diff
@@ -459,6 +466,22 @@ func (c *StaticConfig) IsRequireTLS() bool {
...
+}
+
+// GetTLSCipherSuitesConfig returns the tls_cipher_suites value from TOML/CLI only.
+// Environment variable TLS_CIPHER_SUITES is not applied here; callers pass this
+// fallback to tlsutil.BuildTLSConfig, which applies env overrides at runtime.
+// validateTLSSettings resolves env overrides when validating at startup/reload.
+func (c *StaticConfig) GetTLSCipherSuitesConfig() []string {
+	return c.TLSCipherSuites
+}
+
```
Cali0707: IMO we should actually check the env values here, and return whatever the TLS min version or cipher suites are (accounting for overrides)

Long term that feels less error prone than requiring these to go through tlsutils to have the overrides take effect

cajieh: Addressed, thank you for the feedback.

pkg/oauth/state.go
Cali0707: @cajieh you will need to capture and compare the tls values in here as we use those to create the http client - otherwise these wouldn't pick up on any changes

pkg/tlsutil/tlsutil.go
```diff
@@ -0,0 +1,170 @@
...
+	if len(result) == 0 {
+		return nil, nil
+	}
+	return result, nil
+}
+
+// ResolveTLSMinVersion returns the TLS minimum version, with env var taking precedence over config.
+func ResolveTLSMinVersion(configValue string) string {
+	if envValue := os.Getenv(EnvTLSMinVersion); envValue != "" {
+		klog.V(2).Infof("Using TLS min version from %s environment variable: %s", EnvTLSMinVersion, envValue)
```
Cali0707: Let's not always log this (same with where you resolve the cipher suites).

This will be super chatty in the logs, and not the best signal

Maybe we can in the pkg/config.go log somewhere that this value is coming from the env?

IMO this is also an argument for why this resolution should not live in pkg/tlsutil but in pkg/config - it keeps the config values all centralized. 

Maybe we can just add some accessor methods to the config struct to get these values (and resolve the env vs. config struct internally there) ?

pkg/tlsutil/tlsutil.go
```diff
@@ -0,0 +1,170 @@
...
+	if envValue := os.Getenv(EnvTLSCipherSuites); envValue != "" {
+		klog.V(2).Infof("Using TLS cipher suites from %s environment variable", EnvTLSCipherSuites)
+		suites := strings.Split(envValue, ",")
+		for i, suite := range suites {
+			suites[i] = strings.TrimSpace(suite)
+		}
+		return suites
+	}
+	return configValue
+}
```
Cali0707: Same as above

pkg/tlsutil/tlsutil.go
```diff
@@ -0,0 +1,170 @@
...
+		return nil, fmt.Errorf("failed to parse TLS cipher suites: %w", err)
+	}
+
+	tlsConfig := &tls.Config{
+		MinVersion: parsedMinVersion,
+	}
+
+	if len(parsedCipherSuites) > 0 {
+		tlsConfig.CipherSuites = parsedCipherSuites
+		klog.V(2).Infof("TLS configured with %d cipher suites", len(parsedCipherSuites))
```
Cali0707: Here and everywhere else, klog should not be used globally

Instead use `klogutil.FromContext(ctx)` to get a logger

pkg/config/config.go
```diff
@@ -460,6 +467,18 @@ func (c *StaticConfig) IsRequireTLS() bool {
...
+// env var taking precedence over the TOML/CLI value.
+func (c *StaticConfig) GetTLSMinVersionConfig() string {
+	return tlsutil.ResolveTLSMinVersion(c.TLSMinVersion)
+}
+
+// GetTLSCipherSuitesConfig returns the effective tls_cipher_suites, with
+// TLS_CIPHER_SUITES env var taking precedence over the TOML/CLI value.
+func (c *StaticConfig) GetTLSCipherSuitesConfig() []string {
+	return tlsutil.ResolveTLSCipherSuites(c.TLSCipherSuites)
+}
```
Cali0707, line 505: These should have the actual implementations within these methods, not just passing through the call to tlsutil

IMO those functions should not exist in tlsutil at all

pkg/tlsutil/tlsutil.go
```diff
@@ -0,0 +1,170 @@
...
+// tlsCipherSuiteMap maps cipher suite names to their IDs
+var tlsCipherSuiteMap = func() map[string]uint16 {
+	m := make(map[string]uint16)
+	for _, suite := range tls.CipherSuites() {
+		m[suite.Name] = suite.ID
+	}
+	// Also include insecure cipher suites for completeness (users may need them for legacy systems)
+	for _, suite := range tls.InsecureCipherSuites() {
+		m[suite.Name] = suite.ID
+	}
```
Cali0707: IMO let's not include these for now. If someone needs it, we can consider adding back

As a default first pass, let's not use them though

Cali0707 (commented):
Thanks for tackling this @cajieh !

Left a few comments

pkg/config/config.go
cajieh: Oh, I see what you mean now.

pkg/oauth/state.go
cajieh: Good catch. I've updated this section to explicitly compare the `TLSMinVersion` and `TLSCipherSuites` fields.

pkg/config/config.go
cajieh: Updated! The `GetTLSMinVersionConfig()` and `GetTLSCipherSuitesConfig()` now have their actual implementation inline.

pkg/config/config.go
```diff
@@ -460,6 +471,44 @@ func (c *StaticConfig) IsRequireTLS() bool {
...
+		for i, suite := range suites {
+			suites[i] = strings.TrimSpace(suite)
+		}
+		return suites
+	}
+	return c.TLSCipherSuites
+}
+
+// logTLSConfigOverrides logs once at startup/reload when TLS env vars override TOML.
+func (c *StaticConfig) logTLSConfigOverrides(ctx context.Context) {
```
Cali0707: nit: this method exists only to log and is called right before validateTLSSettings - both read the same env vars via the same getters

Let's fold these together to avoid a method just for logging a startup message?

pkg/oauth/state.go
```diff
@@ -97,6 +103,9 @@ func CreateOIDCProviderAndClient(cfg *config.StaticConfig) (*oidc.Provider, *htt
 	ctx := context.Background()
 	var httpClient *http.Client
```
Cali0707: this was a var before since the original branching logic constructed the http client on each branch. Now it is done once after the branching, so we should be able to just define it there with a `:=` assignment

pkg/kiali/kiali.go
```diff
@@ -101,11 +107,15 @@ func (k *Kiali) createHTTPClient(ctx context.Context) *http.Client {
 		caPEM, err := os.ReadFile(caValue)
 		if err != nil {
 			logger.Error(err, "failed to read CA certificate from file, proceeding without custom CA", "ca_file", caValue)
+			tlsConfig, buildErr := tlsutil.BuildTLSConfig(k.tlsMinVersion, k.tlsCipherSuites, tlsOpts...)
+			if buildErr != nil {
+				return nil, fmt.Errorf("failed to build TLS config: %w", buildErr)
+			}
```
Cali0707: we can probably simplify this branch as this early return branch duplicates the normal exit path.

Since the ReadFile failed, no WithRootCAs option gets appended to tlsOpts, so falling through to the shared BuildTLSConfig produces the same result.

I think it would be best to restructure so that ReadFile error logs and skips the rest of the CA block rather than duplicating the client construction

pkg/oauth/state.go
```diff
@@ -106,21 +115,23 @@ func CreateOIDCProviderAndClient(cfg *config.StaticConfig) (*oidc.Provider, *htt
...
 		}
-		if caCertPool.Equal(x509.NewCertPool()) {
-			caCertPool = nil
-		}
-		var transport http.RoundTripper = &http.Transport{
-			TLSClientConfig: &tls.Config{
-				MinVersion: tls.VersionTLS12,
-				RootCAs:    caCertPool,
-			},
+		if !caCertPool.Equal(x509.NewCertPool()) {
```
Cali0707: I don't think we need this check anymore, since we should early return above if we fail to append certs

Cali0707: so we should be fine to just add the tls opt unconditionally

pkg/tlsutil/tlsutil.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, nil
+	}
+	return result, nil
+}
+
+// BuildTLSConfig creates a tls.Config from TLS settings returned by
+// config.GetTLSMinVersionConfig / GetTLSCipherSuitesConfig.
+// Options (e.g., RootCAs, InsecureSkipVerify) are applied last.
+func BuildTLSConfig(minVersion string, cipherSuites []string, opts ...TLSConfigOption) (*tls.Config, error) {
+	minVersionStr := strings.TrimSpace(minVersion)
```
Cali0707: doesn't ParseTLSVersion also trim this string? IMO we don't need to trim twice

Cali0707 (commented):
Thanks @cajieh - just a few nits then this should be good to go!

cajieh:
> Thanks @cajieh - just a few nits then this should be good to go!

@Cali0707 I believe I've addressed all the feedback. Let me know if anything else needs tweaking!

Cali0707:
Thanks @cajieh one more thing - we recently merged a new package that has another outbound client in pkg/netobserv/netobserv.go around lines 99-107

Mind adding support for that client as well in this PR then we can merge?

Cali0707 (approved):
LGTM, thanks for all your work on this @cajieh!!

PR #1272, opened 2026-07-08
feat(kiali): remove patch version from Kiali

josunect:
Update KIALI_VERSION from v2.27.0 to v2.27 so the local Kiali setup tracks the latest patch release in the 2.27 line (e.g. v2.27.1 when available), instead of pinning to a specific patch.

Kiali publishes a floating minor tag on Quay (v2.27) that is updated to point to the newest patch release.


aljesusg (approved):
LGFM

manusa (approved):
LGTM, thx!

PR #1275, opened 2026-07-09
refactor(ci): use timeline API for mcpchecker SHA pinning

Cali0707:
This moves to using the timeline API to track the last SHA before the mcpchecker comment, rather than fetching the sha directly from the PR in the workflow

### DIFF
--- a/.github/workflows/mcpchecker.yaml
+++ b/.github/workflows/mcpchecker.yaml
@@ -58,6 +58,11 @@ jobs:
   check-trigger:
     name: Check if evaluation should run
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      actions: read
+      # Needed to comment on the PR when the TOCTOU check fails
+      pull-requests: write
     # Never run on forks: the evaluation needs model/judge secrets that forks
     # don't have, so it would only ever fail there.
     if: |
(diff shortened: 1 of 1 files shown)


.github/workflows/mcpchecker.yaml
```diff
@@ -69,52 +74,122 @@ jobs:
...
+                  // After a force-push the old commit SHAs are invalid; reset
+                  // and let subsequent committed events repopulate.
+                  lastCommitSha = null;
+                } else if (event.event === 'commented' && event.id === commentId) {
+                  seenComment = true;
+                }
+              }
+
+              if (!seenComment) {
+                core.warning('Could not locate the triggering comment in the PR timeline; ' +
```
grokspawn: This should probably be fatal.  The timeline API is populated asynchronously, so this narrows but does not eliminate the concern window.  
Making it fatal via 
```suggestion
                core.setOutput('should-run', 'false');
                return
```
will make it impossible to run the task outside the appropriate conditions.

.github/workflows/mcpchecker.yaml
```diff
@@ -69,52 +74,122 @@ jobs:
...
       pr-number: ${{ steps.check.outputs.pr-number }}
       pr-sha: ${{ steps.check.outputs.pr-sha }}
       is-pr: ${{ steps.check.outputs.is-pr }}
     steps:
       - name: Check trigger conditions
         id: check
-        # All event-derived values are passed through `env` and referenced as shell
-        # variables. They are never interpolated directly into the `run` script, so
-        # untrusted input (e.g. a comment body) cannot alter the script that executes.
+        uses: actions/github-script@v9
```
grokspawn: pin v9?

.github/workflows/mcpchecker.yaml
```diff
@@ -69,52 +74,123 @@ jobs:
...
+                  lastCommitSha = null;
+                } else if (event.event === 'commented' && event.id === commentId) {
+                  seenComment = true;
+                }
+              }
+
+              if (!seenComment) {
+                core.setOutput('should-run', 'false');
+                return
+                             'TOCTOU check was skipped');
```
grokspawn: My orig comment missed dropping this line.  I suspect the resulting syntax is broken if you don't.  Sorry!

Cali0707:
@manusa I'm merging this one now - feel free to iterate with a follow up PR

PR #1276, opened 2026-07-09
feat(http): add bind address and warnings when on 0.0.0.0 with no tls/auth

### LINKED ISSUE #1264: Add `--bind-address` flag/config field
Currently we always listen on `0.0.0.0`. While this makes sense when running in a containerized environment like k8s, we may sometimes want to bind to localhost directly.

This issue involves adding a config option to pkg/config as well as a cli flag, and wiring all of that into pkg/kubernetes-mcp-server/cmd/ so that this will work correctly, along with a test case verifying that it binds to the new address when specified. By default, the bind address should _not_ change.

Note that this idea was originally discussed with me by @novice-22, and I am sharing it on their behalf here

### LINKED ISSUE #1265: Add startup warn log if using HTTP without auth
It is possible to run the MCP server on HTTP without auth. This is similar to how you can run valkey or nginx without auth.

We want the auth layer to always be explicitly opt in (similar to how http is opt in), as it currently is. However, if http is enabled and no auth is enabled, we should have a warning log that this is the case.

Note that this idea was originally discussed with me by @novice-22, and I am sharing it on their behalf here

Cali0707:
Fixes #1264
Fixes #1265

These are some hardenings to improve the http server, enabling specifying a bind address and clarifying that if you run on 0.0.0.0 without any auth/tls that you get a warning

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -35,12 +35,13 @@ type ToolOverride struct {
 type StaticConfig struct {
 	DeniedResources []api.GroupVersionKind `toml:"denied_resources"`
 
-	LogLevel   int    `toml:"log_level,omitzero"`
-	LogFile    string `toml:"log_file,omitempty"`
-	Port       string `toml:"port,omitempty"`
-	SSEBaseURL string `toml:"sse_base_url,omitempty"`
-	KubeConfig string `toml:"kubeconfig,omitempty"`
-	ListOutput string `toml:"list_output,omitempty"`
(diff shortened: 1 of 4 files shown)


2uasimojo:
/lgtm

Seems right to me :)

manusa (approved):
LGTM, thx!

I'll follow up with a PR to address some minor findings and improve hardening, but nothing blocking this iteration 🚀

PR #1279, opened 2026-07-10
fix(nodes): omit tailLines for file log queries

### LINKED ISSUE #1262: nodes_log fails on file queries: /var/log prefix double-join and tailLines rejected by kubelet
## Summary

The `nodes_log` tool has two defects that make **file-based** node log queries fail against the kubelet's node-log endpoint (`/api/v1/nodes/{name}/proxy/logs`). Both exist on `main` today. They were surfaced while reviewing #419 (closing separately) and are independent of that PR.

## 1. The documented `/var/log/...` example double-prefixes the path

The `nodes_log` tool description tells callers to query files like `"/var/log/kubelet.log"`. The kubelet resolves a file query with [`SecureJoin(nodeLogDir, query)`](https://github.com/kubernetes/kubernetes/blob/70d3cc986aa8221cd1dfb1121852688902d3bf53/pkg/kubelet/kubelet_server_journal.go#L231-L232), where [`nodeLogDir = "/var/log/"`](https://github.com/kubernetes/kubernetes/blob/70d3cc986aa8221cd1dfb1121852688902d3bf53/pkg/kubelet/kubelet_server_journal.go#L48). So `/var/log/kubelet.log` resolves to `/var/log/var/log/kubelet.log` and returns a 404.

The correct form is a path relative to `/var/log` (and, because the kubelet [classifies a query as a file only when it contains a slash](https://github.com/kubernetes/kubernetes/blob/70d3cc986aa8221cd1dfb1121852688902d3bf53/pkg/kubelet/kubelet_server_journal.go#L130), with a leading slash), e.g. `/kubelet.log`.

## 2. `tailLines` on a file query is rejected by the kubelet

The kubelet's `validate()` [rejects options on a file query](https://github.com/kubernetes/kubernetes/blob/70d3cc986aa8221cd1dfb1121852688902d3bf53/pkg/kubelet/kubelet_server_journal.go#L229-L230): `case len(n.Files) == 1 && n.options != (options{}): "cannot specify file with options"`, and `options` includes `TailLines` (per the kubelet, `tailLines` is only supported for *service* logs).

Our tool advertises `tailLines` with a default of `100`, and `Core.NodesLog` forwards it whenever `tailLines > 0`:

```go
// pkg/kubernetes/nodes.go
if tailLines > 0 {
    req.Param("tailLines", fmt.Sprintf("%d", tailLines))
}
```

So a file query that carries `tailLines` gets a 406 with a cryptic error. A caller must explicitly pass `tailLines: 0` for a file query to succeed.

## Proposed fix

1. Correct the `nodes_log` description examples to the real relative form (`/kubelet.log`, not `/var/log/kubelet.log`).
2. Only forward `tailLines` for service-style queries, mirroring the kubelet's own slash heuristic:

```go
// pkg/kubernetes/nodes.go
if tailLines > 0 && !strings.ContainsAny(query, `/\`) {
    req.Param("tailLines", fmt.Sprintf("%d", tailLines))
}
```

3. Add suite coverage: a service query forwards `tailLines`; a file query omits it and succeeds.

ychampion:
## Summary

- stop forwarding `tailLines` for file-based node log queries
- document file paths relative to `/var/log` and cover service/file behavior

## Why

The kubelet rejects options on file queries, so `nodes_log` returned 406 whenever a file path was used with `tailLines`. The existing examples also double-prefixed `/var/log`.

Fixes #1262.

## Validation

- `go test -p 1 ./pkg/mcp -run 'TestNodes/TestNodesLog' -count=1`
- `go test -p 1 -count=1 ./...`
- `make lint`
- `make build`

### DIFF
--- a/pkg/kubernetes/nodes.go
+++ b/pkg/kubernetes/nodes.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -14,10 +15,10 @@ import (
 func (c *Core) NodesLog(ctx context.Context, name string, query string, tailLines int64) (string, error) {
 	// Use the node proxy API to access logs from the kubelet
 	// https://kubernetes.io/docs/concepts/cluster-administration/system-logs/#log-query
-	// Common log paths:
-	// - /var/log/kubelet.log - kubelet logs
-	// - /var/log/kube-proxy.log - kube-proxy logs
-	// - /var/log/containers/ - container logs
+	// File paths are relative to /var/log and must contain a slash:
+	// - /kubelet.log - kubelet logs
+	// - /kube-proxy.log - kube-proxy logs
+	// - /containers/ - container logs
 
 	if _, err := c.CoreV1().Nodes().Get(ctx, name, metav1.GetOptions{}); err != nil {
 		return "", fmt.Errorf("failed to get node %s: %w", name, err)
@@ -27,8 +28,8 @@ func (c *Core) NodesLog(ctx context.Context, name string, query string, tailLine
 		Get().
 		AbsPath("api", "v1", "nodes", name, "proxy", "logs")
 	req.Param("query", query)
-	// Query parameters for tail
-	if tailLines > 0 {
+	// The kubelet only supports tailLines for service log queries.
+	if tailLines > 0 && !strings.ContainsAny(query, `/\`) {
 		req.Param("tailLines", fmt.Sprintf("%d", tailLines))
 	}
 
--- a/pkg/mcp/nodes_test.go
+++ b/pkg/mcp/nodes_test.go
@@ -2,7 +2,6 @@ package mcp
 
 import (
 	"net/http"
-	"strconv"
 	"testing"
 
 	"github.com/BurntSushi/toml"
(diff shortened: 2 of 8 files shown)


Cali0707:
@2uasimojo mind reviewing this one?

pkg/kubernetes/nodes.go
```diff
@@ -27,8 +28,8 @@ func (c *Core) NodesLog(ctx context.Context, name string, query string, tailLine
 		Get().
 		AbsPath("api", "v1", "nodes", name, "proxy", "logs")
 	req.Param("query", query)
-	// Query parameters for tail
-	if tailLines > 0 {
+	// The kubelet only supports tailLines for service log queries.
+	if tailLines > 0 && !strings.ContainsAny(query, `/\`) {
```
2uasimojo, line 32: Do you think we should at least warn when ignoring an input parameter? (I'm actually curious why we don't just error.)

pkg/mcp/nodes_test.go
```diff
@@ -51,18 +50,24 @@ func (s *NodesSuite) TestNodesLog() {
 			query := req.URL.Query().Get("query")
 			var logContent string
 			switch query {
+			case "kubelet":
+				if req.URL.Query().Get("tailLines") != "2" {
+					w.WriteHeader(http.StatusNotAcceptable)
+					return
+				}
+				logContent = "Line 4\nLine 5\n"
```
2uasimojo, line 58: I'm not fully understanding how this test is structured, but this seems tightly coupled to the test case by requiring an exact `tailLines` value. Would it be too painful to do something like (pseudo):

```
logLines = "Line 1\nLine 2\nLine 3\nLine 4\nLine 5\n".Split("\n")
switch tailLines, err := atoi(req.URL.Query().Get("tailLines")) {
case (err != nil, tailLines <= 0): // Is zero acceptable?
  w.WriteHeader(http.StatusNotAcceptable) // invalid numeric input
  return
case tailLines > len(logLines):
  tailLines = len(logLines)
  fallthrough
default:
  logContent = logLines[len(logLines)-tailLines:].Join("\n") // Possible off-by-one; I didn't try too hard :)
}
```

2uasimojo (commented):
This looks right to me based on the issue. Couple questions inline.

PR #1280, opened 2026-07-10
fix(http): extend bind address warning to IPv6 and add tests

manusa:
Follow-up to #1276.

- Extend the all-interfaces warning to also fire for IPv6 `::` (not just `0.0.0.0`)
- Add `TestBindAddress` covering bind address wiring, warning presence for `0.0.0.0`
  and `::`, and warning absence for `127.0.0.1` and OAuth-enabled configs
- Document `bind_address` / `--bind-address` in the TOML reference table, CLI flags
  table, and complete example in `docs/configuration.md`

Refs #1264, #1265


Cali0707 (approved):
Thanks for adding the test cases and the IPV6 bits @manusa 

LGTM

PR #1282, opened 2026-07-10
fix(evals): install claude-agent-acp locally instead of globally

### LINKED ISSUE #1281: make claude-agent-acp does not add install location to PATH
## Problem

`make claude-agent-acp` in `build/evals.mk` installs the binary via `npm install -g`, but if the npm global bin directory (e.g. `~/.npm-global/bin`) is not already on the user's `PATH`, the subsequent `make run-evals AGENT=claude-code` fails with:

```
exec: "claude-agent-acp": executable file not found in $PATH
```

The user has to manually find the binary and add it to their PATH, which is a confusing experience.

## Suggestion

The `claude-agent-acp` target (or the `run-evals` target) could detect the npm global bin directory and either:
- Warn the user if it's not on PATH
- Prepend it to PATH for the eval run
- Document the PATH requirement in the target's help text

## Context

Discovered while running evals locally on a Fedora dev box where npm global prefix was `~/.npm-global`.

dhensel-rh:
## Summary
- Install `claude-agent-acp` under `_output/tools/` (alongside `mcpchecker`) instead of using `npm install -g`
- `run-evals` target prepends the local bin directory to PATH automatically
- Eliminates PATH issues entirely — no global install, no manual PATH setup

Closes #1281

## Test plan
- [ ] Run `make claude-agent-acp` — installs to `_output/tools/node_modules/.bin/claude-agent-acp`
- [ ] Run `make run-evals AGENT=claude-code SUITE=core` — finds `claude-agent-acp` without manual PATH changes
- [ ] Verify `_output/tools/` can be cleaned with `rm -rf _output` as usual

### DIFF
--- a/build/evals.mk
+++ b/build/evals.mk
@@ -45,7 +45,6 @@ claude-agent-acp: ## Install the claude-agent-acp adapter for the claude-code ev
 	@[ -f $(CLAUDE_AGENT_ACP) ] || { \
 		set -e ;\
 		echo "Installing claude-agent-acp@$(CLAUDE_AGENT_ACP_VERSION) to $(CLAUDE_AGENT_ACP)..." ;\
-		mkdir -p $(shell dirname $(CLAUDE_AGENT_ACP))/.. ;\
 		npm install --prefix $(shell pwd)/_output/tools @agentclientprotocol/claude-agent-acp@$(CLAUDE_AGENT_ACP_VERSION) ;\
 		echo "✅ claude-agent-acp installed" ;\
 	}
--- a/build/evals.mk
+++ b/build/evals.mk
@@ -45,7 +45,14 @@ claude-agent-acp: ## Install the claude-agent-acp adapter for the claude-code ev
 		set -e ;\
 		echo "Installing claude-agent-acp@$(CLAUDE_AGENT_ACP_VERSION) (npm install -g)..." ;\
 		npm install -g @agentclientprotocol/claude-agent-acp@$(CLAUDE_AGENT_ACP_VERSION) ;\
-		echo "✅ claude-agent-acp installed" ;\
+		if command -v claude-agent-acp >/dev/null 2>&1; then \
+			echo "✅ claude-agent-acp installed" ;\
+		else \
+			NPM_BIN=$$(npm prefix -g)/bin ;\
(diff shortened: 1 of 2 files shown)


build/evals.mk
```diff
@@ -37,20 +38,21 @@ mcpchecker:
...
+# the claude-code eval agent (evals/claude-code/agent.yaml runs `claude-agent-acp`).
 .PHONY: claude-agent-acp
 claude-agent-acp: ## Install the claude-agent-acp adapter for the claude-code eval agent
-	@command -v claude-agent-acp >/dev/null 2>&1 || { \
+	@[ -f $(CLAUDE_AGENT_ACP) ] || { \
 		set -e ;\
-		echo "Installing claude-agent-acp@$(CLAUDE_AGENT_ACP_VERSION) (npm install -g)..." ;\
-		npm install -g @agentclientprotocol/claude-agent-acp@$(CLAUDE_AGENT_ACP_VERSION) ;\
+		echo "Installing claude-agent-acp@$(CLAUDE_AGENT_ACP_VERSION) to $(CLAUDE_AGENT_ACP)..." ;\
+		mkdir -p $(shell dirname $(CLAUDE_AGENT_ACP))/.. ;\
```
Cali0707: Do we need this mkdir line? I think `npm install --prefix` handles that

Cali0707 (commented):
Thanks @dhensel-rh !

Would you mind:
1. addressing the comment I added
2. updatine evals/README to reflect that this doesn't install claude-agent-acp globally anymore, but installs it locally?

build/evals.mk
dhensel-rh: We do not need this. I have removed that line.

dhensel-rh: @Cali0707 I have addressed both comments. Please have a look and see.

Cali0707:
Looks good @dhensel-rh ! I think you just need to update your last commit to be signed off so that the DCO check passes then this is good to merge

Thanks for your work here!

dhensel-rh:
> Looks good @dhensel-rh ! I think you just need to update your last commit to be signed off so that the DCO check passes then this is good to merge
> 
> Thanks for your work here!

@Cali0707  I have fixed the DCO check.  I think it is ready to merge.

Cali0707 (approved):
LGTM

PR #1285, opened 2026-07-10
fix(kubernetes): bound discovery requests during startup

### LINKED ISSUE #1284: Server hangs at startup when kubeconfig contains an unreachable cluster (v0.0.64)
first of all thanks for this great MCP server! ❤️

Since v0.0.64, `kubernetes-mcp-server` hangs indefinitely at startup when the kubeconfig has multiple contexts and one of the clusters is offline/unreachable. The MCP client's startup timeout (e.g. 30s) kills the server before it ever completes initialization.

I think this is caused by #1196, which made GVK discovery against **all** kubeconfig contexts unconditional. The discovery HTTP client has no timeout, so a request to an unreachable cluster blocks forever.

### Environment

- **Version:** `v0.0.64`
- **OS:** Ubuntu 26.04 (arm64)
- **Transport:** stdio
- **Config:**
  ```toml
  read_only = true
  stateless = true
  toolsets = ["core", "config", "helm"]
  cluster_provider_strategy = "kubeconfig"
  # disable_multi_cluster is unset (multi-cluster mode enabled)
  ```
- **Kubeconfig:** 3 contexts — `prod-cluster` (online, current-context), `test-cluster` (online), `dev-cluster` (offline at `10.10.1.100:6443`)

### Reproduction

1. Create a kubeconfig with at least 2 contexts where one cluster is unreachable (host powered off / network disconnected — TCP SYN gets no response).
2. Start the server:
   ```bash
   kubernetes-mcp-server --config config.toml --kubeconfig kubeconfig.yaml
   ```
3. The server hangs. It never responds to the MCP `initialize` handshake.

<details><summary>Logs</summary>
<p>

With `--log-level 9 --log-file`, the server successfully discovers GVKs against the two reachable clusters, then hangs on the third:

```
I0710 17:27:14.784178  discovery_client.go:255] "Request Body" body=""
I0710 17:27:14.784336  round_trippers.go:527] "Request" curlCommand=<
        curl -v -XGET ... 'https://kube.mydomain.tld:6443/api'        # ← prod-cluster: 200 OK (55ms)
 >
I0710 17:27:14.863853  discovery_client.go:255] "Request Body" body=""
I0710 17:27:14.863908  round_trippers.go:527] "Request" curlCommand=<
        curl -v -XGET ... 'https://192.168.1.5:6443/api'        # ← test-cluster: 200 OK (28ms)
 >
I0710 17:27:14.907916  discovery_client.go:255] "Request Body" body=""
I0710 17:27:14.908050  round_trippers.go:527] "Request" curlCommand=<
        curl -v -XGET ... 'https://10.10.1.100:6443/api'        # ← dev-cluster: HANGS HERE (no response, no timeout)
 >
```

The log ends here — the process is killed by the MCP client's 30s timeout.

</p>
</details>

### Research

PR #1196 replaced the single-target `IsOpenShift()` check with `AnyTargetHasGVKs()`, which iterates over **all** targets from `GetTargetManagers()` and calls `HasGVKs()` on each.

The "permissive on error" design in `AnyTargetHasGVKs()` only helps if `HasGVKs()` **returns** an error. When the cluster endpoint is unreachable and packets are silently dropped (no TCP RST), the HTTP client blocks indefinitely with no timeout — `HasGVKs()` never returns at all, so the fallback never fires.

### What I have tried

| Workaround | Result |
|---|---|
| `timeout: "5s"` in the kubeconfig cluster entry | Log is identical with and without it. The discovery HTTP client does not use the kubeconfig `timeout` field. |
| `disable_multi_cluster = true` | Works, but **defeats multi-cluster mode**. The per-call `context` argument is no longer exposed, which is the reason I enabled multi-cluster mode. |
| Removing the offline context from the kubeconfig | **This is my current workaround** ... works, but requires manual intervention each time a cluster goes on/offline. |

### Possible solutions

1. Apply a timeout to discovery requests (e.g. respect the kubeconfig `timeout` field, or use a sensible default like 5–10s), so that unreachable clusters fail fast and the "assume yes on error" fallback in `AnyTargetHasGVKs()` can fire.
2. Run GVK discovery concurrently with a per-target timeout, so one slow/unreachable target doesn't block the entire startup
3. Make the discovery lazy until a tool that needs it is actually called, rather than blocking startup for all targets.

Thanks in advance

ychampion:
## Summary

- apply a 10-second default to discovery requests when no timeout is configured
- keep the shared Kubernetes HTTP client unchanged for watches and streaming operations

## Why

`NewDiscoveryClientForConfigAndClient` receives a pre-built zero-timeout HTTP client, so client-go's discovery default never reaches the request. An unreachable kubeconfig context can therefore block synchronous GVK filtering during startup indefinitely.

Fixes #1284.

## Validation

- `go test ./pkg/kubernetes -count=1`
- `go test ./pkg/api ./pkg/toolsets/core -count=1`
- `go vet ./pkg/kubernetes`
- `_output/tools/bin/golangci-lint run ./pkg/kubernetes/...`
- `git diff --check`

### DIFF
--- a/pkg/kubernetes/kubernetes_test.go
+++ b/pkg/kubernetes/kubernetes_test.go
@@ -1,0 +1,56 @@
+package kubernetes
+
+import (
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/containers/kubernetes-mcp-server/internal/test"
+	"github.com/containers/kubernetes-mcp-server/pkg/config"
+	"github.com/stretchr/testify/suite"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+type KubernetesTestSuite struct {
+	suite.Suite
+}
+
+func (s *KubernetesTestSuite) TestDiscoveryRequestsHaveDefaultTimeout() {
+	called := false
+	requestTimeout := time.Duration(0)
+	discoveryHandler := test.NewDiscoveryClientHandler()
+	transport := &mockRoundTripper{
+		called: &called,
+		onRequest: func(w http.ResponseWriter, r *http.Request) {
+			deadline, ok := r.Context().Deadline()
+			s.True(ok, "Expected discovery request context to have a deadline")
+			if ok {
+				requestTimeout = time.Until(deadline)
+			}
+			discoveryHandler.ServeHTTP(w, r)
+		},
+	}
+
+	rawConfig := clientcmdapi.NewConfig()
+	k, err := NewKubernetes(
+		s.T().Context(),
+		&config.StaticConfig{},
+		clientcmd.NewDefaultClientConfig(*rawConfig, &clientcmd.ConfigOverrides{}),
+		&rest.Config{Host: "https://cluster.example", Transport: transport},
+	)
+	s.Require().NoError(err)
+	s.T().Cleanup(k.close)
+
+	_, err = k.DiscoveryClient().ServerResourcesForGroupVersion("v1")
+	s.NoError(err)
+	s.True(called, "Expected discovery request to reach the transport")
+	s.GreaterOrEqual(requestTimeout, 9*time.Second)
+	s.LessOrEqual(requestTimeout, 10*time.Second)
+	s.Zero(k.httpClient.Timeout, "Expected the shared Kubernetes client to remain unbounded")
+}
+
+func TestKubernetes(t *testing.T) {
+	suite.Run(t, new(KubernetesTestSuite))
+}
--- a/pkg/kubernetes/accesscontrol_round_tripper_test.go
+++ b/pkg/kubernetes/accesscontrol_round_tripper_test.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/BurntSushi/toml"
@@ -20,11 +21,15 @@ import (
 )
 
 type mockRoundTripper struct {
+	mu        sync.Mutex
 	called    *bool
 	onRequest func(w http.ResponseWriter, r *http.Request)
 }
 
 func (m *mockRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
 	*m.called = true
 	rec := httptest.NewRecorder()
 	m.onRequest(rec, req)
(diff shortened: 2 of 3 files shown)


2uasimojo:
Thanks for addressing this.

What we want is: requests to a down KAS should report an error (after timeout) to HasGVKs(), resulting in permissive-by-default behavior -- i.e. tools being shown rather than hidden -- rather than hanging or killing the server.*

At a preliminary glance I can't tell whether that's what this is doing, but if it is, +1

*For bonus points: thread the requests to the targets in parallel to minimize response time.

pkg/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+type KubernetesTestSuite struct {
+	suite.Suite
+}
+
+func (s *KubernetesTestSuite) TestDiscoveryRequestsHaveDefaultTimeout() {
```
2uasimojo, line 20: The race failures here can be resolved by synchronizing the `mockRoundTripper`:

```diff
diff --git a/pkg/kubernetes/accesscontrol_round_tripper_test.go b/pkg/kubernetes/accesscontrol_round_tripper_test.go
index c1df2275..31ec20fd 100644
--- a/pkg/kubernetes/accesscontrol_round_tripper_test.go
+++ b/pkg/kubernetes/accesscontrol_round_tripper_test.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/BurntSushi/toml"
@@ -20,11 +21,15 @@ import (
 )
 
 type mockRoundTripper struct {
+	mu        sync.Mutex
 	called    *bool
 	onRequest func(w http.ResponseWriter, r *http.Request)
 }
 
 func (m *mockRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
 	*m.called = true
 	rec := httptest.NewRecorder()
 	m.onRequest(rec, req)
```

ychampion: Thanks. Confirmed the synchronized round tripper is on the branch, and `go test -race ./pkg/kubernetes` passes.

ychampion:
Yes, that is the path here: discovery gets a 10-second HTTP timeout, `HasGVKs` returns the timeout error, and `AnyTargetHasGVKs` keeps the permissive fallback. The race fix is also on the branch now.

2uasimojo:
@ychampion With #1293, the target-specific filtering behavior will become opt-in. Please join us in [CNCF slack](https://cloud-native.slack.com/archives/C0AHQJVR725/p1784046077860409)* to help design a more permanent, robust config option.

*See https://communityinviter.com/apps/cloud-native/cncf for an invitation

PR #1293, opened 2026-07-14
Config option for toolset prefiltering

2uasimojo:
A recent change (#1196) introduced the framework for prefiltering toolsets based on the compatibility of the configured target(s). The initial implementation included a filter to expose the `projects_list` tool based on whether any target is OpenShift or not. This functionality was always on, which was problematic in certain setups.

This change gates this functionality behind a config option:

```
experimental_enable_target_compatibility_tool_filters = true
```

As implied by the name, we consider the feature experimental, and will rename/redesign/remove the config option in the future.

Co-Authored-By: claude


Cali0707 (approved):
LGTM, thanks for making this behaviour opt in again @2uasimojo !

PR #1297, opened 2026-07-15
fix(netobserv): harden OpenShift detection and TLS handling

### LINKED ISSUE #1291: netobserv: follow-up hardening (OpenShift detection, TLS, minor cleanups)
Follow-up to #1159. None of these block that PR; grouping them here so they are tracked.

### OpenShift detection

`NewNetObserv` calls `clusterIsOpenShift` (a discovery lookup) on every tool call and ignores the `FilteringProvider` passed to `GetTools`. Detection fails closed to "not OpenShift" on any discovery error, which flips the synthesized URL from `https://` to `http://` and would send the ServiceAccount bearer token in cleartext on a real OpenShift cluster if discovery ever errors. Reuse the OpenShift signal the framework already resolves (the `FilteringProvider` / cluster-state watcher) instead of re-detecting per call. This also aligns with the plan to retire `IsOpenshift`.

### TLS hardening

- No `CheckRedirect` is set, so the token-bearing GET follows redirects. A compromised or MITM'd plugin could 3xx it to an internal target. Restrict or disable redirects.
- A configured `certificate_authority` is unioned with the system trust pool rather than pinned, and a CA read failure proceeds with the system pool instead of failing closed. That weakens pinning for a public-DNS `url`.
- Consider `additionalProperties: false` on the tool input schemas so arbitrary extra query params cannot be forwarded to the backend.

### Minor cleanups

- `applyDefaultsWithStat` in `pkg/netobserv/config.go` logs via `klog.Background()` (no context), so that line has no trace/span correlation. Thread a context through.
- `aggregateBy` in `get_flow_metrics` is both `Required` and carries a `Default`, which is contradictory. Drop the default.
- `stringArg` in `pkg/netobserv/query.go` always returns `true` for its bool, so the ok value is dead.

jpinsonneau:
Fixes https://github.com/containers/kubernetes-mcp-server/issues/1291

### OpenShift detection
- Added `FilteringProvider` as a named field on `ToolHandlerParams` and wired it from the MCP server in `tools_gosdk.go`.
- Replaced per-call discovery in `NewNetObserv` with `isOpenShiftFromProvider()`, which uses `FilteringProvider.AnyTargetHasGVKs()` (same GVK signal as the cluster-state watcher).
- On discovery errors, the framework fails open to OpenShift (HTTPS), avoiding cleartext token leakage from the old fail-closed path.

### TLS hardening
- **Redirects disabled**: `CheckRedirect` returns an error so token-bearing GETs cannot follow 3xx responses.
- **CA pinning**: when `certificate_authority` is set, trust is limited to that CA only (no system pool union).
- **Fail closed**: unreadable or unparseable CA files cause client creation to fail instead of silently falling back.

### Minor cleanups
- `applyDefaultsWithStat` now takes a `context.Context` and logs via `klogutil.FromContext(ctx)`.
- Removed the contradictory `Default` on required `aggregateBy` in `get_flow_metrics`.
- Simplified `stringArg` to return `string` only (the `ok` bool was always true).

### Schema hardening
- Added `additionalProperties: false` on all three NetObserv tool input schemas via a shared `toolInputSchema()` helper.

### Tests
- Updated OpenShift, config, and HTTP client tests.
- Added tests for CA fail-closed behavior and redirect rejection.

### DIFF
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -166,6 +166,7 @@ type ToolHandlerParams struct {
 	context.Context
 	BaseConfig
 	KubernetesClient
+	FilteringProvider FilteringProvider
 	ToolCallRequest
 	ListOutput output.Output
 	Elicitor
--- a/pkg/mcp/tools_gosdk.go
+++ b/pkg/mcp/tools_gosdk.go
@@ -83,12 +83,13 @@ func ServerToolToGoSdkTool(s *Server, tool api.ServerTool) (*mcp.Tool, mcp.ToolH
 		}
 
 		result, err := tool.Handler(api.ToolHandlerParams{
-			Context:          ctx,
-			BaseConfig:       cfg,
-			KubernetesClient: k,
-			ToolCallRequest:  toolCallRequest,
-			ListOutput:       cfg.ListOutput(),
-			Elicitor:         &sessionElicitor{},
+			Context:           ctx,
+			BaseConfig:        cfg,
+			KubernetesClient:  k,
+			FilteringProvider: s.p,
+			ToolCallRequest:   toolCallRequest,
+			ListOutput:        cfg.ListOutput(),
+			Elicitor:          &sessionElicitor{},
 		})
 		if err != nil {
 			return nil, err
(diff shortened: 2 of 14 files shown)


manusa (approved):
LGTM, thx!

PR #1299, opened 2026-07-15
fix!: internalize enabled check for target filtering

Cali0707:
In #1293 we added a config option to disable target filtering. However, currently we wire that interface and check throughout the codebase.

This PR tries to simplify this code and prep for a possibly more expressive flag that goes beyond just enabling/disabling by internalizing the check within the actual filtering provider method.


2uasimojo:
IIUC this is moving the responsibility of checking the config option down into the filtering methods. It seems like that will result in more duplication of logic (when we have more varieties of filter) and I’m not sure what advantage we’re gaining. Perhaps a prototype of the change to the `all/default/none` config design would make it clearer?

PR #1300, opened 2026-07-15
ci(mcpchecker): switch PR trigger to pull_request_review + workflow_run 

Cali0707:
In #1275 we switched to using the timeline API for commit ordering

That approach seems to have some issues (including the fact that the ordering of commit events is not server controlled, while the rest of the events are).

This PR switches to a different trigger (pull request review) as that has an associated commit sha which we can leverage. However given that pull request targets run fork code, we have to split the trigger into a new workflow that triggers the actual mcpchecker action itself

### DIFF
--- a/.github/workflows/mcpchecker.yaml
+++ b/.github/workflows/mcpchecker.yaml
@@ -5,9 +5,10 @@ on:
   schedule:
     - cron: '0 9 * * 1'
 
-  # Manual trigger via PR comments
-  issue_comment:
-    types: [created]
+  # PR evaluation trigger via review comments (chained from mcpchecker-trigger.yaml)
+  workflow_run:
+    workflows: ["mcpchecker MCP Evaluation - Trigger"]
+    types: [completed]
 
   # Allow manual workflow dispatch for testing
   workflow_dispatch:
@@ -35,16 +36,17 @@ on:
         type: boolean
         default: false
 
-# Minimal permissions - no write access to PRs/issues
-# This workflow checks out and runs potentially untrusted PR code
+# Minimal permissions at workflow level. The check-trigger job escalates to
+# pull-requests: write for posting rejection comments on PRs.
 permissions:
   contents: read
   actions: read
 
 concurrency:
   # Only run once for latest commit per ref and cancel other (previous) runs.
-  # For issue_comment events, use PR number as group to avoid different PRs canceling each other.
-  group: ${{ github.workflow }}-${{ github.event_name == 'issue_comment' && format('pr-{0}', github.event.issue.number) || github.ref }}
+  # For workflow_run events, use head_branch to group by PR branch so different
+  # PRs don't cancel each other.
+  group: ${{ github.workflow }}-${{ github.event_name == 'workflow_run' && format('branch-{0}', github.event.workflow_run.head_branch) || github.ref }}
   cancel-in-progress: true
 
 env:
@@ -62,7 +64,7 @@ jobs:
     permissions:
       contents: read
       actions: read
-      # Needed to comment on the PR when the TOCTOU check fails
+      # Needed to comment on the PR when the evaluation is rejected
       pull-requests: write
     # Never run on forks: the evaluation needs model/judge secrets that forks
     # don't have, so it would only ever fail there.
@@ -70,9 +72,8 @@ jobs:
       github.repository == 'containers/kubernetes-mcp-server' &&
       (github.event_name == 'schedule' ||
        github.event_name == 'workflow_dispatch' ||
-       (github.event_name == 'issue_comment' &&
-        github.event.issue.pull_request &&
-        contains(github.event.comment.body, '/run-mcpchecker')))
+       (github.event_name == 'workflow_run' &&
+        github.event.workflow_run.conclusion == 'success'))
     outputs:
       should-run: ${{ steps.check.outputs.should-run }}
       toolsets: ${{ steps.suite.outputs.toolsets }}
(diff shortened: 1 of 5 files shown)


Cali0707:
FYI @grokspawn @manusa

.github/workflows/mcpchecker.yaml
```diff
@@ -249,9 +253,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v7.0.0
         with:
-          # Use pinned SHA to prevent TOCTOU attacks
-          # For PRs: the SHA captured when maintainer commented
-          # For other triggers: the current commit SHA
+          # For PRs: review.commit_id — the server-assigned SHA from the
+          # maintainer's review (immutable, not derived from forgeable metadata).
+          # For other triggers: the current commit SHA.
           ref: ${{ needs.check-trigger.outputs.pr-sha }}
```
manusa, line 264: `actions/checkout@v7.0.0` will refuse this checkout for fork PRs. v7's fork-PR guard (`src/unsafe-pr-checkout-helper.ts`, called from `input-helper.ts`) throws when the event is `workflow_run`, `workflow_run.event` starts with `pull_request` (true for `pull_request_review`), the head repo is a fork, and the checked-out commit matches `workflow_run.head_sha` / `head_commit.id`. When the `currentHead === reviewSha` gate passes, `pr-sha` equals `head_sha`, so this matches and the step fails. There is no `allow-unsafe-pr-checkout: true` here. Triggering the workflow directly on `pull_request_review` (the event is then `pull_request_review`, which the guard does not cover) avoids this and the lookup issue below in one move.

.github/workflows/mcpchecker.yaml
```diff
@@ -88,97 +89,105 @@ jobs:
...
+            if (context.eventName === 'workflow_run') {
+              // PR evaluation path: the trigger workflow fired on
+              // pull_request_review.  We independently verify everything from
+              // the GitHub API — the trigger workflow is untrusted (runs from
+              // the merge ref, so forks can modify it).
+
+              // Find the PR using server-controlled workflow_run metadata.
+              const headBranch = context.payload.workflow_run.head_branch;
+              const headRepoOwner = context.payload.workflow_run.head_repository.owner.login;
+              const { data: prs } = await github.rest.pulls.list({
```
manusa: This reconstructs the PR from `workflow_run.head_branch` / `head_repository`, which is fragile for fork PRs. If `head_repository` is the base repo, `head: '<base-owner>:<fork-branch>'` matches no PR and the eval silently skips. If it is the fork, the checkout is refused instead (see the checkout step). Reading `github.event.pull_request.number` and `github.event.review.commit_id` from a direct `pull_request_review` trigger removes this reconstruction entirely. Also note `head_repository` can be null for a deleted fork, which would throw at `.owner.login` on the line above.

manusa (changes requested):
**Request changes: the fork PR evaluation path will not run as written**

Thanks for iterating on this. The security core is solid: gating on a write/admin reviewer plus pinning checkout to the immutable `review.commit_id` is a clean, non-forgeable replacement for the old timeline walk, and I could not find a way for a non-maintainer to get unreviewed code executed.

My blocking concern is functional, not security. As written, the PR-evaluation path does not run for **fork** PRs, which are the primary case here (this PR is itself from a fork). It also cannot be fully validated until it is on the default branch, because both `pull_request_review` and `workflow_run` read their workflow files and metadata from the default branch.

### Why it breaks (holds either way)

The design derives the PR and arms the checkout from `workflow_run.head_*` metadata of a run whose source event is `pull_request_review`. For a fork PR this fails no matter how GitHub populates that metadata:

1. **If `workflow_run.head_repository` is the fork**, `actions/checkout@v7.0.0` refuses the checkout. v7 added a guard (`src/unsafe-pr-checkout-helper.ts`, called unconditionally from `src/input-helper.ts`) that throws when the event is `workflow_run`, `workflow_run.event` starts with `pull_request` (and `pull_request_review` does), the head repo is a fork, and the checked-out commit matches `workflow_run.head_sha` / `head_commit.id`. A bare 40 character `ref` is parsed into that commit (`input-helper.ts`), and when the `currentHead === reviewSha` gate passes, `pr-sha` equals `head_sha`, so the match is guaranteed. There is an explicit test for this (`__test__/unsafe-pr-checkout-helper.test.ts`: "refuses workflow_run fork PR head_commit.id checkout"). The opt-out `allow-unsafe-pr-checkout: true` is not set here, so the checkout step fails.

2. **If `workflow_run.head_repository` is the base repo**, the guard does not fire, but `pulls.list({ head: '<base-owner>:<fork-branch>', state: 'open' })` matches nothing (the fork PR's head is `<fork-owner>:<fork-branch>`), so `prs.length === 0` and the evaluation never starts.

Either branch means fork PR evals do not run. Same-repo PRs and scheduled runs are unaffected.

I checked real `pull_request_review` workflow runs across several public repos: `workflow_run.head_branch` is the PR's head branch, not the base branch, which is why the lookup is not simply always empty. I could not directly observe a fork-sourced review run's `head_repository`, and that value is exactly what decides which of the two failure modes above you hit. That uncertainty is why this needs a real fork-PR test on the default branch before it can be trusted.

### Suggested fix (removes the dependency entirely)

Trigger `mcpchecker.yaml` directly on `pull_request_review` instead of the trigger plus `workflow_run` split, and read the PR number from `github.event.pull_request.number` and the SHA from `github.event.review.commit_id` (both trusted base-context fields, reliably populated for fork PRs). That:

- drops all reliance on `workflow_run.head_*`, so the lookup cannot miss,
- makes the event `pull_request_review`, which the checkout guard does not act on (it only covers `pull_request_target` and `workflow_run`), so no `allow-unsafe-pr-checkout` is needed,
- keeps the write-access reviewer plus immutable `commit_id` SHA pin unchanged.

`pull_request_review` runs from the default branch with secrets available even for fork PRs, so a fork cannot alter the workflow that executes. That also means the split is not required for safety, and the SECURITY NOTE about running "from the merge ref" is not accurate for this event.

If you would rather keep the `workflow_run` split, it needs `allow-unsafe-pr-checkout: true` on the eval checkout **and** a confirmation on a real fork PR that `head_repository` resolves to the fork so the lookup succeeds. That is strictly more moving parts for no benefit that I can see.

### How maintainers trigger it

Worth pinning down the convention, since the command has to live in a PR **review summary body**, not a plain comment and not an inline line comment:

- **Web UI:** Files changed, click **Review changes**, put `/run-mcpchecker [suite]` in the review comment box, choose a review type, then **Submit review**.
- **CLI:** `gh pr review <num> --comment --body "/run-mcpchecker core"` (or `--approve` when you also mean to approve someone else's PR).
- A plain Conversation **comment** no longer triggers anything, so the netobserv README and `docs/NETOBSERV.md` need updating, and a short "how to trigger" note in the workflow header would help.
- Suggested convention: use a **Comment-type** review as the default trigger. It is neutral and it works on your own PRs too (GitHub does not allow approving your own PR). Reserve **Approve + command** for when you actually mean to approve.
- Editing a review does not re-fire it (only `submitted`), so to re-run, submit a fresh review.

### Smaller notes

I have a few non-blocking items (concurrency key on `head_branch`, a `DISMISSED` review-state casing bug, the unanchored suite regex, and the docs that still say "comment") that I will leave as separate inline comments. None of them block on their own.

Cali0707:
@manusa I don't think this is correct from looking through the docs:
> pull_request_review runs from the default branch with secrets available even for fork PRs, so a fork cannot alter the workflow that executes. That also means the split is not required for safety, and the SECURITY NOTE about running "from the merge ref" is not accurate for this event.

Per the docs:
> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository. The GITHUB_TOKEN has read-only permissions in pull requests from forked repositories. For more information, see [Use GITHUB_TOKEN for authentication in workflows](https://docs.github.com/en/actions/tutorials/authenticate-with-github_token).

So our auth to the model would fail in this scenario...

I will see if there is another way to resolve the issue you were calling out around functionality

manusa (approved):
LGTM, thx!

You're right about the secrets.
That was the only unverified claim the agent did and for some reason that section slipped into my review.

PR #1303, opened 2026-07-16
test: verify that the mcp server works with kuadrant mcp gateway

Cali0707:
Kuadrant MCP gateway is a common gateway users of this MCP server will run the MCP server behind on kubernetes.

This adds a set of e2e tests to verify that we can actually connect to the mcp server through the gateway and see tools.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -34,11 +34,8 @@ jobs:
         with:
           go-version-file: go.mod
 
-      - name: Create Kind cluster
-        run: make kind-create-cluster
-
-      - name: Build and load e2e image
-        run: make e2e-image
+      - name: Setup cluster and build e2e image
+        run: make e2e-full-setup
 
       - name: Run e2e tests
         env:
--- a/Makefile
+++ b/Makefile
@@ -167,6 +167,29 @@ local-env-setup-tekton: ## Setup complete local development environment with Kin
 	@echo "Tekton Pipelines is now available!"
 	@echo "Check status with: make tekton-status"
 
+.PHONY: local-env-setup-kuadrant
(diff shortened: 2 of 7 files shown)


manusa (commented):
Thanks for this, testing against Kuadrant behind CI is a solid addition and nice that it's cheap to run.

A few things I'd like sorted, only the first is really a blocker:

- **Pin Istio** (`build/kuadrant.mk:24,28`). It's the only unpinned install in the file, right under the "Pinned version" comment. `build/kiali.mk` already pins `ISTIO_VERSION`, and leaving these floating means an Istio release can redden unrelated PRs with nothing to bisect.
- **Narrow the `except Exception` -> skip** (`conftest.py:456`). Right now any failure (API timeout, a renamed gateway service) turns into a green skip, and since CI always installs Kuadrant now, the test could silently stop running and nobody'd notice. A 404-only skip that hard-fails in CI would keep the local convenience.
- **Set the httpx client's timeout/redirects** (`conftest.py:125`). Passing your own `http_client` makes the SDK skip `create_mcp_http_client`, so you lose `follow_redirects=True` and the 300s read timeout and fall back to httpx's 5s default. Worth matching the direct-server path before it flakes on a slow run.
- **`kuadrant-uninstall` deletes `istio-system`** (`kuadrant.mk:92`), which `kiali.mk` also installs into. Anyone with a Kiali env loses it. Dropping the namespace deletes (the `helm uninstall`s already do the work) would scope it.

One optional idea: the chart already has an `httpRoute` template that renders the route you're building by hand in the test, and the `extra_values` hook you added makes it a one-liner to use. Would give that template its only e2e coverage. Fine as a follow-up though.

manusa (approved):
LGTM, thx!

PR #1306, opened 2026-07-17
feat(kubevirt): add pause and unpause actions to vm_lifecycle tool

### LINKED ISSUE #1061: feat(kubevirt): add pause/unpause support to vm_lifecycle tool
## Summary

The `vm_lifecycle` tool currently supports `start`, `stop`, and `restart` actions but not `pause` or `unpause`. KubeVirt's pause/unpause functionality freezes a running VMI in place (preserving memory/CPU state), which is distinct from stop (which shuts down the VM entirely).

## Problem

The KubeVirt pause/unpause API uses the `subresources.kubevirt.io` API group rather than the main `kubevirt.io` group:

```
PUT /apis/subresources.kubevirt.io/v1/namespaces/{ns}/virtualmachineinstances/{name}/pause
PUT /apis/subresources.kubevirt.io/v1/namespaces/{ns}/virtualmachineinstances/{name}/unpause
```

The `AccessControlRoundTripper` in `pkg/kubernetes/accesscontrol_round_tripper.go` validates all outgoing requests against the REST mapper. Since `subresources.kubevirt.io` is a virtual API group that doesn't appear in standard API discovery, the round tripper rejects requests to it with:

```
Validation Error [RESOURCE_NOT_FOUND]: Resource virtualmachineinstances.subresources.kubevirt.io does not exist in the cluster
```

This blocks the dynamic client from reaching the pause/unpause endpoints, even though the endpoints exist and work when called directly (e.g. via `kubectl proxy`).

## Possible approaches

1. **Allowlist `subresources.kubevirt.io` in the AccessControlRoundTripper** -- skip REST mapper validation for known KubeVirt subresource API groups
2. **Bypass the round tripper for subresource calls** -- use a separate HTTP client/REST client that doesn't go through the access control layer
3. **Use the KubeVirt client library directly** -- import `kubevirt.io/client-go` and use its typed client for pause/unpause

## Context

The `pause-vm` eval task was removed in #1054 because the tool doesn't support pause. Once this is implemented, the eval task should be re-added.

/cc @ksimon1

lyarwood:
## Summary

Adds `pause` and `unpause` actions to the `vm_lifecycle` tool using KubeVirt's subresource API (`subresources.kubevirt.io/v1`).

With #1361 merged, the `AccessControlRoundTripper` natively resolves subresources without a `Kind` field via the discovery client fallback, allowing requests to KubeVirt's subresource endpoints without requiring explicit API group bypass declarations.

## Changes

- Add `pause` and `unpause` actions to `vm_lifecycle` tool
- Implement `PauseVM` and `UnpauseVM` helper functions in `pkg/kubevirt` using REST calls to the VMI subresource endpoints (`/pause` and `/unpause`)
- Add unit tests for `PauseVM`, `UnpauseVM`, and MCP tool invocation error cases
- Add `pause-vm` and `unpause-vm` mcpchecker eval tasks
- Update tool snapshot and regenerated documentation (`README.md` and `docs/configuration.md`)

## Related

- Enabled by #1361
- Supersedes #884
- Resolves #1061
- JIRA: [CNV-93080](https://redhat.atlassian.net/browse/CNV-93080)

### DIFF
--- a/evals/tasks/kubevirt/unpause-vm/task.yaml
+++ b/evals/tasks/kubevirt/unpause-vm/task.yaml
@@ -1,0 +1,90 @@
+kind: Task
+apiVersion: mcpchecker/v1alpha2
+metadata:
+  labels:
+    suite: kubevirt
+    project: kubevirt
+    requires: kubevirt
+  annotations:
+    project-name: "KubeVirt"
+    project-url: "https://kubevirt.io"
+  name: "unpause-vm"
+  difficulty: easy
+spec:
+  requires:
+    - extension: kubernetes
+      as: k8s
+  setup:
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-unpause
+        ignoreNotFound: true
+    - k8s.create:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-unpause
+    - k8s.create:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachine
+        metadata:
+          name: test-vm
+          namespace: vm-test-unpause
+        spec:
+          runStrategy: Always
+          startStrategy: Paused
+          template:
+            spec:
+              domain:
+                devices:
+                  disks:
+                    - name: containerdisk
+                      disk:
+                        bus: virtio
+                memory:
+                  guest: 2Gi
+              terminationGracePeriodSeconds: 0
+              volumes:
+                - name: containerdisk
+                  containerDisk:
+                    image: quay.io/containerdisks/fedora:latest
+    - k8s.wait:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachineInstance
+        metadata:
+          name: test-vm
+          namespace: vm-test-unpause
+        condition: Paused
+        timeout: 60s
+  verify:
+    - k8s.wait:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachineInstance
+        metadata:
+          name: test-vm
+          namespace: vm-test-unpause
+        condition: Ready
+        timeout: 120s
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          # The Paused condition is removed entirely when unpaused so
+          # k8s.wait cannot be used. Verify the condition is absent.
+          PAUSED=$(kubectl get vmi test-vm -n vm-test-unpause -o jsonpath='{.status.conditions[?(@.type=="Paused")].status}')
+          if [[ "$PAUSED" == "True" ]]; then
+            echo "✗ VirtualMachineInstance still has Paused=True condition"
+            exit 1
+          fi
+          echo "✓ VirtualMachineInstance is unpaused"
+  cleanup:
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-unpause
+        ignoreNotFound: true
+  prompt:
+    inline: |
+      The virtual machine named test-vm in the vm-test-unpause namespace is currently paused. Please unpause it so it resumes running.
--- a/evals/tasks/kubevirt/unpause-vm/task.yaml
+++ b/evals/tasks/kubevirt/unpause-vm/task.yaml
@@ -1,0 +1,101 @@
+kind: Task
+apiVersion: mcpchecker/v1alpha2
+metadata:
+  labels:
+    suite: kubevirt
+    project: kubevirt
+    requires: kubevirt
+  annotations:
+    project-name: "KubeVirt"
+    project-url: "https://kubevirt.io"
+  name: "unpause-vm"
+  difficulty: easy
+spec:
+  requires:
+    - extension: kubernetes
+      as: k8s
+  setup:
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-unpause
+        ignoreNotFound: true
+    - k8s.create:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-unpause
+    - k8s.create:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachine
+        metadata:
+          name: test-vm
+          namespace: vm-test-unpause
+        spec:
+          runStrategy: Always
+          template:
+            spec:
+              domain:
+                devices:
+                  disks:
+                    - name: containerdisk
+                      disk:
+                        bus: virtio
+                memory:
+                  guest: 2Gi
+              terminationGracePeriodSeconds: 0
+              volumes:
+                - name: containerdisk
+                  containerDisk:
+                    image: quay.io/containerdisks/fedora:latest
+    - k8s.wait:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachine
+        metadata:
+          name: test-vm
+          namespace: vm-test-unpause
+        condition: Ready
+        timeout: 300s
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          virtctl pause vm test-vm -n vm-test-unpause
+    - k8s.wait:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachineInstance
+        metadata:
+          name: test-vm
+          namespace: vm-test-unpause
+        condition: Paused
+        timeout: 60s
+  verify:
+    - k8s.wait:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachineInstance
+        metadata:
+          name: test-vm
+          namespace: vm-test-unpause
+        condition: Ready
+        timeout: 120s
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          # The Paused condition is removed entirely when unpaused so
+          # k8s.wait cannot be used. Verify the condition is absent.
+          PAUSED=$(kubectl get vmi test-vm -n vm-test-unpause -o jsonpath='{.status.conditions[?(@.type=="Paused")].status}')
+          if [[ "$PAUSED" == "True" ]]; then
+            echo "✗ VirtualMachineInstance still has Paused=True condition"
+            exit 1
+          fi
+          echo "✓ VirtualMachineInstance is unpaused"
+  cleanup:
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-unpause
+        ignoreNotFound: true
+  prompt:
+    inline: |
+      The virtual machine named test-vm in the vm-test-unpause namespace is currently paused. Please unpause it so it resumes running.
--- a/pkg/kubevirt/vm.go
+++ b/pkg/kubevirt/vm.go
@@ -6,7 +6,11 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
 )
 
 // RunStrategy represents the run strategy for a VirtualMachine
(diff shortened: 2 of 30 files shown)


evals/tasks/kubevirt/unpause-vm/task.yaml
```diff
@@ -0,0 +1,101 @@
...
+        kind: VirtualMachine
+        metadata:
+          name: test-vm
+          namespace: vm-test-unpause
+        condition: Ready
+        timeout: 300s
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          virtctl pause vm test-vm -n vm-test-unpause
```
ksimon1: IINM virtctl is not vailable in this project at the moment

pkg/kubevirt/vm.go
```diff
@@ -144,6 +148,55 @@ func CloneVM(ctx context.Context, dynamicClient dynamic.Interface, namespace, so
...
+	cfg := rest.CopyConfig(restConfig)
+	cfg.GroupVersion = &schema.GroupVersion{Group: "subresources.kubevirt.io", Version: "v1"}
+	cfg.APIPath = "/apis"
+	cfg.NegotiatedSerializer = serializer.NewCodecFactory(runtime.NewScheme())
+	return rest.RESTClientFor(cfg)
+}
+
+// PauseVM pauses a running VirtualMachineInstance via the KubeVirt subresource API
+// and returns the parent VirtualMachine
+func PauseVM(ctx context.Context, dynamicClient dynamic.Interface, restConfig *rest.Config, namespace, name string) (*unstructured.Unstructured, error) {
```
ksimon1, line 160: Would be possible to add unit tests for these functions?

pkg/kubevirt/vm.go
```diff
@@ -144,6 +148,55 @@ func CloneVM(ctx context.Context, dynamicClient dynamic.Interface, namespace, so
 	return result, nil
 }
 
+// newSubresourceClient creates a REST client for the KubeVirt subresources API group
+func newSubresourceClient(restConfig *rest.Config) (rest.Interface, error) {
```
ksimon1, line 150: this function is almost the same as https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/kubevirt/guestagent.go#L41. Would it make sense to unite the code?

pkg/kubevirt/vm.go
```diff
@@ -144,6 +148,55 @@ func CloneVM(ctx context.Context, dynamicClient dynamic.Interface, namespace, so
...
+func PauseVM(ctx context.Context, dynamicClient dynamic.Interface, restConfig *rest.Config, namespace, name string) (*unstructured.Unstructured, error) {
+	client, err := newSubresourceClient(restConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create subresource client: %w", err)
+	}
+	result := client.Put().
+		Namespace(namespace).
+		Resource("virtualmachineinstances").
+		Name(name).
+		SubResource("pause").
```
codingben, line 169: Can you handle the situation that this command is executed, but pause is being done async and the response will be 200? I mean, is it really paused after response arrived, or you need to check if it's really paused before returning here the result?

codingben: What if there a migration in progress? AI model will check the state of the VMI before calling to pause?

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
```diff
@@ -116,8 +118,22 @@ func lifecycle(params api.ToolHandlerParams) (*api.ToolCallResult, error) {
 		}
 		message = "# VirtualMachine restarted successfully\n"
 
+	case ActionPause:
+		vm, err = kubevirt.PauseVM(params.Context, dynamicClient, params.RESTConfig(), namespace, name)
```
codingben, line 122: Do you want to check if VMI is running before calling this?

pkg/kubevirt/vm.go
```diff
@@ -144,6 +148,55 @@ func CloneVM(ctx context.Context, dynamicClient dynamic.Interface, namespace, so
 	return result, nil
 }
 
+// newSubresourceClient creates a REST client for the KubeVirt subresources API group
+func newSubresourceClient(restConfig *rest.Config) (rest.Interface, error) {
+	cfg := rest.CopyConfig(restConfig)
+	cfg.GroupVersion = &schema.GroupVersion{Group: "subresources.kubevirt.io", Version: "v1"}
+	cfg.APIPath = "/apis"
+	cfg.NegotiatedSerializer = serializer.NewCodecFactory(runtime.NewScheme())
```
Cali0707: Is there a reason we don't use a similar pattern in the negotiated serializer as we do in `guestagent.go` here? Since this is diverging from there, I'm wondering if intentional (maybe a comment why if that is the case) or if we can consolidate

pkg/kubevirt/vm.go
Cali0707: +1 on this, @lyarwood can we consolidate this?

pkg/kubevirt/vm_test.go
```diff
@@ -327,3 +329,121 @@ func TestRestartVMNotFound(t *testing.T) {
...
+			serverStatus:  http.StatusInternalServerError,
+			initialVM:     createTestVM("test-vm", "default", RunStrategyAlways),
+			wantError:     true,
+			errorContains: "failed to pause VirtualMachineInstance",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
Cali0707, line 383: Maybe here we can validate a little bit that the path and method look sort of like how we expect? Minimally, something like:
```go
if r.Method != http.MethodPut {
    t.Errorf("expected PUT, got %s", r.Method)
}
if !strings.HasSuffix(r.URL.Path, "/pause") {
    t.Errorf("unexpected path: %s", r.URL.Path)
}
```

pkg/kubevirt/vm_test.go
```diff
@@ -327,3 +329,121 @@ func TestRestartVMNotFound(t *testing.T) {
...
+			serverStatus:  http.StatusInternalServerError,
+			initialVM:     createTestVM("test-vm", "default", RunStrategyAlways),
+			wantError:     true,
+			errorContains: "failed to unpause VirtualMachineInstance",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
Cali0707, line 471: same for this test server as the other one

pkg/mcp/kubevirt_test.go
```diff
@@ -668,6 +668,19 @@ func (s *KubevirtSuite) TestVMLifecycle() {
...
+		for _, action := range []string{"pause", "unpause"} {
+			s.Run("action="+action, func() {
+				toolResult, err := s.CallTool("vm_lifecycle", map[string]interface{}{
+					"name":      "non-existent-vm",
+					"namespace": "default",
+					"action":    action,
+				})
+				s.Nilf(err, "call tool failed %v", err)
+				s.Truef(toolResult.IsError, "expected call tool to fail for non-existent VM")
+				s.Truef(strings.Contains(toolResult.Content[0].(*mcp.TextContent).Text, "failed to"),
```
Cali0707: Can we be a bit more specific on this assertion, since we do know a bit more about the structure we expect it to have? e.g. `"failed to pause VirtualMachineInstance"` or something like that?

pkg/kubevirt/vm.go
Cali0707, line 160: should we add a pre-check here to see if the VM is already paused? or will that not cause any issues? 

For example, if a VM is already paused, would a precheck maybe give us a better error/response message? I'd assume pausing an already paused VM could be succesful in a MCP tool sense if it is already paused, just with something indicating it was already paused

pkg/kubevirt/vm.go
```diff
@@ -144,6 +148,55 @@ func CloneVM(ctx context.Context, dynamicClient dynamic.Interface, namespace, so
...
+		Do(ctx)
+	if err := result.Error(); err != nil {
+		return nil, fmt.Errorf("failed to pause VirtualMachineInstance: %w", err)
+	}
+	return GetVirtualMachine(ctx, dynamicClient, namespace, name)
+}
+
+// UnpauseVM unpauses a paused VirtualMachineInstance via the KubeVirt subresource API
+// and returns the parent VirtualMachine
+func UnpauseVM(ctx context.Context, dynamicClient dynamic.Interface, restConfig *rest.Config, namespace, name string) (*unstructured.Unstructured, error) {
```
Cali0707, line 180: Same as with pause, do we maybe want to check if it is already runnning?

pkg/kubevirt/vm.go
lyarwood: Done — `newSubresourceClient` now lives in `vm.go` and `getVMISubresource` in `guestagent.go` delegates to it, removing the duplicate REST client setup. Also unified the codec to use `subresourcesCodec.WithoutConversion()` in both paths.

pkg/kubevirt/vm.go
lyarwood: Addressed — `newSubresourceClient` now uses `subresourcesCodec.WithoutConversion()` consistent with the pattern in `guestagent.go`, and `guestagent.go` itself now calls `newSubresourceClient` rather than duplicating the setup inline.

pkg/kubevirt/vm.go
lyarwood: Done, see the reply above.

pkg/kubevirt/vm.go
lyarwood: The pause operation is indeed async — the API returns 200 when the request is accepted, not when the VMI is actually paused. Rather than adding polling/wait logic here, we trust the API's response and return the current VM state. The message could be updated to say "pause requested" but for consistency with how `start`/`stop` behave (they return after the runStrategy update, not after the VMI reaches the target state) we leave it as-is.

lyarwood: KubeVirt already validates this — if a migration is in progress, the subresource endpoint returns `409 Conflict` with message `"cannot pause VMI with a live migration in progress"`. That error is propagated directly to the model as `"failed to pause VirtualMachineInstance: cannot pause VMI with a live migration in progress"`. We've added a unit test case for this.

pkg/toolsets/kubevirt/vm/lifecycle/tool.go
lyarwood: Intentionally not adding a pre-check here — KubeVirt already validates the VMI state before accepting the request and returns appropriate errors (404 if the VMI doesn't exist, 409 if it can't be paused). Duplicating those checks would be redundant. The errors are propagated directly to the model with the KubeVirt-provided reason.

pkg/kubevirt/vm_test.go
lyarwood: Done — both test server handlers now assert `r.Method == http.MethodPut` and that the path has the correct suffix (`/pause` or `/unpause`).

pkg/kubevirt/vm_test.go
lyarwood: Done, same pattern applied to the unpause test server.

pkg/mcp/kubevirt_test.go
lyarwood: Done — the assertion now uses `fmt.Sprintf("failed to %s VirtualMachineInstance", action)` so it matches `"failed to pause VirtualMachineInstance"` or `"failed to unpause VirtualMachineInstance"` depending on the action under test.

pkg/kubevirt/vm.go
lyarwood: Not adding a pre-check — KubeVirt returns `409 Conflict` with `"VM is already paused"` if you try to pause an already-paused VMI, and that error message is propagated directly to the model. We've added a unit test case for this. Adding an explicit pre-check would duplicate what the API already enforces.

pkg/kubevirt/vm.go
lyarwood: Same — KubeVirt returns `409 Conflict` with `"VMI is not paused"` if you try to unpause a running VMI. Unit test case added for this too.

evals/tasks/kubevirt/unpause-vm/task.yaml
lyarwood: Good catch — the `virtctl` step has been replaced. The eval task now creates the VM with `startStrategy: Paused` and waits for the VMI's `Paused` condition to be true before prompting the model to unpause it.

pkg/kubevirt/vm.go
lyarwood: Tests are now in `pkg/kubevirt/vm_test.go` covering: successful pause/unpause, VMI not found (404), VM already paused (409), migration in progress (409), and VM not paused (409) — each with a proper Kubernetes `Status` JSON body so the REST client parses the error reason correctly.

lyarwood:
@Cali0707 thanks for the review again, hopefully everything is addressed now.

2uasimojo (commented):
LGTM

PR #1308, opened 2026-07-17
feat(kubernetes): add AllowedAPIGroups support to toolset API

### LINKED ISSUE #1307: feat(kubernetes): add AllowedAPIGroups support to toolset API for AccessControlRoundTripper
## Summary

Toolsets that need to interact with API groups not discoverable via the REST mapper (e.g. `subresources.kubevirt.io`) are currently blocked by the `AccessControlRoundTripper`, which rejects requests to unknown groups with `RESOURCE_NOT_FOUND`.

The current blocking behaviour is a security requirement ([ref](https://github.com/containers/kubernetes-mcp-server/pull/884#discussion_r2995105983)). Rather than weakening the round tripper by passing through unknown groups, toolsets should be able to explicitly declare which additional API groups they need.

## Proposal

Add an `AllowedAPIGroups` method or field to the `Toolset` interface that returns a list of API groups the toolset needs access to. The `AccessControlRoundTripper` would check these declarations before rejecting requests to groups not found in the REST mapper.

For example:

```go
type Toolset interface {
    // ... existing methods ...
    
    // AllowedAPIGroups returns API groups that should be allowed through
    // the AccessControlRoundTripper even if they are not discoverable via
    // the REST mapper (e.g. virtual subresource API groups).
    AllowedAPIGroups() []schema.GroupVersion
}
```

The kubevirt toolset would then declare:

```go
func (t *KubeVirtToolset) AllowedAPIGroups() []schema.GroupVersion {
    return []schema.GroupVersion{
        {Group: "subresources.kubevirt.io", Version: "v1"},
    }
}
```

The `AccessControlRoundTripper` would collect allowed groups from all enabled toolsets at startup and check against them before rejecting a request for an unknown group. The `denied_resources` list would still take precedence over allowed groups.

## Why this is needed

- KubeVirt's pause/unpause API uses `subresources.kubevirt.io`, a virtual API group not in standard API discovery
- The guest agent queries also use this group
- PR #884 / #1306 attempted to add pause/unpause support but the round tripper blocks the subresource requests
- Other toolsets may face similar issues with non-standard API groups in the future

## Design considerations

- Allowed groups should be explicit and auditable — no blanket pass-through
- `denied_resources` must take precedence over allowed groups
- The mechanism should work with the existing `denied_resources` field-level redaction (#1155)
- Consider whether this should be at the `Toolset` level or `ServerTool` level

## Related

- #1061 — pause/unpause support blocked by AccessControlRoundTripper
- #884 — previous attempt (pass-through approach, rejected by reviewers)
- #1306 — draft PR for pause/unpause (blocked on this issue)

lyarwood:
## Summary

Add an `AllowedAPIGroupsProvider` interface that allows toolsets to explicitly declare API groups that should bypass REST mapper validation in the `AccessControlRoundTripper`. This addresses the [reviewer feedback](https://github.com/containers/kubernetes-mcp-server/pull/884#discussion_r2995105983) from @Cali0707 and @manusa that exceptions to the round tripper's security model should be explicit and declared through the toolset API, not a blanket pass-through.

## How it works

- `Toolset` interface gains `GetAllowedAPIGroups() []string`
- `AllowedAPIGroupsProvider` interface follows the `DeniedResourcesProvider` pattern exactly
- At startup, allowed groups are collected from all enabled toolsets and stored on `StaticConfig`
- The `AccessControlRoundTripper` checks allowed groups before rejecting requests to API groups not in the REST mapper
- `denied_resources` still takes precedence for groups that ARE in the REST mapper
- Requests to allowed virtual groups also bypass HTTP validators (schema, RBAC pre-check, and confirmation rules) since validators require a resolved GVK. The API server still enforces its own authz.
- Empty-string groups are rejected defensively in both `collectAllowedAPIGroups` and `isAPIGroupAllowed` to prevent accidental matches against core API resources.

## KubeVirt use cases

This PR adds the infrastructure. The kubevirt toolset currently returns `nil` from `GetAllowedAPIGroups()` — concrete use cases are in follow-up PRs:

- **#1306** — adds pause and unpause actions to the vm_lifecycle tool using KubeVirt's subresource API
- **#1331** — declares `subresources.kubevirt.io` for guest agent queries (`guestosinfo`, `filesystemlist`, `userlist`, `interfacelist`) and pause/unpause endpoints
- **#1321** — `vm_create_from_template` tool using `subresources.template.kubevirt.io` (should add its group via the same mechanism rather than bypassing the round tripper)

## Design notes

- The pattern is intentionally identical to `DeniedResourcesProvider` for consistency.
- `AllowedAPIGroups` has `toml:"-"` so it cannot be set via config files — it is computed purely from toolset code.

## Testing

- New test cases in `accesscontrol_round_tripper_test.go`:
  - Allowed group not in REST mapper → passes through
  - Empty-string allowed group does not match core API resources
  - Unknown group not in allowed list → rejected (existing behaviour preserved)
- Unit tests for `collectAllowedAPIGroups` covering nil input, unknown toolsets, dedup, and empty-string filtering

## Related

- Resolves #1307
- Unblocks #1306 (pause/unpause draft PR)
- Addresses feedback from #884
- Follow-up: #1331 (kubevirt guest agent), #1321 (virt-template)
- JIRA: [CNV-93085](https://redhat.atlassian.net/browse/CNV-93085)

### DIFF
--- a/pkg/kubernetes-mcp-server/cmd/root.go
+++ b/pkg/kubernetes-mcp-server/cmd/root.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/cli-runtime/pkg/genericiooptions"
 	"k8s.io/kubectl/pkg/util/i18n"
 	"k8s.io/kubectl/pkg/util/templates"
@@ -233,6 +234,8 @@ func (m *MCPServerOptions) Complete(ctx context.Context, cmd *cobra.Command) err
 		klogutil.FromContext(ctx).Error(otelLogErr, "Failed to create OTel log provider, log export disabled")
 	}
 
+	m.StaticConfig.AllowedAPIGroups = collectAllowedAPIGroups(m.StaticConfig.Toolsets)
+
 	if m.StaticConfig.RequireOAuth && m.StaticConfig.Port == "" {
 		// RequireOAuth is not relevant flow for STDIO transport
 		m.StaticConfig.RequireOAuth = false
@@ -437,6 +440,8 @@ func (m *MCPServerOptions) setupSIGHUPHandler(
 				continue
 			}
 
+			newConfig.AllowedAPIGroups = collectAllowedAPIGroups(newConfig.Toolsets)
+
 			// Apply the new configuration to the MCP server first — if this fails,
 			// we skip the OAuth state and config state updates to avoid inconsistent state.
 			if err := mcpServer.ReloadConfiguration(ctx, newConfig); err != nil {
@@ -489,4 +494,36 @@ func (m *MCPServerOptions) setupSIGHUPHandler(
 		close(sigHupCh)
 		<-done // Wait for goroutine to finish
 	}
+}
+
+// collectAllowedAPIGroups collects the union of API groups declared by all
+// enabled toolsets. If every toolset returns nil (no restrictions declared),
+// the result is nil, meaning any API group may pass through the REST mapper
+// NoMatchError path. Once any toolset returns a non-nil slice the result is
+// non-nil and restricted to the declared groups.
+func collectAllowedAPIGroups(toolsetNames []string) []string {
+	seen := sets.New[string]()
+	var groups []string
+	anyDeclared := false
+	for _, name := range toolsetNames {
+		ts := toolsets.ToolsetFromString(name)
+		if ts == nil {
+			continue
+		}
+		toolsetGroups := ts.GetAllowedAPIGroups()
+		if toolsetGroups == nil {
+			continue
+		}
+		anyDeclared = true
+		for _, group := range toolsetGroups {
+			if !seen.Has(group) {
+				seen.Insert(group)
+				groups = append(groups, group)
+			}
+		}
+	}
+	if !anyDeclared {
+		return nil
+	}
+	return groups
 }
--- a/pkg/kubernetes-mcp-server/cmd/root.go
+++ b/pkg/kubernetes-mcp-server/cmd/root.go
@@ -233,6 +233,8 @@ func (m *MCPServerOptions) Complete(ctx context.Context, cmd *cobra.Command) err
 		klogutil.FromContext(ctx).Error(otelLogErr, "Failed to create OTel log provider, log export disabled")
 	}
 
+	m.StaticConfig.AllowedAPIGroups = collectAllowedAPIGroups(m.StaticConfig.Toolsets)
+
 	if m.StaticConfig.RequireOAuth && m.StaticConfig.Port == "" {
 		// RequireOAuth is not relevant flow for STDIO transport
 		m.StaticConfig.RequireOAuth = false
@@ -437,6 +439,8 @@ func (m *MCPServerOptions) setupSIGHUPHandler(
 				continue
 			}
 
+			newConfig.AllowedAPIGroups = collectAllowedAPIGroups(newConfig.Toolsets)
+
 			// Apply the new configuration to the MCP server first — if this fails,
 			// we skip the OAuth state and config state updates to avoid inconsistent state.
 			if err := mcpServer.ReloadConfiguration(ctx, newConfig); err != nil {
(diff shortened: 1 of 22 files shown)


lyarwood:
@manusa @Cali0707 I'm taking another swing at allowing this to unblock the pause/unpause use cases with kubevirt. I've hopefully understood your feedback from https://github.com/containers/kubernetes-mcp-server/pull/884 but let me know if not!

codingben (approved):
/lgtm

Can we get this PR merged to unblock others?

ksimon1:
@lyarwood would you please fix the formatting so the tests are happy?

codingben:
@Cali0707 @manusa Can you please review it?

pkg/kubernetes-mcp-server/cmd/root.go
```diff
@@ -502,3 +494,24 @@ func (m *MCPServerOptions) setupSIGHUPHandler(
 		<-done // Wait for goroutine to finish
 	}
 }
+
+func collectAllowedAPIGroups(toolsetNames []string) []string {
+	seen := make(map[string]struct{})
```
2uasimojo: Consider using [Set](https://github.com/containers/kubernetes-mcp-server/blob/77b4ce2ac70db7282a4e3eee09576f8cb9092a62/vendor/k8s.io/apimachinery/pkg/util/sets/set.go#L25) instead of rolling your own.

2uasimojo (commented):
How does this interact with denied resources?

I don't understand how this will ever allow core resources to come through.

pkg/kubernetes/accesscontrol_round_tripper.go
```diff
@@ -103,6 +106,9 @@ func (rt *AccessControlRoundTripper) RoundTrip(req *http.Request) (*http.Respons
 	gvk, err := restMapper.KindFor(gvr)
 	if err != nil {
 		if meta.IsNoMatchError(err) {
+			if rt.isAPIGroupAllowed(gvr.Group) {
+				return rt.delegate.RoundTrip(req)
+			}
 			return nil, &api.ValidationError{
 				Code:    api.ErrorCodeResourceNotFound,
 				Message: fmt.Sprintf("Resource %s does not exist in the cluster", api.FormatResourceName(&gvr)),
```
2uasimojo, line 114: I know there's a philosophy in security-land that distinguishing between, "You're not allowed to do that," and "Daaah, I have no idea what you're talking about," supposedly gives up some kind of information that could be exploited in theory. Is that really applicable here? I can see the phone ringing with, "WDYM doesn't exist in the cluster? `oc get` shows it right here!"

lyarwood:
> How does this interact with denied resources?
>
> I don't understand how this will ever allow core resources to come through.

`AllowedAPIGroups` and `denied_resources` operate at completely different levels in the round tripper:

- **`AllowedAPIGroups`** only kicks in when the REST mapper returns a `NoMatchError` — i.e. the resource isn't in the cluster's REST mapper at all. This is specifically for virtual/subresource API groups like `subresources.kubevirt.io` that never appear in the mapper. If the group bypasses that check, the request goes straight to the API server (which still enforces its own authz).
- **`denied_resources`** is checked via `isAllowed()` *after* a successful GVK resolution, so it only applies to resources that *are* in the REST mapper. These two paths are mutually exclusive.

As for core API resources (pods, services, etc.) — their group is the empty string `""`. `isAPIGroupAllowed` explicitly returns `false` for empty-string groups, so there's no way for a toolset to accidentally whitelist core resources through this mechanism.

2uasimojo:
> As for core API resources (pods, services, etc.) — their group is the empty string `""`. `isAPIGroupAllowed` explicitly returns `false` for empty-string groups, so there's no way for a toolset to accidentally whitelist core resources through this mechanism.

Right, this is my point: the default behavior appears to be to exclude (blacklist) core resources. Which might make sense for kubevirt, but is definitely Bad™ for e.g. the "core" toolset.

I guess I would have expected a design like:
- distinguish between implemented and not-implemented (e.g. `nil` vs `[]`)
- if not implemented, there are no restrictions (including core resources)
- if implemented, allow only specified groups, which *can* include `""` (core)

lyarwood:
Closing in favour of #1361, which solves the same problem at the round tripper level via a discovery fallback — no toolset changes required. I've tested the pause/unpause use case from #1306 on top of #1361 and it works correctly on OCP. There is one issue with #1361 on clusters that use aggregated discovery (OCP 4.x) that I've commented on there.

PR #1310, opened 2026-07-17
chore: always set read-only to false in evals server

Cali0707:
Our evals are a little inconsistent in different environments as they rely on default configuration in the MCP server to have read only be false. That is not always true and can lead to the evals incorrectly failing, as they generally require write access


manusa (approved):
LGTM, thx!

PR #1312, opened 2026-07-20
feat(Kiali): Kiali multicluster evals

josunect:
## Summary

Adds multicluster Kiali mcpchecker evals against a primary-remote Kind setup (`east` home cluster, `west` remote cluster).

- **Evals:** 6 tasks under `evals/tasks/kiali/multicluster/` covering mesh status, listing mesh clusters, remote metrics/workloads/topology, and traces (agent must identify the remote cluster; judge checks for `west`).
- **Makefile (`build/kiali.mk`):** `setup-kiali-multicluster`, `run-evals-multicluster`, `redeploy-kiali-multicluster-dev`, `kind-delete-multicluster`. Uses Kiali dev/master (`KIALI_MC_VERSION=dev`) for `list_clusters`. CI fix: `SHELL := /bin/bash` for `pipefail` recipes.
- **CI:** New `run-evaluation-multicluster` job when suite is `kiali`: spins up east/west Kind, starts MCP with multicluster config, runs the 6 tasks.
- **Docs:** `docs/KIALI.md` updated with setup/run instructions and troubleshooting.

** Requires the changes in https://github.com/containers/kubernetes-mcp-server/pull/1224 to work

### DIFF
--- a/.github/workflows/mcpchecker.yaml
+++ b/.github/workflows/mcpchecker.yaml
@@ -80,7 +80,6 @@ jobs:
       pr-number: ${{ steps.check.outputs.pr-number }}
       pr-sha: ${{ steps.check.outputs.pr-sha }}
       is-pr: ${{ steps.check.outputs.is-pr }}
-      suite: ${{ steps.suite.outputs.suite }}
     steps:
       - name: Check trigger conditions
         id: check
@@ -443,7 +442,7 @@ jobs:
     needs: check-trigger
     if: |
       needs.check-trigger.outputs.should-run == 'true' &&
-      needs.check-trigger.outputs.suite == 'kiali'
+      needs.check-trigger.outputs.label-selector == 'suite=kiali'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
@@ -474,6 +473,7 @@ jobs:
           MCP_CONFIG_DIR: dev/config/mcp-configs-multicluster
 
       - name: Run mcpchecker multicluster evaluation
+        id: mcpchecker
         uses: mcpchecker/mcpchecker/.github/actions/mcpchecker-action@v0.0.18
         with:
           eval-config: 'evals/tasks/kiali/multicluster/eval.yaml'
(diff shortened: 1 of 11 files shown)


.github/workflows/mcpchecker.yaml
```diff
@@ -417,6 +417,142 @@ jobs:
           path: eval-context/
           retention-days: 1
 
+  run-evaluation-multicluster:
+    name: Run MCP Multicluster Evaluation
+    needs: check-trigger
+    if: |
+      needs.check-trigger.outputs.should-run == 'true' &&
+      needs.check-trigger.outputs.label-selector == 'suite=kiali'
```
cajieh: Curious about this gate — does `check-trigger` currently expose a `label-selector` output? On `main` I’m only seeing `should-run`, `matrix`, `pr-number`, `pr-sha`, and `is-pr`. If `label-selector` isn’t set, would this job always be skipped?
Also for weekly `schedule` runs, suite selection looks like a matrix (`schedule-matrix`) rather than a single `suite=kiali` — how is multicluster supposed to be triggered in that case?

.github/workflows/mcpchecker.yaml
```diff
@@ -509,3 +645,90 @@ jobs:
...
+
+  # Kiali multicluster results only (suite=kiali scheduled runs)
+  commit-multicluster-results:
+    name: Commit Kiali Multicluster Results
+    needs: [check-trigger, run-evaluation-multicluster]
+    if: |
+      always() &&
+      github.event_name == 'schedule' &&
+      needs.check-trigger.outputs.label-selector == 'suite=kiali' &&
+      needs.run-evaluation-multicluster.result == 'success'
```
cajieh: Same question as on `run-evaluation-multicluster`: if `label-selector` isn’t an output of `check-trigger`, would this scheduled commit job ever run?

.github/workflows/mcpchecker.yaml
```diff
@@ -509,3 +645,90 @@ jobs:
...
+          cp "$MC_RESULTS_FILE" "evals/results/openai-agent-multicluster-latest.json"
+
+      - name: Create Pull Request
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TRIGGER: ${{ github.event_name }}
+          COMMIT_SHA: ${{ needs.check-trigger.outputs.pr-sha }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: |
+          BRANCH="chore/update-eval-results"
```
cajieh: Is it intentional to reuse `chore/update-eval-results` here? The existing results committer uses the same branch with `git push -f` — could weekly kiali + multicluster race and overwrite each other? Would a separate branch name be safer?

cajieh (commented):
Added a few comments, but I still need to do another pass. 
Als, should these 6 commits be squashed, or is keeping them separate intentional?

.github/workflows/mcpchecker.yaml
josunect: Good catch. That was written against the pre-matrix workflow. After rebasing onto `main`, `check-trigger` no longer exposes `label-selector`.

Updated to a `run-multicluster` output set when the selected suite is `kiali`, `all`, or `schedule-matrix` (weekly runs). The multicluster job now gates on `run-multicluster == 'true'`, so it still runs on weekly schedules and on `/run-mcpchecker kiali`.

.github/workflows/mcpchecker.yaml
josunect: Agreed on the race risk with two jobs force-pushing the same branch.

To avoid that, multicluster results are now copied inside the existing `commit-results` job (single branch / single force-push) instead of a second `commit-multicluster-results` job.

josunect:
@cajieh thanks for the review! Related the commits, I'll squash on merge as usual.

josunect:
Tested here: https://github.com/josunect/kiali-mcp-server/actions/runs/31079711001/job/92549385437

PR #1314, opened 2026-07-20
chore(evals): refactor to suite specific eval configs

Cali0707:
This continues on initial work a while ago that started to introduce the concept of per agent/suite eval files, rather than one large eval config file.

The other diff here is that we want to always run the core + config tests against each toolset as well, to make sure no tools introduce confusion


Cali0707:
@manusa my plan is to follow this up with a change to the weekly job to run the different toolset evals in a matrix, so that our job comparisons go against the toolset baseline, not the core baseline

Currently, if I run mcpchecker for e.g. kubevirt it would compare that result against `core`, which is not correct

manusa (approved):
LGTM, thx!

My review agent tells me:

Double checked the symlinks and they resolve the agent config from their own dir (not the target), so the acp/anthropic/google dirs still run their own agents. Per-suite assertions match what the old combined eval ran too, and making the CSI install unconditional is the right call now that core runs in every suite.

Couple of follow-ups:
1. The weekly job still defaults to the `core` suite, so it only regenerates the `core` baseline and the other suites won't have a `builtin-openai-<suite>-latest.json` to diff against (the diff step just skips). The matrix change you mentioned will cover that.
2. `evals/README.md` and `build/evals.mk` still point at `evals/openai-agent/eval.yaml` as what CI runs, but CI is on `core-eval-testing/builtin-openai/` now. Worth repointing those (and dropping the stale `openai-agent-latest.json` baseline) so local `make run-evals` lines up with CI.

PR #1316, opened 2026-07-20
fix(http): reject unknown wellknown paths in proxy

Cali0707:
This is a defense in depth hardening improvement, that makes the handler for the wellknown proxy reject urls that are not on the allowlist.

Note that when require oauth is enabled, these requests would get rejected earlier anyways.

### DIFF
--- a/pkg/http/wellknown.go
+++ b/pkg/http/wellknown.go
@@ -157,8 +157,8 @@ func (w *WellKnown) ServeHTTP(writer http.ResponseWriter, request *http.Request)
 
 	// Validate that the request targets a known well-known endpoint to prevent
 	// proxying arbitrary paths to the upstream authorization server.
-	if !isKnownWellKnownPath(requestPath) {
-		http.Error(writer, "Unknown well-known endpoint", http.StatusNotFound)
+	if !isWellKnownPath(requestPath) {
+		http.Error(writer, "Not a well-known endpoint", http.StatusNotFound)
 		return
 	}
 
@@ -415,9 +415,13 @@ func copyMap(src map[string]interface{}) map[string]interface{} {
 	return dst
 }
 
-// isKnownWellKnownPath reports whether path matches one of the known
+// isWellKnownPath reports whether path matches one of the known
 // well-known endpoints (exact match or sub-path, e.g. /sse suffix).
-func isKnownWellKnownPath(path string) bool {
+// Paths containing ".." are rejected to prevent path traversal bypasses.
+func isWellKnownPath(path string) bool {
+	if strings.Contains(path, "..") {
+		return false
+	}
 	for _, ep := range WellKnownEndpoints {
 		if path == ep || strings.HasPrefix(path, ep+"/") {
 			return true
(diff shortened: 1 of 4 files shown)


grokspawn (approved):

pkg/http/wellknown.go
```diff
@@ -155,6 +155,13 @@ func (w *WellKnown) ServeHTTP(writer http.ResponseWriter, request *http.Request)
 
 	requestPath := request.URL.EscapedPath()
 
+	// Validate that the request targets a known well-known endpoint to prevent
+	// proxying arbitrary paths to the upstream authorization server.
+	if !isKnownWellKnownPath(requestPath) {
+		http.Error(writer, "Unknown well-known endpoint", http.StatusNotFound)
```
2uasimojo: This may very well be standard phrasing, but "[un]known well-known" reads weird/redundant to me. Would something like
```suggestion
	// Validate that the request targets a well-known endpoint to prevent
	// proxying arbitrary paths to the upstream authorization server.
	if !isWellKnownPath(requestPath) {
		http.Error(writer, "Not a well-known endpoint", http.StatusNotFound)
```
...be acceptable?

pkg/http/wellknown.go
```diff
@@ -408,6 +415,17 @@ func copyMap(src map[string]interface{}) map[string]interface{} {
 	return dst
 }
 
+// isKnownWellKnownPath reports whether path matches one of the known
+// well-known endpoints (exact match or sub-path, e.g. /sse suffix).
+func isKnownWellKnownPath(path string) bool {
```
2uasimojo: Is there an opportunity to DRY [this chunk](https://github.com/containers/kubernetes-mcp-server/blob/bc4aba54e406f8aaf4fc9a2f11b988b06c5a0ea5/pkg/http/authorization.go#L64-L66) with this func?

pkg/http/wellknown_test.go
```diff
@@ -257,6 +257,47 @@ func (s *WellknownSuite) TestOAuthScopesOverride() {
 	})
 }
 
+func (s *WellknownSuite) TestUnknownWellKnownPath() {
+	s.StaticConfig.RequireOAuth = true
```
2uasimojo: Having trouble reconciling this with your PR message, which implies that when this setting is on, your new code either isn't exercised or is redundant. Did I misunderstand?

pkg/http/wellknown.go
Cali0707: Fixed

pkg/http/wellknown.go
Cali0707: Nice catch, yes this works well there

pkg/http/wellknown_test.go
Cali0707: Ah so this is where I found that the defense in depth was actually happening - I moved this into the authorization test suite as it makes more logical sense there.

The main thing this new change is protecting is for when the require oauth setting is false, but I wasn't aware of this oauth protection until I enabled it and I didn't see it being explicitly tested elsewhere. As such, keeping the test, just moving the test suite.

2uasimojo (commented):
/lgtm

pkg/http/wellknown_test.go
2uasimojo: Looks like you also removed that setting, so :+1:

PR #1320, opened 2026-07-21
feat(e2e): rework e2e suite and add full OAuth scenario coverage

dlom:
* Replace the Python/pytest e2e test suite with native Go tests using the
mcp-go SDK, eliminating the Python toolchain dependency (pyproject.toml,
uv.lock, conftest.py). The new Go tests cover smoke, image spec,
Keycloak OIDC, and Kuadrant gateway scenarios.

* Migrate the local dev environment from Kind to Minikube: remove
build/kind.mk and dev/config/kind/, add build/minikube.mk. Simplify
the Keycloak setup by replacing individual JSON config files
(clients, client-scopes, mappers, realm, users) with a single
realm-import.yaml and a TOML config. Extract shared tool-install
logic into build/tools.mk.

* Update CI workflows, Makefile targets, and documentation to reflect
the new test runner and cluster tooling.

* Add six new e2e tests covering the OAuth surface end-to-end against a
real Keycloak authorization server:

	- TestOAuthOIDCFlows: rejection of missing, malformed, expired, and
	  wrong-audience tokens; unauthenticated well-known metadata endpoints;
	  scope-not-enforced guard
	- TestOAuthServerModes: server behavior across OIDC / non-OIDC modes
	- TestOAuthForwardedIdentity: forwarded-token passthrough
	- TestOAuthSTSVariants: STS token-exchange variations
	- TestOAuthGroupRBAC: OIDC groups claim driving Kubernetes RBAC
	- TestOAuthSTSAssertion: RFC 7523 private_key_jwt client authentication

* Extend the Keycloak realm with a mcp-server-jwt client (private_key_jwt
auth), an mcp-viewers group with a non-admin user, and a group-based
ClusterRoleBinding for RBAC tests. Add a keypair generation script
(gen-sts-assertion-keypair.sh) that renders the public cert into the
realm at install time.

* Extract shared test helpers (oidcServerConfig, mcpInitProbe,
fetchExchangedToken, mintJWT) from inline code into reusable functions.


test/e2e/test_auth_feature_flags.py
```diff
@@ -0,0 +1,126 @@
+"""E2e tests for MCP server feature flags.
+
+Each test deploys the server via Helm with a specific TOML config and
+verifies the resulting behavior through the MCP protocol.
+"""
+
+import pytest
+
+
+WRITE_TOOLS = {"apply_resource", "delete_resource", "patch_resource", "create_resource"}
```
dlom: **Bug (CONFIRMED):** `WRITE_TOOLS` is a hand-maintained subset of 4 tool names, but the server's read-only filtering is annotation-driven (`ReadOnlyHint`). This set is already out of sync — tools like `run_task`, `cancel_task_run`, and `create_vm` are also mutating but not listed here. A new mutating tool missing its `ReadOnlyHint` annotation would not be caught by this test.

Consider dynamically deriving the expected write-tool set from the server's full tool list (tools present in unrestricted mode but absent in read-only mode) rather than maintaining a static allowlist:
```suggestion
# Derived dynamically below — do not maintain a static set.
# WRITE_TOOLS is computed as tools_unrestricted - tools_readonly in the test body.
```

test/e2e/test_auth_entra.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+    async with server.connect_mcp_with_auth(token) as session:
+        result = await session.list_tools()
+        assert len(result.tools) > 0
+
+
+async def test_entra_oidc_fallback(deploy_server, entra_mock):
+    """Server falls back to openid-configuration when oauth-authorization-server returns 404.
+
+    Looks redundant with test_entra_obo_exchange because the mock always
```
dlom: **Efficiency (CONFIRMED):** This test deploys a full server (~30-60s) but exercises the exact same code path as `test_entra_obo_exchange` — the mock unconditionally returns 404 for the `oauth-authorization-server` endpoint, so the OIDC fallback is always triggered. The test's own docstring acknowledges this redundancy.

To add real value, the mock would need to conditionally return 200 for the OBO test and 404 for the fallback test, so they exercise genuinely different server-side paths. As-is, this is wasted CI time without additional coverage.

test/e2e/conftest.py
```diff
@@ -378,6 +460,133 @@ async def _wait_for_healthz(url: str, timeout: float = 30.0) -> None:
...
+
+def ca_namespace_setup(ca_cert_path):
+    """Return an async callable that creates the Keycloak CA cert Secret."""
+
+    async def setup(core_v1, namespace):
+        await create_ca_cert_secret(core_v1, namespace, ca_cert_path)
+
+    return setup
+
+
```
dlom: **Cleanup (PLAUSIBLE):** `merge_extra_values` shallow-merges dict values (`{**result[key], **val}`), which means if a caller passes `extra_values={"image": {"repository": "foo"}}` intending to fully override the base image config, the base keys (`registry`, `pullPolicy`, `version`) would leak through. No current caller triggers this, but the semantics are surprising — `update()` replaces the whole key, while this silently merges sub-keys.

Consider documenting the merge semantics or using `val` directly for dict values when the caller intends a full override.

test/e2e/test_auth_entra.py
```diff
@@ -0,0 +1,174 @@
...
+# ---------------------------------------------------------------------------
+
+_ENTRA_OAUTH_KNOWN_KEYS = frozenset({
+    "require_oauth", "authorization_url", "oauth_audience",
+    "skip_jwt_verification", "token_exchange_strategy",
+    "sts_client_id", "sts_client_secret",
+})
+
+
+def _entra_oauth_toml(
```
dlom: **Cleanup (PLAUSIBLE):** `_entra_oauth_toml`, `_oauth_toml` (in `test_auth_keycloak.py:30`), and the inline TOML block in `test_multicluster.py:305` all follow the same structural pattern (known-keys frozenset, conflict check, build lines list, extend with extras, join). While the field sets diverge (Entra-specific vs Keycloak-specific vs multicluster+RFC8693), the boilerplate (conflict checking, `toml_extra_lines` integration) could be factored into a shared base builder in `conftest.py`, with each site providing only its unique fields.

dlom:
**Bug (CONFIRMED) — `test/e2e/conftest.py:652`:** The port-forward cleanup at line 652 uses bare `proc.wait()` without a timeout after `proc.kill()`, unlike all three other cleanup sites (lines 231, 279, 470) which pass `timeout=10`. This inconsistency could cause the test teardown to hang indefinitely if the killed process fails to exit.

This line is outside the PR diff but was introduced in this commit's changes to the kuadrant gateway teardown path.

```python
# Current (line 652):
            proc.wait()

# Should be:
            proc.wait(timeout=10)
```

dlom:
so that's what that button does

test/e2e/fixtures/entra_mock.py
```diff
@@ -0,0 +1,249 @@
...
+            return web.Response(status=404)
+        return web.json_response(self._discovery_response())
+
+    async def _handle_openid_configuration(self, request: web.Request):
+        return web.json_response(self._discovery_response())
+
+    async def _handle_jwks(self, request: web.Request):
+        return web.json_response({"keys": [self._jwk]})
+
+    async def _handle_token(self, request: web.Request):
```
nader-ziada: The handler validates grant_type, assertion, and requested_token_use but never checks client_id/client_secret, so test_entra_obo_auth_style can't actually verify that sts_auth_style="params" vs "header" places credentials correctly

test/e2e/test_auth_entra.py
```diff
@@ -0,0 +1,164 @@
...
+    client_id = "00000000-0000-0000-0000-000000000000"
+    client_secret = os.environ["AZURE_CLIENT_SECRET"]
+
+    token = entra_mock.get_user_token()
+    server = await deploy_server(
+        "entra-bad-client",
+        _entra_oauth_toml(issuer, client_id, client_secret),
+    )
+    status = await server.raw_mcp_request(token)
+    assert status in (401, 403, 500), (
```
nader-ziada: accepting `500` means a server crash passes the test

test/e2e/fixtures/keycloak.py
```diff
@@ -0,0 +1,158 @@
...
+
+    Skips all dependent tests if Keycloak is not reachable.
+    """
+    client = KeycloakClient(keycloak_ca_cert_path)
+    try:
+        await client.get_user_token()
+    except Exception as exc:
+        await client.close()
+        pytest.skip(
+            f"Keycloak not reachable at {KEYCLOAK_URL}: {exc}"
```
nader-ziada: Any exception including TypeError, AttributeError, KeyError from a bug in KeycloakClient itself gets caught and presented as "Keycloak not reachable," silently skipping all Keycloak tests. Maybe narrow to connectivity errors:
`except (aiohttp.ClientError, ssl.SSLError, OSError, asyncio.TimeoutError) as exc:`

test/e2e/test_auth_feature_flags.py
```diff
@@ -0,0 +1,133 @@
...
+        tool_names = [t.name for t in result.tools]
+        assert "delete_resource" not in tool_names, (
+            "delete_resource should not be listed when it is disabled"
+        )
+        assert len(tool_names) > 0, (
+            "expected other tools to still be present"
+        )
+
+
+async def test_stateless_mode(deploy_server):
```
nader-ziada: The test asserts len(result.tools) > 0 but doesn't verify any observable stateless behaviour. Maybe check that stateful tools (e.g., configuration_set_context) are absent from the tool list, similar to how test_read_only_mode verifies write tools are filtered.

test/e2e/test_startup_validation.py
```diff
@@ -0,0 +1,226 @@
...
+
+def test_kubeconfig_mode_with_sts(server_binary):
+    _assert_fails_with(
+        server_binary,
+        """\
+cluster_auth_mode = "kubeconfig"
+sts_audience = "openshift"
+""",
+        "incompatible with cluster_auth_mode",
+    )
```
nader-ziada: Tests cover invalid cluster_auth_mode (line 86), invalid sts_auth_style (line 119), and invalid confirmation_fallback (line 197), but not invalid token_exchange_strategy. The Go validation code at pkg/config/config.go:662 does validate this

nader-ziada (commented):
great addition to the testing, added a few comments

test/e2e/fixtures/entra_mock.py
dlom: The mock now extracts creds from either the params or the headers and validates both correctly. The test also verifies that `sts_auth_style` places the creds in the right location

test/e2e/test_auth_feature_flags.py
dlom: Added some assertions for verifying observable stateless behavior (checking `listChanged` in stateless mode)

test/e2e/test_startup_validation.py
dlom: Added a test case for this

dlom:
@Cali0707 @nader-ziada can/should I get GH action workflow automatic approval?

Cali0707:
@dlom I think the actions will run automatically after your first contribution merges, not sure how to make that easier...

Cali0707:
@dlom looks like we had some failures in the new e2e code, mind looking into it?

dlom:
close/reopen to trigger GH actions properly...

dlom:
yeah idk, trying in another PR maybe

dlom:
#1350

build/e2e.mk
```diff
@@ -48,17 +52,56 @@ helm:
...
+	$(KIND) load image-archive _output/e2e-image.tar --name $(KIND_CLUSTER_NAME)
+	@rm -f _output/e2e-image.tar
 
 .PHONY: e2e-full-setup
-e2e-full-setup: kind-create-cluster kuadrant-setup e2e-image ## Create Kind cluster, install Kuadrant MCP Gateway, and build e2e image
+e2e-full-setup: build kind-create-cluster ## Build binary, create Kind clusters, install all infra, and build e2e images
+	-$(MAKE) kind-create-secondary-cluster kind-multicluster-kubeconfig
+	@# Restore kubectl context to primary cluster (secondary creation switches it)
+	-kubectl config use-context kind-$(KIND_CLUSTER_NAME) >/dev/null 2>&1
+	-$(MAKE) keycloak-install
```
Cali0707: I think currently this is also being called in the github action e2e workflow, so this is being double installed?

build/kind.mk
```diff
@@ -69,10 +88,87 @@ kind-create-cluster: kind kind-create-certs
 	$(KIND) export kubeconfig --name $(KIND_CLUSTER_NAME) --kubeconfig _output/kubeconfig; \
 	echo "✅ Kubeconfig exported to _output/kubeconfig"
 
+SECONDARY_CLUSTER_NAME = $(KIND_CLUSTER_NAME)-2
+
+.PHONY: kind-create-secondary-cluster
+kind-create-secondary-cluster: kind kind-create-certs ## Create a secondary Kind cluster for multi-cluster testing
```
Cali0707: If I understand the ordering of everything here (pls push back if I am incorrect), the secondary cluster here would be created with the placeholder CA (as `_output/cert-manager-ca/ca.crt` is generated by `hack/generate-placeholder-ca.sh` at this point)`

`keycloak-install` runs later and overwrites that file with the real cert-manager CA, but the secondary cluster's kube-apiserver has already started and cached the placeholder CA for OIDC validation. Unlike the primary cluster (which gets an explicit apiserver restart inside keycloak-install), there's no corresponding restart for the secondary cluster.

If OIDC-authenticated requests against the secondary are made, the api server would reject them since it would still trust the placeholder cert, not the real one

Cali0707 (commented):
@dlom mind fixing the merge conflict?

Cali0707:
E2E tests passed 🎉

@dlom the unit test failure seems like a flake, don't worry about that one

dlom:
/retitle feat(e2e): rework e2e suite and add full oauth scenario coverage

Cali0707 (approved):
LGTM, thanks for all the had work iterating on these tests @dlom !!

dlom:
(just a rebase)

manusa:
I'll try to review this one later.
I have a question on the motivation to replace Kind with Minikube. I'm a regular user of Minikube myself (https://github.com/manusa/actions-setup-minikube :)), but I'm unsure why this change was needed.

Cali0707:
@manusa iirc the main reasoning for minikube is that once we add multi cluster networking for the multi cluster scenarios, it was getting difficult to have it all work with kind

Cali0707:
@manusa I'm going to merge this one now to start getting more e2e signals on other PRs. Feel free to add more comments here/open issues and @dlom  and i can tackle as follow ups

dlom:
Yes, I'm running a rootless podman setup and kind's networking on my host was completely borked. Switching to minikube eliminated all the networking-related issues I was having, especially concerning DNS and certs

PR #1321, opened 2026-07-21
feat(kubevirt): add vm_create_from_template tool

codingben:
Add MCP tool to create VirtualMachines from
VirtualMachineTemplates via the virt-template
process subresource API, with eval tasks
and tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,7 +273,7 @@ and only needed for the project-specific scenarios noted.
 | [Istio](https://istio.io) | `kiali` | 5 |
 | [Kiali](https://kiali.io) | `kiali` | 16 |
 | [Kubernetes](https://kubernetes.io) | - | 32 |
-| [KubeVirt](https://kubevirt.io) | `kubevirt`, `tekton` | 19 |
+| [KubeVirt](https://kubevirt.io) | `kubevirt`, `tekton` | 24 |
 | [NetObserv](https://netobserv.io) | `netobserv` | 4 |
 | [Tekton](https://tekton.dev) | `tekton` | 9 |
 
@@ -565,6 +565,11 @@ In case multi-cluster support is enabled (default) and you have access to multip
   - `action` (`string`) **(required)** - The lifecycle action to perform: 'start' (changes runStrategy to Always), 'stop' (changes runStrategy to Halted), or 'restart' (stops then starts the VM)
   - `name` (`string`) **(required)** - The name of the virtual machine
   - `namespace` (`string`) **(required)** - The namespace of the virtual machine
+
+- **vm_create_from_template** - Create a VirtualMachine from a VirtualMachineTemplate (virt-template) on KubeVirt. Processes the template server-side to substitute parameters (required values, defaults, and auto-generated values like passwords), then creates the resulting VirtualMachine in the same namespace. Cross-namespace template usage is not supported.
+  - `namespace` (`string`) **(required)** - The namespace of the VirtualMachineTemplate and the resulting VirtualMachine (must be the same)
+  - `parameters` (`object`) - Parameter values to substitute in the template. Keys are parameter names (e.g. VM_NAME), values are strings. Required parameters must be provided; optional parameters use their defaults if omitted; parameters with generate/from will be auto-generated if not provided.
+  - `template_name` (`string`) **(required)** - The name of the VirtualMachineTemplate to create the VM from
 
 - **vm_troubleshoot** - Diagnose KubeVirt VirtualMachine issues with automated root-cause detection. Collects VM status, VMI status, volumes, DataVolume/PVC state, cloud-init configuration, pod state, logs, and events, then runs heuristic checks to identify specific problems and suggest fixes. Returns a 'Detected Issues' section with CRITICAL/WARNING findings and actionable remediation steps, followed by raw diagnostic data. Use this tool FIRST whenever a user asks why a VM is not starting, stuck in Provisioning, crashlooping, failing to migrate, or exhibiting unexpected behavior. Automatically detects: missing StorageClasses, invalid PVC specs, dangerous cloud-init commands (shutdown/halt), nodeSelector migration blockers, failed migrations, and pod crashloops. If the user asks to fix or remediate the issue, use the Suggested Fixes from the report with vm_lifecycle (restart) or resources_create_or_update.
   - `name` (`string`) **(required)** - The name of the VirtualMachine to troubleshoot
--- a/README.md
+++ b/README.md
@@ -273,7 +273,7 @@ and only needed for the project-specific scenarios noted.
 | [Istio](https://istio.io) | `kiali` | 5 |
 | [Kiali](https://kiali.io) | `kiali` | 16 |
 | [Kubernetes](https://kubernetes.io) | - | 32 |
-| [KubeVirt](https://kubevirt.io) | `kubevirt`, `tekton` | 19 |
+| [KubeVirt](https://kubevirt.io) | `kubevirt`, `tekton` | 24 |
 | [NetObserv](https://netobserv.io) | `netobserv` | 4 |
 | [Tekton](https://tekton.dev) | `tekton` | 9 |
 
@@ -565,6 +565,11 @@ In case multi-cluster support is enabled (default) and you have access to multip
   - `action` (`string`) **(required)** - The lifecycle action to perform: 'start' (changes runStrategy to Always), 'stop' (changes runStrategy to Halted), or 'restart' (stops then starts the VM)
   - `name` (`string`) **(required)** - The name of the virtual machine
   - `namespace` (`string`) **(required)** - The namespace of the virtual machine
+
+- **vm_create_from_template** - Create a VirtualMachine from a VirtualMachineTemplate (virt-template) on KubeVirt. Processes the template server-side to substitute parameters (required values, defaults, and auto-generated values like passwords), then creates the resulting VirtualMachine in the same namespace. Cross-namespace template usage is not supported.
+  - `namespace` (`string`) **(required)** - The namespace of the VirtualMachineTemplate and the resulting VirtualMachine (must be the same)
+  - `parameters` (`object`) - Parameter values to substitute in the template. Keys are parameter names (e.g. VM_NAME), values are strings. Required parameters must be provided; optional parameters use their defaults if omitted; parameters with generate/from will be auto-generated if not provided.
+  - `template_name` (`string`) **(required)** - The name of the VirtualMachineTemplate to create the VM from
 
 </details>
 
--- a/evals/tasks/kubevirt/create-template-from-vm/task.yaml
+++ b/evals/tasks/kubevirt/create-template-from-vm/task.yaml
@@ -1,0 +1,104 @@
+kind: Task
+apiVersion: mcpchecker/v1alpha2
+metadata:
+  labels:
+    suite: kubevirt
+    project: kubevirt
+    requires: kubevirt
+  annotations:
+    project-name: "KubeVirt"
+    project-url: "https://kubevirt.io"
+  name: "create-template-from-vm"
+  difficulty: medium
+spec:
+  requires:
+    - extension: kubernetes
+      as: k8s
+  setup:
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vmt-test-from-vm
+        ignoreNotFound: true
+    - k8s.create:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vmt-test-from-vm
+    - k8s.create:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachine
+        metadata:
+          name: source-web-server
+          namespace: vmt-test-from-vm
+          labels:
+            app: web-server
+            tier: frontend
+        spec:
+          runStrategy: Halted
+          template:
+            metadata:
+              labels:
+                app: web-server
+            spec:
+              domain:
+                cpu:
+                  cores: 2
+                memory:
+                  guest: 2Gi
+                devices:
+                  disks:
+                    - name: containerdisk
+                      disk:
+                        bus: virtio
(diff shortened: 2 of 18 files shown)


pkg/kubevirt/template.go
```diff
@@ -0,0 +1,57 @@
...
+	if len(parameters) > 0 {
+		processOptions["parameters"] = parameters
+	}
+
+	body, err := json.Marshal(processOptions)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal ProcessOptions: %w", err)
+	}
+
+	result := &unstructured.Unstructured{}
```
0xFelix, line 42: You could perhaps use the types from virt-template-api package instead of unstructured

pkg/kubevirt/template.go
```diff
@@ -0,0 +1,57 @@
...
+	err = restClient.Post().
+		Namespace(namespace).
+		Resource("virtualmachinetemplates").
+		Name(templateName).
+		SubResource("process").
```
0xFelix: If you want to create directly, you could also use the `create` subresource. Would avoid having to create from the returned YAML manually in the tool definition

pkg/toolsets/kubevirt/vm/template/tool.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return api.NewToolCallResult("", fmt.Errorf("failed to create VirtualMachine from template: %w", err)), nil
+	}
+
+	marshalledYaml, err := output.MarshalYaml(resources)
+	if err != nil {
+		return api.NewToolCallResult("", fmt.Errorf("failed to marshal created VirtualMachine: %w", err)), nil
+	}
+
+	return api.NewToolCallResult("# VirtualMachine created from template successfully\n"+marshalledYaml, nil), nil
```
0xFelix: With the `create` subresource, you could drop these lines

pkg/toolsets/kubevirt/vm/template/tool.go
```diff
@@ -0,0 +1,140 @@
...
+							},
+						},
+					},
+					Required: []string{"namespace", "template_name"},
+				},
+				Annotations: api.ToolAnnotations{
+					Title:           "Virtual Machine: Create from Template",
+					ReadOnlyHint:    ptr.To(false),
+					DestructiveHint: ptr.To(true),
+					IdempotentHint:  ptr.To(true),
```
0xFelix: Might not be idempotent? Depends on what happens when an existing VM is updated

pkg/kubevirt/template.go
ksimon1: We currently do not have any kubevirt package imported in this project. Might be worth checking if virt-template package is easy to import in this PR

README.md
```diff
@@ -273,7 +273,7 @@ and only needed for the project-specific scenarios noted.
 | [Istio](https://istio.io) | `kiali` | 5 |
 | [Kiali](https://kiali.io) | `kiali` | 16 |
 | [Kubernetes](https://kubernetes.io) | - | 32 |
-| [KubeVirt](https://kubevirt.io) | `kubevirt`, `tekton` | 19 |
+| [KubeVirt](https://kubevirt.io) | `kubevirt`, `tekton` | 24 |
```
ksimon1, line 276: Is this number correct? Since you added just a single tool

pkg/kubevirt/template.go
```diff
@@ -0,0 +1,57 @@
+package kubevirt
```
ksimon1, line 1: would be possible to add unit tests for this file?

pkg/kubevirt/gvr.go
```diff
@@ -103,6 +103,31 @@ var (
...
+var (
+	// VirtualMachineTemplateGVK is the GroupVersionKind for VirtualMachineTemplate resources
+	VirtualMachineTemplateGVK = schema.GroupVersionKind{
+		Group:   "template.kubevirt.io",
+		Version: "v1beta1",
+		Kind:    "VirtualMachineTemplate",
+	}
+
+	// VirtualMachineTemplateGVR is the GroupVersionResource for VirtualMachineTemplate resources
+	VirtualMachineTemplateGVR = schema.GroupVersionResource{
```
ksimon1: is this GVR used anywhere?

ksimon1:
You should probably add similar test as is here: https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/mcp/toolsets_test.go#L129

pkg/kubevirt/gvr.go
```diff
@@ -103,6 +103,31 @@ var (
...
+		Group:    "template.kubevirt.io",
+		Version:  "v1beta1",
+		Resource: "virtualmachinetemplates",
+	}
+)
+
+// HasVirtualMachineTemplate returns a TargetCompatibilityFilter that checks whether any
+// target cluster has the VirtualMachineTemplate GVK registered.
+func HasVirtualMachineTemplate(p api.FilteringProvider) func() bool {
```
ksimon1, line 132: You should add unit test for this fn

evals/tasks/kubevirt/create-template-from-vm/task.yaml
```diff
@@ -0,0 +1,104 @@
+kind: Task
```
ksimon1, line 1: Is possible to share some code (store it in evals/tasks/kubevirt/helpers) in these eval tasks in other eval tasks introduced by this PR?

README.md
codingben: It's auto-generated by `make update-readme-tools` and represents the amount of KubeVirt eval scenarios and not the tools.

pkg/kubevirt/template.go
codingben: Yes.

pkg/kubevirt/gvr.go
codingben: Nope.

pkg/kubevirt/gvr.go
codingben: Done.

evals/tasks/kubevirt/create-template-from-vm/task.yaml
codingben: Yes.

pkg/toolsets/kubevirt/vm/template/tool.go
codingben: Might be.

pkg/toolsets/kubevirt/vm/template/tool.go
codingben: Awesome thanks

pkg/kubevirt/template.go
codingben: Yes would be better.

pkg/kubevirt/template.go
codingben: @lyarwood can drop your opinion about this?

lyarwood: Something for a follow up but I'd assume we want to stay unstructured for now.

lyarwood (commented):
Thanks for this @codingben, a few pieces of feedback:

## Rebase on #1308 and route through `AccessControlRoundTripper`

`CreateVMFromTemplate` in `pkg/kubevirt/template.go` creates its own REST client via `rest.CopyConfig(restConfig)`, which means the request to `subresources.template.kubevirt.io` completely bypasses the `AccessControlRoundTripper`. This is the same pattern as the existing guest agent code in `guestagent.go`, but I think we should take the opportunity to fix this going forward.

#1308 adds `AllowedAPIGroups` support to the toolset API — toolsets can declare virtual API groups (like `subresources.kubevirt.io`) that should be allowed through the `AccessControlRoundTripper` even though they're not in standard API discovery. This PR would be a great second use case for that mechanism.

I'd suggest:
1. Rebase this PR on #1308
2. Add `"subresources.template.kubevirt.io"` to the kubevirt toolset's `GetAllowedAPIGroups()` return value
3. Rework `CreateVMFromTemplate` to go through the standard Kubernetes client (which has the `AccessControlRoundTripper` installed) rather than building its own `rest.RESTClientFor`

That way the request goes through the same access control path as everything else — `denied_resources`, confirmation rules, and any future validators would all apply. The `AllowedAPIGroups` mechanism lets it past the REST mapper check, and the API server still enforces RBAC on the user's credentials.

## Evals in a separate commit

I'd like to see the eval tasks (the 5 task YAMLs + `verify-vm.sh` helpers) in their own commit, separate from the tool implementation. This way the evals demonstrate and justify the need for the new `vm_create_from_template` tool before the implementation commit adds it. It also makes it easier to review the evals and tool independently.

Note that only 2 of the 5 evals (`create-vm-from-template`, `template-required-params`) actually exercise the new tool — the other 3 (`list-vm-templates`, `inspect-vm-template`, `create-template-from-vm`) test the agent's ability to work with template CRDs via existing generic tools, which is fine for broader coverage but worth calling out.

codingben:
@lyarwood I'd like to avoid rebasing this PR on yours, can we get the PR you opened merged faster instead? It seems not much changes, no reason to keep it open for longer.

codingben:
@lyarwood Your PR is closed now. How we should proceed here?

codingben:
This PR is now blocked by: https://github.com/containers/kubernetes-mcp-server/pull/1361

codingben:
PR rebased. @lyarwood PTAL

evals/tasks/kubevirt/create-template-from-vm/task.yaml
```diff
@@ -0,0 +1,104 @@
...
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vmt-test-from-vm
+        ignoreNotFound: true
+  prompt:
+    inline: |
+      A VirtualMachine named source-web-server exists in the vmt-test-from-vm namespace.
+
+      Create a VirtualMachineTemplateRequest to convert this VM into a reusable template.
```
lyarwood: Can you use something less leading and more natural like `create a template from this vm`

lyarwood (commented):
lgtm thanks Ben, a quick eval nit and it looks like  there are uncommitted changes after running tests, can you amend?

```
$ make test
[..]
$ git diff
diff --git a/pkg/mcp/testdata/toolsets-kubevirt-tools.json b/pkg/mcp/testdata/toolsets-kubevirt-tools.json
index 18b16a6f..3f4da3b7 100644
--- a/pkg/mcp/testdata/toolsets-kubevirt-tools.json
+++ b/pkg/mcp/testdata/toolsets-kubevirt-tools.json
@@ -161,7 +161,9 @@
   {
     "annotations": {
       "destructiveHint": true,
+      "idempotentHint": false,
       "openWorldHint": false,
+      "readOnlyHint": false,
       "title": "Virtual Machine: Create from Template"
     },
     "description": "Create a VirtualMachine from a VirtualMachineTemplate (virt-template) on KubeVirt. Processes the template server-side to substitute parameters (required values, defaults, and auto-generated values like passwords), then creates the resulting VirtualMachine in the same namespace. Cross-namespace template usage is not supported.",
```

evals/tasks/kubevirt/create-template-from-vm/task.yaml
codingben: Done.

codingben:
@lyarwood Sorry I missed it. It should be okay now.

lyarwood (approved):
Looks good to me, Thanks @codingben!

codingben:
@manusa @Cali0707 Can you please review and merge if all good?

pkg/kubevirt/template.go
```diff
@@ -0,0 +1,57 @@
...
+// CreateVMFromTemplate calls the create subresource on a VirtualMachineTemplate
+// to perform server-side parameter substitution and create the resulting
+// VirtualMachine in a single atomic operation.
+func CreateVMFromTemplate(ctx context.Context, restConfig *rest.Config, namespace, templateName string, parameters map[string]string) (map[string]any, error) {
+	config := rest.CopyConfig(restConfig)
+
+	gv := schema.GroupVersion{Group: "subresources.template.kubevirt.io", Version: "v1beta1"}
+	config.GroupVersion = &gv
+	config.APIPath = "/apis"
+	config.NegotiatedSerializer = subresourcesCodec.WithoutConversion()
```
Cali0707, line 22: nit: maybe we can move the `subresourcesCodec` var into a more intuitive/central file in `pkg/kubevirt` than its current definition in `pkg/kubevirt/guestagent.go` now that it is being used in multiple places?

internal/test/envtest.go
```diff
@@ -102,6 +102,9 @@ func EnvTest() *envtest.Environment {
 				CRD("tekton.dev", "v1", "taskruns", "TaskRun", "taskrun", true),
 				CRD("pipelinesascode.tekton.dev", "v1alpha1", "repositories", "Repository", "repository", true),
 				CRD("operator.tekton.dev", "v1alpha1", "tektonconfigs", "TektonConfig", "tektonconfig", false),
+				// virt-template
+				CRD("template.kubevirt.io", "v1beta1", "virtualmachinetemplates", "VirtualMachineTemplate", "virtualmachinetemplate", true),
+				CRD("template.kubevirt.io", "v1beta1", "virtualmachinetemplaterequests", "VirtualMachineTemplateRequest", "virtualmachinetemplaterequest", true),
```
Cali0707: do we need this one? as far as I can tell none of the tools create/manage this, so is it needed in the envtest? note that this is only required for the unit tests to run

Cali0707 (commented):
Overall looks good @codingben !

Left a few nits then lets try to get this in

pkg/kubevirt/template.go
codingben: Okay.

internal/test/envtest.go
codingben: Seems like only `virtualmachinetemplaterequests` unused. I've removed it.

Cali0707 (approved):
LGTM

PR #1322, opened 2026-07-21
chore: cleanup readme generator to reuse mcp imports

Cali0707:
Currently whenever adding new toolsets we need to add an import for the toolset into both pkg/mcp and internal/tools/update-readme

This is error prone and overly complex.

This PR moves to just importing pkg/mcp, which then in turn imports each toolset, leading to init functions running as expected


manusa (approved):
LGTM, thx!

PR #1323, opened 2026-07-21
fix(http): use const url instead of user provided url in wellknown proxy

Cali0707:
This continues on #1316 

Snyk still complains despite the validation added there, as the taint detection does not account for us validating the values - it wants the value that is propagated to be one from our code.

This switches two things:
1. No more subpath matching on the wellknown endpoints (as far as I know, this is fine/safe for the IdPs we support)
2. We pass our value of the allowed url through, rather than the one coming from the user (this is what lets snyk pass)


2uasimojo (commented):
Ah, nice, that feels tighter.

/lgtm

PR #1324, opened 2026-07-22
ci(mcpchecker): move weekly evals to matrix

Cali0707:
Follow up on #1314 

This moves to running the scheduled job as a matrix where we run all suites, and then individual runs are compared against that


manusa (changes requested):
The matrix approach looks good overall, but the `schedule-matrix` path has a bug that will break weekly runs.

The heredoc produces a multi-line string in `$MATRIX`. When you write it with `echo "matrix=${MATRIX}" >> "$GITHUB_OUTPUT"`, GitHub Actions only reads the first line as the value. The matrix output ends up as `{"include":[` (truncated), and `fromJSON()` will fail.

You can reproduce this locally:

```bash
MATRIX=$(cat <<'EOFM'
{"include":[
  {"suite":"core","eval-config":"eval-core.yaml","toolsets":"core,config"},
  {"suite":"helm","eval-config":"eval-helm.yaml","toolsets":"core,config,helm"}
]}
EOFM
)
OUTFILE=$(mktemp)
echo "matrix=${MATRIX}" >> "$OUTFILE"
head -1 "$OUTFILE"
# matrix={"include":[
```

The single-suite branches are fine because `single_matrix()` uses `printf` (no trailing newline, single line).

Fix: use the multi-line delimiter syntax for GITHUB_OUTPUT:

```bash
{
  echo "matrix<<EOFMATRIX"
  echo "$MATRIX"
  echo "EOFMATRIX"
} >> "$GITHUB_OUTPUT"
```

Or compact it after the heredoc: `echo "matrix=$(echo "$MATRIX" | jq -c .)" >> "$GITHUB_OUTPUT"`

manusa (approved):
LGTM, thx!

PR #1325, opened 2026-07-22
cleanup: evals don't point at old configs

Cali0707:
Follow up on #1314, to make docs and make targets point to the new config files and get rid of old eval.yaml files


manusa (approved):
LGTM, thx!

PR #1328, opened 2026-07-23
cleanup: trim down AGENTS.md

Cali0707:
Currently our AGENTS.md is approximately half full with instructions on writing tests.

However, agents do not always need to write tests when working in this code base. As such, we are having a large context hit that is not always useful

This moves to a progressive discovery pattern where we load in the testing docs when tests are being changed


manusa (approved):
LGTM, thx!

PR #1331, opened 2026-07-24
refactor(kubevirt): use shared subresource client for guest agent queries

lyarwood:
## Summary

Refactor `getVMISubresource` in `pkg/kubevirt/guestagent.go` to extract and use a `newSubresourceClient` helper function instead of setting up the REST client inline.

Now that #1361 has merged into `main`, requests to `subresources.kubevirt.io` are natively resolved by the `AccessControlRoundTripper` via discovery fallback, without requiring explicit API group bypasses.

## Changes

- Extract `newSubresourceClient` helper in `pkg/kubevirt/guestagent.go`
- Refactor `getVMISubresource` to use `newSubresourceClient`

### DIFF
--- a/pkg/kubevirt/guestagent.go
+++ b/pkg/kubevirt/guestagent.go
@@ -5,8 +5,6 @@ import (
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/rest"
 )
 
@@ -16,13 +14,6 @@ const (
 	subresourceFilesystemList = "filesystemlist"
 	subresourceUserList       = "userlist"
 	subresourceInterfaceList  = "interfacelist"
-)
-
-var (
-	// subresourcesScheme is the scheme for subresources API
-	subresourcesScheme = Scheme
-	// subresourcesCodec is the codec for encoding/decoding subresources
-	subresourcesCodec = serializer.NewCodecFactory(subresourcesScheme)
 )
 
 // AllGuestInfo holds all guest agent information with error tracking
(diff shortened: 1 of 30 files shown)


pkg/kubevirt/vm.go
```diff
@@ -144,6 +148,55 @@ func CloneVM(ctx context.Context, dynamicClient dynamic.Interface, namespace, so
 	return result, nil
 }
 
+// newSubresourceClient creates a REST client for the KubeVirt subresources API group
+func newSubresourceClient(restConfig *rest.Config) (rest.Interface, error) {
+	cfg := rest.CopyConfig(restConfig)
+	cfg.GroupVersion = &schema.GroupVersion{Group: "subresources.kubevirt.io", Version: "v1"}
+	cfg.APIPath = "/apis"
+	cfg.NegotiatedSerializer = serializer.NewCodecFactory(runtime.NewScheme())
```
ksimon1: would this make sense to extract as package level variable? So it is not initialized every time newSubresourceClient is called?

pkg/kubevirt/guestagent.go
```diff
@@ -39,21 +30,11 @@ type AllGuestInfo struct {
...
 // getVMISubresource retrieves a VMI subresource using the REST client
 func getVMISubresource(ctx context.Context, restConfig *rest.Config, namespace, vmiName, subresource string) (map[string]any, error) {
-	// Create a copy to avoid mutating the original config
-	config := rest.CopyConfig(restConfig)
-
-	// Create a REST client configured for the subresources.kubevirt.io API group
-	gv := schema.GroupVersion{Group: "subresources.kubevirt.io", Version: "v1"}
-	config.GroupVersion = &gv
-	config.APIPath = "/apis"
-	config.NegotiatedSerializer = subresourcesCodec.WithoutConversion()
```
ksimon1, line 49: This line is missing in the new code. Is it not required anymore?

PR #1335, opened 2026-07-28
Allow env var to specify config file path

2uasimojo:
With this change, if `--config` is not given on the CLI, we will look for an environment variable specifying a path to the main config file.

The variable is spelled `$K8S_MCP_CONFIG_PATH` here, but this change includes an override mechanism so downstreams can choose their own spelling.


Cali0707 (approved):
LGTM, thanks @2uasimojo !

PR #1336, opened 2026-07-28
fix: handle TLS version format from MCPLO

Cali0707:
In #1270 we added support for configuring tls min version and cipher suites through env vars. This was done in large part to work with https://github.com/kubernetes-sigs/mcp-lifecycle-operator/pull/257, which will allow for setting tls properties across all MCP servers in a k8s cluster.

However, we did not actually support env vars for the tls min version set with the format provided by the MCPLO. This PR adds support for those, while keeping support for the simpler "1.0"-"1.3" strings, as that is likely easier for non MCPLO users to configure

### DIFF
--- a/pkg/tlsutil/tlsutil.go
+++ b/pkg/tlsutil/tlsutil.go
@@ -14,10 +14,14 @@ import (
 
 // tlsVersionMap maps version strings to tls.Version constants
 var tlsVersionMap = map[string]uint16{
-	"1.0": tls.VersionTLS10,
-	"1.1": tls.VersionTLS11,
-	"1.2": tls.VersionTLS12,
-	"1.3": tls.VersionTLS13,
+	"1.0":          tls.VersionTLS10,
+	"1.1":          tls.VersionTLS11,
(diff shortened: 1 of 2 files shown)


pkg/tlsutil/tlsutil.go
```diff
@@ -14,10 +14,14 @@ import (
 
 // tlsVersionMap maps version strings to tls.Version constants
 var tlsVersionMap = map[string]uint16{
-	"1.0": tls.VersionTLS10,
-	"1.1": tls.VersionTLS11,
-	"1.2": tls.VersionTLS12,
-	"1.3": tls.VersionTLS13,
+	"1.0":          tls.VersionTLS10,
```
2uasimojo, line 17: We can’t get these mappings from the tls lib? (I haven’t looked yet.)

Cali0707: looks like the tls package only provides:

1. the constants for each version, as a uint16
2. A function to convert that to a string, not from a string
so we need our own mapping from string -> uint16

PR #1339, opened 2026-07-30
feat(kiali): API based filtering for Kiali

josunect:
- URL probe: If [toolset_configs.kiali].url is set, Kiali tools are registered only when GET {url}/api/status succeeds.
- Auto-discovery: If url is empty, the server discovers the in-cluster Kiali Service URL from the Kiali CR (with well-known DNS fallback), probes /api/status, and uses the working URL for tool calls.
- Opt-in: Probe and discovery run only when experimental_enable_target_compatibility_tool_filters = true. With the flag off (default), Kiali tools are always registered (unchanged behavior).
- Tool registration: All Kiali tools share a single HasKiali reachability filter via tools.All(p).

Tests: 
-  `go test ./pkg/kiali/... ./pkg/toolsets/kiali/...`
- Tools hidden without `kialis.kiali.io` (with `experimental_enable_target_compatibility_tool_filters = true`)
- Tools visible when the Kiali CRD is present

To test the url auto discovery, the mcp must be deployed inside the cluster: 

**Prerequisites**

- Kiali Operator installed and a Kiali CR (kialis.kiali.io) in Running state.
- Kiali Service reachable from inside the cluster (GET http://<kiali-svc>:<port>/api/status returns {"status":...}).
- A local image built from this branch (the published chart image may not include these changes).

**Deploy MCP server in-cluster**

```
make build
docker build -t kubernetes-mcp-server:dev .
```
 
load into your cluster if using Kind:
```
kind load docker-image kubernetes-mcp-server:dev --name <cluster-name>
helm upgrade -i kubernetes-mcp-server charts/kubernetes-mcp-server \
  -n mcp-system --create-namespace \
  -f charts/kubernetes-mcp-server/examples/values-kiali-discovery.yaml \
  --set openshift=false \
  --set ingress.enabled=false \
  --set image.registry=docker.io/library \
  --set image.repository=kubernetes-mcp-server \
  --set image.version=dev \
  --set image.pullPolicy=Never
```

where values-kiali-discovery.yaml:

```
ingress:
  enabled: false

serviceAccount:
  create: true
  automountToken: true

rbac:
  create: true
  extraClusterRoles:
    - name: kiali-discovery
      rules:
        - apiGroups: ["kiali.io"]
          resources: ["kialis"]
          verbs: ["get", "list", "watch"]
  extraClusterRoleBindings:
    - name: view
      roleRef:
        name: view
        external: true
    - name: kiali-discovery
      roleRef:
        name: kiali-discovery

config:
  toolsets:
    - core
    - kiali
  experimental_enable_target_compatibility_tool_filters: true
  cluster_auth_mode: kubeconfig

  toolset_configs:
    kiali:
      insecure: true
      # Omit url to enable in-cluster discovery from the Kiali CR.
```

Key config (already in the example values):

- experimental_enable_target_compatibility_tool_filters: true
- toolsets: [core, kiali]
- No [toolset_configs.kiali].url — discovery uses the Kiali CR
- RBAC to list/get/watch kialis.kiali.io

**Verify**
```
kubectl -n mcp-system rollout status deployment/kubernetes-mcp-server
kubectl -n mcp-system port-forward svc/kubernetes-mcp-server 18080:8080
```
In MCP Inspector (http://localhost:18080/mcp) or via API:

- tools/list → all 11 kiali_* tools should appear.
- kiali_get_mesh_status → should return mesh status from the discovered Kiali URL.

Negative checks
- No Kiali CR → kiali_* tools should not appear in tools/list.
- Flag disabled (experimental_enable_target_compatibility_tool_filters: false) → tools always listed, no probe/discovery.

**Troubleshooting**
- Confirm the Kiali CR exists: kubectl get kiali -A
- Confirm MCP can list CRs: kubectl auth can-i list kialis.kiali.io --as=system:serviceaccount:mcp-system:<sa-name> -n <kiali-namespace>
- Discovery logs are at verbosity 1–2: check pod logs with -v=2 if tools are missing.

### DIFF
--- a/pkg/kiali/config.go
+++ b/pkg/kiali/config.go
@@ -27,8 +27,15 @@ func (c *Config) Validate() error {
 	if c == nil {
 		return errors.New("kiali config is nil")
 	}
-	if c.Url == "" {
-		return errors.New("url is required")
+	// URL is optional: when empty, HasKiali may discover an in-cluster Service URL
+	// from a Kiali CR and inject it into this config at runtime.
+	if strings.TrimSpace(c.Url) == "" {
+		if caValue := strings.TrimSpace(c.CertificateAuthority); caValue != "" {
+			if _, err := os.Stat(caValue); err != nil {
+				return fmt.Errorf("certificate_authority must be a valid file path: %w", err)
+			}
+		}
+		return nil
 	}
 	if u, err := url.Parse(c.Url); err != nil || u.Scheme == "" || u.Host == "" {
 		return errors.New("url must be a valid URL")
@@ -61,8 +68,9 @@ func kialiToolsetParser(ctx context.Context, primitive toml.Primitive, md toml.M
 		// If it's already absolute or configDir is empty, use as-is
 	}
 
-	// Validate TLS settings when require_tls is enabled
-	if config.RequireTLSFromContext(ctx) {
+	// Validate TLS settings when require_tls is enabled (only when a URL is set;
+	// auto-discovered in-cluster URLs use http and are validated at filter time).
+	if config.RequireTLSFromContext(ctx) && strings.TrimSpace(cfg.Url) != "" {
 		if err := config.ValidateURLRequiresTLS(cfg.Url, "Kiali URL"); err != nil {
 			return nil, err
 		}
--- a/pkg/kiali/config.go
+++ b/pkg/kiali/config.go
@@ -8,10 +8,12 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/BurntSushi/toml"
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"github.com/containers/kubernetes-mcp-server/pkg/config"
+	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
 )
 
 // Config holds Kiali toolset configuration
@@ -80,6 +82,34 @@ func kialiToolsetParser(ctx context.Context, primitive toml.Primitive, md toml.M
 	}
 
 	return &cfg, nil
+}
+
+var warnEmptyURLWithoutDiscoveryOnce sync.Once
+
+// WarnIfEmptyURLWithoutDiscovery logs once when Kiali has no configured URL and
+// target-compatibility filtering is disabled, so in-cluster auto-discovery will not run.
+func WarnIfEmptyURLWithoutDiscovery(p api.FilteringProvider) {
+	if p == nil || p.IsTargetCompatibilityToolFiltersEnabled() {
+		return
+	}
+	cfgProvider, ok := p.(api.ExtendedConfigProvider)
+	if !ok || cfgProvider == nil {
+		return
+	}
+	ext, ok := cfgProvider.GetToolsetConfig("kiali")
+	if !ok {
+		return
+	}
+	kc, ok := ext.(*Config)
+	if !ok || kc == nil || strings.TrimSpace(kc.Url) != "" {
+		return
+	}
+	warnEmptyURLWithoutDiscoveryOnce.Do(func() {
+		klogutil.LogWarn(
+			klogutil.FromContext(context.Background()),
+			"Kiali url is empty and experimental_enable_target_compatibility_tool_filters is disabled; Kiali tools will not auto-discover an in-cluster URL",
+		)
+	})
 }
 
 func init() {
--- a/pkg/kiali/config.go
+++ b/pkg/kiali/config.go
@@ -8,12 +8,10 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"sync"
 
 	"github.com/BurntSushi/toml"
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	"github.com/containers/kubernetes-mcp-server/pkg/config"
-	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
 )
 
 // Config holds Kiali toolset configuration
@@ -82,34 +80,6 @@ func kialiToolsetParser(ctx context.Context, primitive toml.Primitive, md toml.M
 	}
 
 	return &cfg, nil
-}
-
-var warnEmptyURLWithoutDiscoveryOnce sync.Once
-
-// WarnIfEmptyURLWithoutDiscovery logs once when Kiali has no configured URL and
-// target-compatibility filtering is disabled, so in-cluster auto-discovery will not run.
-func WarnIfEmptyURLWithoutDiscovery(p api.FilteringProvider) {
-	if p == nil || p.IsTargetCompatibilityToolFiltersEnabled() {
-		return
-	}
-	cfgProvider, ok := p.(api.ExtendedConfigProvider)
-	if !ok || cfgProvider == nil {
-		return
-	}
-	ext, ok := cfgProvider.GetToolsetConfig("kiali")
-	if !ok {
-		return
-	}
-	kc, ok := ext.(*Config)
-	if !ok || kc == nil || strings.TrimSpace(kc.Url) != "" {
-		return
-	}
-	warnEmptyURLWithoutDiscoveryOnce.Do(func() {
-		klogutil.LogWarn(
-			klogutil.FromContext(context.Background()),
-			"Kiali url is empty and experimental_enable_target_compatibility_tool_filters is disabled; Kiali tools will not auto-discover an in-cluster URL",
-		)
-	})
 }
 
 func init() {
--- a/pkg/kiali/discover.go
+++ b/pkg/kiali/discover.go
@@ -1,0 +1,124 @@
+package kiali
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
+)
+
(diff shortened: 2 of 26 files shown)


pkg/kiali/gvr.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+)
+
+// KialiGVK is the GroupVersionKind for the Kiali custom resource installed by
+// the Kiali Operator (kialis.kiali.io CRD).
+var KialiGVK = schema.GroupVersionKind{
+	Group:   "kiali.io",
+	Version: "v1alpha1",
+	Kind:    "Kiali",
```
cajieh, line 32: Is filtering on the Kiali Operator CRD (kiali.io/v1alpha1) intentional for OSSM, or should Helm/standalone installs without that CRD still expose tools when` [toolset_configs.kiali].url` is set?

cajieh (commented):
Could we add an MCP ListTools test similar to kubevirt’s [TestKubevirtToolsFilteredWithoutCRDs](https://github.com/containers/kubernetes-mcp-server/blame/main/pkg/mcp/toolsets_test.go#L129), so we assert Kiali tools are hidden when the experimental flag is on and the Kiali CRD is missing?

Added comments. Otherwise, it looks good. Thank you!

pkg/kiali/gvr.go
josunect: @cajieh Thanks for the review. Added an MCP ListTools test similar to TestKubevirtToolsFilteredWithoutCRDs, asserting Kiali tools are hidden when the experimental flag is on and the Kiali Operator CRD is missing.

On the CRD question: filtering on kiali.io/v1alpha1 alone is intentional for OSSM/operator installs, but Helm/standalone installs do not register that CRD and talk to Kiali only via [toolset_configs.kiali].url. Because the tools call the Kiali HTTP API (not the Operator CR), HasKiali now treats Kiali as available if either the Operator GVK is present or a non-empty URL is configured. The MCP server wraps the filtering provider with the live config so that URL check sees current toolset_configs without changing the GetTools API.

josunect:
@manusa @cajieh @Cali0707 this is ready to review. Thanks!

pkg/kiali/config.go
```diff
@@ -27,8 +27,15 @@ func (c *Config) Validate() error {
...
-		return errors.New("url is required")
+	// URL is optional: when empty, HasKiali may discover an in-cluster Service URL
+	// from a Kiali CR and inject it into this config at runtime.
+	if strings.TrimSpace(c.Url) == "" {
+		if caValue := strings.TrimSpace(c.CertificateAuthority); caValue != "" {
+			if _, err := os.Stat(caValue); err != nil {
+				return fmt.Errorf("certificate_authority must be a valid file path: %w", err)
+			}
+		}
+		return nil
```
Cali0707, line 38: So HasKiali will only be called if `experimental_enable_target_compatibility_tool_filers=true` which is not the default. Which means by default this may allow some configs to pass validation that it should not...

Maybe we can at a minimum log a warning in this case?

pkg/kiali/gvr.go
```diff
@@ -0,0 +1,192 @@
...
+		return false
+	}
+	injectDiscoveredURL(cfg, cfgOK, url)
+	return true
+}
+
+func injectDiscoveredURL(cfg *Config, cfgOK bool, url string) {
+	setDiscoveredURL(url)
+	if cfgOK && cfg != nil {
+		cfg.Url = url
```
Cali0707, line 134: This write technically isn't synchronized wrt to reads as far as I can tell

pkg/kiali/gvr.go
```diff
@@ -0,0 +1,192 @@
...
+// Service URL can be discovered and probed.
+//
+// The returned closure memoizes the result (sync.Once). Callers that register
+// the filter on many tools should share one closure (see Toolset.GetTools).
+func HasKiali(p api.FilteringProvider) func() bool {
+	var once sync.Once
+	var available bool
+	return func() bool {
+		once.Do(func() {
+			available = evaluateKialiAvailability(context.TODO(), p)
```
Cali0707, line 66: If we need to use `context.TODO()` in places just because the TargetCompatibilityFilters api doesn't pass a ctx parameter in, maybe we should add the ctx paramter to that method?

cc @2uasimojo for input

pkg/kiali/status.go
```diff
@@ -0,0 +1,96 @@
...
+
+	probeCtx, cancel := context.WithTimeout(ctx, statusProbeTimeout)
+	defer cancel()
+
+	client, err := k.createHTTPClient(probeCtx)
+	if err != nil {
+		klogutil.FromContext(ctx).V(2).Info("kiali status probe: failed to create HTTP client", "error", err)
+		return false
+	}
+	client.Timeout = statusProbeTimeout
```
Cali0707, line 58: I don't _think_ we need the timeout explicitly set since there is the context deadline on `probeCtx` here

pkg/kiali/status.go
```diff
@@ -0,0 +1,96 @@
...
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
+)
+
+const (
+	statusPath         = "/api/status"
+	statusProbeTimeout = 3 * time.Second
```
Cali0707, line 16: I'm a little worried about this timeout slowing things down, as this will ultimately be called whenever the `WatchTargets` cluster-state changes (which can happen a lot in multi cluster envs).

Since this all runs synchronously each time it fires (and at startup) and we potentially run this against 6 urls, that is a 18 second delay.


Maybe we can:
1. make these probes in parallel
2. Tighten this timeout a decent amount?

Cali0707 (commented):
Thanks for starting on this @josunect !

I left some comments inline, but they may or may not be valid long term as I think there is a larger architectural consideration we need to work through first.

Fundamentally the `FilteringProvider` was designed to hide cluster-specific details behind a stable abstraction, so that we can swap things to e.g. leverage ACM search instead of iterating over clusters in some setups. However, Kiali is now fundamentally `ClusterAware = false` after https://github.com/containers/kubernetes-mcp-server/pull/1224, as it is a mesh control that can span clusters. 

This leads to the type assertion from the `FilteringProvider` to `kubernetes.Provider`, which IMO we shouldn't be doing. We should either:
1. explicitly pass that provider in
2. don't assert and expect it to be there

A key reason for this is that doing this and then looking at the default provider only or hand rolling the multi cluster checks by iterating providers blocks us from making optimizations based on the specifics of the multi cluster setup.

I wonder if maybe the correct fix is we make a kube client that internally checks every target rather than exposing a single method like `HasGVKs` which iterates over the targets? Then you could have a more expressive API and we could still make changes based on the multi cluster setup?

cc @2uasimojo since you have put a lot of thought into this part of the codebase

pkg/kiali/discover.go
```diff
@@ -0,0 +1,124 @@
...
+
+const defaultKialiPort int64 = 20001
+
+// internalServiceURLs builds candidate in-cluster Service base URLs for a Kiali CR.
+// Prefer values from status/spec; fall back to CR metadata and Kiali defaults.
+func internalServiceURLs(cr *unstructured.Unstructured) []string {
+	if cr == nil {
+		return nil
+	}
+	instanceName := nestedString(cr.Object, "status", "deployment", "instanceName")
```
2uasimojo, line 31: I'll again mention my distaste for `Unstructured` and generic struct walking like this. I hope there's a really good reason we're not importing the schema and unmarshaling into a real typed struct.

pkg/kiali/config.go
2uasimojo: > So HasKiali will only be called if `experimental_enable_target_compatibility_tool_filers=true`

I don't see that.

Which is good. When that option is `false` or absent, we should absolutely be defaulting to the path that's _more_ likely to expose the tools.

But (I intend to write this up in detail in the main review comment) I don't think that option is applicable at this level.
- It's supposed to filter individual tools, not an entire toolset -- that's the charter of the `toolsets` in the config.
- I don't hate the idea of overriding that ⏫ in cases where the kiali hub can't be reached for whatever reason (URL unset and no discoverable path; URL set but unreachable; whatever). But since that operation is O(1), *and* it'll need to be done at some point anyway, I'm not sure it makes sense to hide it behind the config option that's intended to guard against expensive fan-out queries.

pkg/kiali/gvr.go
2uasimojo: > maybe we should add the ctx paramter to that method?

(FMR, the [method def](https://github.com/containers/kubernetes-mcp-server/blob/aaab9a036dca4fd538d81d88e98de0e20782d418/pkg/api/toolsets.go#L19))

I don't object to adding a context param there, but the prototype is super generic for a reason. And we generally expect the impl to be defined in a closure like we've got here. So if we need a context param, it could be plumbed through `HasKiali()` itself. WDYT?

2uasimojo (commented):
I like the part of this that makes it possible to run without a `url`.

I think it's fine to hide all the kiali tools if discovery/validation of the endpoint fails.

But I don't think TargetCompatibilityToolFilters (the ServerTool field)  is the right mechanism for that. **If** we decide to condition the expensive discovery/validation behind the `experimental_enable_target_compatibility_tool_filters` option, it should be more like:

```
func GetTools(p) []api.ServerTool {
	if p.IsTargetCompatibilityToolFiltersEnabled() && !HasKiali() {
		return []
	}

	tools := slices.Concat(
	  ...
	)
	for i := range tools {
		// *No* TargetCompatibilityToolFilters here
		tools[i].ClusterAware = ptr.To(false)
	}
	return tools
}
```

The ServerTool field is intended for filtering individual tools in a toolset.

josunect:
@Cali0707 @2uasimojo Thanks for the reviews. I've addressed the feedback in the latest commits.

There's still an open question on the overall approach, I'm happy to adjust once that's decided.

Please let me know if you need anything else from my side.

PR #1344, opened 2026-08-04
chore: always set disable-destructive to false in evals server

Cali0707:
Follow up on #1310 to help keep the set of tools enabled in a eval run consistent in different envs


2uasimojo:
Per discussion, superseded by #1345

PR #1346, opened 2026-08-04
chore: add python dependabot config

Cali0707:
We recently added e2e tests in python, but no dependabot config to bump deps there. This PR adds that config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,6 +16,16 @@ updates:
       golang-x:
         patterns:
           - "golang.org/x/*"
+  - package-ecosystem: "pip"
+    directory: "/test/e2e"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 5
+    versioning-strategy: "lockfile-only"
+    groups:
+      python-test:
+        patterns:
+          - "*"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:


2uasimojo (commented):
/lgtm

Proxying agent review:

Does it do what's stated?
Yes. It adds a single new pip ecosystem entry to .github/dependabot.yml targeting the /test/e2e directory. This will have Dependabot start tracking and bumping Python dependencies upstream, which should address the downstream-only dep bump situation mentioned above.

Indentation & structure:
The new block uses the same 2-space YAML indentation as the existing gomod and github-actions entries. All nested keys (directory, schedule, open-pull-requests-limit, versioning-strategy, groups) are consistently indented to match the rest of the file. No issues.

Spelling:
All keywords are correctly spelled — package-ecosystem, versioning-strategy, lockfile-only, open-pull-requests-limit, group name python-test, etc. All are valid Dependabot configuration keys/values.

A few observations (not issues):
- open-pull-requests-limit is set to 5 here vs 10 for the other ecosystems — looks like a deliberate choice to keep Python dep PRs manageable.
- versioning-strategy: "lockfile-only" is a sensible pick for test dependencies — it updates the lockfile without modifying the requirements file directly.
- The "*" wildcard pattern in the python-test group means all Python dep bumps will be batched into a single PR, which keeps things tidy.

Everything checks out — clean, minimal change that does exactly what's described.

PR #1348, opened 2026-08-05
chore: expand run-server target to accept additional arguments

grokspawn:
provide the capability to pass additional arguments to run-server.  This allows us to neither mutate nor override the mcp-server config while still being able to adapt it to different runtime environments (for e.g. passing `--kubeconfig filename`).

### DIFF
--- a/build/evals.mk
+++ b/build/evals.mk
@@ -78,7 +78,7 @@ diff-evals: mcpchecker ## Diff latest mcpchecker results against baseline
 .PHONY: run-server
 run-server: build ## Start MCP server in background and wait for health check
 	@echo "Starting MCP server on port $(MCP_PORT)..."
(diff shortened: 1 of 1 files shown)


Cali0707 (approved):

2uasimojo:
I'm late to the party, but: I was expecting to see this new param getting *set* somewhere. Does that happen somewhere outside of this repo? Do we need to be concerned about escapes? (`MCP_SERVER_EXTRA_ARGS="; rm -f /tables/students;"` [ref](https://xkcd.com/327/))

Cali0707:
I guess in terms of worrying about escapes it depends on where we expect the value to come from, if it is from a workflow file we control I'm not too worried, but if it is coming from user input then we should think about this some more

This is useful for local dev as well, e.g. just pointing evals at a different kubeconfig locally

manusa:
The makefile (and includes) should be dev-facing, they aren't part of the project distribution.
They are here for dev and CI convenience. I wouldn't worry too much about command injection unless this is exposed somehow in the CI where it could be used for sensitive information disclosure (secrets et al)

Update: Just saw https://github.com/containers/kubernetes-mcp-server/pull/1349 which covers concerns.

PR #1357, opened 2026-08-10
feat(dev): add compose + MCP Inspector support

slashpai:
Adds a compose.yaml and `make inspect` target to run the server and MCP Inspector as containers (Docker or Podman). The Makefile recipe auto-rewrites Kind kubeconfigs for container access (host.docker.internal + insecure-skip-tls-verify).

No Node.js/npx needed on the host useful for container-only setups.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,8 @@ LD_FLAGS = -s -w \
 	-X '$(PACKAGE)/pkg/version.BinaryName=$(BINARY_NAME)' \
 	-X '$(PACKAGE)/pkg/version.WebsiteURL=$(WEBSITE_URL)'
 COMMON_BUILD_ARGS = -ldflags "$(LD_FLAGS)"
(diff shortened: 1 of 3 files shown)


slashpai:
@manusa @Cali0707 Please review when you get a chance

slashpai:
@Cali0707 Gentle ping for review

Cali0707:
@slashpai I guess I don't get the intended use case here? Why would you want to run the kubernetes mcp server locally without a local kubernetes setup like kind? could we not just have the target install the mcp server and inspector into kind and then port forward the inspector deployment?

slashpai:
> @slashpai I guess I don't get the intended use case here? Why would you want to run the kubernetes mcp server locally without a local kubernetes setup like kind? could we not just have the target install the mcp server and inspector into kind and then port forward the inspector deployment?

I have the setup to avoid installing node.js/npx locally not kind cluster

slashpai:
@Cali0707 Is this good to merge?

PR #1360, opened 2026-08-11
fix(bugs): A few minor bugfixes

dlom:
- Return tool result errors instead of protocol errors for invalid multi-cluster targets (tools_gosdk.go)
- Skip OIDC client ID check when oauth_audience is empty instead of passing empty string to the verifier (authorization.go)
- Default SubjectTokenType to access_token in token exchange config so keycloak accepts the subject token (provider_token_exchange.go, token_exchange.go)

### DIFF
--- a/pkg/http/authorization.go
+++ b/pkg/http/authorization.go
@@ -218,9 +218,11 @@ func (c *JWTClaims) ValidateOffline(audience string) error {
 // ValidateWithProvider validates the JWT claims against the OIDC provider.
 func (c *JWTClaims) ValidateWithProvider(ctx context.Context, audience string, provider *oidc.Provider) error {
 	if provider != nil {
-		verifier := provider.Verifier(&oidc.Config{
-			ClientID: audience,
-		})
+		cfg := &oidc.Config{ClientID: audience}
+		if audience == "" {
+			cfg.SkipClientIDCheck = true
+		}
+		verifier := provider.Verifier(cfg)
 		_, err := verifier.Verify(ctx, c.Token)
 		if err != nil {
 			return fmt.Errorf("OIDC token validation error: %w", err)
--- a/pkg/kcp/provider.go
+++ b/pkg/kcp/provider.go
@@ -189,7 +189,7 @@ func (p *kcpClusterProvider) managerForWorkspace(ctx context.Context, workspace 
 	}
 
 	if _, exists := p.managers[workspace]; !exists {
-		return nil, fmt.Errorf("workspace %s not found", workspace)
+		return nil, fmt.Errorf("workspace %s not found: %w", workspace, kubernetes.ErrUnknownTarget)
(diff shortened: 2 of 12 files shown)


manusa (changes requested):
Hi @dlom,

Thanks for digging these out, all three diagnoses look correct to me. I verified the OIDC one against go-oidc v3.20.0 (`oidc/verify.go:251-259`) and you're right that an empty `oauth_audience` with `authorization_url` set was rejecting every single request, so that's a genuine outage fix.

One blocker before this can merge though.

**CI is red.** The `tools_gosdk.go` change breaks `TestAuthorizationUnauthorizedTokenExchangeFailure`:

```
--- FAIL: TestAuthorization/TestAuthorizationUnauthorizedTokenExchangeFailure/Protected_resource/Call_tool_returns_error_when_token_exchange_fails
        Error:    An error is expected but got nil.
        Messages: Expected tool call to fail when token exchange fails
```

The root cause is that the change is broader than the description suggests. `GetDerivedKubernetes` returns a lot more than invalid target errors: token exchange and authentication failures (`provider_token_exchange.go:57-60`, `manager.go:181`), configuration errors (`unknown cluster_auth_mode`, `token_exchange.go:83`), and transient client construction failures (`manager.go:212,217`). Turning all of those into tool results means that when a user's IdP session expires mid conversation, the model receives a retryable looking string, retries, then starts guessing other target names, while the MCP client never sees a transport failure and can't trigger re-auth. That test was added on purpose, see the comment at `authorization_mcp_test.go:320-321`.

Could you narrow the conversion so that only the unknown target case becomes a tool result? An `ErrUnknownTarget` sentinel in `pkg/kubernetes` plus an `errors.Is` check at the call site would do it, with every other error class keeping `return nil, err`. That way the existing test passes unmodified.

If you'd rather change the contract for all error classes, I'm open to it, but let's discuss it separately and update the test with the reasoning instead of just flipping the assertion.

Two smaller things while you're in there:

1. Could you add a regression test for the `subject_token_type` fix? `provider_token_exchange_test.go` already drives that exact path and asserts the posted form, so it's about two lines: add a `subjectTokenType` field to `observedTokenRequest` (line 29) and record `r.PostForm.Get(tokenexchange.FormKeySubjectTokenType)` alongside the others at line 341. As it stands nothing in the suite would catch a regression there.

2. The same empty `subject_token_type` bug still exists on the per target path (`token_exchange.go:41`, configs coming from `GetTokenExchangeConfig`), since `Validate()` only checks `AuthStyle`. Defaulting it once inside the exchangers (`if cfg.SubjectTokenType == "" { ... = TokenTypeAccessToken }`) would fix all the call sites at once and let `sts.go:51` use the constant instead of the raw literal.

Thanks!

manusa (approved):
LGTM, thx!

PR #1361, opened 2026-08-11
fix(kubernetes): roundtripper handles subresources/resources without Kind set

Cali0707:
This is another attempt at solving the root cause of #1308 

The problem there is that kubevirt subresources do not have a `Kind` field set, so the rest mapper errors when trying to find the matching GVK.

This moves to see if the resource can be matched through the discovery interface as a fallback. From the unit tests added, this seems to work for the kubevirt resources.

I also tested the raw go code against a cluster with kubevirt installed and was able to have the discovery client approach work to find kubevirt subresources

### DIFF
--- a/pkg/kubernetes/accesscontrol_round_tripper.go
+++ b/pkg/kubernetes/accesscontrol_round_tripper.go
@@ -24,6 +24,7 @@ type AccessControlRoundTripper struct {
 	delegate                http.RoundTripper
 	deniedResourcesProvider api.DeniedResourcesProvider
 	restMapperProvider      func() meta.RESTMapper
+	discoveryProvider       func() discovery.DiscoveryInterface
 	apiPathPrefix           string
 	validators              []api.HTTPValidator
 }
@@ -58,6 +59,7 @@ func NewAccessControlRoundTripper(ctx context.Context, cfg AccessControlRoundTri
 		delegate:                cfg.Delegate,
 		deniedResourcesProvider: cfg.DeniedResourcesProvider,
 		restMapperProvider:      cfg.RestMapperProvider,
+		discoveryProvider:       cfg.DiscoveryProvider,
 		apiPathPrefix:           apiPathPrefix,
 	}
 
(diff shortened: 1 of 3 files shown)


lyarwood (commented):
Great approach — using the discovery API as a fallback is cleaner than requiring each toolset to explicitly declare its API groups, and it correctly handles the denied_resources interaction via the partial GVK.

I built this on top of a pause/unpause PR ([lyarwood/kubernetes-mcp-server#kubevirt-pause-1361](https://github.com/containers/kubernetes-mcp-server/commit/79ad0e07bbe8032534b114471b74170915de4d1a)) and tested it end-to-end against a real OCP cluster with KubeVirt installed. **Pause and unpause work correctly** — the VM transitions `Running → Paused → Running` as expected.

However, I hit one issue that won't be caught by the unit tests alone: **`resourceExistsInDiscovery` returns false on OCP clusters that use the aggregated discovery API.**

The root cause: `k.discoveryClient` is a `memory.NewMemCacheClient`. When the cluster supports aggregated discovery (which OCP 4.x does), the memory cache populates itself via `GroupsAndMaybeResources()` rather than per-group `ServerResourcesForGroupVersion` calls. The aggregated discovery endpoint filters out resources with an empty `Kind` — which is exactly what `subresources.kubevirt.io/v1` advertises. So the cache stores an empty `APIResourceList` for that group. When `resourceExistsInDiscovery` then calls `disc.ServerResourcesForGroupVersion("subresources.kubevirt.io/v1")`, it gets back 0 resources (no error), finds no match, and returns false — blocking the request.

I confirmed this with debug logging:
```
DEBUG: ServerResourcesForGroupVersion("subresources.kubevirt.io/v1") returned 0 resources
```
But the raw endpoint returns 31:
```bash
$ oc get --raw /apis/subresources.kubevirt.io/v1 | jq '.resources | length'
31
```

The fix I used: store the non-cached discovery client alongside the memory-cached one in the `Kubernetes` struct, and pass that to `DiscoveryProvider` in the round tripper. The non-cached client always hits the API server directly and returns the full resource list:

```go
// pkg/kubernetes/kubernetes.go
type Kubernetes struct {
    ...
    discoveryClient    discovery.CachedDiscoveryInterface
    rawDiscoveryClient discovery.DiscoveryInterface  // non-cached, used by round tripper fallback
}

// After creating the discovery client:
k.rawDiscoveryClient = discoveryClient
k.discoveryClient = memory.NewMemCacheClient(discoveryClient)

// In the Wrap closure:
DiscoveryProvider: func() discovery.DiscoveryInterface { return k.rawDiscoveryClient },
```

With that change applied, pause/unpause work correctly on OCP. Happy to open a follow-up if you'd prefer to keep the fix separate, or I can push a suggestion here.

codingben:
This PR blocks: https://github.com/containers/kubernetes-mcp-server/pull/1321

Cali0707:
@lyarwood mind re-checking this PR?

lyarwood (approved):
Tested with my pause PR and this worked for me thanks!

pkg/kubernetes/accesscontrol_round_tripper.go
```diff
@@ -155,6 +154,53 @@ func (rt *AccessControlRoundTripper) RoundTrip(req *http.Request) (*http.Respons
...
+	restMapper := rt.restMapperProvider()
+	if restMapper == nil {
+		return schema.GroupVersionKind{},
+			fmt.Errorf("failed to make request: AccessControlRoundTripper restMapper not initialized")
+	}
+
+	gvk, err := restMapper.KindFor(gvr)
+	if err == nil {
+		return gvk, nil
+
```
2uasimojo:
```suggestion
```

2uasimojo (commented):
/lgtm

Clean.

Minor whitespace damage was the worst thing my agent could find :P

PR #1365, opened 2026-08-12
Option to run /metrics on a different port

### LINKED ISSUE #1364: Proposal: Make Prometheus metrics port configurable
Currently, `/metrics` is served on the same port as the MCP server itself. When running in Kubernetes, this makes things awkward, since if you want to put network policy in front of the MCP server to protect it, you would need to use policy that supports layer 7 rules (e.g. allow HTTP:8080/metrics ingress from Prometheus, but don't allow any other URL path) instead of layer 4 rules.

If we could configure an alternate port, say the typical `9090`, for `/metrics`, then we could use purely layer 4 rules (e.g. allow HTTP:9090 ingress from Prometheus, allow HTTP:8080 ingress from nothing).

2uasimojo:
To facilitate simpler network policies, add an option `--metrics-port`/`metrics_port`, which may only be provided in http mode (i.e. with `--port`), which causes the `/metrics`, `/stats`, and a separate `/healthz` endpoint to run in a separate server on the specified port (which must differ from `--port`).

Note that this configuration setting is in the subset that *cannot* be reloaded via SIGHUP. A full restart is required.

Almost-Exclusively-Authored-By: claude

Closes: #1364

### DIFF
--- a/pkg/http/authorization.go
+++ b/pkg/http/authorization.go
@@ -57,14 +57,15 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			logger := klogutil.FromContext(r.Context())
-			// Skip auth for infrastructure endpoints (health, metrics) and well-known endpoints
-			if slices.Contains(infraPaths, r.URL.Path) || isWellKnownPath(r.URL.EscapedPath()) {
-				next.ServeHTTP(w, r)
-				return
-			}
 			// Load the latest config snapshot on every request so that
 			// SIGHUP-reloaded auth settings take effect immediately.
 			staticConfig := cfgState.Load()
+			// Skip auth for infrastructure endpoints (health, metrics) and well-known endpoints.
+			// When metrics are on a separate port, only /healthz is exempt on the main port.
+			if slices.Contains(infraPaths(staticConfig.MetricsPort != ""), r.URL.Path) || isWellKnownPath(r.URL.EscapedPath()) {
+				next.ServeHTTP(w, r)
+				return
+			}
 			if !staticConfig.RequireOAuth {
 				next.ServeHTTP(w, r)
 				return
--- a/pkg/http/http.go
+++ b/pkg/http/http.go
@@ -51,9 +51,16 @@ const (
 )
 
 var (
-	// infraPaths contains infrastructure endpoints which should not have oauth applied
-	infraPaths = []string{healthEndpoint, metricsEndpoint, statsEndpoint}
+	infraPathsMetricsSeparate = []string{healthEndpoint}
+	infraPathsMetricsTogether = []string{healthEndpoint, metricsEndpoint, statsEndpoint}
 )
+
+func infraPaths(metricsOnSeparatePort bool) []string {
+	if metricsOnSeparatePort {
+		return infraPathsMetricsSeparate
+	}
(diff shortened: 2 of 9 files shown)


pkg/http/authorization.go
```diff
@@ -58,14 +58,15 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
...
-			if slices.Contains(infraPaths, r.URL.Path) || isWellKnownPath(r.URL.EscapedPath()) {
-				next.ServeHTTP(w, r)
-				return
-			}
 			// Load the latest config snapshot on every request so that
 			// SIGHUP-reloaded auth settings take effect immediately.
 			staticConfig := cfgState.Load()
+			// Skip auth for infrastructure endpoints (health, metrics) and well-known endpoints.
+			// When metrics are on a separate port, only /healthz is exempt on the main port.
+			if slices.Contains(infraPaths(staticConfig.MetricsPort != ""), r.URL.Path) || isWellKnownPath(r.URL.EscapedPath()) {
```
Cali0707, line 65: This now allocates a slice on every request, which even though it is a very small slice can probably be avoided

Maybe we can just have two constants `infraPathsAll` and `infraPathsHealthOnly` and we switch which constant slice we check on?

pkg/http/http.go
Cali0707: Maybe here we need another branch to log a warning if `BindAddress == "0.0.0.0"` as now even if TLS/OAuth is enabled, that can be skipped on metrics endpoints, which e.g. could expose `/stats` without auth

Cali0707:
@2uasimojo gentle ping, mind fixing the merge conflicts here when you have a chance?

2uasimojo:
> mind fixing the merge conflicts

Rebase done. Will address your comments in a separate push for review ease.

pkg/http/authorization.go
2uasimojo: Done. (I chose slightly different names for better future-proofing. Or so I'm telling myself.)

pkg/http/http.go
2uasimojo: ...like this? :arrow_down:

pkg/http/authorization.go
```diff
@@ -57,14 +57,15 @@ func AuthorizationMiddleware(cfgState *config.StaticConfigState, oauthState *oau
...
```
Cali0707, line 65: @2uasimojo looking at this a bit more, I think we may have one more thing to fix here (sorry for not catching first time):

I believe that if someone were to change the metrics port (despite the fact it is documented as not reloadable), this could change whether or not auth is applied to the metrics enpdoint

Maybe we want to capture the metrics port once at startup and use that value for the comparison here to avoid that issue?

2uasimojo: Good catch!

I'd like to fix this by *not* changing the configured value (so this code path continues to use the original) and logging a warning when the SIGHUP reload happens.

In the new config architecture (#1404) I think we would do something similar for all non-reloadable settings.

WFY?

2uasimojo: Reading your original comment again, maybe that's exactly what you were suggesting :facepalm:

2uasimojo:
=> 5b54ca9 rebase only.

pkg/http/authorization.go
2uasimojo: fixed in [99800cf](https://github.com/containers/kubernetes-mcp-server/commit/99800cfe5c0c39c4a6e224cf9f67f9fbe27e066c)

Cali0707 (approved):
LGTM

PR #1367, opened 2026-08-13
fulfill jq eval dependency

grokspawn:
core evals require the presence of jq but make rules do not presently supply them:
-  create-canary-deployment
- create-network-policy
- horizontal-pod-autoscaler

Though jq is quite common, the build environment shouldn't presuppose the presence of critical binaries in users' environments, so this PR pulls jq (IFF missing) into the environment similarly as we currently pull mcpchecker, but with a pinned version since it is expected to be more stable.

### DIFF
--- a/build/evals.mk
+++ b/build/evals.mk
@@ -9,6 +9,16 @@ MCPCHECKER = $(shell pwd)/_output/tools/bin/mcpchecker
 MCPCHECKER_VERSION ?= latest
 CLAUDE_AGENT_ACP = $(shell pwd)/_output/tools/node_modules/.bin/claude-agent-acp
 CLAUDE_AGENT_ACP_VERSION ?= latest
+JQ = $(shell pwd)/_output/tools/bin/jq
+JQ_VERSION ?= 1.7.1
+# Derive the jq release asset for the host from the Go toolchain (already a
+# dependency of this eval flow). jq names assets macos/i386 where Go says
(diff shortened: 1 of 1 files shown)


build/evals.mk
```diff
@@ -27,6 +29,18 @@ EVAL_LABEL_SELECTOR ?= suite=$(SUITE)
...
 EVAL_VERBOSE ?= false
 
+# Download and install jq static binary if not already installed
+.PHONY: jq
+jq:
+	@[ -f $(JQ) ] || { \
+		set -e ;\
+		echo "Installing jq $(JQ_VERSION) to $(JQ)..." ;\
+		mkdir -p $(shell dirname $(JQ)) ;\
+		curl -fsSL "https://github.com/jqlang/jq/releases/download/jq-$(JQ_VERSION)/jq-linux-amd64" \
```
Cali0707: maybe don't assume `amd-64` ? if we are trying to ensure this works in different envs, this could fail in e.g. arm

Cali0707 (approved):

build/evals.mk
grokspawn: Oversight on my part.  Addressed in [d35c3f3dd58c2f3827cdcdeac9f90da7682a8078](https://github.com/containers/kubernetes-mcp-server/commit/d35c3f3dd58c2f3827cdcdeac9f90da7682a8078)

PR #1376, opened 2026-08-18
feat!(http): remove SSE MCP transport

### LINKED ISSUE #1370: Drop final mentions of SSE
We don't want to support SSE as a server anymore, but still have mentions of SSE in the help text in `pkg/kubernetes-mcp-server/cmd/root.go` and in a single flag there for the CLI.

We should clean this up and remove those (and any other references found)

Shaurya2k06:
## What

- Stop serving the MCP SSE transport. HTTP mode now exposes Streamable HTTP at `/mcp` only (plus `/healthz`, `/stats`, `/metrics`, and well-known OAuth paths).
- Remove `ServeSse()`, the `/sse` and `/message` routes, `--sse-base-url`, and the unused `sse_base_url` TOML field (`SSEBaseURL` was never applied to the SSE handler).
- Update CLI help/examples, README, configuration docs, and OTEL examples so they no longer describe SSE as a server transport.
- Replace SSE client tests with a 404 check on `/sse`. Keep Streamable HTTP auth/token-exchange coverage.

This is a breaking change for clients still connecting to `/sse`. Existing configs that still set `sse_base_url` continue to load; unknown TOML keys are ignored.

## Why

SSE is no longer a supported server transport. The remaining flag, config field, and help text advertised a path that should not exist.

## Testing

- [x] `go test -race -count=1 ./pkg/http ./pkg/config ./pkg/mcp ./pkg/kubernetes-mcp-server/cmd`
- [x] `make test` (`go test -race -count=1 -v ./...`)

Closes #1370


Cali0707 (approved):
LGTM, thanks @Shaurya2k06 !

PR #1378, opened 2026-08-19
feat(auth): make RFC 8693 token-type params configurable

zruchi:
## Summary

Split out of #1181 so each piece is scoped independently — this is the token-type half, `sts_token_url` follows separately.

- `requested_token_type` is hardcoded to `access_token` in the rfc8693 exchanger.
- `subject_token_type` is only reachable via the per-target `TargetTokenExchangeConfig`, which no production provider populates — so the global exchange path can't set it either.

Both are now configurable via `sts_subject_token_type` / `sts_requested_token_type`, plumbed through `StsConfigProvider` to both global exchange call sites.

Since #1181 was opened, #1360 landed the `"" -> access_token` default in the rfc8693 and keycloak-v1 exchangers. That fixes the empty-`subject_token_type` bug described there, and is what makes this PR a pure no-op when the new keys are unset.

## New TOML keys

| Key | Default | Purpose |
|---|---|---|
| `sts_subject_token_type` | `urn:ietf:params:oauth:token-type:access_token` | RFC 8693 `subject_token_type` form param |
| `sts_requested_token_type` | `urn:ietf:params:oauth:token-type:access_token` | RFC 8693 `requested_token_type` form param |

## Notable behavior changes

None — with both keys unset the emitted request is byte-identical. The call sites swap a hardcoded `TokenTypeAccessToken` for the config getter, which returns `""`, and the exchangers already default that back to `access_token`. `entra-obo` reads neither param.

- Corrects a wrong claim from #1181: [2.1](https://datatracker.ietf.org/doc/html/rfc8693#section-2.1) makes `requested_token_type` OPTIONAL and leaves the issued type to the AS when unspecified — it does not "default to `access_token`". Fixed in `pkg/config/config.go`, `pkg/tokenexchange/config.go` and `docs/configuration.md`. Thanks @theJC for catching it.
- Still *sends* `access_token` when unset rather than omitting the param, to preserve current behavior. Omitting would be closer to the spec — say the word if you'd prefer that instead.
- New `rfc8693_exchanger_test.go` asserts the exact form params against a recording server, including both defaults.

## Example: cross-realm exchange requiring a JWT subject

```toml
authorization_url        = "https://idp.example.com/realms/users"
token_exchange_strategy  = "rfc8693"
sts_client_id            = "mcp-backend"
sts_client_secret        = "..."
sts_audience             = "kubernetes-api"
sts_subject_token_type   = "urn:ietf:params:oauth:token-type:jwt"
sts_requested_token_type = "urn:ietf:params:oauth:token-type:jwt"
```


zruchi:
@Cali0707 can you have a look? thanks

theJC:
@zruchi --  Thank you so much for driving this.


We need this mcp server's RFC compliance fix in order to be able to integrate with Okta and use it as our AS.

pkg/kubernetes/provider_token_exchange.go
Cali0707, line 160: shouldn't SubjectTokenType/RequestedTokenType get plumbed into the cache keys here?

zruchi: Done

pkg/kubernetes/token_exchange.go
```diff
@@ -174,12 +174,13 @@ func strategyBasedTokenExchange(
...
 			ClientSecret:       baseConfig.GetStsClientSecret(),
 			Audience:           baseConfig.GetStsAudience(),
-			SubjectTokenType:   tokenexchange.TokenTypeAccessToken,
 			Scopes:             baseConfig.GetStsScopes(),
 			AuthStyle:          authStyle,
 			ClientCertFile:     baseConfig.GetStsClientCertFile(),
 			ClientKeyFile:      baseConfig.GetStsClientKeyFile(),
 			FederatedTokenFile: baseConfig.GetStsFederatedTokenFile(),
+			SubjectTokenType:   baseConfig.GetStsSubjectTokenType(),
+			RequestedTokenType: baseConfig.GetStsRequestedTokenType(),
```
theJC, line 183: These getters are only reached when token_exchange_strategy is non-empty. With the valid sts_audience-only configuration, exchangePassthroughToken instead calls NewFromConfig, whose externalaccount path still hardcodes both token types to access_token. The accepted settings are also partially or fully ignored for other strategies: keycloak-v1 ignores RequestedTokenType, and entra-obo ignores both fields. That means a valid configuration can silently fail to apply these new options. Could we either route the supported configuration through the RFC 8693 exchanger, or validate and document the exact strategy combinations in which each option is effective?

pkg/tokenexchange/rfc8693_exchanger_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+// newRecordingServer returns an httptest.Server that decodes the form body of the
+// inbound request into the supplied url.Values pointer and replies with a valid
+// token exchange response.
+func (s *RFC8693ExchangerTestSuite) newRecordingServer(captured *map[string]string) *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		s.Equal(http.MethodPost, r.Method)
+		s.Equal(ContentTypeXWWWFormUrlEncoded, r.Header.Get(HeaderContentType))
+
+		s.Require().NoError(r.ParseForm())
```
theJC: require assertions call FailNow, but this handler runs in the httptest server goroutine rather than the test goroutine. If ParseForm fails, only the handler goroutine exits, so the test can report misleading follow-on behavior or potentially hang. Could the handler send the parse error back through the captured result or a channel and let the test goroutine assert it? The assertions on lines 22 and 23 also run from the handler, although those are non-fatal Equal calls.

pkg/kubernetes/token_exchange.go
Cali0707: we've been meaning to pull out the sts only config for a while (at least internally), since the externalaccount path is basically an opinionated subset of the full rfc 8693 exchanger. IMO I wouldn't worry about documenting/routing this through for now, as this may just highlight that it is time to pull out the sts bits (as a separate PR)

pkg/tokenexchange/rfc8693_exchanger_test.go
zruchi: I fixed that @theJC. Thanks for catching that.

pkg/kubernetes/token_exchange.go
theJC: Cool, thanks for the context @Cali0707 -- makes sense!

theJC (approved):
Re-reviewed the latest head. 

The test-goroutine issue is fixed, and the token-exchange scope is now clear to me. My concerns are addressed. 

This is a non-maintainer code-review approval; full CI should still be rerun before merge.

zruchi:
@Cali0707 could you have another look? keen to merge this so I can also add the option to configure another AS in the next PR. ta

Cali0707 (approved):
LGTM thanks for this work @zruchi !!

And thanks @theJC for helping review this one

zruchi:
It seems we need approval from @manusa before we can merge this PR.

manusa (approved):
LGTM, thx!

PR #1379, opened 2026-08-19
fix(nodes): fix file query paths and tailLines for nodes_log tool

### LINKED ISSUE #1262: nodes_log fails on file queries: /var/log prefix double-join and tailLines rejected by kubelet

fuleinist:
Fixes #1262

Two bugs in nodes_log:

**1. Double-prefix on file paths**
The docstring examples showed `/var/log/kubelet.log` but the kubelet SecureJoin normalizes the query against `/var/log`, so `/var/log/kubelet.log` resolves to `/var/log/var/log/kubelet.log` and returns 404. Correct form is the relative path `/kubelet.log`.

**2. tailLines forwarded on file queries**
The kubelet rejects options on file queries (validate() at https://github.com/kubernetes/kubernetes/blob/70d3cc9/pkg/kubelet/kubelet_server_journal.go#L229-L230). Our tool advertised `tailLines` with a default of 100, which caused a 406 on every file query. A caller must explicitly pass `tailLines: 0` for a file query to succeed.

## Fix

- Update comments/docs to use correct relative paths (e.g. `/kubelet.log` not `/var/log/kubelet.log`)
- Only forward `tailLines` for service-style queries (no `/` or `\` in query)
- Add test: service query `kubelet` with `tailLines=2` returns tail
- Add test: file query `/kubelet.log` with `tailLines=2` ignores tailLines and returns full log

## Verification

```bash
go test ./pkg/mcp/ -run TestNodes -v
# 69 passed
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ In case multi-cluster support is enabled (default) and you have access to multip
 
 - **nodes_log** - Get logs from a Kubernetes node (kubelet, kube-proxy, or other system logs). This accesses node logs through the Kubernetes API proxy to the kubelet
   - `name` (`string`) **(required)** - Name of the node to get logs from
-  - `query` (`string`) **(required)** - query specifies services(s) or files from which to return logs (required). Example: "kubelet" to fetch kubelet logs, "/<log-file-name>" to fetch a specific log file from the node (e.g., "/var/log/kubelet.log" or "/var/log/kube-proxy.log")
+  - `query` (`string`) **(required)** - query specifies the service or file to fetch logs from. Service names (e.g. "kubelet", "kube-proxy") return service logs. File paths are relative to `/var/log` on the node (e.g. "/kubelet.log", not "/var/log/kubelet.log"). See https://kubernetes.io/docs/concepts/cluster-administration/system-logs/#log-query
   - `tailLines` (`integer`) - Number of lines to retrieve from the end of the logs (Optional, 0 means all logs)
 
 - **nodes_stats_summary** - Get detailed resource usage statistics from a Kubernetes node via the kubelet's Summary API. Provides comprehensive metrics including CPU, memory, filesystem, and network usage at the node, pod, and container levels. On systems with cgroup v2 and kernel 4.20+, also includes PSI (Pressure Stall Information) metrics that show resource pressure for CPU, memory, and I/O. See https://kubernetes.io/docs/reference/instrumentation/understand-psi-metrics/ for details on PSI metrics
--- a/pkg/kubernetes/nodes.go
+++ b/pkg/kubernetes/nodes.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/containers/kubernetes-mcp-server/pkg/api"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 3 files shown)


fuleinist:
Ping: this fixes #1262 — two bugs where file log queries were double-prefixing paths (`/var/log/kubelet.log` → resolved to ) and forwarding `tailLines` on file queries (kubelet rejects options on file queries with a 406). Both issues are fixed with ~10 lines of changes.

fuleinist:
Re-ping: this is a 3-file fix for #1262. CI is green. Could use a maintainer review when convenient.

Cali0707:
There is already a PR open to resolve this issue. Closing this one for now in favour of that one (just waiting on author answers to a few questions there)

If that for some reason doesn't get in in the next few weeks, we can re-open this one

PR #1381, opened 2026-08-19
fix(jwt): mint fresh assertions every time

dlom:
Fix a correctness bug in JWT client assertions: remove the assertion
cache from TargetTokenExchangeConfig and mint a fresh assertion (with a
unique jti) on every token exchange. OIDC private_key_jwt requires
single-use jti values, and Keycloak rejects reused assertions with
"Token reuse detected". Also add the iat claim, which strict validators
require to bound token age. Rename GetOrBuildAssertion → BuildAssertion
to reflect the new semantics.


Cali0707 (approved):

pkg/kubernetes/provider_token_exchange.go
Cali0707: shouldn't this also include SubjectTokenType/RequestedTokenType?

PR #1382, opened 2026-08-19
cleanup(http): remove unused code after sse removal

Cali0707:
Removes unnecessary http auth header propagation to context, as the streamable http transport correctly propagates those on the context. 

Also adds a test case to verify `/messages` 404s as that is also an unused enpoint post SSE removal

### DIFF
--- a/pkg/mcp/middleware.go
+++ b/pkg/mcp/middleware.go
@@ -169,7 +169,7 @@ func authHeaderPropagationMiddleware(next mcp.MethodHandler) mcp.MethodHandler {
 			}
 		}
 
-		// If no auth header in RequestExtra, context may already have it from HTTP middleware.
+		// No auth header found; pass through unchanged (stdio transport or unauthenticated request).
 		return next(ctx, method, req)
 	}
 }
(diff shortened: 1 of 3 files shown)


Cali0707:
Let's wait on #1320 to make sure this doesn't break any auth

pkg/mcp/middleware.go
```diff
@@ -169,7 +169,7 @@ func authHeaderPropagationMiddleware(next mcp.MethodHandler) mcp.MethodHandler {
 			}
 		}
 
-		// If no auth header in RequestExtra, context may already have it from HTTP middleware.
+		// No auth header found; pass through unchanged (stdio transport or unauthenticated request).
```
matzew, line 172: is that related? :)

matzew: ah - clarification of _transport_

Cali0707: so before the "already have it from HTTP middleware" was needed because SSE made us need this middleware

What I added here is a comment why it still makes sense to pass the value through (e.g. unauthenticated or stdio transport), as the old one doesn't make sense anymore

dlom (approved):

PR #1384, opened 2026-08-20
feat(e2e): add MCP client connectivity smoke tests

dlom:
Add the official MCP conformance suite (modelcontextprotocol/conformance) as a lightweight CI check that runs against a local server with a fake kubeconfig — no Minikube required. Baseline 22 expected failures (5 unimplemented features, 17 fixture mismatches).

xref: [OCPMCP-295](https://redhat.atlassian.net/browse/OCPMCP-295)

### DIFF
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -1,0 +1,43 @@
+name: MCP Conformance Tests
+
+on:
+  pull_request:
+    branches: ['main']
+    paths-ignore:
+      - '.gitignore'
+      - 'LICENSE'
+      - '**/*.md'
+      - 'docs/**'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  conformance:
+    name: MCP Protocol Conformance
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Setup Go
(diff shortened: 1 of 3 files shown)


test/e2e/clients_test.go
```diff
@@ -0,0 +1,96 @@
...
+				{"both", test.WithClientCapabilities(&mcp.ClientCapabilities{
+					RootsV2: &mcp.RootCapabilities{ListChanged: true},
+					Elicitation: &mcp.ElicitationCapabilities{
+						Form: &mcp.FormElicitationCapabilities{},
+					},
+				})},
+			}
+
+			for _, ci := range clientInfos {
+				for _, cap := range capabilities {
```
matzew: nit: rename cap to avoid shadowing the builtin..

matzew:
LGTM

overall!

test/e2e/clients_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/containers/kubernetes-mcp-server/internal/test"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
```
Cali0707: I'm not super sure this test has a ton of value over e.g. the conformance ones since as far as I can tell this is only asserting that the client from the go sdk can call a server also from the go sdk. IMO that is something the SDK is probably already checking.

I'd be inclined to either:
1. drop this test
2. Rewrite this test to somehow leverage the SDK from a different language to do the same thing (e.g. python or JS) to make sure that we work across N>1 SDKs

dlom: Yeah you're right, I think that the conformance test should cover everything that's covered here. I'm going to drop these entirely in favor of just the conformance tests

Cali0707 (approved):
LGTM, thanks @dlom !

PR #1392, opened 2026-08-26
feat(auth): decouple the STS endpoint from authorization_url

zruchi:
## Summary

Second piece of #1181, split out so each change reviews independently — companion to #1378 (token types).

- The STS token endpoint could only ever be the OIDC-discovered endpoint from `authorization_url`. That breaks cross-realm setups, where the STS gateway is a different service than the IDP that issued the user's token, and it breaks token exchange entirely when `skip_jwt_verification=true` (no OIDC provider means nothing to discover an endpoint from).
- Add `sts_token_url` so the STS endpoint can be set explicitly. `resolveStsTokenURL` prefers it when set, falls back to OIDC discovery otherwise — same fallback upstream already relied on.
- `SecurityTokenService.IsEnabled()` now keys off `TokenURL != ""` instead of `Provider != nil` — the real question is "do I have an endpoint to POST to," not "did OIDC discovery succeed." This also means the plain passthrough (`externalaccount`) path now respects `sts_token_url` too, not just the `rfc8693`/`keycloak-v1`/`entra-obo` strategy path — relevant to the gap @theJC flagged on #1378.

## New TOML key

| Key | Default | Purpose |
|---|---|---|
| `sts_token_url` | *(falls back to OIDC discovery)* | Explicit STS/token-exchange endpoint |

## Notable behavior changes

None when unset — `resolveStsTokenURL` falls back to `oidcProvider.Endpoint().TokenURL`, which is exactly what both call sites did before.

- Validates `sts_token_url` as `http`/`https` with a host; rejects scheme-only inputs like `https://` that `url.Parse` would otherwise accept and let fail later with a confusing error at exchange time.
- Warns (doesn't reject) on `http`.
- New integration test spins up two separate OIDC servers to prove the exchange POST hits the gateway, not the issuer — `pkg/http/authorization_mcp_test.go`.

## Example: separate gateway from issuer

```toml
require_oauth     = true
authorization_url = "https://idp.example.com/realms/users"     # validates user JWT
oauth_audience    = "kubernetes-mcp-server"

token_exchange_strategy = "rfc8693"
sts_token_url           = "https://sts-gateway.internal/oauth/token" # mints cluster token
sts_client_id           = "mcp-backend"
sts_client_secret       = "..."
sts_audience            = "kubernetes-api"
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -551,6 +551,7 @@ Configure OAuth/OIDC authentication for HTTP mode deployments.
 | `sts_federated_token_file` | string | `""` | Path to a JWT file from an external identity provider, e.g., SPIRE JWT-SVID (for `federated` auth style). |
 | `sts_subject_token_type` | string | `"urn:ietf:params:oauth:token-type:access_token"` | `subject_token_type` form parameter for RFC 8693 token exchange. Override to `urn:ietf:params:oauth:token-type:jwt` for cross-realm flows. |
 | `sts_requested_token_type` | string | `"urn:ietf:params:oauth:token-type:access_token"` | `requested_token_type` form parameter for RFC 8693 token exchange. The parameter is OPTIONAL per RFC 8693 §2.1, which leaves the issued token type to the authorization server's discretion when unspecified; this server sends `access_token` when the key is unset. Override to `urn:ietf:params:oauth:token-type:jwt` when the STS must mint a fresh JWT rather than echo the subject token shape. |
+| `sts_token_url` | string | `""` | Explicit token endpoint for token exchange. When set, takes precedence over the endpoint discovered from the OIDC provider at `authorization_url`. Use this to point token exchange at a separate STS gateway, or to enable token exchange when no `authorization_url` is configured (e.g., with `skip_jwt_verification=true`). |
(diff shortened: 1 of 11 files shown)


zruchi:
@Cali0707 this will complete our requirement re: defining a different token endpoint for rfc8693.

zruchi:
/cc @Cali0707 @manusa 
keen to merge this changes. let me know your thoughts.

Cali0707:
Hey @zruchi - will come back to this PR in a bit. I've actually been meaning to consolidate away from the STS approach into just the RFC 8693 token exchanger (STS is mostly a subset of the 8693 exchange, it just also has some basic auth bits we probably need to add to the 8693 exchange). Currently working on a PR that changes some of the config and internals around for this. 


Will ping you on that PR to take a look and see how it fits with what you are trying to do here

zruchi:
> Hey @zruchi - will come back to this PR in a bit. I've actually been meaning to consolidate away from the STS approach into just the RFC 8693 token exchanger (STS is mostly a subset of the 8693 exchange, it just also has some basic auth bits we probably need to add to the 8693 exchange). Currently working on a PR that changes some of the config and internals around for this.
> 
> Will ping you on that PR to take a look and see how it fits with what you are trying to do here

Sounds good, will hold off and wait for your PR to see if it fits our use case.

PR #1399, opened 2026-08-31
feat(kubevirt): add VM console screenshot tool

codingben:
Add the vm_console_screenshot tool for capturing
running VM graphical (VNC) console as PNG
images.

**Note:** This is ported from [codingben/kubevirt-console-mcp](https://github.com/codingben/kubevirt-console-mcp) repository which is the original source of it.

### DIFF
--- a/evals/tasks/kubevirt/vm-console-screenshot/task.yaml
+++ b/evals/tasks/kubevirt/vm-console-screenshot/task.yaml
@@ -1,0 +1,82 @@
+kind: Task
+apiVersion: mcpchecker/v1alpha2
+metadata:
+  labels:
+    suite: kubevirt
+    project: kubevirt
+    requires: kubevirt
+  annotations:
+    project-name: "KubeVirt"
+    project-url: "https://kubevirt.io"
+  name: "vm-console-screenshot"
+  difficulty: easy
+spec:
+  requires:
+    - extension: kubernetes
+      as: k8s
+  setup:
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-console-screenshot
+        ignoreNotFound: true
+    - k8s.create:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-console-screenshot
+    - k8s.create:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachine
+        metadata:
+          name: test-vm
+          namespace: vm-test-console-screenshot
+        spec:
+          runStrategy: Always
+          template:
+            spec:
+              domain:
+                devices:
+                  disks:
+                    - name: containerdisk
+                      disk:
+                        bus: virtio
+                memory:
+                  guest: 5Gi
+              terminationGracePeriodSeconds: 0
+              volumes:
+                - name: containerdisk
+                  containerDisk:
+                    image: quay.io/containerdisks/fedora:latest
+    - k8s.wait:
+        apiVersion: kubevirt.io/v1
+        kind: VirtualMachine
+        metadata:
+          name: test-vm
+          namespace: vm-test-console-screenshot
+        condition: Ready
+        timeout: 300s
+  verify:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          set -e
+          if ! kubectl get vm test-vm -n vm-test-console-screenshot &>/dev/null; then
+            echo "ERROR: VirtualMachine 'test-vm' not found"
+            exit 1
+          fi
+          echo "Verification passed: VM exists"
+          exit 0
+    - llmJudge:
+        contains: "successfully captured a screenshot image from the VM's graphical console and describes what is visible on it"
+  cleanup:
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-console-screenshot
+        ignoreNotFound: true
+  prompt:
+    inline: |
+      Take a screenshot of the graphical console for the virtual machine "test-vm" in the vm-test-console-screenshot namespace, so I can see what's currently displayed on its screen.
--- a/pkg/api/toolsets.go
+++ b/pkg/api/toolsets.go
@@ -9,6 +9,35 @@ import (
 	"github.com/google/jsonschema-go/jsonschema"
 )
 
+// ToolContentType distinguishes text from image content blocks.
+type ToolContentType string
+
+const (
+	ToolContentTypeText  ToolContentType = "text"
+	ToolContentTypeImage ToolContentType = "image"
+)
+
+// ToolContent represents a single content block in a multi-content tool result.
+// Used by tools that return both text and images (e.g., vm_console_screenshot).
+type ToolContent struct {
+	Type     ToolContentType
+	Text     string
+	Data     []byte // for image content, a copy of the image bytes
+	MIMEType string // for image content, e.g. "image/png"
+}
+
+// NewTextToolContent creates a text content block.
+func NewTextToolContent(text string) *ToolContent {
+	return &ToolContent{Type: ToolContentTypeText, Text: text}
+}
+
+// NewImageToolContent creates an image content block, copying the byte slice for safety.
+func NewImageToolContent(data []byte, mimeType string) *ToolContent {
+	dataCopy := make([]byte, len(data))
+	copy(dataCopy, data)
+	return &ToolContent{Type: ToolContentTypeImage, Data: dataCopy, MIMEType: mimeType}
+}
+
 type ServerTool struct {
 	Tool               Tool
 	Handler            ToolHandlerFunc
@@ -69,6 +98,9 @@ type ToolCallResult struct {
 	// When set, it is passed as structuredContent in the MCP CallToolResult alongside Content.
 	// Must be completely omitted (nil) when not used.
 	StructuredContent any
+	// ContentBlocks are SDK-neutral content blocks (text and/or images) for multi-content tools.
+	// When non-empty, these are used instead of the legacy Content field.
+	ContentBlocks []*ToolContent
 	// Error (non-protocol) to send back to the LLM.
 	Error error
 }
@@ -77,6 +109,16 @@ type ToolCallResult struct {
 // Use this for tools that return human-readable text output.
 func NewToolCallResult(content string, err error) *ToolCallResult {
 	return NewToolCallResultFull(content, nil, err)
+}
+
+// NewToolCallResultWithContent creates a ToolCallResult with content blocks (text and/or images).
+// Use this for multi-content tools like vm_console_screenshot.
+func NewToolCallResultWithContent(blocks []*ToolContent, structured any, err error) *ToolCallResult {
+	return &ToolCallResult{
+		ContentBlocks:     blocks,
+		StructuredContent: structured,
+		Error:             err,
+	}
 }
 
 // NewToolCallResultFull creates a ToolCallResult with both human-readable text
--- a/pkg/kubevirt/console_client.go
+++ b/pkg/kubevirt/console_client.go
@@ -1,0 +1,44 @@
+package kubevirt
+
+import (
+	"context"
+	"fmt"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/dynamic"
+)
+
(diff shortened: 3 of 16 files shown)


codingben:
@lyarwood @ksimon1 What do you think about splitting this PR into two PRs?

One PR to add `vm_console_screenshot` and another one `vm_console_capture` which both mostly important ones.

ksimon1:
@codingben Please divide this PR into 2 PRs and divide evals and functionality code into separate commits.

codingben:
@ksimon1 Yeah it was huge previously. Let's start by adding one tooling.

pkg/mcp/multi_content_result.go
```diff
@@ -0,0 +1,42 @@
+package mcp
+
+import (
+	"fmt"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+func NewMultiContentResult(blocks []*api.ToolContent, structured any, err error) *mcp.CallToolResult {
```
ksimon1, line 10: structured parameter is not used anywhere

evals/tasks/kubevirt/vm-console-screenshot/task.yaml
```diff
@@ -0,0 +1,82 @@
...
+  verify:
+    - script:
+        inline: |-
+          #!/usr/bin/env bash
+          set -e
+          if ! kubectl get vm test-vm -n vm-test-console-screenshot &>/dev/null; then
+            echo "ERROR: VirtualMachine 'test-vm' not found"
+            exit 1
+          fi
+          echo "Verification passed: VM exists"
```
ksimon1, line 69: Verification should test, that the image was retrieved

pkg/kubevirt/console_screenshot.go
```diff
@@ -0,0 +1,87 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+)
+
+// ConsoleMaxScreenshotBytes bounds the size of a VNC screenshot PNG read into
+// memory. Responses larger than this are rejected rather than buffered whole.
+const ConsoleMaxScreenshotBytes = 4 * 1024 * 1024 // 4 MiB
+
+func Screenshot(ctx context.Context, dynamicClient dynamic.Interface, restConfig *rest.Config, namespace, name string, wakeScreen bool) ([]byte, string, error) {
```
ksimon1, line 21: This is very long function, please divide it into more functions.

pkg/kubevirt/console_client.go
```diff
@@ -0,0 +1,48 @@
...
+func getVMI(ctx context.Context, dynamicClient dynamic.Interface, namespace, name string) (*unstructured.Unstructured, error) {
+	vmi, err := dynamicClient.Resource(VirtualMachineInstanceGVR).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		switch {
+		case apierrors.IsNotFound(err):
+			return nil, &ConsoleError{
+				Code:    ConsoleCodeVMINotFound,
+				Message: fmt.Sprintf("VirtualMachineInstance %q not found in namespace %q", name, namespace),
+			}
+		case apierrors.IsForbidden(err) || apierrors.IsUnauthorized(err):
```
ksimon1: Is this case possible? If account doesn't have access to VMIs, then I think kubevirt is unusable

ksimon1 (changes requested):
Please divide the eval test into separate commit than code

codingben:
> Please divide the eval test into separate commit than code

There is only one yaml file, why it deserves a separate commit in this PR? usually we kept it within the same commit in previous PRs.

pkg/mcp/multi_content_result.go
codingben: Fixed.

evals/tasks/kubevirt/vm-console-screenshot/task.yaml
codingben: Fixed.

pkg/kubevirt/console_screenshot.go
codingben: Done.

pkg/kubevirt/console_client.go
codingben: I've addressed it.

pkg/kubevirt/console_errors.go
```diff
@@ -0,0 +1,35 @@
+package kubevirt
+
+import "fmt"
+
+type ConsoleErrorCode string
+
+const (
+	ConsoleCodeVMINotFound           ConsoleErrorCode = "vmi_not_found"
```
ksimon1, line 8: Since you introduced this new file, do you plan in follow up to adjust already existing code?

codingben: No plans, this is only for console at the moment

evals/tasks/kubevirt/vm-console-screenshot/task.yaml
ksimon1, line 69: The script still does not verify, that the screenshot was taken

pkg/kubevirt/console_client_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+	dynamicfake "k8s.io/client-go/dynamic/fake"
+)
+
+func newTestVMI(namespace, name, phase string, autoattachGraphics *bool) *unstructured.Unstructured {
+	devices := map[string]any{}
+	if autoattachGraphics != nil {
+		devices["autoattachGraphicsDevice"] = *autoattachGraphics
+	}
+	return &unstructured.Unstructured{Object: map[string]any{
```
ksimon1, line 22: Nit: since multiple other tests are already creating the VM manifest, I think it would make sense to create fn e.g. like GetVMI(metadata, spec, status map[string]any), which will prefill ApiVersion, Kind, ...
Feel free to ignore this comment

ksimon1 (commented):
Last two more comments

pkg/toolsets/kubevirt/vm/console/tool_test.go
```diff
@@ -0,0 +1,60 @@
+package console
+
+import (
+	"slices"
+	"testing"
+
+	"k8s.io/utils/ptr"
+)
+
+func TestTools(t *testing.T) {
```
ksimon1, line 10: Can you convert this to a testify suite? All the other tool tests under pkg/toolsets/kubevirt use suites (see vm/create and vm/guestagent), and docs/dev/testing.md asks for it. Same for pkg/mcp/multi_content_result_test.go.

pkg/kubevirt/console_client_test.go
```diff
@@ -0,0 +1,118 @@
...
+	var ce *ConsoleError
+	if !errors.As(err, &ce) {
+		t.Fatalf("expected *ConsoleError, got %T: %v", err, err)
+	}
+	if ce.Code != want {
+		t.Fatalf("expected code %q, got %q (%s)", want, ce.Code, ce.Message)
+	}
+}
+
+func TestGetVMI(t *testing.T) {
```
ksimon1, line 62: These test unexported functions. Screenshot() already covers all of these paths through the public API in console_screenshot_test.go, so I would just drop this file.

pkg/kubevirt/console_client.go
```diff
@@ -0,0 +1,44 @@
...
+func getVMI(ctx context.Context, dynamicClient dynamic.Interface, namespace, name string) (*unstructured.Unstructured, error) {
+	vmi, err := dynamicClient.Resource(VirtualMachineInstanceGVR).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, &ConsoleError{
+				Code:    ConsoleCodeVMINotFound,
+				Message: fmt.Sprintf("VirtualMachineInstance %q not found in namespace %q", name, namespace),
+			}
+		}
+		return nil, &ConsoleError{Code: ConsoleCodeInternal, Message: "failed to retrieve the VirtualMachineInstance", Cause: err}
```
ksimon1, line 22: You already have ConsoleCodePermissionDenied and use it in mapScreenshotStreamError. Can you do the same forbidden/unauthorized check here? Right now a missing permission to get the VMI surfaces as 'internal', which is not very helpful for the user.

pkg/kubevirt/console_screenshot.go
```diff
@@ -0,0 +1,110 @@
...
+	data, err := fetchScreenshot(ctx, restConfig, namespace, name, wakeScreen)
+	if err != nil {
+		return nil, "", err
+	}
+
+	if err := validatePNG(data); err != nil {
+		return nil, "", err
+	}
+
+	return data, "Screenshot captured from the VNC graphical console.", nil
```
ksimon1, line 40: This string is always the same, so it is presentation rather than cluster logic. I would drop the second return value and build the text in the tool handler - then it could also mention namespace/name or the image size from DecodeConfig.

pkg/kubevirt/console_screenshot.go
```diff
@@ -0,0 +1,110 @@
...
+	if len(data) == 0 {
+		return nil, &ConsoleError{Code: ConsoleCodeScreenshotUnavailable, Message: "the VNC screenshot subresource returned no data"}
+	}
+
+	return data, nil
+}
+
+func validatePNG(data []byte) error {
+	if _, err := png.DecodeConfig(bytes.NewReader(data)); err != nil {
+		return &ConsoleError{Code: ConsoleCodeScreenshotUnavailable, Message: "the screenshot subresource did not return a valid PNG image"}
```
ksimon1, line 100: Missing Cause: err. Every other ConsoleError in this file sets it, and the decode error would be useful in the logs.

pkg/toolsets/kubevirt/toolset.go
```diff
@@ -28,6 +29,7 @@ func (t *Toolset) GetDescription() string {
 
 func (t *Toolset) GetTools(p api.FilteringProvider) []api.ServerTool {
 	return slices.Concat(
+		vm_console.Tools(p),
```
ksimon1, line 32: nit: the rest of the list is alphabetical, please move this below vm_clone. It also makes the generated README list vm_console_screenshot before vm_clone, while the testdata json has it after.

pkg/api/toolsets.go
```diff
@@ -9,6 +9,35 @@ import (
...
+	MIMEType string // for image content, e.g. "image/png"
+}
+
+// NewTextToolContent creates a text content block.
+func NewTextToolContent(text string) *ToolContent {
+	return &ToolContent{Type: ToolContentTypeText, Text: text}
+}
+
+// NewImageToolContent creates an image content block, copying the byte slice for safety.
+func NewImageToolContent(data []byte, mimeType string) *ToolContent {
```
ksimon1: nit: is the copy really needed? The only caller passes a freshly read buffer that is never reused, and this can be up to 4 MiB. If it is meant as an API contract, please mention it in the godoc.

evals/tasks/kubevirt/vm-console-screenshot/task.yaml
```diff
@@ -0,0 +1,82 @@
...
+  cleanup:
+    - k8s.delete:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: vm-test-console-screenshot
+        ignoreNotFound: true
+  prompt:
+    inline: |
+      Take a screenshot of the graphical console for the virtual machine "test-vm" in the vm-test-console-screenshot namespace, so I can see what's currently displayed on its screen.
```
Cali0707, line 82: What is the expected output/utility of this tool?

If it just streams an image back, do MCP hosts generally display that to the user? Would this maybe make more sense as an MCP app component?

Or are models able to understand what is in the console image from the file sent back from the tool?

codingben: > What is the expected output/utility of this tool?

The tools calls to KubeVirt API subresource to take screenshot of the VM console. Please watch the [demo](https://github.com/codingben/kubevirt-console-mcp#demo) I prepared for it and you can read the [article](https://codingben.io/building-an-mcp-server-for-kubevirt-guest-consoles/) about it.

> If it just streams an image back, do MCP hosts generally display that to the user? Would this maybe make more sense as an MCP app component?

No. The purpose of this MCP tool is to help AI analyze and understand the failure behind the VM, that is reported as running – because **VM that is running** can be on kernel panic or any other error that is not visible in k8s object's statuses.

> Or are models able to understand what is in the console image from the file sent back from the tool?

Most modern AI models (e.g. GPT-5.6 Luna, Claude, etc.) are multimodal and can process both text and image inputs. Therefore, if the tool returns a screenshot of the console, the model can directly analyze the image and interpret the console output.

PR #1401, opened 2026-09-01
feat(tokenexchange)!: migrate from sts_* to explicit config block

Cali0707:
As we added the tokenexchange work, we ended up in a state where we had a legacy STS exchange and the new strategy based tokenexchange. This consolidates into the token exchange, with an explicit config block and no more legacy STS implementation.

This should be safe since STS is a subset of the full RFC 8693 spec, which we already supported through the token exchange.


test/e2e/keycloak_test.go
```diff
@@ -64,7 +64,7 @@ func TestKeycloakOIDC(t *testing.T) {
 			//    mounted so it can trust Keycloak's TLS. The CA secret is created
 			//    by the copyKeycloakCASecret pre-install hook before Helm install.
 			dep := deployServer(ctx, t, cfg, "keycloak-oidc",
-				withConfig(oidcServerConfig(nil)),
+				withConfig(oidcServerConfig(nil, "client_secret_post")),
```
matzew: should this be constant? I see it often - I guess won't change - but just wondering 😅

nader-ziada (approved):
great job@Cali0707 all looks good to me

The migration from the flat sts_* keys to the structured [token_exchange] / [token_exchange.client_auth] block is solid, just want to call out that we have to document this as a  breaking config change in the release notes

manusa (commented):
Thanks for taking this on, @Cali0707. Collapsing the legacy STS implementation and the strategy-based exchange into one explicit block is the right call, and it also closes two fail-open paths that were worth removing on their own: previously, an `IsEnabled()` false case (missing `sts_client_id` or `sts_audience`) and an unregistered strategy both forwarded the user's token to the cluster unchanged, with no error. The removed-key detector is a nice touch too, most projects would just break on upgrade.

I reviewed this personally and also ran it through an agent-based review, so the list below is a mix of both. Items I'd like addressed:

1. `docs/KEYCLOAK_OIDC_SETUP.md:182`: `certificate_authority` now sits after `[token_exchange.client_auth]`, so TOML parses it as `token_exchange.client_auth.certificate_authority`. There's no field by that name and the decoder is not strict, so it is silently discarded and anyone copying this snippet gets a server that cannot trust Keycloak's CA, with no error to go on. The real template gets this right (`dev/config/keycloak/config.toml:6` keeps it above the tables), and it's the same hazard you already fixed in `oauth_flows_test.go:174`.

2. The migrated configs change client authentication from HTTP Basic to form-body. Configs with `sts_client_id`/`sts_client_secret`/`sts_audience` and no `token_exchange_strategy` took the built-in STS path, which sent credentials as `Authorization: Basic` (`externalaccount.NewTokenSource` hard-codes `AuthStyleInHeader`, and the `sts_test.go` deleted here asserted exactly that). `client_secret_post` maps to `AuthStyleParams`, i.e. body credentials. No capability is lost and nobody breaks silently, since a stale `sts_*` config now fails at startup, but an operator who hits that error, copies `client_secret_post` from our docs and starts cleanly will then get `401 invalid_client` on every call if their IdP client is registered for `client_secret_basic`. Please switch the configs that replace legacy STS setups to `client_secret_basic` (`dev/config/keycloak/config.toml:14`, `docs/KEYCLOAK_OIDC_SETUP.md:178`, `docs/configuration.md:570`). I would leave the `oidcServerConfig` default on `client_secret_post`: before this PR no e2e exercised body credentials at all, since C5 went through the legacy Basic path and D3 was Basic too, so the current split of C5 on `client_secret_post`, D3 on `client_secret_basic` and D4 on `private_key_jwt` covers more than before and is worth keeping as is. The `entra-obo` snippets are already mapped correctly, they ran `params` before.

3. `pkg/config/config.go:792-817`: since this is the choke point every migrating user passes through, please have it carry the value-level mapping as well. The key rename is only half the migration for `sts_auth_style`, so the message should say `params` becomes `client_secret_post`, `header` becomes `client_secret_basic`, `assertion` becomes `private_key_jwt`, `federated` becomes `jwt_file`. The legacy built-in STS shape is also detectable from key presence alone: if `sts_client_id` is present and `token_exchange_strategy` is absent, that deployment was using HTTP Basic, so the error can point straight at `client_secret_basic`. That turns the whole class of post-migration 401s into a startup message, which reaches people who never read release notes.

4. `pkg/config/config.go:807`: the check only matches top-level keys. `md.Undecoded()` returns dotted paths while `locations` holds bare names, so a half-migrated config (`[token_exchange]` with `strategy` set plus a leftover `sts_audience` inside the table) yields `token_exchange.sts_audience`, matches nothing, and is silently dropped, leaving `audience` empty and the exchange requesting the wrong target at runtime. Matching on the final path segment would catch it. More generally, since the decoder is not strict, a typo in any of the 11 new nested keys is silently ignored, so rejecting unknown keys under `token_exchange` would be a cheap safety net for a breaking migration.

5. `pkg/config/config.go:398`: this is the only validation inside `ReadToml`, while everything else lives in `Validate(ctx)`. I understand it cannot simply move, `toml.MetaData` only exists here and `Validate` sees a struct that no longer has these fields, so decode time is right. Two alternatives worth considering: detect in `readAndMergeFiles`, where each file is decoded separately and the filename is known, so the error can name which file holds the stale key (today the merge flattens the main config and drop-ins into one map before re-encoding, so it cannot); or give metadata-level policy a single entry point so the unknown-key check from the previous item has a home instead of becoming a third scattered check.

6. `pkg/config/config.go:656-683`: a public client can no longer be configured. Before, `strategy` plus `sts_client_id` with no secret was valid and sent `client_id` with no `client_secret`, since `injectClientAuth` only sets the secret when non-empty. Now any client-auth field makes `authConfigured()` true, which forces a `method`, and every method requires a second credential. Dropping the `client_auth` block is not equivalent, `applyClientAuth` returns early, `ClientID` stays empty and `injectClientAuth` bails at `cfg.ClientID == ""`, so `client_id` is never sent at all. If this is intentional it's worth calling out, otherwise it needs a way back.

7. Test coverage dropped in a few places that matter for a breaking auth change: `normalizeTokenExchange` (`config.go:760-776`) is now entirely untested, all nine whitespace-trimming subtests were deleted without replacement, and `method = " client_secret_basic "` from a mounted Secret is exactly the case it guards; `validateTokenExchangeFile`'s `os.Stat` branch (`config.go:786-788`) has no test, only the empty-path case, and that's the check that catches an unmounted key or token file at boot rather than per request; the `default` invalid-method arm (`config.go:681-682`) and the `client_id is required` branch (`config.go:662`) are unreached by any test; and only 2 of the 11 migration mappings are asserted (`config_test.go:1168-1169`), which given the non-strict decoder means a dropped map entry would silently disable token exchange rather than report it.

8. Related to the above, no test anywhere asserts the `client_secret_basic` wire format any more. `sts_test.go` was the only one that sent a request and checked for `Authorization: Basic`, and its replacements only assert config mapping. I would not resurrect that file, it tested the deleted `SecurityTokenService` type, but `rfc8693_exchanger_test.go:29-51` already has the right harness: adding an `authorization` field to `recordedRequest` and capturing `r.Header.Get(HeaderAuthorization)` would let one subtest per auth style cover this, and would also close the missing `AuthStyle` assertions for `private_key_jwt` and `jwt_file`.

9. `docs/configuration.md:543`: the row lost the list of valid values that `token_exchange_strategy` used to carry, and this is now the one key that is mandatory (`config.go:652` rejects empty or unregistered) yet it is shown with "Default: none". Please restore `rfc8693`, `keycloak-v1`, `entra-obo`, which is also the order the validation error prints, mark it required, and mention that the block now requires `require_oauth = true` and `authorization_url`. The latter is a new startup rejection for configs that previously started. A short `sts_*` to `token_exchange.*` migration table in this section would help too, and as @nader-ziada noted, this needs to be called out in the release notes.

10. `pkg/kubernetes/token_exchange.go:44-45`: `global` can be nil here, the branch just above tests for exactly that, so `global.GetStrategy()` panics if a caller passes a non-nil `globalConfig` with a base config that has no `[token_exchange]`. The in-repo caller is safe because both values derive from the same `baseConfig`, but this is exported and the coupling is invisible in the signature. Hoisting `global` above the `globalConfig == nil` check removes it. While there, the untyped-nil contract of `GetTokenExchangeConfig` and `GetClientAuth` (`config.go:445-449` and `config.go:53-58`) is load-bearing and undocumented: returning a typed nil instead would make the interface non-nil and panic on every request. Worth a comment on the `api` interfaces, and note that `s.Nil` at `config_test.go:1158` does not catch it, testify unwraps the pointer.

11. `pkg/kubernetes/provider_token_exchange.go:125-137`: the switch has no `default`, and the guard that used to catch an unmapped method is gone with the `cfg.Validate()` call this PR removes. An unrecognized method leaves `AuthStyle` empty, which `injectClientAuth` treats as `AuthStyleParams`, so it degrades to `client_secret_post` with `ClientID` and `ClientSecret` already assigned two lines earlier. `StaticConfig.Validate` blocks this for TOML-loaded configs so it is defense in depth, but note that `TargetTokenExchangeConfig.Validate()` is now referenced nowhere, which leaves the per-target `TokenExchangeProvider` path with no validation at all.

12. `pkg/config/config.go:40-41`: `Strategy` and `Audience` are the only fields in the struct without `omitempty`, and the flat keys they replace all had it. `mergeConfig` (`config_default.go:39-48`) encodes an override to TOML and decodes it onto the base, so a downstream override that sets `TokenExchange` to adjust only `Scopes` will also emit `strategy = ""` and `audience = ""` and clobber them.

13. A few diagnostics regressed. `getOrBuildTokenExchangeConfig` returns at `provider_token_exchange.go:74-76` before taking the mutex, so removing `[token_exchange]` on reload strands the previous config and its memoized HTTP client with its idle connections to the IdP, and `Close()` (`:207-209`) only delegates. The warning that used to fire when the OIDC provider returned an empty token URL (`:84-86`) is gone, so that condition is now only visible in the client-facing error. And every global-path failure is now reported as `token exchange failed for target ""`, since target is empty for single cluster, where the previous message named the strategy.

14. `pkg/config/config.go:648-651`: with the fallback to `tokenexchange.GetRegisteredStrategies()`, `WithTokenExchangeStrategies` no longer changes any outcome, both production callers pass exactly that list. Its doc comment (`:512-520`) also still refers to `token_exchange_strategy`. Either drop it or update the contract. Same kind of drift in `test/e2e/README.md:106`, which still documents the one-argument `oidcServerConfig(scopes)`.

For what it's worth, CI is green including e2e, and I did not find anything that breaks a valid current deployment without telling the operator about it.

Cali0707:
FYI @zruchi this is the relevant tokenexchange/sts refactor PR

zruchi:
@Cali0707 Nice, this is a solid consolidation.

One thing to flag — right now the token URL still comes only from the OIDC provider's discovery endpoint, so it assumes the issuer and the rc8693 provider are the same service. We actually need those decoupled for cross-realm RFC 8693 setup. I can make a follow up after this PR lands.

zruchi (approved):

docs/configuration.md
```diff
@@ -536,35 +536,58 @@ Configure OAuth/OIDC authentication for HTTP mode deployments.
...
-| `sts_federated_token_file` | string | `""` | Path to a JWT file from an external identity provider, e.g., SPIRE JWT-SVID (for `federated` auth style). |
-| `sts_subject_token_type` | string | `"urn:ietf:params:oauth:token-type:access_token"` | `subject_token_type` form parameter for RFC 8693 token exchange. Override to `urn:ietf:params:oauth:token-type:jwt` for cross-realm flows. |
-| `sts_requested_token_type` | string | `"urn:ietf:params:oauth:token-type:access_token"` | `requested_token_type` form parameter for RFC 8693 token exchange. The parameter is OPTIONAL per RFC 8693 §2.1, which leaves the issued token type to the authorization server's discretion when unspecified; this server sends `access_token` when the key is unset. Override to `urn:ietf:params:oauth:token-type:jwt` when the STS must mint a fresh JWT rather than echo the subject token shape. |
+| `token_exchange.strategy` | string | required | Registered token exchange strategy: `rfc8693`, `keycloak-v1`, or `entra-obo`. The block enables global exchange and requires `require_oauth = true` and `authorization_url`. |
```
2uasimojo: nit: 
```
 | required | ... |
```
is not super clear (at least to someone like me who doesn't know things about token exchanging): Does it mean the default value is `"required"` or that I am required to supply a value in this field?

Assuming the latter (which is what my agent tells me is the case), presumably the qualifier is that it's required IFF the `[token_exchange]` section is provided.

If this is obvious to the savvy reader, no change is needed; otherwise perhaps it would play better as:
```
 | none | Required when `[token_exchange]` is used. Valid values are ... |

File: docs/configuration.md [hunk 3]
```diff
@@ -536,35 +536,58 @@ Configure OAuth/OIDC authentication for HTTP mode deployments.
...
+| `token_exchange.client_auth.client_id` | string | `""` | OAuth client ID. May be configured without a method or secret for a public client. |
+| `token_exchange.client_auth.client_secret` | string | `""` | Required by the client-secret methods. |
+| `token_exchange.client_auth.certificate_file` | string | `""` | Certificate PEM required by `private_key_jwt`. |
+| `token_exchange.client_auth.private_key_file` | string | `""` | Private-key PEM required by `private_key_jwt`. |
+| `token_exchange.client_auth.token_file` | string | `""` | JWT file required by `jwt_file`. |
 | `cluster_auth_mode` | string | `""` | Cluster auth mode: `passthrough` (forward Authorization header when present, fall back to kubeconfig when absent) or `kubeconfig` (always use kubeconfig credentials). Defaults to `passthrough`. |
 | `certificate_authority` | string | `""` | Path to CA certificate for validating authorization server connections. |
 | `server_url` | string | `""` | Public URL of the MCP server (used for OAuth metadata). |
 
+**Breaking-change migration:** The legacy top-level `token_exchange_strategy` and `sts_*` settings are removed and rejected at startup. Migrate them as follows:
```
2uasimojo: Eh, it would be nice if this doc chunk lived in a separate "release notes" space (intended to represent movement) rather than the main doc (intended to represent "how things work today").

Otherwise we kinda have to mark this section for removal "later". (When? You tell me 🤷)

Also, Calum and I were tossing around the idea of some sort of best-effort "config migrator" tool. No action necessary on that; just something to noodle.

docs/configuration.md
```diff
@@ -536,35 +536,58 @@ Configure OAuth/OIDC authentication for HTTP mode deployments.
...
+| `token_exchange.strategy` | string | required | Registered token exchange strategy: `rfc8693`, `keycloak-v1`, or `entra-obo`. The block enables global exchange and requires `require_oauth = true` and `authorization_url`. |
+| `token_exchange.audience` | string | `""` | Audience for the exchanged token. |
+| `token_exchange.scopes` | string[] | `[]` | Scopes for the exchanged token. |
+| `token_exchange.subject_token_type` | string | `"urn:ietf:params:oauth:token-type:access_token"` | RFC 8693 `subject_token_type`. |
+| `token_exchange.requested_token_type` | string | `"urn:ietf:params:oauth:token-type:access_token"` | RFC 8693 `requested_token_type`. |
+| `token_exchange.client_auth.method` | string | none | Required when client credentials are configured: `client_secret_basic`, `client_secret_post`, `private_key_jwt`, or `jwt_file`. May be omitted when only `client_id` is set for a public client. |
+| `token_exchange.client_auth.client_id` | string | `""` | OAuth client ID. May be configured without a method or secret for a public client. |
+| `token_exchange.client_auth.client_secret` | string | `""` | Required by the client-secret methods. |
```
2uasimojo, line 550: [Latent] IIUC the value here is an actual secret (sensitive data). *Somewhere* it would be worth calling out that, when standing up the MCP server in a cluster, when configs contain sensitive data, they should be mounted from Secrets rather than ConfigMaps.

And some kind of visual indicator in the table for such fields would be nice.

pkg/config/config.go
```diff
@@ -35,6 +35,45 @@ type ToolOverride struct {
...
+func (c *TokenExchangeConfig) GetClientAuth() api.TokenExchangeClientAuth {
+	if c.ClientAuth == nil {
+		return nil
+	}
+	return c.ClientAuth
+}
+
+// TokenExchangeClientAuth is the TOML configuration for token exchange client authentication.
+type TokenExchangeClientAuth struct {
+	Method          api.TokenExchangeClientAuthMethod `toml:"method"`
```
2uasimojo: nit: in theory omitting `omitempty` would cause an unspecified `method` in an _override_ to blank out the base. Not a thing here, and probably not a thing ever, but technically possible if e.g. a downstream decided to set a default in the base config.

...whereas there's not a significant downside to including `omitempty`.

2uasimojo (approved):
With the caveat that I'm not versed in the actual technology in play (token exchange things)...

I really had to stretch to find things to say here; according to my agent, y'all (and your agents) have wrapped this up nicely.

A few minor issues noted inline, which could be addressed in a followon, or ignored.

/lgtm

docs/configuration.md
Cali0707: IMO let's do this as a follow up

2uasimojo (commented):
/lgtm

Thanks for the update!

PR #1402, opened 2026-09-01
fix: minikube setup on mac/podman has working dns

Cali0707:
Without this, the gateway installs in a bad state as it cannot resolve istiod


dlom:
LGTM

Issue #1374, opened 2026-08-17
Consolidate Config Loading and Precedence


Cali0707:
Currently config is sprawled throughout the codebase. While almost all of it lives in `pkg/config`, some of the logic around how the config is used sprawls. In particular, resolution from env vars vs. config value vs. cli flag lives in many places.

We should:
- [ ] Make a unified config parser that consistently handles cli flags, config values, and env vars in a pre defined precedence
  - [ ] Log at startup what is actually running (and ideally where it came from, but this isn't needed for the first iteration)
- [ ] Look throughout the repo and pull anything that doesn't match our handling into the shared parser (e.g. otel, some start up values, etc.)
- [ ] Reduce the number of cli flags to only the minimum needed to startup/do one off overrides for different envs

Cali0707:
cc @2uasimojo I believe you had some thoughts on this one

2uasimojo:
The only thing I would add from our discussion was the notion of using a struct/class-with-methods to denote the characteristics of a config option. Kind of a reinvention of argparse.
- Availability via the various mechanisms (default, CLI opt, env, config opt).
- Spelling of the option in each of the above. (Potential for automatic mapping, but perhaps better explicit?)
- Description / help text (feeds generated README, manpage, etc.)
- Internal data type.
- Validation pattern/func.
- Default value.
- Is it SIGHUP-reloadable?

...and then once we've done all the processing:
- The value we're gonna use.
- Where that value came from (default, CLI, env, config).
- One or more methods to dump the value. (`portOpt.String()` => `"0.0.0.0"`; `portOpt.Describe()` => `"0.0.0.0" (value from config file /path/to/foo.toml)` kind of thing)

PR #1404, opened 2026-09-02
[feat] Design spec for config consolidation

2uasimojo:
Typed-Out-By: cursor/grok
Architected-By: moi
Starts-To-Address: #1374

### DIFF
--- a/docs/specs/config.md
+++ b/docs/specs/config.md
@@ -1,0 +1,1641 @@
+# Config Option Framework
+
+Status: **Proposed**
+
+This spec is the implemented configuration engine, originally sketched in
+[containers/kubernetes-mcp-server#1374](https://github.com/containers/kubernetes-mcp-server/issues/1374).
+There is no deprecation window and no compatibility shims.
+
+Implementation has not started. Do not treat this file as a description of
+current code. Current behavior lives in [`docs/configuration.md`](../configuration.md).
+
+The [master table](#master-table) records which options have which vectors
+(TOML / env) and how those are spelled. Empty TOML / Env / CLI cells mean
+that vector is **not available**. CLI cells are empty for every runtime
+option; the only flags are bootstrap (`--config`, `--config-dir`,
+`--version`).
+
+## Problem (previous surface)
+
+Before this engine, config was defined and applied in several places:
+
+- TOML + defaults in [`pkg/config`](../../pkg/config) (`StaticConfig`)
+- CLI copies in [`pkg/kubernetes-mcp-server/cmd/root.go`](../../pkg/kubernetes-mcp-server/cmd/root.go)
+  (`loadFlags`, only if cobra `Changed`)
+- Getter-time `os.Getenv` for TLS and OTEL
+- Direct `os.Getenv` bypasses (`KUBE_CLIENT_QPS`, debounce windows, …)
+
+There was no provenance, SIGHUP did not re-apply CLI overrides, and help
+text / docs / flags were duplicated by hand.
+
+## Design
+
+- Type is `Config` (renamed from `StaticConfig`; `ConfigState` from
+  `StaticConfigState`).
+- Runtime values live on `Config` as `Option[T]`. Callers use
+  `cfg.Port.Get()`, `cfg.Port.Describe()`, `cfg.Port.Source()`.
+- **Source is a string**, not an enum. See [Source](#source).
+- **Vectors are data, not engine policy.** Each option opts into TOML /
+  env by having a non-empty name for that vector. Empty = not available.
+  There is no CLI vector on `Option`.
+- **Precedence among enabled vectors is fixed:** default &lt; file (main,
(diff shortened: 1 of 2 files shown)


docs/specs/config.md
```diff
@@ -0,0 +1,1524 @@
...
+    Description                string
+    Default                    T
+    ParseEnv                   func(string) (T, error) // required if EnvName != ""
+    ParseTOML                  func(any) (T, error)    // required if TOMLKey != ""
+    Validate                   func(T) error           // per-value; cross-field stays on Config.Validate
+    Reloadable                 bool
+    Sensitive                  bool // redact in String / Describe / startup dump
+    HiddenFlag                 bool // cobra hidden, if FlagName set
+    value                      T
+    source                     Source // token or winning file path
```
Cali0707, line 70: Could you add a quick example for how nested config values should be defined? Maybe something like:
```go
type Config struct {
    TokenExchange TokenExchangeConfig `toml:"token_exchange"`
}

type TokenExchangeConfig struct {
    Strategy     Option[string] `toml:"strategy"`
    ClientSecret Option[string] `toml:"client_secret"`
}

func newTokenExchangeConfig() TokenExchangeConfig {
    return TokenExchangeConfig{
        Strategy: StringOption(
            "",
            "Token exchange strategy",
            LiveReload,
        ),
        ClientSecret: StringOption(
            "",
            "Token exchange client secret",
            LiveReload,
            Sensitive(),
        ),
    }
}
```

docs/specs/config.md
```diff
@@ -0,0 +1,1524 @@
...
+
+const (
+    SourceDefault Source = "<Default>"
+    SourceEnv     Source = "<Env>"
+    SourceCLI     Source = "<CLI>"
+    SourceTest    Source = "<Test>"
+)
+
+type Option[T any] struct {
+    TOMLKey, EnvName, FlagName string // empty => that vector is off
```
Cali0707, line 61: Do we need TOMLKey, or can we stick with struct toml tags?

docs/specs/config.md
```diff
@@ -0,0 +1,1524 @@
...
+    Description                string
+    Default                    T
+    ParseEnv                   func(string) (T, error) // required if EnvName != ""
+    ParseTOML                  func(any) (T, error)    // required if TOMLKey != ""
+    Validate                   func(T) error           // per-value; cross-field stays on Config.Validate
+    Reloadable                 bool
+    Sensitive                  bool // redact in String / Describe / startup dump
+    HiddenFlag                 bool // cobra hidden, if FlagName set
```
Cali0707: Do we need all the CLI machinery for hiding flags?

If part of the proposal is to drop all hidden flags, maybe we can just not include support for any hidden flags to start?

docs/specs/config.md
```diff
@@ -0,0 +1,1524 @@
...
+<td>—</td>
+<td>—</td>
+<td>no</td>
+<td>no</td>
+<td>DROP <code>--require-tls</code></td>
+</tr>
+<tr>
+<td>tls_min_version</td>
+<td>string</td>
+<td><code>""</code> (effective TLS 1.2)</td>
```
Cali0707, line 920: Maybe for this (and other "" is effectively some value), we can just explicitly set that value rather than having some aliasing between an empty string and a default?

2uasimojo: IME that locks us into a losing maintenance burden: if default behaviors change underneath us, we have to duplicate to match.

This is exacerbated by the fact that not every default assignment is a simple static value. Sometimes there's logic. Now instead of staying current with an external value, we're duplicating function code etc. Gets hairy.

Above applies to defaults we don't "own" (and I think TLS qualifies, since we get it from the cluster?). Can look for ones we do own and see if it makes sense to define explicit defaults. Though again, sometimes they would be computed and/or dependent on other values, so it may not be feasible.

docs/specs/config.md
2uasimojo: Looks like we can't use tags because of the way these are going to need to be unmarshaled now. Words added.

docs/specs/config.md
Cali0707: Oh yeah that is a good point I hadn't considered

2uasimojo:
@Cali0707 Updated!

2uasimojo:
/retest

?

Cali0707:
LGTM from my end on the design here

2uasimojo:
[ce3a462](https://github.com/containers/kubernetes-mcp-server/pull/1404/commits/ce3a46255a540f56a308357e126f01355430160c): We'll enforce strict toml parsing: unknown fields result in an error.

2uasimojo:
01a1087: Rebase to pick up [metrics_port](https://github.com/containers/kubernetes-mcp-server/pull/1365) and [token exchange](https://github.com/containers/kubernetes-mcp-server/pull/1401) updates.

2uasimojo:
Merged under #1423

PR #1405, opened 2026-09-03
feat: support cluster CA served over HTTPS via kubeconfig extension

george-angel:
When a cluster's API-server CA is served over HTTPS, you can now put the URL in the kubeconfig instead of copying the certificate in.

Today, running the server against multiple clusters means pasting certificate-authority-data into every kubeconfig by hand. The CA changes whenever a cluster is rebuilt, so that copy goes stale and API calls start failing until someone updates the kubeconfig again.

Now a cluster entry can point at the URL that serves its CA:

```yaml
clusters:
  - name: production
    cluster:
      server: https://kubernetes.default.svc
      extensions:
        - name: kubernetes-mcp-server
          extension:
            caURL: https://ca.production.example.com/
```

The server fetches the CA when it starts, saves it to a cache file and verifies the cluster against that file. When the CA changes, the server picks up the new one automatically, so nothing breaks when a cluster is rebuilt. A background task keeps the cache fresh on a schedule you control with ca_refresh_interval (24h by default). With 0s the background task is disabled, and the CA is then refreshed when a manager is rebuilt or via the immediate re-fetch that SIGHUP config reload now triggers, so a rotated CA can be picked up without a restart in either mode.

The endpoint serving caURL must itself present a certificate the MCP process already trusts (system roots or SSL_CERT_FILE), not the cluster CA being fetched; fetches follow only https redirects, and a CA that cannot be fetched over https fails closed at startup. Other tools reading the kubeconfig ignore the extension, and the CA cache lives in a subdirectory of the system temp dir (honoring `$TMPDIR`), which needs to be writable - with a read-only root filesystem, set `$TMPDIR` to writable storage or mount an emptyDir at `/tmp`.

The Duration/omitzero config pattern on ca_refresh_interval is intentional: a zero value means "not set" and the config merge keeps the default, while an explicit "0s" in a config file or via the --ca-refresh-interval flag is honored. The one exception is build-time default overrides, which cannot express the disabled value; that limitation is documented in config.go.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -238,6 +238,8 @@ rate_limit_burst = 10        # allow bursts of up to 10 requests
 |-------|------|---------|-------------|
 | `kubeconfig` | string | `""` | Path to the Kubernetes configuration file. If not provided, the server uses the in-cluster configuration or the default kubeconfig location (`~/.kube/config`). |
 | `cluster_provider_strategy` | string | auto-detect | How the server finds clusters. Valid values: `kubeconfig`, `in-cluster`, `kcp`, `disabled`. |
+| `ca_cache_dir` | string | `""` | Directory where CA certificates fetched via the `caURL` kubeconfig cluster extension are cached. Empty means the default: a subdirectory of the system temp dir. The directory must be writable; with a read-only root filesystem, mount an emptyDir there. |
+| `ca_refresh_interval` | duration | `"24h"` | How often cached CA certificates are re-fetched so a rotated cluster CA is picked up without a restart. Use `0s` to disable re-fetching; the CA is then only refreshed when a manager is rebuilt (e.g. kubeconfig change or restart). |
 
 **Example:**
 ```toml
@@ -313,6 +315,33 @@ If the server starts but operations fail (e.g., `failed to list namespaces: unkn
 2. **Kubeconfig validity** — Ensure the kubeconfig contains valid credentials (token, client certificate, etc.) and points to the correct API server address.
 3. **Permissions** — The credentials in the kubeconfig must have sufficient RBAC permissions on the target cluster.
 4. **TLS certificates** — If the external cluster uses a private CA, the CA certificate must be included in the kubeconfig or mounted separately.
+
+#### CA served over HTTPS
+
+When a cluster's API-server CA is served over HTTPS (e.g. an endpoint mirroring the cluster CA for cross-cluster consumers), a kubeconfig cluster entry can point at that URL instead of embedding the CA:
+
+```yaml
+apiVersion: v1
+clusters:
+- name: production
+  cluster:
+    server: https://kubernetes.default.svc
+    extensions:
+    - name: kubernetes-mcp-server
+      extension:
+        caURL: https://ca.production.example.com/
+```
+
+At manager creation the server fetches the CA from `caURL`, caches it to a file and points the Kubernetes client at that file. Because the client reloads CA files it is pointed at, a rotated CA (e.g. after the cluster is re-provisioned) is picked up without a restart; `ca_refresh_interval` controls how often the cache is re-fetched (default `24h`, `0s` disables). Other tools ignore the extension entirely.
+
+Notes:
+
+- The URL must be `https`. A trust root fetched over cleartext could be swapped by a network attacker, who could then impersonate the cluster the CA is meant to authenticate.
+- The response must be a PEM CA certificate; other content fails fast with a clear error.
+- client-go re-checks the cached file every few minutes, so a rotation converges once the refresher writes the new CA; setting `ca_refresh_interval` below that cadence does not speed it up.
+- The cache directory (system temp dir by default, or `ca_cache_dir`) must be writable. With a read-only root filesystem, mount an emptyDir at the cache location — e.g. `emptyDir: {}` mounted at `/tmp`.
+- The initial fetch happens at manager creation; if it fails (unreachable endpoint, non-PEM response), manager creation fails rather than silently using an empty trust store. For the default context that is startup; lazily-created contexts (other targets) surface the error on first use.
+- Applies to every cluster provider strategy backed by a kubeconfig (`kubeconfig` and `kcp`).
 
 ### Access Control
 
@@ -775,6 +804,8 @@ The following options can be set via command-line arguments. CLI arguments overr
 | `--toolsets` | Comma-separated list of toolsets to enable |
 | `--disable-multi-cluster` | Disable multi-cluster support |
 | `--cluster-provider` | Cluster provider strategy (`kubeconfig`, `in-cluster`, `kcp`, `disabled`) |
+| `--ca-cache-dir` | Directory where CA certificates fetched via the `caURL` kubeconfig cluster extension are cached (default: subdirectory of the system temp dir) |
+| `--ca-refresh-interval` | How often cached CA certificates are re-fetched so a rotated cluster CA is picked up without a restart; `0` disables re-fetching (default: `24h`) |
 | `--tls-cert` | Path to TLS certificate file for HTTPS (must be used with `--tls-key`) |
 | `--tls-key` | Path to TLS private key file for HTTPS (must be used with `--tls-cert`) |
 | `--require-tls` | Enforce TLS for server and all outbound connections |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -238,7 +238,6 @@ rate_limit_burst = 10        # allow bursts of up to 10 requests
 |-------|------|---------|-------------|
 | `kubeconfig` | string | `""` | Path to the Kubernetes configuration file. If not provided, the server uses the in-cluster configuration or the default kubeconfig location (`~/.kube/config`). |
 | `cluster_provider_strategy` | string | auto-detect | How the server finds clusters. Valid values: `kubeconfig`, `in-cluster`, `kcp`, `disabled`. |
-| `ca_cache_dir` | string | `""` | Directory where CA certificates fetched via the `caURL` kubeconfig cluster extension are cached. Empty means the default: a subdirectory of the system temp dir. The directory must be writable; with a read-only root filesystem, mount an emptyDir there. |
 | `ca_refresh_interval` | duration | `"24h"` | How often cached CA certificates are re-fetched so a rotated cluster CA is picked up without a restart. Use `0s` to disable automatic re-fetching; the CA is then only refreshed when a manager is rebuilt (e.g. kubeconfig change or restart) or via an on-demand SIGHUP re-fetch. |
 
 **Example:**
@@ -341,7 +340,7 @@ Notes:
 - The response must be a PEM CA certificate; other content fails fast with a clear error.
 - client-go re-checks the cached file roughly every 5 minutes (the `ClientsAllowCARotation` feature, beta and enabled by default in client-go 1.36+), so a rotation converges once the refresher writes the new CA; setting `ca_refresh_interval` below that cadence does not speed it up. With that feature disabled (e.g. `KUBE_FEATURE_ClientsAllowCARotation=false`), the client no longer re-reads the file, so a changed CA is only picked up when a manager is rebuilt.
 - Sending SIGHUP to reload configuration also triggers an immediate CA re-fetch, so a rotated CA can be picked up without waiting out `ca_refresh_interval` or restarting.
-- The cache directory (system temp dir by default, or `ca_cache_dir`) must be writable. With a read-only root filesystem, mount an emptyDir at the cache location — e.g. `emptyDir: {}` mounted at `/tmp`.
+- The cache directory (a subdirectory of the system temp dir, which honors `$TMPDIR`) must be writable. With a read-only root filesystem, mount an emptyDir at `/tmp` or set `$TMPDIR` to writable storage.
 - The initial fetch happens at manager creation; if it fails (unreachable endpoint, non-PEM response), manager creation fails rather than silently using an empty trust store. For the default context that is startup; lazily-created contexts (other targets) surface the error on first use.
 - Applies to every cluster provider strategy backed by a kubeconfig (`kubeconfig` and `kcp`).
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -239,7 +239,7 @@ rate_limit_burst = 10        # allow bursts of up to 10 requests
 | `kubeconfig` | string | `""` | Path to the Kubernetes configuration file. If not provided, the server uses the in-cluster configuration or the default kubeconfig location (`~/.kube/config`). |
 | `cluster_provider_strategy` | string | auto-detect | How the server finds clusters. Valid values: `kubeconfig`, `in-cluster`, `kcp`, `disabled`. |
 | `ca_cache_dir` | string | `""` | Directory where CA certificates fetched via the `caURL` kubeconfig cluster extension are cached. Empty means the default: a subdirectory of the system temp dir. The directory must be writable; with a read-only root filesystem, mount an emptyDir there. |
-| `ca_refresh_interval` | duration | `"24h"` | How often cached CA certificates are re-fetched so a rotated cluster CA is picked up without a restart. Use `0s` to disable re-fetching; the CA is then only refreshed when a manager is rebuilt (e.g. kubeconfig change or restart). |
+| `ca_refresh_interval` | duration | `"24h"` | How often cached CA certificates are re-fetched so a rotated cluster CA is picked up without a restart. Use `0s` to disable automatic re-fetching; the CA is then only refreshed when a manager is rebuilt (e.g. kubeconfig change or restart) or via an on-demand SIGHUP re-fetch. |
 
 **Example:**
 ```toml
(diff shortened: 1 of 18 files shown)


docs/configuration.md
```diff
@@ -238,6 +238,8 @@ rate_limit_burst = 10        # allow bursts of up to 10 requests
 |-------|------|---------|-------------|
 | `kubeconfig` | string | `""` | Path to the Kubernetes configuration file. If not provided, the server uses the in-cluster configuration or the default kubeconfig location (`~/.kube/config`). |
 | `cluster_provider_strategy` | string | auto-detect | How the server finds clusters. Valid values: `kubeconfig`, `in-cluster`, `kcp`, `disabled`. |
+| `ca_cache_dir` | string | `""` | Directory where CA certificates fetched via the `caURL` kubeconfig cluster extension are cached. Empty means the default: a subdirectory of the system temp dir. The directory must be writable; with a read-only root filesystem, mount an emptyDir there. |
```
2uasimojo, line 241: YAGNI?

Is there any usage scenario under which this *must* be set? Or even where setting it makes something easier? If not, let's consider not making a new config knob for it.

docs/configuration.md
```diff
@@ -238,6 +238,8 @@ rate_limit_burst = 10        # allow bursts of up to 10 requests
...
+| `ca_refresh_interval` | duration | `"24h"` | How often cached CA certificates are re-fetched so a rotated cluster CA is picked up without a restart. Use `0s` to disable re-fetching; the CA is then only refreshed when a manager is rebuilt (e.g. kubeconfig change or restart). |
```
2uasimojo: It looks like SIGHUP will not refresh. It should (and it sounds like this wouldn't be terribly hard to implement). This gives me a way to shortcut unbreaking my server in less than whatever remains of `ca_refresh_interval` without a full restart.

If you decide against it, then these config options need to be included in the documented list of "not reloadable".

pkg/kubernetes/ca.go
```diff
@@ -0,0 +1,359 @@
...
+	for attempt := 1; ; attempt++ {
+		data, retryable, err := fetchCAOnce(ctx, client, caURL)
+		if err == nil {
+			return data, nil
+		}
+		if !retryable || attempt >= caFetchAttempts {
+			return nil, err
+		}
+		select {
+		case <-time.After(time.Duration(attempt) * time.Second):
```
2uasimojo: This does not fall through. Is that the right behavior?

pkg/kubernetes/ca.go
```diff
@@ -0,0 +1,359 @@
...
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
+	}
+}
+
+func fetchCAOnce(ctx context.Context, client *http.Client, caURL string) (data []byte, retryable bool, err error) {
```
2uasimojo, line 244: This all seems really low-level, like a helper function should already exist that we can reuse?

2uasimojo (commented):
First pass, human. (Forthcoming: agent review.)

pkg/kcp/provider.go
```diff
@@ -114,6 +128,10 @@ func (p *kcpClusterProvider) reset(ctx context.Context) error {
 		}
 	}
 
+	// Close managers from the previous reset so their background CA
+	// refreshers and HTTP transports are released before being replaced.
+	closeManagers(p.managers)
```
2uasimojo, line 143: Cursor Grok 4.6 (local review of this branch):

`NewKubeconfigManager` has already started the CA refresher by this point. `closeManagers` only runs on the success path after discovery.

If `discoverWorkspaces` and `workspacesFromKubeconfig` both fail just above, or `Derived()` fails just below after `baseManager` is stored, `reset` returns without `Close()`. Provider construction then fails with a live goroutine and HTTP transport left behind.

`defer baseManager.Close()` until the new manager is installed in `p.managers` (and close it on those error returns).

pkg/kcp/provider.go
```diff
@@ -330,10 +350,26 @@ func (p *kcpClusterProvider) WatchTargets(ctx context.Context, reload kubernetes
...
+		if m != nil {
+			m.Close()
+		}
+	}
+}
+
+func (p *kcpClusterProvider) Close() {
+	p.closeWatchers()
+	closeManagers(p.managers)
+}
```
2uasimojo, line 387: Cursor Grok 4.6 (local review of this branch):

This correctly closes kcp managers (and their refreshers) on `Close()`. `kubeconfigClusterProvider.Close()` was not given the same treatment — it still only stops watchers.

`Server.Close()` → `provider.Close()` therefore leaves per-context refresh loops running for the kubeconfig strategy. `reset()` closes old managers, so reload is fine; shutdown and tests that only call `Close()` are not.

Please close kubeconfig managers in `Close()` the same way (that file is unchanged in this PR, but this is the hole this change creates).

pkg/kubernetes/ca.go
```diff
@@ -0,0 +1,359 @@
...
+	// caMaxBodyBytes bounds CA response size; bundles are a few KB.
+	caMaxBodyBytes = 1 << 20 // 1 MiB
+)
+
+// CAFetchClientFactory builds the HTTP client used to fetch CA certificates
+// from caURL endpoints. It is a variable so tests can pin a TLS test
+// server's certificate without touching the process-wide system pool.
+var CAFetchClientFactory = func() *http.Client {
+	return &http.Client{Timeout: caFetchTimeout}
+}
```
2uasimojo, line 103: Cursor Grok 4.6 (local review of this branch):

Default `http.Client` verifies `caURL` against the **system** trust store, not the cluster CA. That is the right chicken-and-egg design, but it is not documented.

If the CA endpoint uses a cert signed by the cluster CA (or any private CA not already in the process trust store), manager creation fails. Tests have to inject `CAFetchClientFactory` for exactly this reason.

Please spell that out in `docs/configuration.md` next to the https-only note.

pkg/kubernetes/ca.go
```diff
@@ -0,0 +1,359 @@
...
+		}
+	}
+}
+
+func fetchCAOnce(ctx context.Context, client *http.Client, caURL string) (data []byte, retryable bool, err error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, caURL, nil)
+	if err != nil {
+		return nil, false, err
+	}
+	resp, err := client.Do(req)
```
2uasimojo, line 249: Cursor Grok 4.6 (local review of this branch):

`validateCAURL` requires `https` on the kubeconfig value, but this client follows redirects, including https→http.

A network attacker cannot inject that on a real TLS session, but a compromised or misconfigured CA host can downgrade, and a hostile `caURL` is an SSRF primitive.

Set `CheckRedirect` to reject non-https (and ideally to cap redirects), or re-validate the scheme on the final URL.

pkg/kubernetes/ca.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+	if err := validateCAURL(caURL); err != nil {
+		return nil, err
+	}
+
+	cacheFile := caCacheFilePath(caCacheDir(config.GetCACacheDir()), caURL)
+	client := CAFetchClientFactory()
+	fetchCtx, cancel := context.WithTimeout(ctx, caFetchTimeout)
+	defer cancel()
+	data, err := fetchCA(fetchCtx, client, caURL)
```
2uasimojo, line 197: Cursor Grok 4.6 (local review of this branch):

This 15s context wraps **all** of `fetchCA`, and `CAFetchClientFactory` also sets `Timeout: caFetchTimeout` (15s). A hung first attempt consumes the whole budget, so the documented DNS-race retries (`caFetchAttempts = 3`) only help when the failure is fast (5xx/429).

Give each attempt its own timeout, or make the outer deadline `caFetchAttempts * caFetchTimeout` plus backoff.

pkg/kubernetes/ca.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+	if err := writeCAFile(cacheFile, data); err != nil {
+		return nil, fmt.Errorf("cache CA for context %q: %w", contextName, err)
+	}
+	logger.V(2).Info("Using CA fetched from kubeconfig extension", "context", contextName, "ca_url", caURL, "ca_file", cacheFile)
+
+	// The fetched CA wins over any certificate-authority-data so the cached
+	// file is the single source client-go watches for rotation.
+	restConfig.CAFile = cacheFile
+	restConfig.CAData = nil
```
2uasimojo, line 209: Cursor Grok 4.6 (local review of this branch):

If the kubeconfig also has `insecure-skip-tls-verify: true`, client-go later errors with “specifying a root certificates file with the insecure flag is not allowed.”

Fail here with a kubeconfig-specific message, or clear `Insecure` when `caURL` wins.

pkg/kubernetes/ca_test.go
```diff
@@ -0,0 +1,349 @@
...
+		s.Equal(currentCA, got, "cached CA should match the initially served CA")
+
+		// Rotate the served CA and wait for the refresher to pick it up.
+		mu.Lock()
+		currentCA = rotatedCA
+		mu.Unlock()
+		require.Eventually(s.T(), func() bool {
+			got, err := os.ReadFile(caFile)
+			return err == nil && string(got) == string(rotatedCA)
+		}, 5*time.Second, 50*time.Millisecond)
```
2uasimojo, line 329: Cursor Grok 4.6 (local review of this branch):

This asserts the cache **file** follows rotation, not that a TLS handshake to the API server trusts the new CA.

client-go 0.36 defaults `ClientsAllowCARotation=true` and re-reads `CAFile` every 5 minutes (`atomicTransportHolder`). The “no restart” claim is plausible, but nothing here proves it.

Worth either a handshake-level test or a docs note that this depends on that feature gate and the ~5 minute poll (`KUBE_FEATURE_ClientsAllowCARotation`).

pkg/config/http_config.go
```diff
@@ -10,7 +10,7 @@ const DefaultRateLimitBurst = 10
 type HTTPConfig struct {
 	// ReadHeaderTimeout is the amount of time allowed to read request headers.
 	// This is the primary defense against Slowloris attacks.
-	ReadHeaderTimeout Duration `toml:"read_header_timeout,omitempty"`
+	ReadHeaderTimeout Duration `toml:"read_header_timeout,omitzero"`
```
2uasimojo, line 13: Cursor Grok 4.6 (local review of this branch):

Drive-by: same `Duration` / `mergeConfig` `omitzero` footgun as `ca_refresh_interval`, but this is not part of the CA-URL feature.

Fine if intentional — please call it out in the commit/PR body so it does not look accidental.

docs/configuration.md
```diff
@@ -314,6 +316,33 @@ If the server starts but operations fail (e.g., `failed to list namespaces: unkn
...
+    - name: kubernetes-mcp-server
+      extension:
+        caURL: https://ca.production.example.com/
+```
+
+At manager creation the server fetches the CA from `caURL`, caches it to a file and points the Kubernetes client at that file. Because the client reloads CA files it is pointed at, a rotated CA (e.g. after the cluster is re-provisioned) is picked up without a restart; `ca_refresh_interval` controls how often the cache is re-fetched (default `24h`, `0s` disables). Other tools ignore the extension entirely.
+
+Notes:
+
+- The URL must be `https`. A trust root fetched over cleartext could be swapped by a network attacker, who could then impersonate the cluster the CA is meant to authenticate.
```
2uasimojo: Cursor Grok 4.6 (local review of this branch):

Please add here: the HTTPS server that hosts `caURL` must present a certificate the MCP process already trusts (system roots / `SSL_CERT_FILE`), **not** the cluster CA being fetched. That is a different trust domain from “the URL must be https.”

2uasimojo (changes requested):
Agent review from Cursor Grok 4.6 on the local `feat/ca-via-url` branch (posted by me). Not a rubber-stamp of a human pass.

The design is sound: opt-in kubeconfig extension, HTTPS-only fetch, atomic cache file, `CAFile` + empty `CAData`, fail-closed startup / fail-open refresh. Two lifecycle bugs should be fixed before merge; the rest are docs, client policy, and test gaps.

Inline comments are all from that same agent review. I did not include later discussion (cache-dir knob, SIGHUP refetch, helper replacements).

docs/configuration.md
george-angel: Kept — read-only root filesystems are the scenario that requires it: the system temp dir is often not writable in hardened images, so operators must point the cache at an emptyDir/PV. That rationale is in the config reference row and the PR description.

docs/configuration.md
george-angel: Implemented: SIGHUP config reload now re-fetches cached CAs immediately (Server.ReloadConfiguration calls a new CARefresherProvider.RefreshCAs on the kubeconfig and kcp providers). This works even with ca_refresh_interval = 0s, where the background loop is disabled but the refresher is kept for on-demand fetches. Documented in the configuration.md notes and the ca_refresh_interval row.

pkg/kubernetes/ca.go
george-angel: Partially reused: CA validation now uses client-go's certutil.ParseCertsPEM instead of the manual pem.Decode + AppendCertsFromPEM dance. writeCAFile stays hand-rolled — atomic temp+rename has no stdlib helper, and atomicity is load-bearing since client-go re-reads the cache file for rotation.

pkg/kubernetes/ca.go
george-angel: Documented in configuration.md (CA served over HTTPS notes) and the PR description: the caURL endpoint must present a certificate the MCP process already trusts (system roots / SSL_CERT_FILE), not the cluster CA being fetched.

pkg/kubernetes/ca_test.go
george-angel: Chose the docs-note option: configuration.md now states convergence is bounded by client-go's CA-file re-read (ClientsAllowCARotation, ~5 min poll, beta on by default) and what happens if that feature is disabled. A handshake-level test through the real rotation is impractical here because the poll interval is not configurable, and upstream client-go covers transport rotation in its own tests.

pkg/config/http_config.go
george-angel: Intentional, and called out in the PR description: zero means "not set" (the merge keeps the default), while an explicit "0s" in a config file or via --ca-refresh-interval is honored. The one exception — build-time default overrides — is documented in the CARefreshInterval comment in config.go.

docs/configuration.md
george-angel: Covered by the same configuration.md note: the HTTPS server hosting caURL must present a certificate the process already trusts — a different trust domain from "the URL must be https".

george-angel:
Hello @2uasimojo - thank you so much for taking the time to look at and review this PR.

I think I have addressed all your comments, please let me know if there is anything outstanding.

Thank you!

docs/configuration.md
2uasimojo: In the RORFS scenario can't I mount my `emptyDir` over `/tmp`, or mount somewhere else and set `$TMPDIR` to point to it?

The reason I'm being particular about it is that we are getting very close to a generally-available release of this thing, after which walking back or semantically changing config options will be much more difficult. @Cali0707 @manusa what is your opinion on this (assuming I'm correct about the RORFS scenario)?

pkg/kubernetes/provider_kubeconfig.go
```diff
@@ -213,10 +213,38 @@ func (p *kubeConfigClusterProvider) WatchTargets(ctx context.Context, reload Mcp
...
+			m.Close()
+		}
+	}
+}
+
+// RefreshCAs re-fetches every live manager's cluster CA now, so a rotated
+// CA is picked up on SIGHUP without waiting out ca_refresh_interval. The
+// snapshot keeps the fetch (bounded by caFetchMaxDuration) outside the
+// lock so a reset triggered meanwhile is not blocked on the network.
+func (p *kubeConfigClusterProvider) RefreshCAs() {
```
2uasimojo, line 238: Duplicates logic (kcp and kubeconfig providers are the same afaics), sad-face. Me wishing golang had a less awkward way to implement a method in an abstract base class. No change requested, just an observation.

2uasimojo (commented):
Thanks for the fixups @george-angel.

One small point I'd like to defer to Calum/Marc. Otherwise LGTM!

docs/configuration.md
Cali0707: +1 on just mounting an empty dir where needed (i.e. to /tmp) over having another config option that we don't 100% need

george-angel: Good point - $TMPDIR and an emptyDir at /tmp cover the read-only-root scenario without a config knob, and keeping the config surface small is the right call. I have removed ca_cache_dir (and the --ca-cache-dir flag); the cache always lives in a subdirectory of the system temp dir. configuration.md and the PR description now say that, and note the $TMPDIR/emptyDir options for read-only filesystems.

george-angel:
`ca_cache_dir` flag removed :+1:

2uasimojo (commented):
/lgtm

Thanks!

pkg/kubernetes/ca.go
```diff
@@ -0,0 +1,420 @@
...
+	if err != nil || len(served) == 0 {
+		return nil, false, fmt.Errorf("CA URL %s response is not a PEM CA certificate: %w", caURL, err)
+	}
+	return body, false, nil
+}
+
+// writeCAFile atomically writes data to path so readers (client-go's CA-file
+// rotation) never observe a partially written certificate.
+func writeCAFile(path string, data []byte) error {
+	if err := os.MkdirAll(filepath.Dir(path), 0o700); err != nil {
```
Cali0707: `MkdirAll(..., 0700)` only applies the mode when creating the directory, it does not secure an existing directory. Because the default path to this directory is predictable beneath the shared system temp directory, another local user could pre-create it as writable or as a symlink and subsequently replace this cached CA. 

Please reject symlinked and group/world-writable cache directories, or use a reliably private cache directory. Also please add tests covering a pre-existing unsafe directory and symlink

pkg/kubernetes/provider_kubeconfig.go
```diff
@@ -213,10 +213,38 @@ func (p *kubeConfigClusterProvider) WatchTargets(ctx context.Context, reload Mcp
...
 	for _, w := range []watcher.Watcher{p.kubeconfigWatcher, p.clusterStateWatcher} {
 		if !reflect.ValueOf(w).IsNil() {
 			w.Close()
 		}
 	}
+	for _, m := range p.managers {
+		if m != nil {
+			m.Close()
+		}
+	}
```
Cali0707, line 231: you will need something equivalent in `provider_single.go` that will work in the case that constructs a provider from a kubeconfig

This will need a bit of careful looking at the `reset` function in that file as well, as that currently uses `p.Close()` to close all the watchers, and just directly adding the `manager.Close` call in here while leaving that unchanged will result in the manager being closed right upon each update. So we probably need to introduce something like a `closeWatchers` method on the `singleClusterProvider`

pkg/kubernetes/provider.go
```diff
@@ -34,6 +34,14 @@ type Provider interface {
 	Close()
 }
 
+// CARefresherProvider is implemented by providers that hold managers with
+// background CA refreshers. SIGHUP config reload calls RefreshCAs so a
+// rotated cluster CA (e.g. after the cluster is re-provisioned) can be
+// picked up without waiting out the refresh interval or restarting.
+type CARefresherProvider interface {
+	RefreshCAs()
+}
```
Cali0707: Do we really want this to be an optional provider capability? I would worry about providers that may own a manager created by `NewKubeconfigManager` missing implementing this method, and silently missing the SIGHUP refresh.

Note that this is already occurring, as both `provider_single.go` and `provider_tokenexchange.go` will not handle SIGHUP correctly.

pkg/config/http_config.go
Cali0707: Can we pull this out into a separate PR that is explicitly marked as a breaking change? This changes config serialization, and increases the regression surface for an already security-sensitive and wide reaching change

Cali0707 (commented):
Thanks for working on this @george-angel !

Left a few comments

pkg/kubernetes/ca.go
george-angel: Implemented. The cache directory is now checked on every write and rejected if it is a symlink, group/world-writable, or owned by another user.

I kept world-readable directories working: the CA files themselves are 0600, so reading cannot swap a trust anchor, only writing can. Tests cover a pre-created unsafe directory, a symlinked directory, and that renaming over a pre-created file symlink replaces it rather than writing through it.

pkg/config/http_config.go
george-angel: Taken. I dropped the read_header_timeout encoding change from this PR entirely, so existing config serialization is untouched. ca_refresh_interval is a new key and keeps the omitzero pattern; if you would still like the omitzero cleanup done as its own explicitly breaking change, I am happy to open that separately.

PR #1407, opened 2026-09-03
feat: add required rbac into _meta for tools/prompts/resources

Cali0707:
This adds a mechanism for tools/prompts/resources to declare what k8s rbac they require to run. This is useful in some environments where a sandbox wants to create a serviceaccount/role/rolebinding dynamically with least permissions, as we tell them what permissions the tool would need to execute.

These are purely advisory and aim to provide upper bounds on what is required. The k8s api server is still the ultimate source of truth for all RBAC

### DIFF
--- a/docs/tool-rbac-metadata.md
+++ b/docs/tool-rbac-metadata.md
@@ -1,0 +1,369 @@
+# Tool RBAC Metadata
+
+Kubernetes MCP Server tools can advertise the Kubernetes RBAC permissions they may require in the `_meta` field returned by `tools/list`. Clients can use this metadata with planned tool arguments to prepare permissions before calling a tool.
+
+The metadata is a planning aid. It does not grant access, replace Kubernetes authorization, or guarantee that a call will succeed. The Kubernetes API server remains authoritative.
+
+Client support for this metadata is optional. Clients that do not dynamically prepare or narrow Kubernetes permissions can ignore it and call tools using their existing credentials. The metadata is intended for clients that want to derive scoped RBAC for a planned call, such as an agent sandbox that creates a short-lived ServiceAccount, Role, or binding before execution.
+
+## Metadata Key
+
+RBAC metadata is published under the `io.kubernetes-mcp-server/rbac` key:
+
+```json
+{
+  "_meta": {
+    "io.kubernetes-mcp-server/rbac": {
+      "version": "v1alpha1",
+      "none": {}
+    }
+  }
+}
+```
+
+The metadata key being absent means that the tool has not declared its RBAC requirements. It must not be interpreted as requiring no RBAC.
+
+## Declaration States
+
+Every declaration has a `version` and exactly one of the following properties:
+
+| Property | Meaning |
+|---|---|
+| `none` | The tool does not require Kubernetes API authorization. |
+| `bounded` | The tool's possible RBAC requirements are described by `requirements`. |
+| `unbounded` | The requirements cannot be determined from the tool declaration and call arguments. |
+
+### No RBAC
+
+Tools that do not access the Kubernetes API declare `none` explicitly:
+
+```json
+{
+  "version": "v1alpha1",
+  "none": {}
+}
+```
+
+An optional `reason` can provide additional context:
+
+```json
+{
+  "version": "v1alpha1",
+  "none": {
+    "reason": "Reads the local kubeconfig without accessing the Kubernetes API"
+  }
+}
+```
+
+### Bounded RBAC
+
+A bounded declaration contains one or more requirements:
+
+```json
+{
+  "version": "v1alpha1",
+  "bounded": {
+    "requirements": [
+      {
+        "verbs": ["get"],
+        "target": {
+          "resource": {
+            "resource": "pods"
+          }
+        }
+      }
+    ]
+  }
+}
+```
+
+`bounded` describes a finite, conservative upper bound. After resolving arguments and manifests, every Kubernetes API authorization check that the tool may perform must be covered by the declaration. The declaration may include permissions that a particular invocation does not use, for example when behavior depends on an optional argument, cluster state, API discovery, or transport negotiation.
+
+Clients combine all resolved requirements when preparing RBAC. Declaring too few permissions is a metadata error and can cause a tool call to fail with an authorization error. Declaring a slightly broader finite set is permitted and avoids requiring a conditional-expression language in the metadata contract.
+
+Bounded declarations should identify concrete resources after argument or manifest resolution. Broad wildcard permissions such as all API groups, all resources, or all subresources should not be used as a substitute for `unbounded`.
+
+### Unbounded RBAC
+
+Tools whose permissions cannot be determined from their arguments declare `unbounded`:
+
+```json
+{
+  "version": "v1alpha1",
+  "unbounded": {
+    "reason": "Permissions depend on resources contained in the referenced Helm chart"
+  }
+}
+```
+
+Clients should apply their own policy to unbounded tools, such as requesting separate approval or rejecting the planned call.
+
+## Requirements
+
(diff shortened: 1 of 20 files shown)


Cali0707:
I plan to add follow up PRs for other toolsets, this PR will just focus on the interface, the structure of the published metadata, and the core toolset as a first consumer of this. I'm worried about it ballooning even further if I start adding the rest of the toolsets here

pkg/api/rbac.go
2uasimojo, line 1: Why aren't we using [upstream types](https://github.com/kubernetes/api/blob/master/rbac/v1/types.go)?

Cali0707: This is because many of our rbac types are not 1:1 with the upstream k8s ones (e.g. where we want to derive the rbac from a manifest, or from tool parameters)

It seemed cleaner to have our own set of types here rather than have a mix of our types/upstream types

cajieh:
Once RBAC metadata is added to all toolsets, should we add a completeness test (e.g., `TestAllToolsHaveRBACMetadata`) that iterates all registered `ServerTool` instances and asserts `RBAC != nil`? That would prevent future tools from silently shipping without RBAC metadata.

Cali0707:
@cajieh that is a good idea - mind opening an issue for that ?

pkg/toolsets/core/resources.go
```diff
@@ -122,7 +137,10 @@ func initResources(p api.FilteringProvider) []api.ServerTool {
 				IdempotentHint:  ptr.To(true),
 				OpenWorldHint:   ptr.To(true),
 			},
-		}, Handler: resourcesCreateOrUpdate},
+		}, RBAC: api.RBACBounded(api.RBACRequirement{
+			Verbs:  []string{"patch"},
```
cajieh: Since `resources_create_or_update`/`pods_run` use Server-Side Apply and the target resource may not exist yet, does `verbs` need `create` in addition to `patch` here? Per the [K8s SSA docs](https://kubernetes.io/docs/reference/using-api/server-side-apply/#access-control-and-permissions), `patch` alone only covers editing an existing resource — creating a new one via SSA also needs `create`.

cajieh: Same SSA patch-vs-create question as on `pods_run` in pods.go — does this one need create too?

pkg/toolsets/core/pods.go
```diff
@@ -259,7 +333,26 @@ func initPods(p api.FilteringProvider) []api.ServerTool {
...
+				Target:    api.RBACTarget{Resource: &api.RBACResourceTarget{Resource: "pods"}},
+				Namespace: &api.RBACNamespace{Argument: "namespace"},
+			},
+			api.RBACRequirement{
+				Verbs:     []string{"patch"},
+				Target:    api.RBACTarget{Resource: &api.RBACResourceTarget{Resource: "services"}},
+				Namespace: &api.RBACNamespace{Argument: "namespace"},
+			},
+			api.RBACRequirement{
+				Verbs: []string{"patch"},
```
cajieh: Since `resources_create_`or`_update/pods_run` use Server-Side Apply and the target resource may not exist yet, does verbs need create in addition to patch here?

cajieh (commented):
@Cali0707 Looks good apart from two open comments.
Does that scope include core's prompts (e.g. cluster-health-check, which reads Nodes/Pods/Deployments/StatefulSets/PVCs/ClusterOperators/Events)? If prompts are being deferred, should `cluster-health-check `get `RBACUnbounded` in the meantime given how broad its access is, or is that also fine to leave for later?

cajieh:
> @cajieh that is a good idea - mind opening an issue for that ?
Issue: [Add a test to catch tools that forget to declare RBAC metadata](https://github.com/containers/kubernetes-mcp-server/issues/1435)

Cali0707:
Nice catch on the RBAC verbs @cajieh - updated

cajieh:
LGTM 
~~Two CI checks are failing, but it appears unrelated to this PR changes.~~

xrajesh:
1. I will update OLS spec for meta key name to match  `io.kubernetes-mcp-server/rbac` and schema for bounded.requirements
2. This PR has only added for core tool set. OLS supports "core", "config", "helm", "observability/metrics", "kubevirt" @Cali0707

pkg/mcp/tools_gosdk.go
```diff
@@ -31,11 +32,28 @@ func ServerToolToGoSdkTool(s *Server, tool api.ServerTool) (*mcp.Tool, mcp.ToolH
...
+		if err := tool.RBAC.Validate(); err != nil {
+			return nil, nil, fmt.Errorf("tool %q: invalid RBAC metadata: %w", tool.Tool.Name, err)
+		}
+		meta = maps.Clone(meta)
+		if meta == nil {
+			meta = make(map[string]any)
+		}
+		meta[api.RBACMetadataKey] = tool.RBAC
+	}
+
```
xrajesh, line 51: Can we add a check for  tools with DestructiveHint  = true - not set RBAC None ? @Cali0707

docs/tool-rbac-metadata.md
```diff
@@ -0,0 +1,369 @@
...
+  "verbs": ["patch"],
+  "target": {
+    "manifest": {
+      "argument": "resource"
+    }
+  }
+}
+```
+
+The client is responsible for parsing the manifest and determining its resource identities, names, and namespaces. This includes handling formats such as multiple YAML documents or Kubernetes List objects when supported by the client.
```
xrajesh, line 205: I’m unclear on the client expectations for `target.manifest`. @Cali0707 could you please share the manifest formats clients are expected to handle when deriving RBAC?

pkg/toolsets/core/health_check.go
```diff
@@ -39,6 +39,53 @@ func initHealthChecks() []api.ServerPrompt {
 					},
 				},
 			},
+			RBAC: api.RBACBounded(
+				api.RBACRequirement{
+					Verbs:  []string{"list"},
+					Target: api.RBACTarget{Resource: &api.RBACResourceTarget{Resource: "nodes"}},
+				},
+				api.RBACRequirement{
+					Verbs:  []string{"list"},
```
blublinsky: The `cluster-health-check` handler performs `params.CoreV1().Namespaces().Get(...)` whenever the optional `namespace` argument is provided (see `clusterHealthCheckHandler`), but this metadata only declares `list` for `namespaces`. A client deriving least-privilege RBAC from this declaration can still receive a forbidden error for namespace-scoped health checks, which contradicts the documented requirement that bounded metadata cover every Kubernetes authorization check the capability may perform. Please add a `get` requirement for the `namespaces` resource and account for the optional argument when choosing the resource-name scope.

blublinsky:
Nice-to-have follow-up: add a completeness test such as `TestAllToolsHaveRBACMetadata` once RBAC metadata is declared for all toolsets. The test could iterate over registered `ServerTool` instances and fail when a tool has no `ServerTool.RBAC`, preventing future tools from silently shipping without metadata. This is already tracked in #1435 and is not blocking for the current core-toolset scope.

pkg/toolsets/core/health_check.go
cajieh: @blublinsky Good catch. The namespaces RBAC needs a `get` requirement as well since cluster-health-check calls `Namespaces().Get() ` for the namespace-scoped case.

pkg/mcp/tools_gosdk.go
Cali0707: Tbh I'm not totally sure that would be universally true, for anything that directly interacts with the k8s api it would, but some toolsets do not call the api server for their calls, and may still be destructive.

So for now, will hold off on doing this.

Issue #1288, opened 2026-07-13
Make target filter checks parallel


Cali0707:
As an improvement to the target filtering, we want to thread the requests to the targets in parallel to minimize response time when doing api.FilteringProvider calls.

_Originally posted by @2uasimojo in https://github.com/containers/kubernetes-mcp-server/issues/1285#issuecomment-4959944392_

2uasimojo:
Note that #1413 does *not* parallelize the [nexus filter loop](https://github.com/containers/kubernetes-mcp-server/blob/cd5735d8effb43c94faed3bc5cd4fbb7a71d12ba/pkg/mcp/mcp.go#L82-L86). This can still be done if desired, but I think it will entail plumbing a `context` through `FilteringProvider` so we can cancel those calls when we want to short-circuit.

PR #1413, opened 2026-09-04
Parallelize AnyTargetHasGVKs()

2uasimojo:
The expensive part of target compatibility tool filtering is iterating over all targets querying whether any target has particular GVKs. In serial, one or more slow/unreachable targets could delay the whole operation, potentially until their discovery call times out, greatly slowing the entire startup process.

With this change, we run each target's discovery in its own goroutine, and pop out of the whole thing as soon as any one of them returns `true`. Thus cases where any target returns `true` should respond very quickly, regardless of whether slow/unreachable targets exist elsewhere.

Now the worst case looks like:
- All live targets respond `false`
- One or more targets is slow/unreachable

...whereupon the overall response time is reduced to that of the slowest target.

Note that we preserve the "fail open" behavior: if any target is unreachable, we assume `true` so that we do *not* hide tools. With this change, that behavior extends to when the parent context is canceled for any reason before all goroutines have responded.

Partially-Addresses: #1288
Co-Authored-By: cursor/grok

### DIFF
--- a/pkg/kubernetes/provider_gvk_filter.go
+++ b/pkg/kubernetes/provider_gvk_filter.go
@@ -7,6 +7,7 @@ import (
 	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/klog/v2"
 )
 
 // ProviderGVKFilter provides GVK-based filtering capabilities for providers.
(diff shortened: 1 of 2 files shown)


pkg/kubernetes/provider_gvk_filter.go
```diff
@@ -38,25 +39,59 @@ func (f *ProviderGVKFilter) AnyTargetHasGVKs(ctx context.Context, gvks []schema.
...
+		go func() {
+			results <- targetHasGVKs(ctx, logger, mgr, gvks)
+		}()
+	}
+
+	for range mgrs {
+		if <-results {
+			// One target with all GVKs (or a discovery error) is enough.
+			// Workers that have not reached HasGVKs yet will stop via the deferred cancel() above.
+			// client-go's DiscoveryClient uses context.TODO() (at the time of this writing), so
```
Cali0707, line 57: once we bump to the 0.37.0 client this should flip, so we can always revisit as needed (that bump should land here soon)

Cali0707 (approved):
LGTM

pkg/kubernetes/provider_gvk_filter.go
2uasimojo: Oh, excellent. I didn't look into it, but I was hoping that would come down eventually (hence the hedging in the comment :P )

PR #1419, opened 2026-09-08
test(evals): read only evals

cajieh:
Read-only mode (`--read-only`) is primary safety boundary for write operations, but coverage stopped at unit tests checking that write tools are hidden from the tool list. There was no live evaluation run proving an AI agent operating on a live cluster with `read_only=true` can still get real diagnostic work done using only read tools.

## What this adds

- **`core-readonly` eval suite** (`evals/core-eval-testing/builtin-openai/eval-core-readonly.yaml`): runs the subset of `core`/`config` tasks that only need read tools, selected via a new `readonly: "true"` label. Asserts `toolsUsed` (at least one read tool called) and, as a suite-wide safety net, `toolsNotUsed` on the write-tool pattern (`resources_create_or_update|resources_delete|resources_scale|pods_delete|pods_exec|pods_run`) so the whole suite fails if any task ever triggers a destructive call.
- **`readonly: "true"` labels** on the read-only-compatible tasks that best showcase an agent doing real diagnostics with only read tools: `debug-app-logs`, `filter-events-by-type`, `nodes-top`, plus `list-contexts`/`view-current-config` for viewing configuration. (`filter-namespace-by-name`, `list-images-for-pods`, and `ahoy` were considered but dropped -- they either overlap with the tasks above or, for `ahoy`, don't assert anything about actual tool exposure that the suite-level `toolsNotUsed` check doesn't already cover better.)
- **`READ_ONLY` knob in `build/evals.mk`**: `make run-server READ_ONLY=true` now starts the MCP server with `--read-only`, so `core-readonly` can be exercised against a server actually running in the mode it's meant to validate:
  ```
  make run-server READ_ONLY=true TOOLSETS=core,config
  make run-evals SUITE=core-readonly AGENT=builtin-openai
  ```
- **`TestReadOnlyBlocksWriteToolInvocation`** (`pkg/mcp/mcp_tools_test.go`): whether `read_only=true` actually rejects a write attempt -- as opposed to just hiding write tools from the tool list -- is a deterministic property of the server, not of model behavior, so it's covered by a Go test instead of an eval task. It calls a write tool directly (bypassing whatever the tool list shows) and asserts the call is rejected, plus a ground-truth cluster read confirming the target resource was never mutated. This is cheaper and non-flaky compared to routing the same check through a live LLM.
- Docs updates in `evals/README.md`, `evals/core-eval-testing/README.md`, and `evals/tasks/README.md` describing the new suite, how tasks get reused across suites via labels, and why write-blocking is verified at the Go-test level rather than as an eval.

Co-authored with Cursor and Claude.


cajieh:
The core-readonly isn't wired into CI yet. See issue [#1436](https://github.com/containers/kubernetes-mcp-server/issues/1436) for adding it to the weekly schedule-matrix.

build/evals.mk
```diff
@@ -98,9 +103,10 @@ diff-evals: mcpchecker ## Diff latest mcpchecker results against baseline
 	$(MCPCHECKER) diff --base "$$BASELINE" --current "$$RESULTS_FILE" --output markdown
 
 .PHONY: run-server
-run-server: build ## Start MCP server in background and wait for health check
+run-server: build ## Start MCP server in background and wait for health check (knobs: TOOLSETS, READ_ONLY)
 	@echo "Starting MCP server on port $(MCP_PORT)..."
-	./$(BINARY_NAME) --port $(MCP_PORT) $(if $(TOOLSETS),--toolsets "$(TOOLSETS)") --config-dir $(MCP_CONFIG_DIR) $(if $(MCP_EVAL_KUBECONFIG),--kubeconfig "$(MCP_EVAL_KUBECONFIG)") & echo $$! > .mcp-server.pid
+	@if [ "$(READ_ONLY)" = "true" ]; then echo "Starting server with --read-only (only readOnlyHint=true tools are exposed)"; fi
+	./$(BINARY_NAME) --port $(MCP_PORT) $(if $(TOOLSETS),--toolsets "$(TOOLSETS)") --config-dir $(MCP_CONFIG_DIR) $(if $(MCP_EVAL_KUBECONFIG),--kubeconfig "$(MCP_EVAL_KUBECONFIG)") $(if $(filter true,$(READ_ONLY)),--read-only) & echo $$! > .mcp-server.pid
```
Cali0707: @cajieh maybe we can simplify this by always adding `--read-only=$(READ_ONLY)` ? Note that this will require the `=` character due to how the flags are parsed, while the other args do not.

Cali0707 (commented):
This is a great start @cajieh ! One nit otherwise this should be good to go

PR #1423, opened 2026-09-09
Config rework

### LINKED ISSUE #1374: Consolidate Config Loading and Precedence

2uasimojo:
See the design at docs/specs/config.md

Closes: #1374
Authored-by: cursor/grok


pkg/config/configtest/configtest.go
2uasimojo, line 1: I don't like this file, but I stared at it for a while and getting rid of it is actually uglier. Sad face.

pkg/kubernetes/kubernetes.go
```diff
@@ -74,15 +75,16 @@ func NewKubernetes(
 
 	k.restConfig.Wrap(func(original http.RoundTripper) http.RoundTripper {
```
Cali0707, line 76: This captures at startup denied resources, validation, and confirmation, which are all marked as releasable. 

Unless I'm missing something the provider isn't recreated on SIGHUP though - mind looking into that?

pkg/config/config.go
```diff
@@ -307,564 +672,356 @@ func loadDropInConfigs(ctx context.Context, dropInConfigDir string) ([]string, e
...
-	}
-	return c.TLSMinVersion
+// applyTOML copies present TOML keys onto matching Options.
+func applyTOML(cfg *Config, merged map[string]any, sources map[string]string) error {
+	var applyErr error
+	walkOptions(cfg, func(o option, path string) {
+		if applyErr != nil || path == "" {
+			return
+		}
+		raw, ok := mapLookup(merged, path)
```
Cali0707, line 829: I _may_ be missing a detail here, but I believe this could lead to cases where invalid configuration leads to starting with defaults instead of returning an error. In particular with how the map assertion can fail, leading to a silent return here. 

So if I have config like:
```toml
http = "invalid"
toolset_configs = "invalid"
```

I believe it would just use defaults rather than error

pkg/kubernetes-mcp-server/cmd/root_var_overrides.go
2uasimojo, line 1: pkg/config/config_default_overrides.go now handles this.

pkg/kubernetes/kubernetes.go
2uasimojo: Good eye. Fixed.

pkg/config/config.go
2uasimojo: Good eye. Fixed.

pkg/config/config.go
```diff
@@ -307,564 +672,424 @@ func loadDropInConfigs(ctx context.Context, dropInConfigDir string) ([]string, e
...
-func (c *StaticConfig) GetTokenExchangeConfig() api.TokenExchangeConfig {
-	if c.TokenExchange == nil {
-		return nil
+	if err := rejectUnknownKeys(merged, cfg, sources); err != nil {
+		return nil, err
+	}
+	if err := rejectWrongTableTypes(merged, cfg, sources); err != nil {
+		return nil, err
+	}
+	if err := parseExtensions(ctx, cfg, merged, sources); err != nil {
```
Cali0707: @2uasimojo I think we need to pin the non reloadable bits here.

The extensions read require_tls in parseExtensions, and we only restore that after this

docs/specs/config.md
```diff
@@ -0,0 +1,1644 @@
...
+</tr>
+</tbody>
+</table>
+
+### HTTP / TLS
+
+Inbound listen/TLS settings need a restart. Outbound TLS (OAuth, token
+exchange, Kiali, NetObserv) already re-reads on reload / per call; the
+non-reloadable mark below is for the **inbound** HTTP server.
+
```
Cali0707: @2uasimojo here let's simplify and just make the outbound and inbound tls be consistent with each other? Unless you feel strongly other way?

2uasimojo: IIUC it's just the text here that's wrong: the actual config values are marked non-reloadable, so changed values don't make it to the outbound calls (or anywhere).

Imma just delete this paragraph, yah?

2uasimojo:
> [Build / Build on ubuntu-24.04-arm (pull_request)](https://github.com/containers/kubernetes-mcp-server/actions/runs/35106597989/job/104829997233?pr=1423)Failing after 3m

Extra Warning event generated for reasons unrelated to the test. Rerun. (IDK how to trigger that.)

Cali0707 (commented):
/run-mcpchecker kubevirt

2uasimojo:
[evals](https://github.com/containers/kubernetes-mcp-server/actions/runs/35116898617)

pkg/kubernetes/provider_single.go
Cali0707: @2uasimojo can we capture this manager in a variable and only close it after we know for sure that the newly created manager had no errors? otherwise we can restore it?

2uasimojo: Fixed for single, kubeconfig, and kcp.

2uasimojo:
Something weird happened with review comment responses -- I think they got batched up by accident.

pkg/kubernetes-mcp-server/cmd/root.go
```diff
@@ -43,99 +39,41 @@ kubernetes-mcp-server --version
...
-	flagOAuthAudience        = "oauth-audience"
-	flagAuthorizationURL     = "authorization-url"
-	flagSkipJWTVerification  = "skip-jwt-verification"
-	flagServerUrl            = "server-url"
-	flagCertificateAuthority = "certificate-authority"
-	flagDisableMultiCluster  = "disable-multi-cluster"
-	flagClusterProvider      = "cluster-provider"
-	flagTLSCert              = "tls-cert"
-	flagTLSKey               = "tls-key"
-	flagRequireTLS           = "require-tls"
```
manusa, line 90: For those who run the MCP server locally via the binary or any of the provided wrappers, I would have left a minimum set of this that might be useful to debug or find out problems.
Namely `log-level` and `log-file`.

In any case, simplification beats it especially for maintenance of the project.
So maybe we can add it back in case the users complain about it.

To be clear, it's fine as is ✅

docs/specs/config.md
```diff
@@ -0,0 +1,1657 @@
...
+### Bootstrap
+
+Not `Config` fields. Path selection is its own ladder: `--config` beats
+`$K8S_MCP_CONFIG_PATH` (already true). `--config-dir` has no env equivalent
+and no default. It can be the only file source. Omit it and no directory
+is read. Relative `--config` and `--config-dir` are resolved against the
+working directory, independently of each other.
+
+<table>
```
manusa, line 360: If possible, it'd be better if we could collapse this html-formatted table into a markdown table, otherwise it's very hard to read for review and for future maintenance

manusa: Same applies for any other table in this file or others

docs/configuration-changes.md
manusa, line 1: I believe we never did a release with the revised configuration.
Maybe it's better to revise the entire file to a single configuration migration.
Ideally, once we have all changes for downstream release ready, we could do a 0.1.0 upstream release.
So we could frame this document as 0.0.x to 0.1.x migration instead

pkg/kubernetes/watcher/cluster.go
```diff
@@ -43,24 +41,15 @@ type ClusterState struct {
...
+		pollInterval = DefaultClusterStatePollInterval
 	}
-	if envDebounce := os.Getenv("CLUSTER_STATE_DEBOUNCE_WINDOW_MS"); envDebounce != "" {
-		if ms, err := strconv.Atoi(envDebounce); err == nil && ms > 0 {
-			debounceWindow = time.Duration(ms) * time.Millisecond
-			logger.V(2).Info("Using custom cluster state debounce window", "debounce_window", debounceWindow)
-		}
+	if debounceWindow <= 0 {
+		debounceWindow = DefaultClusterStateDebounceWindow
 	}
```
manusa: Not that these variables were extremely important for tests and to be able to run the suite quickly (not public facing).

2uasimojo:
> [Build / Build on macos-15-intel (pull_request)](https://github.com/containers/kubernetes-mcp-server/actions/runs/35143963959/job/104955794045?pr=1423)Failing after 10m

Looks like a flake: we tried querying the fake server before it was ready.

Cali0707 (approved):
LGTM, but let's also wait for whatever @manusa finds before merging

matzew (approved):
## Review: PR #1423 - Config rework

**Verdict: Approve.** :+1:

This consolidates flags + env + TOML into a single canonical model (TOML as
source of truth, env overlay) and removes the operational CLI flags. The
direction is right - collapsing the flag surface into one config model removes a
whole class of precedence ambiguity, and the codebase is simpler for it.

I reviewed the full branch, including the reload sequencing fixups
(`031f36a3`, `43f5961f`, `49bc72ec`).

**What's solid:**
- Single, unambiguous config precedence - no more silent flag/env/file games.
- Config reload is handled carefully: SIGHUP is serialized, and `ReloadConfig`
  now snapshots and restores the prior config on failure
  (`TestReloadConfigKeepsManagersOnError` covers it). Old managers keep serving
  after a failed reload.
- Deliberate strictness (unknown-key rejection, stdio + `require_oauth` guard) -
  I read these as intentional "fail loud" choices, and I agree with them.

**Non-blocking, optional:**
- On a removed flag (e.g. `--port`), users get cobra's bare `unknown flag`. A
  one-line hint pointing at `docs/configuration.md` would smooth the migration.
  Purely UX, not a compatibility commitment. Take it or leave it.

**One thing to be aware of (not asking for a change):**
- If a reload's provider rebuild succeeds but `applyToolsetsLocked` then fails,
  the provider briefly holds the new access-control snapshot while
  `s.configuration` still reports the old. The reload returns an error and the
  next reload/restart reconciles, so it's narrow and self-healing - just noting
  it for the record.

Nice cleanup.

manusa (commented):
Thanks for the rework! The `Option[T]` model is a real improvement:
- a single reflective walk drives TOML, env, validation, reload pinning and docs, so new fields can't be forgotten;
- unknown keys and wrong table types now fail instead of being silently ignored;
- every non-reloadable option is pinned on SIGHUP (main only pinned `metrics_port`);
- `denied_resources`, validation and confirmation changes finally reach the Kubernetes client on reload.

Below is what I found. References are `file:line` at `49bc72ec` plus a code anchor, since lines will drift.

### Must fix

**Keep `pkg/kubernetes` and the toolset clients decoupled from `pkg/config`**
- **Background:** #566 moved config access behind interfaces in `pkg/api` (`api.BaseConfig` and friends). That way `pkg/kubernetes`, the providers and the toolset clients depend on abstractions, not the concrete config type. On main, no non-test file in `pkg/kubernetes` imports `pkg/config`.
- **What changed:** this PR deletes those interfaces from `pkg/api/config.go`. `NewProvider`, `NewManager`, `NewKubernetes`, `ProviderFactory`, token exchange, `pkg/kcp`, Kiali and NetObserv now take `*config.Config` directly. 8 non-test files in `pkg/kubernetes`, plus `pkg/kcp/provider.go`, now import `pkg/config`.
- **Symptom:** Kiali and NetObserv can no longer get TLS settings through `pkg/api`. `configFromProvider` (`pkg/kiali/kiali.go:221-233`, `pkg/netobserv/netobserv.go:247-259`) type-switches to `*config.Config`. For anything else it falls back to `config.New()`, which means `require_tls=false`. That fallback is a latent fail-open.
- **Fix:** the Option model doesn't require the coupling.
  - `*config.Config` can implement narrow `pkg/api` interfaces with one-line getters over `Option.Get()`, e.g. `func (c *Config) GetDeniedResources() []api.GroupVersionKind { return c.DeniedResources.Get() }`.
  - Watcher timings and QPS can be passed as constructor arguments.
  - A `func() api.BaseConfig` provider backed by the committed config then gives the access-control round tripper live values, without rebuilding managers on every SIGHUP (see the SIGHUP item below).

**Helm chart with `tls.enabled=true` won't start**
- **Where:** `charts/kubernetes-mcp-server/templates/deployment.yaml:56-57` still passes `"--tls-cert=…"` / `"--tls-key=…"`. The file isn't in this diff, so chart CI didn't run, and no values file or e2e test enables TLS.
- **Fix:** move these into the chart's TOML (`tls_cert` / `tls_key`).
- **Docs:** the migration notes should warn that chart `extraArgs` and container `args` using removed flags now fail with `unknown flag`.

**The startup option dump is never printed**
- **Where:** `dumpConfig` logs at `V(1)` and is called inside `resolve` (`pkg/config/config.go:810`, `dumpConfig(ctx, cfg, settings.previous)`). That runs from `Complete` via `config.Read`/`ReadToml` (`pkg/kubernetes-mcp-server/cmd/root.go:122,130`), before `logging.New` sets klog's verbosity (`root.go:150`).
- **At startup:** the dump is dropped even with `log_level = 5`, although `docs/configuration.md:58` says it's logged.
- **On SIGHUP:** it runs before `Validate`, so a rejected reload logs `changed=true` / `previous` for values that never took effect.
- **Fix:** dump after `logging.New`, and on SIGHUP only after the reload succeeds.

**The Docker image ignores `K8S_MCP_CONFIG_PATH`**
- **Where:** `Dockerfile:22` is now `CMD ["--config", "/etc/kubernetes-mcp-server/config.toml"]`, and `root.go:116` (`m.ConfigPath == "" && cp != ""`) only reads the env var when `--config` is empty.
- **Scenario:** `docker run -e K8S_MCP_CONFIG_PATH=/cfg/config.toml …` worked on main and is documented. Now the user's file is silently ignored, and the server starts on `0.0.0.0:8080` without `require_oauth`, `read_only` or `denied_resources`.
- **Fix:** `ENV K8S_MCP_CONFIG_PATH=/etc/kubernetes-mcp-server/config.toml` with no `CMD` keeps both the env var and an explicit `--config` working.

**Dropping the implicit `conf.d` is a silent regression**
- **Main:** loaded `<dir of --config>/conf.d/*.toml` by default, and resolved a relative `--config-dir` against the config file's directory.
- **Now:** `Read` only reads drop-ins when `--config-dir` is passed (`pkg/config/config.go:618`, `if dropInConfigDir != "" {`). It resolves the directory against the working directory, and skips a missing one with a `V(2)` log (`loadDropInConfigs`, `:659`).
- **Scenario:** a deployment with `read_only`, `denied_resources` or `require_oauth` in a drop-in upgrades and starts without them.
- **Fix:** drop-in lookup is independent of the Option engine, so please keep main's behavior. Ideally, also fail when an explicitly passed `--config-dir` doesn't exist.

**Invalid env values are now fatal**
- **Where:** `applyEnv` returns `fmt.Errorf("env %s: %w", …)` (`pkg/config/option.go:135`), which fails startup and every SIGHUP.
- **Main ignored these:**
  - an unparseable `KUBE_CLIENT_QPS` / `KUBE_CLIENT_BURST`;
  - an unparseable or ≤0 `*_DEBOUNCE_WINDOW_MS` / `*_POLL_INTERVAL_MS` (the default was used);
  - an invalid or out-of-range `OTEL_TRACES_SAMPLER_ARG` (logged, default used).
- **Fix:** restore that with a lenient env parser for these options that logs a warning and keeps the file or default value.

**SIGHUP now rebuilds every manager, and can hang the server**
- **Where:** `ReloadConfiguration` calls `s.p.ReloadConfig(ctx, newConfig)` on every SIGHUP (`pkg/mcp/mcp.go:587`), even for a `log_level`-only change. It closes and recreates all managers and watchers under `reloadMu` and the provider lock.
- **Hang:**
  - `watcher.Kubeconfig.Watch` sets `w.started = true` and then returns early if `fsnotify.NewWatcher()` fails (`pkg/kubernetes/watcher/kubeconfig.go:56-65`).
  - `Close()` then blocks forever on `<-w.stoppedCh` (`:131`), and every tool call waits on the provider lock.
  - The bug predates this PR, but it was only reachable at shutdown; now any SIGHUP (for example logrotate) reaches it.
- **kcp:** `resetLocked` runs `DiscoverAllWorkspaces(context.TODO(), …)` (`pkg/kcp/provider.go:151`) while holding the write lock.
- **Failed reload:** `ReloadConfig` commits before `applyToolsetsLocked` (`mcp.go:594`). If the toolset step fails, the reload is reported as failed, but the rebuilt managers already enforce the new access-control values.
- **Fix:**
  - Set `started` only once the goroutine runs.
  - Instead of rebuilding, have the access-control round tripper read `denied_resources`, `validation_enabled` and the confirmation rules per request from the committed config, as it already does with `RestMapperProvider` (see the decoupling item above). Reverting to main isn't an option: there, SIGHUP never reached the Kubernetes client at all.

**Test helpers leak downstream default overrides**
- **Where:** `configtest.OverlayTOML` (`pkg/config/configtest/configtest.go:42`) goes through `ReadToml` → `resolve` → `config.New()`, which applies `defaultOverrides`. `KeepPreviousIfDefault` only keeps previous values whose source isn't `<Default>`.
- **Effect:** anything a suite seeded from `BaseDefault()` without `SetForTest` silently becomes the downstream default.
- **Scope:**
  - 48 call sites in 22 `pkg/mcp` test files use this helper.
  - `validate_test.go:23`, `mcp_config_provider_test.go:156` and `netobserv/config_test.go:70` switched from `BaseDefault()` to `New()`.
  - openshift-mcp-server overrides `read_only`, `disable_destructive`, `port`, `denied_resources` and `disabled_tools`, so its test run would diverge.
- **Fix:** let `ReadToml` start from `BaseDefault()` for tests, and revert those three swaps.

**envtest's QPS/burst no longer reaches the MCP suites**
- **Why these knobs matter:** `KUBE_CLIENT_QPS` / `KUBE_CLIENT_BURST` and the `*_DEBOUNCE_WINDOW_MS` / `*_POLL_INTERVAL_MS` env vars exist mainly so the suites run fast and aren't throttled on the client side.
- **Where:** `internal/test/envtest.go:43-44` sets `os.Setenv("KUBE_CLIENT_QPS", "1000")` / `os.Setenv("KUBE_CLIENT_BURST", "2000")`.
  - On main, `NewManager` read those env vars directly, so every test client got them.
  - Now they're only applied when a config is loaded with `config.Read` / `ReadToml`.
  - `BaseMcpSuite` starts from `config.BaseDefault()` (`pkg/mcp/common_test.go:121`), which ignores env.
- **Effect:**
  - Suites that don't go through `OverlayTOML` run with client-go's default 5 QPS / 10 burst, so this setup is no longer exercised.
  - Any `t.Setenv` before `New()` / `BaseDefault()` silently does nothing.
  - The watcher timing vars still work where a test loads config via `ReadToml` after `t.Setenv` (`provider_watch_test.go:32-34`, `provider_kubeconfig_test.go:340-342`).
- **Fix:**
  - Set the envtest defaults on the test config instead of the process env, e.g. `s.Cfg.KubeClientQPS.SetForTest(1000)` / `s.Cfg.KubeClientBurst.SetForTest(2000)` in `BaseMcpSuite.SetupTest`, or an `internal/test` helper that returns envtest-tuned values.
  - Add a `pkg/config` test that each of these env vars still resolves to its option.

### Should fix

**`McpReloader` is easy to misuse**
- **Where:** it's a new exported struct with three optional funcs (`pkg/kubernetes/provider.go:23`, `type McpReloader struct`).
- **Deadlock:** if `Apply` is nil, `ApplyToolsets()` (`:46`) falls back to `Reapply`, which takes `reloadMu`. But it's called inside `Run`/`Do`, which already holds that lock.
- **Fix:** unexported fields behind a constructor (or a small interface), with no fallback.

**`ReadToml` drops the caller's context**
- **Where:** `resolve(context.Background(), …)` (`pkg/config/config.go:768`).
- **Why it matters now:** main did the same, but this now also runs from `Complete` when no config is given (`root.go:130`) and emits the option dump there, so those logs lose the logger and trace context.
- **Fix:** add a `ctx` parameter.

**Spec and code disagree**
- The spec's `Option` has `ParseTOML func(any) (T, error)` (`docs/specs/config.md:70`), but TOML parsing is hardwired in `parseTyped` (`option.go:212`).
- The spec validates before dumping (`config.md:251`); the code dumps first.
- `toolsetConfigsSource` / `clusterProviderConfigsSource` are written but never read (`config.go:351-353`).
- The `reflect.Struct` branch in `rejectContainerUnknown` (`extended.go:206`) can never run, because `isContainer` only matches maps and slices of structs.

### Tests

**Behavioral tests for the engine.** Config in, observable effect out:
- **Unknown fields inside a registered extension table** (`rejectUnknownStructKeys` call, `extended.go:56`): startup fails. On SIGHUP, `toolset_configs` fails the load and `cluster_provider_configs` keeps the previous table. Deleting that call passes CI today.
- **Option dump:** with `log_level = 1`, every option is logged with its source and `client_secret` is redacted. On SIGHUP, changed values carry `changed=true` / `previous`, and a rejected reload reports nothing as changed.
- **Drop-in provenance:** a key overridden in `conf.d/10-x.toml` reports that file as its source. Cover a scalar, `[http]`, `[[prompts]]` and `tool_overrides`.
- **SIGHUP pinning** for each non-reloadable option type: string, bool, slice, `*bool`, duration and float32.
- **Kubernetes-client reloads:** changes to `validation_enabled` and kube-level `confirmation_rules` reach the client (only `denied_resources` is covered), and kcp `ReloadConfig` keeps its manager on error.
- **Precedence end to end:** file < drop-in < env, with empty env ignored.

**Table-driven tests**
- **Where:** new table tests in `pkg/config/validate_test.go` (`TestListOutput` `:53` with `cases := []struct{…}`, plus `TestToolsets`, `TestClusterProviderStrategy`, `TestAuthorizationURL`, `TestTokenExchangeStrategy`, `TestConfirmationFallback`, `TestSkipJWTVerification`, `TestMetricsPort`) and `config_test.go:167,190`. They replace the explicit `s.Run` subtests main had, and several branch on `if tc.wantErr == ""`.
- **Fix:** please keep explicit `s.Run` scenarios or parameterized helpers (see `pkg/mcp/require_tls_test.go`).

**Assertions that can't fail**
- `TestSIGHUPIgnoredWithoutConfig` sets `Port` after `Complete` (`root_sighup_test.go:683`). The log sink is already in stdio mode writing to `io.Discard`, so the `Never(… "Shutting down HTTP server" …)` check can't fail.
- `config_test.go:1172` has `s.Equal(cfg.TLSCipherSuites.Get(), cfg.TLSCipherSuites.Get())`, which compares a value with itself.
- Coverage of the removed `sts_*` keys dropped from all 11 `removedKeyReplacements` entries to one.

### Docs

**`docs/dev/testing.md` examples no longer compile**
- `toml.Unmarshal(…, s.Cfg)` (`:153`)
- `candidateStatic.KubeConfig = s.Cfg.KubeConfig` (`:209`)
- `s.Cfg.Toolsets = append(...)`
- `map[string]toml.Primitive` (`:165`)

Please update it for `SetForTest` and `configtest`.

**`docs/configuration.md` examples fail the new strict key check**
- **Complete Example:** top-level keys placed after `[http]` (`kubeconfig`, `kube_client_qps = 50`, `read_only`, `toolsets`… at `:913-932`) parse as `http.*`, and `server_instructions` after `[[denied_resources]]` (`:945`) lands in that table.
- **Kiali example:** `token = "your-kiali-token"` (`:819`) isn't a Kiali config field.
- **kcp example:** `[cluster_provider_configs.kcp]` (`:860`) has no registered parser.
- **Wrong reload text:** `:228` says outbound clients pick up TLS changes on reload, but `tls_min_version` / `tls_cipher_suites` are pinned.

### Minor

- **Quoted dotted keys slip through the unknown-key check.**
  - `"http.max_body_bytes" = 1048576` (quoted) is one top-level key. `flattenKeys` (`config.go:1006`) turns it into the string `http.max_body_bytes`, so it passes.
  - `mapLookup` (`:849`) then finds no `http` table and drops the value silently.
  - Fix: reject key segments that contain `.`.
- **`tomlFields` doesn't match BurntSushi's field matching** (`extended.go:115`). Untagged fields match case-insensitively and embedded structs are promoted, so a downstream extension parser can be rejected falsely. Checking `md.Undecoded()` after the parser runs would be exact.
- **`Validate` trims values in place** (`validate.go:25-27`, `replaceValue(strings.TrimSpace(…))`).
  - `tls_cert = " /x "` logs "Ignoring change to non-reloadable option" on every SIGHUP.
  - Per-option validator errors (`validate.go:39`) don't include the option path or source.
- **Options are keyed by joined TOML path** (`optionsByPath`, `config.go:581`). Two options with an empty `TOMLKey` (the documented opt-out) collide, so pinning copies the wrong value.
- **Drop-in provenance for map options:** `deepMerge` (`config.go:729`) doesn't update the table's own source when it recurses, so `tool_overrides` reports the first file, not the last.
- **Override hook limits:** `defaultOverrides` can't rename the table names (`http`, `telemetry`, `token_exchange`, `toolset_configs`, `cluster_provider_configs`). Overriding `ConfigPathEnvName` there writes a package global on every `New()`, including from the SIGHUP goroutine; an `init()` avoids that race.
- **Nit, file split:** `HTTPConfig`, `TelemetryConfig` and `TokenExchangeConfig` (plus their validators) could keep their own files (`http_config.go`, `telemetry_config.go`, `token_exchange_config.go`) as on main, with `validate.go` only for cross-domain rules.
- **Nits:**
  - The error prefix is doubled: `failed to read config X: failed to read config X: …` (`root.go:125` + `config.go:708`).
  - `Makefile:194,197` still prints `--toolsets core,config`.
  - `docs/getting-started-claude-code.md:9,19` repeats "Add the MCP server using the `claude mcp add-json` command:".
  - `pkg/mcp/mcp_reload_test.go:286` has a comment pasted inside the `SetForTest(...)` arguments.

pkg/kubernetes-mcp-server/cmd/root.go
2uasimojo: Yeah, I would rather take a YAGNI approach: it's much harder to remove things than to add them.

docs/specs/config.md
2uasimojo: The choice was deliberate here: I wanted to make it easier for reviewers to comment on individual cells. In retrospect, I probably overthought it :)

I'll convert to native markdown shape.

docs/configuration-changes.md
2uasimojo: Ack.

2uasimojo:
@manusa thanks for the review!

> ### Must fix
> **Keep `pkg/kubernetes` and the toolset clients decoupled from `pkg/config`**
> 
> * **Background:** #566 moved config access behind interfaces in `pkg/api` (`api.BaseConfig` and friends). That way `pkg/kubernetes`, the providers and the toolset clients depend on abstractions, not the concrete config type. On main, no non-test file in `pkg/kubernetes` imports `pkg/config`.
> * **What changed:** this PR deletes those interfaces from `pkg/api/config.go`. `NewProvider`, `NewManager`, `NewKubernetes`, `ProviderFactory`, token exchange, `pkg/kcp`, Kiali and NetObserv now take `*config.Config` directly. 8 non-test files in `pkg/kubernetes`, plus `pkg/kcp/provider.go`, now import `pkg/config`.
> * **Symptom:** Kiali and NetObserv can no longer get TLS settings through `pkg/api`. `configFromProvider` (`pkg/kiali/kiali.go:221-233`, `pkg/netobserv/netobserv.go:247-259`) type-switches to `*config.Config`. For anything else it falls back to `config.New()`, which means `require_tls=false`. That fallback is a latent fail-open.
> * **Fix:** the Option model doesn't require the coupling.
>   
>   * `*config.Config` can implement narrow `pkg/api` interfaces with one-line getters over `Option.Get()`, e.g. `func (c *Config) GetDeniedResources() []api.GroupVersionKind { return c.DeniedResources.Get() }`.
>   * Watcher timings and QPS can be passed as constructor arguments.
>   * A `func() api.BaseConfig` provider backed by the committed config then gives the access-control round tripper live values, without rebuilding managers on every SIGHUP (see the SIGHUP item below).

Getting rid of explicitly-named one-line custom accessor wrappers was one of my main goals with this work. Looking at #566, it feels okay because there's only a small handful of options. But at the scale of the current config footprint, it becomes cluttered and noisy (IMO) and the accessors don't add value.

I agree the type switch is ugly and hacky. But I'd like to solve it by backing out the ExtendedProviderConfig and just passing Config to the relevant methods. I get that it was intended to abstract/separate provider-specific config knobs from a) the main config, and b) other providers' provider-specific configs. But a) is something we actively don't want, as demonstrated here (we still want to be able to see e.g. TLS settings) and b) is artificial, as any provider could `Get{Provider|Toolset}Config()` with params that'll get the other guy's thing.

It seems what's making this awkward is that, today, config imports api, resulting in import cycles when trying to clean and simplify. I'm looking at inverting that (api imports config).

----

> **Helm chart with `tls.enabled=true` won't start**
> 
> * **Where:** `charts/kubernetes-mcp-server/templates/deployment.yaml:56-57` still passes `"--tls-cert=…"` / `"--tls-key=…"`. The file isn't in this diff, so chart CI didn't run, and no values file or e2e test enables TLS.
> * **Fix:** move these into the chart's TOML (`tls_cert` / `tls_key`).
> * **Docs:** the migration notes should warn that chart `extraArgs` and container `args` using removed flags now fail with `unknown flag`.

Good find, fixed ✓

----

> **The startup option dump is never printed**
> 
> * **Where:** `dumpConfig` logs at `V(1)` and is called inside `resolve` (`pkg/config/config.go:810`, `dumpConfig(ctx, cfg, settings.previous)`). That runs from `Complete` via `config.Read`/`ReadToml` (`pkg/kubernetes-mcp-server/cmd/root.go:122,130`), before `logging.New` sets klog's verbosity (`root.go:150`).
> * **At startup:** the dump is dropped even with `log_level = 5`, although `docs/configuration.md:58` says it's logged.
> * **On SIGHUP:** it runs before `Validate`, so a rejected reload logs `changed=true` / `previous` for values that never took effect.
> * **Fix:** dump after `logging.New`, and on SIGHUP only after the reload succeeds.

Fixed, including making `Dump()` a method on `Config`, which seemed to make sense.

----

> **The Docker image ignores `K8S_MCP_CONFIG_PATH`**
> 
> * **Where:** `Dockerfile:22` is now `CMD ["--config", "/etc/kubernetes-mcp-server/config.toml"]`, and `root.go:116` (`m.ConfigPath == "" && cp != ""`) only reads the env var when `--config` is empty.
> * **Scenario:** `docker run -e K8S_MCP_CONFIG_PATH=/cfg/config.toml …` worked on main and is documented. Now the user's file is silently ignored, and the server starts on `0.0.0.0:8080` without `require_oauth`, `read_only` or `denied_resources`.
> * **Fix:** `ENV K8S_MCP_CONFIG_PATH=/etc/kubernetes-mcp-server/config.toml` with no `CMD` keeps both the env var and an explicit `--config` working.

Thanks. I stared at this for a minute, but couldn't remember the nuances of our discussion a couple months ago when we fixed this. Agree with the suggested resolution :+1: 

I'm also renaming the var to drop the `K8S_` prefix. We're really not worried about a conflict, and it will make life easier for downstreams.

----

> **Dropping the implicit `conf.d` is a silent regression**
> 
> * **Main:** loaded `<dir of --config>/conf.d/*.toml` by default, and resolved a relative `--config-dir` against the config file's directory.
> * **Now:** `Read` only reads drop-ins when `--config-dir` is passed (`pkg/config/config.go:618`, `if dropInConfigDir != "" {`). It resolves the directory against the working directory, and skips a missing one with a `V(2)` log (`loadDropInConfigs`, `:659`).
> * **Scenario:** a deployment with `read_only`, `denied_resources` or `require_oauth` in a drop-in upgrades and starts without them.
> * **Fix:** drop-in lookup is independent of the Option engine, so please keep main's behavior. Ideally, also fail when an explicitly passed `--config-dir` doesn't exist.

I definitely don't want to bring implicit conf.d back. I got rid of it because the semantics were "magical" and confusing. However, I can see your point: a consumer who didn't read the upgrade doc and was using implicit conf.d now *isn't*. Best case, the "removal" of the options therein results in an invalid config and we error... but worst case we lose overrides and e.g. become less secure.

So even though this PR is explicitly a breaking change, here it's worth adding a regression check: we'll look for the condition that _would have_ loaded conf.d and error if we find it.

----

> **Invalid env values are now fatal**
> 
> * **Where:** `applyEnv` returns `fmt.Errorf("env %s: %w", …)` (`pkg/config/option.go:135`), which fails startup and every SIGHUP.
> * **Main ignored these:**
>   
>   * an unparseable `KUBE_CLIENT_QPS` / `KUBE_CLIENT_BURST`;
>   * an unparseable or ≤0 `*_DEBOUNCE_WINDOW_MS` / `*_POLL_INTERVAL_MS` (the default was used);
>   * an invalid or out-of-range `OTEL_TRACES_SAMPLER_ARG` (logged, default used).
> * **Fix:** restore that with a lenient env parser for these options that logs a warning and keeps the file or default value.

I don't agree with this. Aggressively failing on bad config is one of the main purposes of this PR.

If we revert to lenient, a misspelled env var value quietly (cause who's reading the logs if there's no error?) results in some other value being used, which could result in unintended and potentially damaging behavior.

Keeping it strict, the server fails to start, *forcing* the user to unveil and fix the problem. Worst case, the env var is being set by something upstream of our invocation; but it's trivial to override in that scenario.

----

> **SIGHUP now rebuilds every manager, and can hang the server**
> 
> * **Where:** `ReloadConfiguration` calls `s.p.ReloadConfig(ctx, newConfig)` on every SIGHUP (`pkg/mcp/mcp.go:587`), even for a `log_level`-only change. It closes and recreates all managers and watchers under `reloadMu` and the provider lock.
> * **Hang:**
>   
>   * `watcher.Kubeconfig.Watch` sets `w.started = true` and then returns early if `fsnotify.NewWatcher()` fails (`pkg/kubernetes/watcher/kubeconfig.go:56-65`).
>   * `Close()` then blocks forever on `<-w.stoppedCh` (`:131`), and every tool call waits on the provider lock.
>   * The bug predates this PR, but it was only reachable at shutdown; now any SIGHUP (for example logrotate) reaches it.
> * **kcp:** `resetLocked` runs `DiscoverAllWorkspaces(context.TODO(), …)` (`pkg/kcp/provider.go:151`) while holding the write lock.
> * **Failed reload:** `ReloadConfig` commits before `applyToolsetsLocked` (`mcp.go:594`). If the toolset step fails, the reload is reported as failed, but the rebuilt managers already enforce the new access-control values.
> * **Fix:**
>   
>   * Set `started` only once the goroutine runs.
>   * Instead of rebuilding, have the access-control round tripper read `denied_resources`, `validation_enabled` and the confirmation rules per request from the committed config, as it already does with `RestMapperProvider` (see the decoupling item above). Reverting to main isn't an option: there, SIGHUP never reached the Kubernetes client at all.

Sigh. Revised this mess *again*. Still low confidence that it's bulletproof.

----

> **Test helpers leak downstream default overrides**
> 
> * **Where:** `configtest.OverlayTOML` (`pkg/config/configtest/configtest.go:42`) goes through `ReadToml` → `resolve` → `config.New()`, which applies `defaultOverrides`. `KeepPreviousIfDefault` only keeps previous values whose source isn't `<Default>`.
> * **Effect:** anything a suite seeded from `BaseDefault()` without `SetForTest` silently becomes the downstream default.
> * **Scope:**
>   
>   * 48 call sites in 22 `pkg/mcp` test files use this helper.
>   * `validate_test.go:23`, `mcp_config_provider_test.go:156` and `netobserv/config_test.go:70` switched from `BaseDefault()` to `New()`.
>   * openshift-mcp-server overrides `read_only`, `disable_destructive`, `port`, `denied_resources` and `disabled_tools`, so its test run would diverge.
> * **Fix:** let `ReadToml` start from `BaseDefault()` for tests, and revert those three swaps.

✓

----

> **envtest's QPS/burst no longer reaches the MCP suites**
> 
> * **Why these knobs matter:** `KUBE_CLIENT_QPS` / `KUBE_CLIENT_BURST` and the `*_DEBOUNCE_WINDOW_MS` / `*_POLL_INTERVAL_MS` env vars exist mainly so the suites run fast and aren't throttled on the client side.
> * **Where:** `internal/test/envtest.go:43-44` sets `os.Setenv("KUBE_CLIENT_QPS", "1000")` / `os.Setenv("KUBE_CLIENT_BURST", "2000")`.
>   
>   * On main, `NewManager` read those env vars directly, so every test client got them.
>   * Now they're only applied when a config is loaded with `config.Read` / `ReadToml`.
>   * `BaseMcpSuite` starts from `config.BaseDefault()` (`pkg/mcp/common_test.go:121`), which ignores env.
> * **Effect:**
>   
>   * Suites that don't go through `OverlayTOML` run with client-go's default 5 QPS / 10 burst, so this setup is no longer exercised.
>   * Any `t.Setenv` before `New()` / `BaseDefault()` silently does nothing.
>   * The watcher timing vars still work where a test loads config via `ReadToml` after `t.Setenv` (`provider_watch_test.go:32-34`, `provider_kubeconfig_test.go:340-342`).
> * **Fix:**
>   
>   * Set the envtest defaults on the test config instead of the process env, e.g. `s.Cfg.KubeClientQPS.SetForTest(1000)` / `s.Cfg.KubeClientBurst.SetForTest(2000)` in `BaseMcpSuite.SetupTest`, or an `internal/test` helper that returns envtest-tuned values.
>   * Add a `pkg/config` test that each of these env vars still resolves to its option.

✓

----

> ### Should fix
> **`McpReloader` is easy to misuse**
> 
> * **Where:** it's a new exported struct with three optional funcs (`pkg/kubernetes/provider.go:23`, `type McpReloader struct`).
> * **Deadlock:** if `Apply` is nil, `ApplyToolsets()` (`:46`) falls back to `Reapply`, which takes `reloadMu`. But it's called inside `Run`/`Do`, which already holds that lock.
> * **Fix:** unexported fields behind a constructor (or a small interface), with no fallback.

Fixed alongside the SIGHUP rework above.

----

> **`ReadToml` drops the caller's context**
> 
> * **Where:** `resolve(context.Background(), …)` (`pkg/config/config.go:768`).
> * **Why it matters now:** main did the same, but this now also runs from `Complete` when no config is given (`root.go:130`) and emits the option dump there, so those logs lose the logger and trace context.
> * **Fix:** add a `ctx` parameter.

✓

----

Will start grinding through the rest tomorrow, but I think this may be all the important things?

charts/kubernetes-mcp-server/README.md
```diff
@@ -113,7 +113,7 @@ Each container accepts any valid Kubernetes container field including `image`, `
 | configFilePath | string | `"/etc/kubernetes-mcp-server/config.toml"` |  |
 | defaultPodSecurityContext | object | `{"seccompProfile":{"type":"RuntimeDefault"}}` | Default Security Context for the Pod when one is not provided |
 | defaultSecurityContext | object | `{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"readOnlyRootFilesystem":true,"runAsNonRoot":true}` | Default Security Context for the Container when one is not provided |
-| extraArgs | list | `[]` | Note: For TLS configuration, use the tls section above instead of extraArgs. |
+| extraArgs | list | `[]` | Only bootstrap flags remain (`--config`, `--config-dir`, `--version`). Runtime flags such as `--tls-cert` fail with `unknown flag`; put those settings in `config` or the `tls` section. |
```
Cali0707: Do we want to use the phrasing "remain" here? rather than just say the current state of the config?

pkg/kcp/provider.go
```diff
@@ -145,9 +144,8 @@ func (p *kcpClusterProvider) resetLocked(ctx context.Context) error {
 
 // discoverWorkspaces queries the kcp tenancy API to discover available workspaces.
 // It recursively discovers nested workspaces in the workspace hierarchy.
-// Runs under lock.
-func (p *kcpClusterProvider) discoverWorkspaces(restConfig *rest.Config, defaultWorkspace string) ([]string, error) {
-	return DiscoverAllWorkspaces(context.TODO(), restConfig, defaultWorkspace)
```
Cali0707: is there a reason this method is just a direct wrapper around another function?

Cali0707 (commented):
Will take another pass tmrw morning, but this is looking pretty good.

Thanks for all the iteration here @2uasimojo !!

manusa:
Thanks for the fast turnaround, and for taking the chart, dump, Dockerfile, test-helper and envtest items straight away. Two things from your replies I'd like to close on before the next pass.

### 1. Strict config: I agree, but the reload path needs to say so out loud

On env specifically I'm with you, and it's even safer than it looks: a running process's environment can't change under it, and nothing in production code calls `os.Setenv` (only `internal/test`). `resolve` re-applies env on every load, so any value present at SIGHUP already parsed at startup. Strictness there costs nothing at runtime. No change wanted.

My concern is what strictness does on the **hot-reload** path, where "fail outright" quietly turns into "keep running on the old config":

- A rejected reload logs one `Error` line and keeps the previous config (`root.go:288-291`). Nothing else surfaces it: `/healthz` is a constant 200 with no `/readyz` (`pkg/http/http.go:159-161`), there's no reload metric in `pkg/metrics`, and MCP clients are never told, because `pkg/mcplog` only emits when a session is in the context and the SIGHUP context has none. With the default `log_level = 0`, a *successful* reload is silent too, since `Dump` is V(1).
- **The result is a landmine:** the running pod is healthy on the old config while the on-disk config would abort `Complete`/`Validate` on the next start. The next restart or rollout CrashLoops with `unknown config key …`, with nothing in between to warn anyone.
- **Two variants are worse, because the reload succeeds.** `pinNonReloadable` (`config.go:844`) replaces the new value with the old one *before* `Config.Validate` runs, so `metrics_port = "abc"`, or a `tls_cert` path that doesn't exist yet, reloads cleanly with a WARN that says "restart the process to apply it" — the one action that will fail. Same shape for an invalid `cluster_provider_configs`, tolerated on reload (`config.go:1072-1084`) but fatal at startup.
- **New in this head:** `cfgState.Store(newConfig)` now runs *before* `ReloadConfiguration` validates it (`root.go:293-302`). `Config.Validate` mutates in place — `TrimSpace` on the CA/TLS fields plus `normalizeTokenExchange` rewriting ten more options — while the auth middleware (`pkg/http/authorization.go`) and token exchange (`pkg/kubernetes/provider_token_exchange.go`) read those same options through `cfgState.Load()`. That's an unsynchronized write against live readers, and during validation a not-yet-validated config is authoritative. On main the store happened after the reload succeeded.

None of the asks below make invalid config lenient:

1. **Fix the two WARN strings** (`config.go:1076`, `1107`). Run the option's own validator on the candidate value in `pinNonReloadable`, and when it fails say so, rather than advising a restart that will abort.
2. **Log the divergence at `Error`**, not V(1): the running config no longer matches what's on disk, and the next start will fail.
3. **Make it machine-readable:** a `config_reload_total{result="ok|rejected"}` counter plus a stale gauge on the existing OTel collector. A `/readyz` that goes not-ready after a rejected reload would stop a rollout before it takes out the fleet, but the metric alone closes most of the gap.
4. **Move `cfgState.Store` after validation**, or normalize in `resolve` so `Validate` doesn't mutate a published object.
5. **Two tests** to pin the intent: a SIGHUP with `metrics_port = "abc"` asserting the new wording, and a bad `cluster_provider_configs` key asserting that the same file still fails a fresh startup.

### 2. `api` → `config`: direction is fine, shape needs work before 0.1.0

I'm not asking for the accessors back — your objection stands. None of the below blocks this PR; it's API shape I'd want settled before 0.1.0 ships, because after that, out-of-tree toolsets compile against it. A follow-up PR is fine by me.

Three observations about the inversion as it landed:
- **`pkg/config` isn't a leaf.** `validate.go:14-17` imports `klogutil`, `output`, `tlsutil` and `tokenexchange`, and `pkg/tokenexchange/registry.go:9` has an `init()`. Every package importing `pkg/api` now links and runs it.
- **It didn't decouple anything.** `pkg/kubernetes`, `pkg/kcp`, Kiali and NetObserv still take `*config.Config` exactly as before; only the edge direction changed.
- **The leak already happened.** `pkg/config/schema.go` now owns `Prompt`, `PromptArgument`, `PromptTemplate`, `ConfirmationRule` (with its `Validate()`) and `GroupVersionKind`. Six packages, four of them toolsets, import `pkg/config` purely to name a type. Non-test packages importing `pkg/config` went from 11 on main to 22.

The part I care most about is that the handler surface is now **mutable**. `ToolHandlerParams.Config` hands every handler the live pointer, and:
- `SetForTest` is exported from production code (`option.go:64`);
- all eight `Option` metadata fields are exported and writable (`option.go:28-38`) — a handler can flip `Reloadable`, replace `Validate`, or clear `Sensitive` so a secret stops being redacted;
- `Get()` shares the backing store for slices and maps, so `params.Config.DeniedResources.Get()[0].Kind = ""` mutates the live denylist, and the same goes for `ToolOverrides`, `Prompts`, `ConfirmationRules` and `Toolsets`;
- `GetTokenExchangeConfig()` returns a pointer into the live config.

`api.BaseConfig` was read-only by construction, so this is new. It's fixable without accessors: guard `SetForTest` with `testing.Testing()`, unexport the metadata fields (only the fluent builders set them, and `defaultOverrides` is in package `config`, so the downstream seam is unaffected), and return copies from `Get()` for container options. That gets immutability in practice while keeping `*config.Config` in the params.

Two smaller things:
- `toolsets.Validate` moved out of `Config.Validate` into two call sites (`cmd/root.go:171`, `mcp.go:582`), so an embedder calling `cfg.Validate(ctx)` alone silently accepts an unknown toolset name. With `WithProviderStrategies` (skipped when not injected) and the `tokenexchange.GetRegisteredStrategies()` fallback (`validate.go:196`), there are now three shapes for the same "validate against a registry" problem — worth picking one.
- `ToolHandlerParams.ClusterProviderStrategy` / `PromptHandlerParams.ClusterProviderStrategy` duplicate `params.Config.ClusterProviderStrategy.Get()` and are only read in tests.

Worth having, whenever you get to it: a `pkg/config/imports_test.go` mirroring the `api` one (whitelist `{klogutil, output, tlsutil}`, drop the `tokenexchange` fallback since both production callers already inject), and a note in `docs/specs/config.md` recording the direction and what `pkg/config` must never import, since the new "Package layout" section doesn't mention it.

### Heads-up: CI is red on `f6868d96`

- `make lint`: `internal/test/config.go:6:2: SA9004: only the first constant in this group has an explicit type (staticcheck)`, which fails every Build job.
- Helm validation: `open ./charts/kubernetes-mcp-server/ci/tls-test-values.yaml: no such file or directory` — looks like the values file didn't get committed with the `helm.mk` change.
- MCP conformance is also failing.

charts/kubernetes-mcp-server/README.md
2uasimojo: fixed, I think.

pkg/kcp/provider.go
2uasimojo: Fixed

2uasimojo:
> **Spec and code disagree**
> 
> * The spec's `Option` has `ParseTOML func(any) (T, error)` (`docs/specs/config.md:70`), but TOML parsing is hardwired in `parseTyped` (`option.go:212`).

Removed from the spec.

> * The spec validates before dumping (`config.md:251`); the code dumps first.

Fixed: always dump, startup x SIGHUP and validate succeed x fail.

> * `toolsetConfigsSource` / `clusterProviderConfigsSource` are written but never read (`config.go:351-353`).

This was a hole in Dump() -- fixed.

> * The `reflect.Struct` branch in `rejectContainerUnknown` (`extended.go:206`) can never run, because `isContainer` only matches maps and slices of structs.

Removed.

Cali0707 (commented):
/run-mcpchecker kubevirt

Cali0707 (approved):
LGTM, but since I pushed a last commit, leaving a review of that to @2uasimojo

PR #1425, opened 2026-09-11
feat(kiali): Filter Kiali tools by configured URL reachability

josunect:
Split from: https://github.com/containers/kubernetes-mcp-server/pull/1339 

Filter Kiali tools by configured URL reachability

When experimental_enable_target_compatibility_tool_filters is enabled, probe GET {url}/api/status and hide the Kiali toolset when the configured URL is missing or unreachable. Expose live toolset config during tool collection so the check works with wrapped kubernetes providers.

When merged, rebase https://github.com/containers/kubernetes-mcp-server/pull/1339

### DIFF
--- a/pkg/kiali/gvr.go
+++ b/pkg/kiali/gvr.go
@@ -1,0 +1,56 @@
+package kiali
+
+import (
+	"context"
+	"strings"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/api"
+	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
+	"github.com/containers/kubernetes-mcp-server/pkg/kubernetes"
+)
+
+// HasKiali reports whether a configured Kiali URL responds to GET /api/status.
+//
+// fp supplies toolset config via ExtendedConfigProvider; kp is the Kubernetes
+// provider used for bearer tokens (may be nil in tests).
+func HasKiali(ctx context.Context, fp api.FilteringProvider, kp kubernetes.Provider) bool {
+	cfg, ok := kialiConfigFromProvider(fp)
+	if !ok || strings.TrimSpace(cfg.Url) == "" {
+		return false
+	}
+
+	token := bearerTokenFromProvider(ctx, kp)
+	ok = probeStatusURL(ctx, cfg.Url, cfg, token)
+	if !ok {
+		klogutil.FromContext(ctx).V(1).Info("configured Kiali URL failed /api/status probe; disabling Kiali tools",
+			"url", cfg.Url)
+	}
+	return ok
+}
+
+func kialiConfigFromProvider(fp api.FilteringProvider) (*Config, bool) {
+	cfgProvider, ok := fp.(api.ExtendedConfigProvider)
+	if !ok || cfgProvider == nil {
+		return nil, false
+	}
+	ext, ok := cfgProvider.GetToolsetConfig("kiali")
+	if !ok {
+		return nil, false
+	}
+	kc, ok := ext.(*Config)
+	if !ok || kc == nil {
+		return nil, false
+	}
+	return kc, true
+}
+
+func bearerTokenFromProvider(ctx context.Context, kp kubernetes.Provider) string {
+	if kp == nil {
+		return ""
+	}
+	k8s, err := kp.GetDerivedKubernetes(ctx, kp.GetDefaultTarget())
+	if err != nil || k8s == nil || k8s.RESTConfig() == nil {
+		return ""
+	}
+	return k8s.RESTConfig().BearerToken
+}
--- a/pkg/kiali/status.go
+++ b/pkg/kiali/status.go
@@ -1,0 +1,94 @@
+package kiali
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/containers/kubernetes-mcp-server/pkg/klogutil"
+)
+
+const (
+	statusPath         = "/api/status"
+	statusProbeTimeout = time.Second
+)
+
+// statusResponse is the subset of Kiali GET /api/status used for reachability checks.
+type statusResponse struct {
+	Status map[string]any `json:"status"`
+}
+
+// probeStatusURL GETs {baseURL}/api/status and returns true when the response
+// is HTTP 2xx and contains a non-empty status object (Kiali is reachable).
+func probeStatusURL(ctx context.Context, baseURL string, cfg *Config, bearerToken string) bool {
+	baseURL = strings.TrimRight(strings.TrimSpace(baseURL), "/")
+	if baseURL == "" {
+		return false
+	}
+	statusURL := baseURL + statusPath
+
+	probeCfg := cfg
+	if probeCfg == nil {
+		probeCfg = &Config{Url: baseURL}
+	} else if strings.TrimSpace(probeCfg.Url) == "" {
+		cp := *probeCfg
+		cp.Url = baseURL
+		probeCfg = &cp
+	}
+
+	k := &Kiali{
+		bearerToken:          bearerToken,
+		kialiURL:             probeCfg.Url,
+		kialiInsecure:        probeCfg.Insecure,
+		certificateAuthority: probeCfg.CertificateAuthority,
+	}
+
+	probeCtx, cancel := context.WithTimeout(ctx, statusProbeTimeout)
+	defer cancel()
+
+	client, err := k.createHTTPClient(probeCtx)
+	if err != nil {
+		klogutil.FromContext(ctx).V(2).Info("kiali status probe: failed to create HTTP client", "error", err)
+		return false
+	}
+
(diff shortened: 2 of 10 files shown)


pkg/kiali/gvr.go
```diff
@@ -0,0 +1,56 @@
...
+	ok = probeStatusURL(ctx, cfg.Url, cfg, token)
+	if !ok {
+		klogutil.FromContext(ctx).V(1).Info("configured Kiali URL failed /api/status probe; disabling Kiali tools",
+			"url", cfg.Url)
+	}
+	return ok
+}
+
+func kialiConfigFromProvider(fp api.FilteringProvider) (*Config, bool) {
+	cfgProvider, ok := fp.(api.ExtendedConfigProvider)
```
Cali0707: Do not get config (or anything else) by widening the FilterProvider interface

This interface is intentional, and it is a bad idea to create an undocumented dependency on the concrete type we are passing in implementing an extra interface. Additionally, the ExtendedConfigProvider is broader than you need here.

Instead, if you need config here, we should solve a way to get that config probably through changing the interface passed in or the toolset interface in some way.

pkg/toolsets/kiali/toolset.go
```diff
@@ -24,28 +31,33 @@ func (t *Toolset) GetDescription() string {
...
 	}
 	return tools
 }
 
+func kialiAvailable(p api.FilteringProvider) bool {
+	var kp kubernetes.Provider
+	if provider, ok := p.(kubernetes.Provider); ok {
+		kp = provider
+	}
+	return kialiclient.HasKiali(context.Background(), p, kp)
```
Cali0707: I'm pretty sure this assertion will fail and always pass a nil `kp` into `HasKiali`, because the `filteringProviderWithConfig` only wraps embeds the api.FilteringProvider type.

pkg/mcp/mcp.go
```diff
@@ -407,6 +411,27 @@ func (s *Server) collectApplicableTools(cfg *Configuration) []api.ServerTool {
 	return tools
 }
 
+// filteringProviderWithConfig embeds FilteringProvider and exposes the live
+// ExtendedConfigProvider so target-compatibility filters can inspect toolset_configs.
+type filteringProviderWithConfig struct {
+	api.FilteringProvider
+	cfg api.ExtendedConfigProvider
+}
```
Cali0707, line 427: I’m concerned about introducing this adapter in the generic tool collection path. It combines filtering and configuration capabilities so that individual toolsets can retrieve config through a runtime interface assertion, even though those responsibilities are otherwise separate.

This also makes behavior dependent on the adapter’s method set: the wrapped Kubernetes provider’s broader methods are no longer exposed, and configuration reloads can mix values from `cfg.StaticConfig` with values retained by `s.p`

Could we step back and define an explicit way for configuration-dependent applicability checks to receive their required inputs? I would prefer that over making `FilteringProvider` indirectly serve as a configuration provider.

pkg/toolsets/kiali/toolset.go
```diff
@@ -24,28 +31,33 @@ func (t *Toolset) GetDescription() string {
...
-	)
-	// Kiali calls a single configured endpoint; mesh scope is selected via meshCluster,
-	// not the provider-level context parameter injected for core Kubernetes tools.
+func (t *Toolset) GetTools(p api.FilteringProvider) []api.ServerTool {
+	if p != nil && !p.IsTargetCompatibilityToolFiltersEnabled() {
+		warnKialiValidationDisabledOnce.Do(func() {
+			klogutil.LogWarn(
+				klogutil.FromContext(context.Background()),
+				"experimental_enable_target_compatibility_tool_filters is disabled; Kiali URL reachability is not validated",
+			)
```
Cali0707: If we are needing context.Background for retrieving the logger (which is normally used for propagating the contextual logging), maybe this is a sign we should be passing context into this method somehow?

cc @2uasimojo

pkg/toolsets/kiali/toolset.go
Cali0707: Here we also really should be passing in a real context, not the background context

pkg/kiali/status.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	if auth := k.authorizationHeader(); auth != "" {
+		req.Header.Set("Authorization", auth)
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		klogutil.FromContext(ctx).V(2).Info("kiali status probe failed", "url", statusURL, "error", err)
+		return false
+	}
```
Cali0707, line 73: is there any way to distinguish transient network errors from other errors here?

In the GVK filtering approach, we try to fail open whenever it may be a transient network error

better to show a few too many tools than to hide a valid tool

pkg/toolsets/kiali/toolset.go
```diff
@@ -60,7 +72,6 @@ func (t *Toolset) GetPrompts() []api.ServerPrompt {
 		kialiPrompts.InitTraceAnalysis(),
 		kialiPrompts.InitIstioConfigReview(),
 	)
```
Cali0707: I beleive a lot of these prompts should ideally be hidden if kiali is unavailable, just like the tools.

@2uasimojo any thoughts on extending your initial tool work to include prompts and resources too?

pkg/kiali/status.go
```diff
@@ -0,0 +1,94 @@
...
+		cp.Url = baseURL
+		probeCfg = &cp
+	}
+
+	k := &Kiali{
+		bearerToken:          bearerToken,
+		kialiURL:             probeCfg.Url,
+		kialiInsecure:        probeCfg.Insecure,
+		certificateAuthority: probeCfg.CertificateAuthority,
+	}
```
Cali0707: This constructs a partial Kiali client rather than using the same configuration path as normal Kiali requests. `createHTTPClient` also uses `tlsMinVersion`, `tlsCipherSuites`, and `requireTLS`, but those fields remain unset here.

As a result, the status probe and actual tool calls can use different TLS policies. For example, the probe may allow a TLS version/cipher rejected by normal requests, or follow a redirect to HTTP that require_tls would reject. Could the probe share the normal Kiali client construction so availability is checked with the same transport configuration used by the tools?

Cali0707 (commented):
Thanks for working on this @josunect !

I think we need to sort out more of how to pass toolset config through into this toolset filtering before this will be ready to merge, but otherwise this seems to be going in a good direction

pkg/kiali/gvr.go
josunect: To avoid getting Kiali config via FilteringProvider type assertions: would you prefer (1) changing the core Toolset interface to accept BaseConfig (touches all toolsets), or (2) an optional toolset extension + explicit registration context from the MCP server (Kiali-only)? Open to either, just want to align before pushing further.

pkg/toolsets/kiali/toolset.go
josunect: Agreed on the design concern. `ctx` is now threaded through `collectApplicableTools` and `collectApplicablePrompts`, but `GetTools`/`GetPrompts` don't accept a context yet so we can't propagate it all the way without changing the core `Toolset` interface. Added a TODO. Happy to tackle that as a separate follow-up — does that work for you?

pkg/toolsets/kiali/toolset.go
josunect: Fixed. The kubernetes.Provider assertion is gone. We now use a purpose-built registrationSource interface (BaseConfig() + BearerToken()) implemented by filteringProviderWithConfig, so the assertion succeeds as designed.

pkg/toolsets/kiali/toolset.go
josunect: Agreed on the design concern. ctx is now threaded through collectApplicableTools and collectApplicablePrompts, but GetTools/GetPrompts don't accept a context yet so we can't propagate it all the way without changing the core Toolset interface. Added a TODO. Happy to tackle that as a separate follow-up. does that work for you?

pkg/kiali/status.go
josunect: Done. We now detect *net.DNSError with IsTemporary: true and return nil from probeStatusURL, which HasKiali interprets as "assume available" (fail open). Added a dedicated test case for it. Not sure for other errors. For example, a time out could indicate a temporary issue, or a wrong url...

pkg/toolsets/kiali/toolset.go
josunect: Good point. Added GetPromptsIfAvailable(p) which applies the same reachability gate as GetTools.

pkg/kiali/status.go
josunect: Good point as well. probeStatusURL now receives api.BaseConfig and extracts the full *Config (including TLS settings) via kialiConfigFromBase before constructing the HTTP client, so the probe and normal tool calls share the same transport configuration.

josunect:
@Cali0707 thanks for the review! I addressed some comments, waiting for https://github.com/containers/kubernetes-mcp-server/pull/1425#discussion_r4005210412 to continue.

PR #1426, opened 2026-09-11
feat(kubevirt): redact inline cloud-init payloads from VM tool responses

lyarwood:
## Summary

`vm_lifecycle` and `vm_create` return the full `VirtualMachine` YAML after acting on the cluster object. Any inline cloud-init data (`cloudInitNoCloud`/`cloudInitConfigDrive` `userData`, `userDataBase64`, `networkData`, `networkDataBase64`) would flow into LLM context windows and audit logs unchanged — passwords, SSH keys, and other credentials included.

- Add `pkg/toolsets/kubevirt/internal/redact` with `VMCloudInitFields()`, which replaces those payload fields with `[REDACTED]` in-place on the unstructured object before marshalling
- Wire it into `vm_lifecycle` (most critical — returns the live cluster VM object) and `vm_create` (defensive)
- Secret/ConfigMap references (`userDataSecretRef` etc.) are preserved as they contain no credential values
- The `vm_troubleshoot` tool's existing text-level cloud-init redaction is unaffected
- `vm_clone` returns a `VirtualMachineClone` object (source/target refs only, no VM spec) — no change needed

The `vm_troubleshoot` tool already applied its own `redactCloudInitSensitiveFields()` for the same reason; this change brings the other write tools into line with that precedent.

Relates to: https://redhat.atlassian.net/browse/OLS-3261

## Test plan

- [ ] `go test ./pkg/toolsets/kubevirt/internal/redact/...` — 8 unit tests covering nil VM, no volumes, no cloud-init, `cloudInitNoCloud` plain/base64, `cloudInitConfigDrive`, preservation of non-payload fields, and multiple volumes
- [ ] `go test ./pkg/toolsets/kubevirt/...` — full KubeVirt toolset suite passes
- [ ] Manually verify that a VM with inline `userData` returns `[REDACTED]` in place of the payload after a lifecycle action

### DIFF
--- a/pkg/toolsets/kubevirt/internal/redact/vm.go
+++ b/pkg/toolsets/kubevirt/internal/redact/vm.go
@@ -1,0 +1,41 @@
+package redact
+
+import "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+// VMCloudInitFields replaces inline cloud-init payloads in a VirtualMachine
+// object with "[REDACTED]". This prevents passwords, SSH keys, and other
+// credentials embedded in userData/networkData from appearing in LLM context
+// windows and audit logs. The object is modified in-place; the function is a
+// no-op for objects that have no cloud-init volumes.
+func VMCloudInitFields(vm *unstructured.Unstructured) {
+	if vm == nil {
+		return
+	}
+	volumes, found, err := unstructured.NestedSlice(vm.Object, "spec", "template", "spec", "volumes")
+	if err != nil || !found {
+		return
+	}
+	changed := false
+	for i, vol := range volumes {
+		volMap, ok := vol.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		for _, ciKey := range []string{"cloudInitNoCloud", "cloudInitConfigDrive"} {
+			ciData, ok := volMap[ciKey].(map[string]interface{})
+			if !ok {
+				continue
+			}
+			for _, field := range []string{"userData", "userDataBase64", "networkData", "networkDataBase64"} {
+				if _, has := ciData[field]; has {
+					ciData[field] = "[REDACTED]"
+					changed = true
+				}
+			}
+		}
+		volumes[i] = volMap
+	}
+	if changed {
+		_ = unstructured.SetNestedSlice(vm.Object, volumes, "spec", "template", "spec", "volumes")
+	}
+}
--- a/pkg/toolsets/kubevirt/internal/redact/vm_test.go
+++ b/pkg/toolsets/kubevirt/internal/redact/vm_test.go
@@ -1,0 +1,154 @@
+package redact
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func vmWithVolumes(volumes []interface{}) *unstructured.Unstructured {
+	return &unstructured.Unstructured{
+		Object: map[string]interface{}{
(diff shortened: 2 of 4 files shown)


pkg/toolsets/kubevirt/internal/redact/vm.go
```diff
@@ -0,0 +1,41 @@
...
+			ciData, ok := volMap[ciKey].(map[string]interface{})
+			if !ok {
+				continue
+			}
+			for _, field := range []string{"userData", "userDataBase64", "networkData", "networkDataBase64"} {
+				if _, has := ciData[field]; has {
+					ciData[field] = "[REDACTED]"
+					changed = true
+				}
+			}
```
ksimon1, line 34: This re-implements the same volume-walking / field-list logic (cloudInitNoCloud/cloudInitConfigDrive -> userData/userDataBase64/networkData/networkDataBase64) that already exists as stripCloudInitBodies in pkg/toolsets/kubevirt/vm/troubleshoot/tool.go:201-234, just with a different placeholder string ([REDACTED] vs <see Cloud-Init Configuration section>). Now the list of "these 4 fields are secrets" lives in two places - if a future KubeVirt release adds another cloud-init payload field, it's easy to update one copy and miss the other. Consider extracting one shared helper in internal/redact that both call sites use, parameterizing the placeholder text if needed.

ksimon1:
vm_create_from_template isn't covered by this fix. pkg/toolsets/kubevirt/vm/template/tool.go:76 marshals the full VM object returned after creating a VM from a VirtualMachineTemplate. That tool's own description says it substitutes "auto-generated values like passwords" - which in OpenShift Virtualization common-templates is frequently wired straight into cloudInitNoCloud.userData (e.g. a generated CLOUD_USER_PASSWORD). This is arguably the tool most likely to leak a live credential, and it isn't redacted here.

result there is a map[string]any (not *unstructured.Unstructured), so applying the fix means either wrapping it (redact.VMCloudInitFields(&unstructured.Unstructured{Object: result})) or widening VMCloudInitFields to accept a plain map.

(Posted as a general comment rather than inline since template/tool.go isn't part of this PR's diff.)

ksimon1:
Worth noting: this only closes the leak on vm_lifecycle/vm_create. The generic resources_get/resources_list tools (pkg/toolsets/core/resources.go) can fetch any GVK unmodified, including VirtualMachine and even Secret - so the same userData/networkData (and  serDataSecretRef targets) are one generic call away, unredacted. There's no centralized redaction layer for tool output, so this is a narrower fix (removes an incidental leak from routine lifecycle/create calls) rather than a real boundary against a client that goes looking for the data. Might be worth calling that out explicitly in the PR description, so it's not read as full protection.

Cali0707:
> Worth noting: this only closes the leak on vm_lifecycle/vm_create. The generic resources_get/resources_list tools (pkg/toolsets/core/resources.go) can fetch any GVK unmodified, including VirtualMachine and even Secret — so the same userData/networkData (and serDataSecretRef targets) are one generic call away, unredacted. There's no centralized redaction layer for tool output, so this is a narrower fix (removes an incidental leak from routine lifecycle/create calls) rather than a real boundary against a client that goes looking for the data. Might be worth calling that out explicitly in the PR description, so it's not read as full protection.

yes +1 here

There is also https://github.com/containers/kubernetes-mcp-server/pull/1155, which I believe @manusa was reviewing - I will try to pick that one up and see if we can move something like it forwards again

PR #1429, opened 2026-09-11
feat(auth): Implement Keycloak V2 OIDC token exchange

dlom:
Add per-target exchange configuration for the kubeconfig cluster provider, enabling different audiences per kubeconfig context — the foundation for ACM spoke cluster token exchange. Keycloak's standard token exchange is RFC 8693 compliant (same-realm, no subject_issuer), so it reuses the existing rfc8693 exchanger rather than a dedicated strategy.

Token exchange:
- KubeConfigProviderConfig: per-target STS overrides via cluster_provider_configs.kubeconfig.targets.<context> in TOML
- kubeConfigClusterProvider implements TokenExchangeProvider, overlaying per-target audience/scopes on global STS defaults
- ExchangeTokenInContext fills TokenURL from OIDC discovery when the per-target config leaves it empty

Keycloak realm:
- spoke client + mcp:spoke scope with audience mapper
- Standard token exchange enabled on mcp-server for spoke audience
- Spoke RBAC bindings (cluster-admin for mcp user, view for mcp-viewers)

Multicluster e2e infrastructure:
- Two-minikube setup with spoke on hub's container network (--network)
- Spoke OIDC configured against shared Keycloak via socat proxy
- Pod-friendly kubeconfig rewriting (hub→kubernetes.default.svc, spoke→container IP)
- Skip-if-absent semantics via E2E_SPOKE_CONTEXT env var

Tests verify: context_list shows both clusters, tool calls succeed against hub (global STS) and spoke (per-target exchange), and RBAC is enforced on the spoke (viewer cannot list secrets).


Cali0707:
@dlom do we really need a new exchanger for keycloak v2? Or does the existing RFC 8639 exchanger with specific config options work?

dlom:
Good point, the RFC 8639 exchanger is basically the same. I've dropped the bespoke keycloak v2 exchanger and switched to just rfc8639 in the hub/spoke configs

PR #1431, opened 2026-09-13
feat(auth): add token_exchange.token_url for cross-realm RFC 8693

zruchi:
## Summary

Add support for cross-realm token exchange. Today the RFC 8693 exchange endpoint can only be the token endpoint discovered from `authorization_url` — the realm that validates the incoming user JWT must also be the one performing the exchange. This adds an explicit `token_exchange.token_url` so the two can be different services: user tokens validated against one IDP, exchange performed at a separate STS gateway.

It also makes token exchange possible with `skip_jwt_verification=true`, where there is no OIDC provider to discover an endpoint from at all.

When unset, behavior is unchanged — the endpoint falls back to OIDC discovery from `authorization_url`.

Supersedes #1392 (same goal, built on the old flat `sts_*` keys).

## New TOML key

| Key | Default | Purpose |
|---|---|---|
| `token_exchange.token_url` | *(falls back to OIDC discovery)* | Explicit RFC 8693 token-exchange endpoint. Decouples the exchange STS from the IDP that issues/validates user tokens (`authorization_url`). |


zruchi:
/cc @Cali0707 
Adding the ability for cross-realm token-exchange

zruchi:
@Cali0707 could you have a look? We require this feature so we could exchange token with different oidc. ta

Issue #1435, opened 2026-09-14
Add a test to catch tools that forget to declare RBAC metadata


cajieh:
Follow-up from #1407, which added the `io.kubernetes-mcp-server/rbac` metadata to tools (starting with the `core` toolset). Other toolsets will get this metadata in follow-up PRs.

**The gap:** right now, if a tool doesn't set `RBAC`, nothing fails — it just silently ships without RBAC metadata. There's no way to tell "nobody added this yet" apart from "this tool intentionally needs no RBAC."

**Proposal:** once all toolsets have RBAC metadata, add a test (e.g. `TestAllToolsHaveRBACMetadata`) that loops through every registered tool and fails if any of them have `RBAC == nil`. This way, if someone adds a new tool later and forgets to set RBAC, CI catches it instead of it slipping through review.

cc @Cali0707

PR #1444, opened 2026-09-16
feat(mcp): add rbac metadata to toolsets

Cali0707:
This adds RBAC metadata to all toolsets that are not core.

For kiali/netobserv the rbac is unbounded since those call out to other repo code. As follow ups, I will ask the maintainers of those toolsets to make these bounded where possible - the goal for now is to just have _something_ for the metadata, and to be conservative where we do not know.


ksimon1:
Kubevirt RBACs looks correct

cajieh (commented):
The change looks solid, aside from a few open questions below:

1. **Docs index gap:** `docs/tool-rbac-metadata.md` isn't referenced in `docs/README.md`'s Advanced Topics table, but neither file is touched by this PR — should we update the Advanced Topics table in the README?
2. **helm_list namespace scope:** Given `all_namespaces=true` is a real cluster-wide code path with no way to conditionally bound it in the current RBAC schema, is `AllNamespaces: true` intended as the permanent conservative choice here, or is a schema enhancement (conditional namespace bounds) something we'd want later?
3. **Prompts/resources coverage:** Prompts look fully covered (10 of 38 files), and none of the affected toolsets currently register any resources (`GetResources()` all return `nil`) — should resource RBAC be a placeholder/tracked item for whenever a toolset first adds one, or is it a non-issue until then?
4. **Test coverage gap:** Already tracked in [#1435](https://github.com/containers/kubernetes-mcp-server/issues/1435).

Cali0707:
@cajieh 
>Docs index gap: docs/tool-rbac-metadata.md isn't referenced in docs/README.md's Advanced Topics table, but neither file is touched by this PR — should we update the Advanced Topics table in the README?

Would you mind opening a PR for this one?

>helm_list namespace scope: Given all_namespaces=true is a real cluster-wide code path with no way to conditionally bound it in the current RBAC schema, is AllNamespaces: true intended as the permanent conservative choice here, or is a schema enhancement (conditional namespace bounds) something we'd want later?

This is something we can consider later, depending on need/request. I'm trying to keep this _relatively_ simple for now, and expand as we have concrete use cases

>Prompts/resources coverage: Prompts look fully covered (10 of 38 files), and none of the affected toolsets currently register any resources (GetResources() all return nil) — should resource RBAC be a placeholder/tracked item for whenever a toolset first adds one, or is it a non-issue until then?

Unset RBAC metadata doesn't imply to the client "no rbac needed", so IMO we are fine here if we don't worry about it for now.

PR #1448, opened 2026-09-18
feat(netobserv): declare bounded RBAC metadata on OpenShift

jpinsonneau:
Follow-up to #1444, which set every netobserv tool's RBAC metadata to `unbounded` because the requirements weren't clear from the code.

The netobserv tools call the console plugin backend over HTTP and forward the caller's bearer token; the plugin runs its own `SubjectAccessReview`. On OpenShift this uses the NetObserv multi-tenancy model, so the requirements are the operator's reader ClusterRoles and can be declared as a conservative **bounded** upper bound:

| Tool | Verb / target | ClusterRole |
|------|---------------|-------------|
| `netobserv_list_flows`, `netobserv_export_flows` | `get network.loki.grafana.com` (resourceName `logs`) | `netobserv-loki-reader` |
| `netobserv_get_flow_metrics` | `create pods.metrics.k8s.io` | `netobserv-metrics-reader` |

The declaration is computed per cluster: `GetTools` receives the `FilteringProvider` and `applyToolsets` recomputes the surface on cluster-state changes, so tools advertise `bounded` RBAC on OpenShift and fall back to `unbounded` (with an accurate per-backend reason) on other clusters, where the plugin's Loki/Prometheus backend isn't derivable. OpenShift detection reuses `IsOpenShiftFromProvider` — the same signal already used to select the HTTPS plugin URL.

No `pkg/mcp/testdata` snapshots change (netobserv isn't enabled in those fixtures). New unit tests cover the bounded/unbounded/nil-provider cases; `docs/NETOBSERV.md` now names the reader roles.


pkg/toolsets/netobserv/tools/rbac.go
```diff
@@ -0,0 +1,59 @@
...
+		return api.RBACBounded(api.RBACRequirement{
+			Verbs: []string{"create"},
+			Target: api.RBACTarget{Resource: &api.RBACResourceTarget{
+				APIGroup: "metrics.k8s.io",
+				Resource: "pods",
+			}},
+		})
+	}
+	return api.RBACUnbounded(unboundedMetricsReason)
+}
```
jpinsonneau, line 59: @jotak WDYT about that approach ? It assume `netobserv` rely on Lokistack & Prometheus while running on Openshift. 

Without that assumption, we must remains `RBACUnbounded` with a description.

jotak: rather than relying on being on openshift, can't we check if LokiStack exists? OpenShift is not the real discriminator, netobserv community can also run with LokiStack if the loki operator is installed

jpinsonneau: In that case, users relying on their own loki solution like zero click would be denied.

An alternative would be checking `FlowCollector.spec` but it seems a bit heavy to me here.

Else I can let `unbounded` for every cases with a detailed text.

PR #1450, opened 2026-09-18
feat(kiali): declare bounded RBAC metadata where possible

josunect:
Follow-up to #1444, which set every Kiali tool/prompt RBAC metadata to `unbounded` because access goes through the Kiali HTTP API.

Kiali receives the caller's bearer token and enforces Kubernetes RBAC (namespace get/list as the main gate; user-token access for logs and Istio/Gateway/Inference writes). Where those checks are derivable from tool arguments or a finite kind enum, tools now advertise a conservative **bounded** upper bound. Mesh-wide aggregation stays **unbounded**.

| Tool / area | Declaration |
|-------------|-------------|
| `kiali_get_logs` | bounded — pods + pods/log |
| `kiali_manage_istio_config` / `_read` | bounded — concrete kinds from the tool enum (create/patch/delete or get/list) |
| `kiali_get_resource_details` | bounded — union of resourceType values; on OpenShift also deploymentconfigs + routes |
| `kiali_get_mesh_traffic_graph` | bounded — namespaces + workloads/services (+ deploymentconfigs on OpenShift) |
| `kiali_get_metrics`, `kiali_list_traces`, `kiali_get_trace_details`, `kiali_get_pod_performance` | bounded — namespace get/list gate |
| `kiali_get_mesh_status`, `kiali_list_mesh_clusters` | unbounded — control-plane / multi-cluster config not expressible as a finite K8s requirement set |

OpenShift detection reuses `project.openshift.io` via `FilteringProvider` (same signal as other toolsets). Prompts reuse the same helpers. `docs/KIALI.md` documents the metadata and points at `kiali-viewer` / `kiali` as the practical ClusterRole bindings.


aljesusg (approved):
LGFM

PR #1453, opened 2026-09-21
ci: add DockSec container scan with SARIF upload

advaitpatel:
## What this adds

A GitHub Actions workflow that scans the Dockerfile with [DockSec](https://github.com/OWASP/DockSec)
and uploads the results to the Security tab as SARIF, so container findings are
annotated inline on pull requests.

DockSec is an OWASP Lab Project (MIT). It wraps Hadolint and Trivy and reports
Dockerfile misconfigurations and image CVEs through a single SARIF upload.

## Why

Running DockSec against the current `Dockerfile` reports two Dockerfile findings
and, with the runtime base image included, six HIGH CVEs:

```
[HIGH] CVE-2026-11352  curl/libcurl: remote denial of service
[HIGH] CVE-2026-11586  curl: denial of service via WebSocket PING flood
[HIGH] CVE-2026-8925   curl: double-free in SASL authentication
...3 more HIGH in the ubi9/ubi-minimal layer

[MEDIUM] Pin the base image to an explicit version instead of :latest (line 1)
[LOW]    Add a HEALTHCHECK instruction
```

Both `FROM` lines use `:latest` (lines 1 and 15), so the set of CVEs in a built
image can change between two builds of the same commit.
A note on scope: this workflow scans the Dockerfile only, so the CVE counts above
are not what it will report on day one. Dockerfile scanning is the part that needs
no build and no registry access, which keeps the job fast and the diff small.
Adding `image:` to the same step turns on the image scan once you decide where the
built image should come from, and I am glad to extend the PR that way if you prefer.

## What it does not do

- No API key and no AI calls: the workflow runs with `scan_only: true`, which is
  fully local to the runner.
- It does not gate merges. There is no `fail_on`, so the job reports and never
  fails the build. Adding a gate later is a one-line change.
- It does not replace existing tooling; findings land in the same Security tab.

## Details

- Runs on pull requests and pushes that touch a Dockerfile, plus a weekly
  scheduled scan so newly disclosed CVEs in the base image surface without a
  code change.
- All three actions are pinned to a commit SHA.
- `security-events: write` is scoped to the single job that needs it; the
  workflow's default permission is `contents: read`.

Happy to adjust the trigger set, drop the schedule, or split it into an existing
workflow instead if you would prefer that.

---

Disclosure: I am the project lead for DockSec, the tool this workflow adds. Happy to close this if it is not a fit - no hard feelings either way.


advaitpatel:
Hi @manusa @Cali0707 can I please get your review here? Happy to share more about the tool if needed. Thanks!

matzew:
Thanks for this, and for being upfront that you lead DockSec - that transparency is appreciated. The tool's useful and the findings are real; we don't lint the Dockerfile today, so it's a fair gap to flag.

I'll be upfront back: I'm cautious about taking on a contributor's own project as a repo dependency, independent of how good it is - we try to keep the dependency surface small. And here the Dockerfile checks are Hadolint underneath, so I'd lean toward wiring up Hadolint directly rather than the wrapper.

Nothing against DockSec, and thanks for flagging the gap so cleanly. Open to your take if you see it differently.

